PR #16, opened 2019-11-04
MacroSync

paberr:
Implements the MacroSync protocol into the client.
I haven't integrated it with the config though, since @jgraef changed the config builder in one of his branches completely and I didn't want to do the work twice.

For now, it always uses the full sync client and the behaviour shouldn't be any different from the current one.

The main two TODOs, for which I will open issues after having merged this PR are:
* Integration into config builder.
* When we receive epoch transactions, we should check that:
    * the transactions in the epoch transactions message are consecutive transactions in the proof and not just any subset,
    * consecutive epoch transactions messages align.
Otherwise a client can send us any subset of the transactions and we will only find out that our vector doesn't give the correct hash when he is done sending us transactions.


blockchain-albatross/src/blockchain.rs
```diff
@@ -1062,12 +1062,12 @@ impl<'env> Blockchain<'env> {
...
+    pub fn get_epoch_transactions<B, F: Fn(&BlockchainTransaction) -> B>(&self, epoch: u32, f: F, txn_option: Option<&Transaction>) -> Option<Vec<B>> {
+        let mut transactions = Vec::new();
 
         let first_block = policy::first_block_of(epoch);
         let first_block = self.chain_store.get_block_at(first_block, true, txn_option)?;
-        hashes.extend(first_block.transactions().unwrap().iter().map(|tx| tx.hash()));
+        transactions.extend(first_block.transactions()?.iter().map(&f));
 
         // Excludes current block and macro block.
         let blocks = self.chain_store.get_blocks(&first_block.hash(), policy::EPOCH_LENGTH - 2, true, Direction::Forward, txn_option);
```
jgraef: The transactions for `first_block` were already added. I think you want `&first_block.main_chain_successor?` here.

blockchain-albatross/src/blockchain.rs
```diff
@@ -1062,12 +1062,12 @@ impl<'env> Blockchain<'env> {
         self.chain_store.get_blocks(start_block_hash, count, include_body, direction, None)
     }
 
-    pub fn get_transactions_root(&self, epoch: u32, txn_option: Option<&Transaction>) -> Option<Blake2bHash> {
-        let mut hashes = Vec::new();
+    pub fn get_epoch_transactions<B, F: Fn(&BlockchainTransaction) -> B>(&self, epoch: u32, f: F, txn_option: Option<&Transaction>) -> Option<Vec<B>> {
```
jgraef: Suggestion: rename `B`. I first thought `B` is some kind of block. It's not constrained by any trait, so it's clear that it can be anything, but at a first glance you might get a wrong idea. I suggest renaming it to `U`

jgraef: The method name doesn't really indicate that you'll have to provide a closure to map the transactions. Maybe return an `impl Iterator<Item=&Transaction>`:

```rust
let iter_first = first_block.transactions()?.iter();
let iter_tail = blocks.into_iter().map(|block| block.transactions.iter()).flatten()
iter_first.chain(iter_tail)
```
Although I think this is not quite possible as we need to get the transactions by ownership and use `into_iter`. This can be achieved with:

```rust
if let Block::Micro(block) = block { block.transactions.into_iter() } else { return None }
```

Alternative is to just return the `Vec` of `Transaction`s. We get the blocks by ownership anyway.

Anyway, the consumer of this method should call `map` on it, instead of passing a closure. But I'm okay with keeping it this way, just to save time.

blockchain-albatross/src/blockchain.rs
jgraef: This should be `&first_block.main_chain_successor?`.

blockchain-albatross/src/blockchain.rs
jgraef: Name type parameter `U` instead of `B`. `B` is used for blocks and is misleading on first glance (`T` is misleading too).

The method name doesn't suggest a function that needs a closure. We can easily return a `impl Iterator` that owns the Transactions. We own the blocks. We just need to match `Block::Micro` and get the transaction `Vec`.

```rust
let iter_first = first_block.transactions().into_iter();
let iter_rest = blocks.into_iter().map(|block| block.transactions()).flatten()
iter_first.chain(iter_rest)
```

Replace `block.transactions()` with something that gets the transactions by ownership. Maybe a method `Block::unwrap_transctions(self) -> Vec<Transaction>` even?
```

File: blockchain-albatross/src/blockchain.rs [hunk 3]
```diff
@@ -1077,9 +1077,15 @@ impl<'env> Blockchain<'env> {
...
         for block in blocks {
-            hashes.extend(block.transactions().unwrap().iter().map(|tx| tx.hash()));
+            transactions.extend(block.transactions()?.iter().map(&f));
         }
 
+        Some(transactions)
+    }
+
+    pub fn get_transactions_root(&self, epoch: u32, txn_option: Option<&Transaction>) -> Option<Blake2bHash> {
+        let mut hashes = self.get_epoch_transactions(epoch, |tx| tx.hash(), txn_option)?;
```
jgraef: Clippy will suggest to use `Transaction::hash` instead of that closure.

blockchain-base/src/lib.rs
```diff
@@ -107,6 +108,9 @@ pub trait AbstractBlockchain<'env>: Sized + Send + Sync {
     fn head_hash_from_store(&self, txn: &ReadTransaction) -> Option<Blake2bHash>;
 
     fn get_accounts_chunk(&self, prefix: &str, size: usize, txn_option: Option<&Transaction>) -> Option<AccountsTreeChunk>;
+
+    // TODO: Currently, we can implement request responses in the ConsensusAgent only for *both* protocols, which is why AbstractBlockchain needs to support this.
+    fn get_epoch_transactions<B, F: Fn(&BlockchainTransaction) -> B>(&self, epoch: u32, f: F, txn_option: Option<&Transaction<'env>>) -> Option<Vec<B>>;
```
jgraef: Let's just define that for Nimiq 1.0 epoch n contains only block n. Problem solved. Not that anyone would use this method than for Nimiq 1.0 though :smile:

consensus/src/consensus_agent/requests.rs
```diff
@@ -106,4 +110,44 @@ impl<B: AbstractBlockchain<'static> + 'static, MA: MessageAdapter<B::Block> + 's
...
+            // TODO: Should we send something?
+            return;
+        }
+
+        let transactions: Option<Vec<Transaction>> = self.blockchain.get_epoch_transactions(get_epoch_transactions_message.epoch, |tx| tx.clone(), None);
+        if transactions.is_none() {
+            debug!("[GET-EPOCH-TRANSACTIONS] Could not determine transactions for epoch {:?}", get_epoch_transactions_message.epoch);
+            return;
+        }
+        let transactions = transactions.unwrap();
```
jgraef: This is unidiomatic. You could use a `if let`. But not important :man_shrugging:

jgraef: You can use the `?` operator here. To still have the `debug!` here, you can use `map_err`.

jgraef: Not idiomatic :man_shrugging:

consensus/src/consensus_agent/requests.rs
```diff
@@ -106,4 +110,44 @@ impl<B: AbstractBlockchain<'static> + 'static, MA: MessageAdapter<B::Block> + 's
...
+        let hashes: Vec<Blake2bHash> = transactions.iter().map(|tx| tx.hash()).collect();
+
+        // Fast integer division ceiling, we want ceil(#txs / MAX_TRANSACTIONS).
+        let num_chunks = (transactions.len() + EpochTransactionsMessage::MAX_TRANSACTIONS - 1) / EpochTransactionsMessage::MAX_TRANSACTIONS;
```
jgraef: Would be useful to put this into `nimiq_utils`. There is a `log2` module in there. We could rename it to `math` and put random functions for calculations in therel

consensus/src/consensus_agent/sync.rs
```diff
@@ -1,72 +1,340 @@
...
+    Finished,
+}
+
+struct MacroBlockSyncState {
+    /// Cache of the next max. 1000 blocks.
+    block_cache: VecDeque<AlbatrossBlock>,
+    /// Transactions of the current block.
+    transactions_cache: Vec<Transaction>,
+    /// The current state of the syncing.
+    state: MacroBlockSyncingState,
```
jgraef: I heard you like states, so I put a syncing state into your sync state :rofl:

consensus/src/consensus_agent/sync.rs
```diff
@@ -1,72 +1,340 @@
...
+    fn new() -> Self {
         Self {
-            blockchain
+            block_cache: VecDeque::new(),
+            transactions_cache: Vec::new(),
+            state: MacroBlockSyncingState::Finished,
+            processing_epoch: false,
         }
     }
 }
```
jgraef: Is there any chance this will have arguments in the future? If not, derive `Default` for the struct and just use that. Also clippy will warn, if you have a constructor without arguments, but no `Default`.

jgraef: Derive `Default` instead of `new`. Clippy will also complain if there is no `Default` for structs with constructors without arguments.

jgraef (changes requested):
Fix `get_epoch_transactions`.

jgraef:
Some are duplicate. I had close the browser and thought they were lost.

blockchain-albatross/src/blockchain.rs
paberr: `get_blocks` does not include the `start_hash` block, so this is actually fine.

consensus/src/consensus_agent/requests.rs
paberr: Cool idea, we use it quite often

paberr:
> Fix `get_epoch_transactions`.

Doesn't need fixing, but I will take care of the style issues.

jgraef (dismissed):
:+1:

jgraef (approved):
Stil :+1: also after rebasing.

PR #25, opened 2019-11-06
PartialMerkleProof

### LINKED ISSUE #19: Make (Get)EpochTransactions more DoS resistant
Once #16 is merged, it still needs to become more DoS resistant.

Currently, the following could be exploited:
- *Peer A* sends a `GetEpochTransactions` request to malicious *Peer B*
- Let's suppose the epoch contained 4 transactions: `[0, 1, 2, 3]`
- *Peer B* is now able to send `EpochTransactions` messages with correct proofs for any subsets. For example, he could send such messages for `[0, 1]`, `[0, 2]`, `[0, 3]`, `[0, 1, 2]`, `[2, 3]`, ...
- All these message would currently pass our checks.
- Only after sending the last message, *Peer A* would try creating the Merkle root of `[0, 1, 0, 2, 0, 3, 0, 1, 2, 2, 3, ...]` and realise that it doesn't match.

How to prevent this?
The best way is to replace the current Merkle proofs with a setting similar to our `AccountsTreeChunks`.
Intuitively, we only provide a proof for the rightmost transaction of our chunk of consecutive transactions. If these transactions are consecutive transactions, we can calculate the proof from there and check.

Consider the following Merkle tree:
```
                   h13
            /               \
          h11                h12
       /      \            /   \
     h7        h8        h9     h10
   /   \     /   \     /   \    |
  h0   h1   h2   h3   h4   h5   h6
   |    |    |    |    |    |    |
  t0   t1   t2   t3   t4   t5   t6
```
We want to split the list of six transactions into three chunks: `[t0, t1, t2]`, `[t3, t4, t5]`, `[t6]`.
The first chunk consists of the transactions `[t0, t1, t2]` and the following *list of hashes* (what we call *proof* in the following): `[h3, h12]`
From the tree picture above, one can see that these two hashes, together with the first three transactions, suffice to verify the root.

The next chunk consists of `[t3, t4, t5]` and the proof is `[h10]`.
In combination with the previously received transactions, this hash again suffices to calculate the root.

The proof for our last chunk of transactions is always empty (even indicating that this is the last set of transactions).

paberr:
This is an implementation of the data structure defined in #19.
Also integrates the data structure and makes use of our new `ceiling_div` function in many places.

Closes #19.

### DIFF
--- a/consensus/src/consensus_agent/requests.rs
+++ b/consensus/src/consensus_agent/requests.rs
@@ -5,7 +5,7 @@ use futures::Future;
 use tokio::prelude::*;
 
 use blockchain_base::AbstractBlockchain;
-use hash::{Blake2bHash, Hash};
+use hash::Blake2bHash;
 use network_messages::{
     AccountsProofMessage,
     AccountsTreeChunkData,
@@ -22,9 +22,8 @@ use network_messages::{
     TransactionReceiptsMessage,
     TransactionsProofMessage,
 };
-use transaction::{Transaction, TransactionsProof};
-use utils::math::CeilingDiv;
-use utils::merkle::Blake2bMerkleProof;
+use transaction::Transaction;
+use utils::merkle::partial::PartialMerkleProofBuilder;
 
 use crate::consensus_agent::ConsensusAgent;
 use crate::ConsensusProtocol;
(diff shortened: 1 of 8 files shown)


jgraef (dismissed):
Looks good

jgraef (approved):
Was already approved, but rebased to albatross

Issue #35, opened 2019-11-08
Invalid Block after finishing sync


jgraef:
With the new client and client API the connection to the seed fails after finishing the sync. This is configured as full sync, but synced up suspiciously fast.

The code for this is on branch `janosch/new-client`.

## Log

```
 INFO  consensus            | Now at block #33000
 INFO  nimiq_client2        | Head: #33072 - 868593894cb7856dd4c1c726c8748645446ec8789ceada7667bca060ed71d952, Peers: 1
 INFO  consensus            | Now at block #33100
 INFO  consensus            | Now at block #33200
 DEBUG consensus            | Finished sync with peer ws://5.0.0.5:8443/1a9c3545e64efca6283c1d2c9c1ba231
 INFO  consensus            | Synced with all connected peers (1), consensus established
 INFO  consensus            | Blockchain at block #33256 [4e81becd8edcbb779a94f448a1f3782ddc9487b2855d5dd58f9ac1c57a8ded97]
 WARN  blockchain           | Failed to determine slots - preceding macro block not found: block_number=33300, view_number=0, state.block_number()=33256
 DEBUG sink                 | Closing connection, reason: InvalidBlock (None)
 DEBUG connection_pool      | Peer left: ws://albatross.nimiq.dev:8444/1a9c3545e64efca6283c1d2c9c1ba231 144.76.172.80 (version=Some(1), closeType=InvalidBlock)
 INFO  consensus            | Disconnected from ws://5.0.0.5:8443/1a9c3545e64efca6283c1d2c9c1ba231
 DEBUG connection_pool      | Connection established (outbound) #0 144.76.172.80 ws://albatross.nimiq.dev:8444/1a9c3545e64efca6283c1d2c9c1ba231
 DEBUG connection_pool      | Peer joined: ws://5.0.0.5:8443/1a9c3545e64efca6283c1d2c9c1ba231 (v1, FULL | VALIDATOR, core-rs/0.1.0 (native; linux x86_64))
 INFO  consensus            | Connected to ws://5.0.0.5:8443/1a9c3545e64efca6283c1d2c9c1ba231
 DEBUG network_agent        | Requesting addresses from ws://5.0.0.5:8443/1a9c3545e64efca6283c1d2c9c1ba231
 DEBUG consensus            | Syncing blockchain with peer ws://5.0.0.5:8443/1a9c3545e64efca6283c1d2c9c1ba231
 INFO  consensus            | Now at block #33257
[...]

PR #36, opened 2019-11-09
Add missing orphan check in push_block

paberr:
While our Nimiq 1.0 blockchain does this check, it was missing in the Albatross code and caused the issue in #35.
Instead of returning an `Orphan` result, the verification will stop when trying to get the block producer and return `InvalidBlock` instead.


jgraef (approved):
LGTM

PR #47, opened 2019-11-09
Move path specification to the root crate

### LINKED ISSUE #30: Remove path from dependencies
Instead of specifying our dependencies like this

```toml
nimiq-database = { path = "../database", version = "0.1" }
```

we can omit the `path` and have a `patch` section in the workspace's `Cargo.toml`:

```toml
[patch.crates-io]
nimiq-database = { path = "../database" }
```

jeffesquivels:
Closes #30

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,3 +49,44 @@ members = [
 [profile.dev.overrides.pairing]
 opt-level = 3
 debug = false
+
+[patch.crates-io]
+beserial = { path = "beserial" }
+beserial_derive = { path = "beserial/beserial_derive" }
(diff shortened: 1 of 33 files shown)


jgraef (approved):
I didn't check every single line, but since it compiles it looks good to me. And thank you so much for this. Never ever do we have to write those paths into the dependencies again :smile:

jgraef:
Okay, there is actually one more thing we could change here. But it's not actually necessary. Insteaf of
```
nimiq-network = { version = "0.1" }
```
we can write
```
nimiq-network = "0.1"
```

PR #51, opened 2019-11-10
WIP: Method to get a fingerprint string for a validator public key

jgraef:
I think there is a good use-case for when we need a shorter identifier for validators. The public key is 96 bytes after all.

I can still improve this PR by having the fingerprint method just return a Blake2bHash. This can then easily be converted to a hexadecimal representation anyway and would totally suffice.


jgraef:
I'm rewriting how `ValidatorInfo`s are exchanged. I think I need this in there and will therefore include this feature in another pull request.

jgraef:
Also `CompressedPublicKey` already implements `Hash`, which is enough to fingerprint a validator.

PR #55, opened 2019-11-10
Prepare accounts tree with generic parameter

paberr:
This is a preparation step for the UTXO tree accounts.
It does simply introduce the generic parameter and a corresponding trait, but does not change logic at all.

### DIFF
--- a/accounts/src/tree/mod.rs
+++ b/accounts/src/tree/mod.rs
@@ -1,6 +1,7 @@
+use std::marker::PhantomData;
 use std::str::FromStr;
 
-use account::Account;
+use account::AccountsTreeLeave;
 use database::{Database, Environment, Transaction, WriteTransaction};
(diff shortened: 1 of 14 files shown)


accounts/src/tree/mod.rs
```diff
@@ -10,32 +11,33 @@ use tree_primitives::accounts_tree_node::{AccountsTreeNode, NO_CHILDREN};
 use tree_primitives::address_nibbles::AddressNibbles;
 
 #[derive(Debug)]
-pub struct AccountsTree {
-    db: Database
+pub struct AccountsTree<A> {
+    db: Database,
+    _account: PhantomData<A>,
 }
```
jgraef: Shouldn't this be:

```rust
pub struct AccountsTree<A: AccountsTreeLeave>
```

jgraef (approved):
Good to merge

PR #68, opened 2019-11-15
Refactored slots data model

jgraef:
## Pull request checklist

Refactored slot data model. See `primitives/src/slot.rs`. Also remove slashing partly, as I didn't want to fix it for my changes, since it's going away anyway.

The part with the tests isn't complete correct. I disabled on test, because @paberr fixed it.

@paberr I can merge your PR first and then rebase this one.


blockchain-albatross/src/chain_store.rs
```diff
@@ -111,20 +111,21 @@ impl ChainStore {
...
-            !chain_info.on_main_chain
-        } {
-            // Loop Body
+
+            // If it's on the main chain we can return from loop
+            if chain_info.on_main_chain {
+                break chain_info;
+            }
+
+            // get next blokc hash
```
paberr:
```suggestion
            // Get next block hash
```

primitives/block-albatross/src/macro_block.rs
```diff
@@ -56,66 +48,52 @@ pub struct MacroHeader {
...
+#[derive(Clone, Debug, Fail)]
+pub enum IntoSlotsError {
+    #[fail(display = "Extrinsics missing in macro block")]
+    MissingExtrinsics,
+}
+
 impl TryInto<Slots> for MacroBlock {
     type Error = IntoSlotsError;
 
+    // FIXME: One validator can have multiple stakes, so slot_addresses is not necessarily the same
```
paberr: Remove FIXME

validator/src/validator_agent.rs
```diff
@@ -162,19 +162,14 @@ impl ValidatorAgent {
         let block_number = fork_proof.block_number();
         let view_number = fork_proof.view_number();
 
-        let producer = self.blockchain.get_block_producer_at(block_number, view_number, None);
-        if producer.is_none() {
-            debug!("[FORK-PROOF] Unknown block producer for #{}.{}", block_number, view_number);
```
paberr, line 167: We don't have an error message for the case we cannot determine the block producer anymore.
Is this a problem?

paberr (changes requested):
I will review the rest of the PR tomorrow.

blockchain-albatross/src/blockchain.rs
```diff
@@ -111,9 +109,11 @@ pub struct BlockchainState {
     macro_head: MacroBlock,
     macro_head_hash: Blake2bHash,
 
-    current_validators: Option<Validators>,
-    last_validators: Option<Validators>,
+    // TODO: Instead of Option, we could use a Cell here and use replace on it. That way we know
+    // at compile-time that there is always a valid value in there.
     current_slots: Option<Slots>,
+
+    // TODO: Rename to `previous_slots`?
```
paberr: Yes, I would actually do that, if you touch the code anyway.

blockchain-albatross/src/blockchain.rs
```diff
@@ -111,9 +109,11 @@ pub struct BlockchainState {
     macro_head: MacroBlock,
     macro_head_hash: Blake2bHash,
 
-    current_validators: Option<Validators>,
-    last_validators: Option<Validators>,
+    // TODO: Instead of Option, we could use a Cell here and use replace on it. That way we know
```
paberr, line 113: Indeed, there must always be a `current_slots`, but not a `previous_slots`.

blockchain-albatross/src/blockchain.rs
```diff
@@ -469,11 +468,11 @@ impl Blockchain {
         };
 
         // Public keys are checked when staking, so this should never fail.
-        let slot: IndexedSlot = self.get_block_producer_at(block.block_number(), block.view_number(), Some(&read_txn))
-            .ok_or(PushError::InvalidSuccessor)?;
+        let (slot, _) = self.get_slot_at(block.block_number(), block.view_number(), Some(&read_txn))
+            .unwrap();
```
paberr, line 470: We had `unwrap` at this point before and the code panicked. If I remember correctly though, that only happened because the check for the previous block being present was missing.

blockchain-albatross/src/blockchain.rs
```diff
@@ -1239,72 +1220,56 @@ impl Blockchain {
...
+                .unwrap_or_else(|| panic!("Missing previous epoch's slots for block {}.{}", block_number, view_number))
+        }
+        else {
             let macro_block = self.chain_store
                 .get_block_at(policy::macro_block_before(block_number), true, Some(&txn))
-                .or_else(|| {
-                    warn!("Failed to determine slots - preceding macro block not found: block_number={}, view_number={}, state.block_number()={}", block_number, view_number, state.block_number());
-                    None
-                })?
+                .unwrap_or_else(|| panic!("Failed to determine slots - preceding macro block not found: block_number={}, view_number={}, state.block_number()={}", block_number, view_number, state.block_number()))
```
paberr: Actually, this *can happen* once we implemented a fast sync method based on zk-proofs!

blockchain-albatross/src/reward_registry/mod.rs
```diff
@@ -300,38 +299,52 @@ impl SlashRegistry {
...
+        // * Use HashRng
+        // * Instead of sampling from the honest "validators" (which are slots really),
+        //   create a list of slot numbers that are still enabled. Then sample from that and
+        //   you'll have the proper slot_number instead of a "honest slot index".
 
-            // Get slots of epoch
-            let slashed_set = self.slashed_set_at(policy::epoch_at(block_number), block_number, txn_option).unwrap();
-            let honest_validators = Vec::from_iter(SlashedSlots::new(&slots, &slashed_set).enabled().cloned());
+        // Get slots of epoch
+        let slashed_set = self.slashed_set_at(policy::epoch_at(block_number), block_number, txn_option).unwrap();
```
paberr: This fails for blocks in historic epochs (< current_epoch - 1).

blockchain-albatross/src/blockchain.rs
```diff
@@ -1239,72 +1220,56 @@ impl Blockchain {
...
-        if block_number >= policy::first_block_of_registry(epoch_number) {
-            return None;
-        }
+            &slots_owned
+        };
 
-        self.chain_store.get_chain_info_at(block_number, false, Some(&txn))
-            .filter(|info| view_number == info.head.view_number())
-            .and_then(|info| info.slot)
+        state.reward_registry.get_slot_at(block_number, view_number, slots, Some(&txn))
```
paberr, line 1240: This will fail for historic epochs as well (see comment in reward_registry.
However, I don't think we use this for any old blocks.

blockchain-albatross/src/blockchain.rs
```diff
@@ -1239,72 +1220,56 @@ impl Blockchain {
...
     }
 
     pub fn state(&self) -> RwLockReadGuard<BlockchainState> {
         self.state.read()
     }
 
     pub fn create_slash_inherents(&self, fork_proofs: &[ForkProof], view_changes: &Option<ViewChanges>, txn_option: Option<&Transaction>) -> Vec<Inherent> {
-        let mut inherents = vec![];
+        /*let mut inherents = vec![];
```
paberr, line 1248: Why do you comment this part out? We still need that.

primitives/src/slot.rs
```diff
@@ -0,0 +1,773 @@
...
+        let mut combined = Vec::new();
+        let mut validator_slots_iter = self.validator_slots.iter().peekable();
+        let mut start_stake_slot_number = 0;
+        let mut start_validator_slot_number = 0;
+
+        for stake_slot in self.stake_slots.iter() {
+            let validator_slot = *validator_slots_iter.peek()
+                .expect("Expected validator slot");
+
+            let end_stake_slot_number = start_stake_slot_number + stake_slot.num_slots();
```
paberr: I would move these two declarations down to the place where they are actually needed.
Makes it a bit more readable IMO.

primitives/src/slot.rs
```diff
@@ -0,0 +1,773 @@
...
+/// Internally this is just a [std::vec::Vec] with the indices into a [std::vec::Vec] with
+/// [SlotBand]s.
+///
+/// # ToDo
+///
+/// * Use a [nimiq_collections::segment_tree::SegmentTree]
+///
+#[derive(Clone)]
+struct SlotIndexTable {
+    index: Vec<u16>
```
paberr, line 634: A SegmentTree might be more efficient here.

paberr: Oh, just saw that you put the same ToDo into your comments above.

primitives/src/slot.rs
```diff
@@ -0,0 +1,773 @@
...
+    const SIZE: usize = ((SLOTS as usize) + 8 - 1) / 8;
+
+    pub fn from_iter<I: Iterator<Item=u16>>(iter: I) -> Result<Self, SerializingError> {
+        let mut bits = BitVec::with_capacity(SLOTS as usize);
+        bits.resize(SLOTS as usize, false);
+
+        let mut i = 0;
+
+        for num_slots in iter {
+            // this is not allowed
```
paberr:
```suggestion
            // This is not allowed
```

paberr (changes requested):
Done

validator/src/validator_agent.rs
jgraef: So the slot is unknown only if it's for a future block. We can keep a debug message. But I'd just ignore them. We should at some point notify the metrics server for such events though.

blockchain-albatross/src/blockchain.rs
jgraef: I'll leave it for now, as I couldn't find a easy way to have the same behavior in regards to `AsRef`.

blockchain-albatross/src/blockchain.rs
jgraef: As long as the previous block (is checked) and the previous macro block exist, we're fine here.

I removed the comment before that line as it doesn't make any sense.

blockchain-albatross/src/blockchain.rs
jgraef: No, this returns None if `RewardRegistry::get_slot_at` returns `None`

blockchain-albatross/src/blockchain.rs
jgraef: Because I didn't want to repeat your work.

primitives/src/slot.rs
jgraef: I think for such small collections `Vec` is more efficient. It's only initialization cost and then constant lookup time.

blockchain-albatross/src/blockchain.rs
paberr: But only after your changes ;)

blockchain-albatross/src/chain_store.rs
```diff
@@ -111,20 +111,21 @@ impl ChainStore {
...
-            !chain_info.on_main_chain
-        } {
-            // Loop Body
+
+            // If it's on the main chain we can return from loop
+            if chain_info.on_main_chain {
+                break chain_info;
+            }
+
+            // Get next blokc hash
```
paberr, line 123:
```suggestion
            // Get next block hash
```

paberr (approved):

PR #70, opened 2019-11-16
Remove slashing of stake and instead park slashed validators

paberr:
## Pull request checklist

Remove slashing of stake and instead park slashed validators
Parking will lead to automatic retirement of the stake after the next epoch
Add unpark transaction to prevent being retired

### DIFF
--- a/accounts/src/accounts.rs
+++ b/accounts/src/accounts.rs
@@ -89,6 +89,9 @@ impl Accounts {
     fn commit_nonfinal(&self, txn: &mut WriteTransaction, transactions: &[Transaction], inherents: &[Inherent], block_height: u32) -> Result<Receipts, AccountError> {
         let mut receipts = Vec::new();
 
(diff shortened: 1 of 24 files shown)


accounts/src/accounts.rs
```diff
@@ -282,10 +290,11 @@ impl Accounts {
         Ok(receipt)
     }
 
-    fn prepare_receipts(receipts: &Receipts) -> (ReceiptsMap, ReceiptsMap, ReceiptsMap, Vec<&PrunedAccount>) {
+    fn prepare_receipts(receipts: &Receipts) -> (ReceiptsMap, ReceiptsMap, ReceiptsMap, ReceiptsMap, Vec<&PrunedAccount>) {
```
jgraef, line 293: Maybe we should make a struct type that we return here instead of that tuple. But that's only a suggestions... one that Clippy is likely to make :wink:

jgraef (approved):
:+1:

PR #81, opened 2019-11-21
Janosch/merge from devnet

jgraef:
## Pull request checklist

Adds `nimiq-client.service` again and fixes small bug related to validator key configuration.

### DIFF
--- a/client/nimiq-client.service
+++ b/client/nimiq-client.service
@@ -3,8 +3,7 @@ Description=Nimiq's Rust client
 After=syslog.target network.target remote-fs.target nss-lookup.target
 
 [Service]
-Type=forking
-PIDFile=/run/nimiq-client.pid
+Type=exec
 ExecStartPre=$(which nimiq-client)
 ExecStart=$(which nimiq-client)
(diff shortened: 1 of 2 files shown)


client/nimiq-client.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Nimiq's Rust client
+After=syslog.target network.target remote-fs.target nss-lookup.target
+
+[Service]
+Type=forking
```
jeffesquivels: For our case, this should probably be changed to `Type=exec` as described [here](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Type=)

client/nimiq-client.service
```diff
@@ -0,0 +1,15 @@
...
+PIDFile=/run/nimiq-client.pid
```
jeffesquivels: I'm not sure if we are creating this file, if not, this should probably be removed.

jeffesquivels (changes requested):

jeffesquivels (approved):

PR #84, opened 2019-11-22
VRF abstraction and RNG

jgraef:
## Pull request checklist

VRF is now more abstracted:

* `VrfSeed`, which encapsulates the BLS signature and provides methods for verifying or signing the next block's seed.
* `VrfRng` takes in a `VrfSeed` and generates randomness  from it using hashes. The hashes use the signature, a counter and a *usage* as input.
* The *usage* is just an enum `VrfUsage` that declares where the RNG is used, i.e. in Validator Selection, Slot selection or Reward distribution (Remainder). I added this as extra input because otherwise some random sequences would be equal. I think the reward distribution and validator selection at the macro block and the slot selection for the first microblock would otherwise all use the same sequence of numbers.

I changed the validator selection to use the Index method with O(n) setup time and O(1) sampling time. The code for the alias method is currently in a seperate crate in my personal Github. I'd suggest to move it into the Nimiq workspace though, as it's very critical to the consensus.

As already mentioned I changed the distribution of the remainder of the epoch reward to be random. The RNG is seeded with the seed in the macro block that triggers the previous epoch to be finalized and not the seed from the macro block from the epoch we finalize.


blockchain-albatross/src/blockchain.rs
```diff
@@ -1424,16 +1406,24 @@ impl Blockchain {
...
+        // Get RNG from last block's seed and build lookup table based on number of eligible slots
+        let mut rng = macro_header.seed
+            .rng(VrfUseCase::RewardDistribution);
+        let lookup = LookupTable::new(num_eligible_slots_for_accepted_inherent);
+
+        // Randomly distribute remainder over accepting slots.
+        while !remainder.is_zero() {
+            let x = rng.next_u64_max(lookup.len() as u64) as usize;
+            let y = rng.next_u64_max(lookup.total() as u64) as u16;
+            let index = lookup.sample(x, y);
```
paberr: In my opinion, it would be nicer to hide generating `x` and `y` coordinates in the `sample` function.

primitives/account/src/staking_contract.rs
```diff
@@ -465,22 +464,22 @@ impl StakingContract {
...
-        //     parallelized.
-        let rng = HashRng::<_, Blake2bHash>::from_seed(seed.clone());
         let mut slots_builder = SlotsBuilder::default();
+        let lookup = LookupTable::new(weights);
+        let mut rng = seed.rng(VrfUseCase::ValidatorSelection);
+
+        for _ in 0 .. policy::SLOTS {
+            let x = rng.next_u64_max(lookup.len() as u64) as usize;
+            let y = rng.next_u64_max(lookup.total());
+            let index = lookup.sample(x, y);
```
paberr: As mentioned above, I don't really like that the user has to take care of generating `x` and `y`.
Also the name `sample` is then misleading as it doesn't really sample from a distribution, but only looks up the coordinates.

paberr (changes requested):

jgraef:
Changes:

 * `Rng` trait
 * Moved alias method into vrf crate, renamed to `AliasMethod`
 * `AliasMethod::sample` takes `Rng` and samples from it.
 * Use RNG for slot selection

paberr (dismissed):

paberr (approved):

PR #92, opened 2019-12-02
RPC server: Add peer_public_key method

riptl:
`peer_public_key` takes no arguments and returns the peer public key as a hex string.

## Pull request checklist

To generate a seed list from a swarm of auto-generated validators, I need some way to get the peer public key from RPC.

Maybe we should create a `peer_config` call though and include more information like peer ID.


jgraef (approved):
:+1:

Issue #97, opened 2019-12-13
Integer overflow in validator timeout


jgraef:
```
DEBUG validator            | Completed view change to #128.1
thread 'tokio-runtime-worker-5' panicked at 'attempt to subtract with overflow', validator/src/validator.rs:377:44
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

jgraef:
I will fix this together with #54

brunoffranca:
The relevant lines for this are https://github.com/nimiq/core-rs-albatross/blob/albatross/validator/src/validator.rs#L515-L516
This should be fixed together with the Tendermint Implementation.
@fiaxh @nibhar

jgraef:
I can't find the error location anymore. Since the validator has been rewritten a lot, I'm going to close this.

Issue #54, opened 2019-11-10
Macro Block PBFT


paberr:
As explained in #53 and #49, there are still some implementation issues around the Macro Block PBFT.
I discussed these with @jgraef and Bruno Franca just now and here's the list of things that still need to be changed:

1. The timeout for view changes should stop when receiving a valid proposal. Even if the network is split at that moment and half of the validators start a view change, they won't be able to complete it and can thus accept the valid proposal after a reunion of the network.
2. A validator could provide two valid, but different proposals. The way to resolve this is to define a strict ordering on proposals (e.g., by their hash) and require honest validators to always choose the lowest.
3. Considering case 2. and a split network, it can happen that half of the validators sends `prepare` for proposal A and the other half for B. So, after a reunion of the network, validators have to be allowed to send another `prepare` for the lower block (e.g., A). But under no circumstances, they will ever send `commit` messages for both blocks. This last requirement is actually already enforced by the rule to only send `commit` if you have 2f+1 prepares.

styppo:
![](https://github.trello.services/images/mini-trello-icon.png) [Fix MacroBlock PBFT](https://trello.com/c/Gwv50bFs/5-fix-macroblock-pbft)

jgraef:
Currently a validator doesn't verify the extrinsics hash in the proposal. I think this was fine, because a macro block with invalid extrinsics wasn't accepted, but view-changed, even after everyone committed to it. Since we only want commits on one block, we need to check that. So when receiving a pBFT proposal, we need to compute the extrinsics and check that the hash in the proposal is correct. We can then also store the extrinsics to get the full block once the proposal is accepted.

paberr:
While reviewing the PR #99, I thought a bit about the following scenario:
- A block is published and 2/3 of the validators send a prepare for this block.
- At this point, the block producer publishes a new block with a lower hash to only a subset of the validators that also sent prepare for the previous block.
- Validators that did not receive the new block will commit on the first block.

What will happen now? The validators who committed on the first block will not be able to commit on the second with the lower hash.
But the rest of the validators also won't be able to finish the second block's PBFT.

brunoffranca:
Analyzing the rules, we know that:

1. Validators never send two commit messages in the same block height and view number.
2. Whenever validators receive 2f+1 prepare messages, they send a commit message.

[...]

PR #99, opened 2019-12-13
Fix MacroBlock pBFT

jgraef:
## Pull request checklist

This is *work in progress*, because the extrinsics root check fails when a validator becomes parked.

 * Fix integer overflow #97 
 * Disable view change timer after we got a valid proposal (and send prepare).
 * Compute state and transactions root and check it.
 * Compute extrinsics, check the extrinsics root and remember extrinsics for when commit is complete.
 * Keep track of current "best" proposal (i.e. smallest hash)
 * Keep track if we already committed to a proposal

### DIFF
--- a/validator/src/validator.rs
+++ b/validator/src/validator.rs
@@ -1,4 +1,3 @@
-use std::collections::HashMap;
 use std::ops::Mul;
 use std::sync::{Arc, Weak};
 use std::time::Duration;
@@ -90,7 +89,14 @@ pub struct ValidatorState {
     fork_proof_pool: ForkProofPool,
     view_number: u32,
     active_view_change: Option<ViewChange>,
-    proposed_extrinsics: HashMap<Blake2bHash, MacroExtrinsics>,
+
+    /// This stores the proposal hash and proposed extrinsics for the current view.
+    ///
+    /// This is cleared when there is a slot change (or the blockchain was extended).
+    proposal: Option<(Blake2bHash, MacroExtrinsics)>,
+
+    /// Whether we already committed to the proposal
+    committed: bool,
 }
 
 impl Validator {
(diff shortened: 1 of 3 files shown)


paberr:
We're still discussing whether https://github.com/nimiq/core-rs-albatross/blob/bbcf2a6739675c7a5e840f279763e5b604dd9744/block-production-albatross/src/lib.rs#L84 makes sense.

validator/src/validator.rs
```diff
@@ -413,19 +424,105 @@ impl Validator {
...
+            return;
+        }
+
+        // Check if this proposal is "better"
+        if let Some((current_proposal_hash, _)) = &state.proposal {
+            if *current_proposal_hash == hash {
+                // We either produced this block, or already started pBFT. Actually if we're not the
+                // producer this should never be called multiple times for the same hash.
+                warn!("Proposal already known: {}", hash);
+                return
```
paberr: The code is a bit inconsistent here with `return` vs `return;`.
I'll make this consistent when rebasing it.

validator/src/validator.rs
```diff
@@ -413,19 +424,105 @@ impl Validator {
...
+            }
+            else {
+                // We already know of another proposal, so check which one we should keep.
+                // We just define that we'll always use the one with the lower hash
+                if *current_proposal_hash < hash {
+                    // The current proposal is "better", so do nothing
+                    info!("Received proposal with greater hash: {}", hash);
+                    return
```
paberr: See https://github.com/nimiq/core-rs-albatross/issues/54#issuecomment-567889016

paberr (dismissed):
We should still discuss the behaviour in my comment in #54, but apart from that this looks good to me.

validator/src/validator.rs
jgraef: Better?

paberr:
TODO: Check which part of the changes still make sense

brunoffranca:
Except for fixing the integer overflow, the rest no longer seems relevant because of the change to Tendermint.

PR #104, opened 2019-12-16
Configurable block timeout

jgraef:
## Pull request checklist

- [ ] All tests pass. Demo project builds and runs.

This builds on top of #54.
This adds a configuration option to set the block timeout for a validator.

The motivation is to be able to easily change the block timeout in a deployment. Since no validator can be forced to wait a fixed time anyway, there is no reason to make it not configurable.

The build fails here too.

### DIFF
--- a/block-production-albatross/src/lib.rs
+++ b/block-production-albatross/src/lib.rs
@@ -122,6 +122,35 @@ impl BlockProducer {
         }
     }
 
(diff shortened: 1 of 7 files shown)


brunoffranca:
@nibhar @fiaxh 
Do you think it is worth it to implement this for Tendermint now?

jgraef:
> @nibhar @fiaxh
> Do you think it is worth it to implement this for Tendermint now?

Our conclusion was that this (and other hidden development config options) are worth having, but we'll delay until the Tendermint is ready.

PR #108, opened 2019-12-16
WIP: Fix timer issues

jgraef:
## Pull request checklist

# `ConsensusAgent`

When the consensus agent finishes a sync, it sets a timer to request the mempool of the peer. This fix also clears the timer when the agent goes out of sync.

# `InventoryManager`

There is a duplicate timer issue when requesting an inventory vector:
```
logs/devnet_validator.4.ul1icdeypqka9t8nergz7fmo0.log: 2019-12-16 21:28:41.574 ERROR timers               | Duplicate delay for key Request(InvVector { ty: Transaction, hash: c2b4a4d11e7c22e2ff8e1a60ac82969ddaeb5757f0083a519f4e977258136cb7 })
```

TODO: Fix this

### DIFF
--- a/consensus/src/consensus_agent/mod.rs
+++ b/consensus/src/consensus_agent/mod.rs
@@ -459,6 +459,7 @@ impl<P: ConsensusProtocol + 'static> ConsensusAgent<P> {
 
     fn out_of_sync(&self) {
         self.timers.clear_delay(&ConsensusAgentTimer::ResyncThrottle);
+        self.timers.clear_delay(&ConsensusAgentTimer::Mempool);
 
         self.state.write().synced = false;
 


terorie:
[Fix timer issues](https://trello.com/c/k3n4zZLs/18-fix-timer-issues)

jgraef:
Regarding the duplicate timer in `InventoryManager`:

There is a race-condition between `InventoryAgent::on_inv` and `InventoryAgent::no_more_data`. Both call `InventoryManager::request_vector` and thus will set the timer. `on_inv` is triggered by an incoming message and is guarded by a mutex. `no_more_data` will try to request the remaining vectors. It is not guarded by the mutex `on_inv` uses. If both run concurrently, they can both call into `request_vector`.

There is also `InventoryAgent::on_not_found`. This will also call into `request_vector`. `on_inv` and `on_not_found` are semantically exclusive, but a peer could send both.

Furthermore we should also take a look at `InventoryAgent::on_block`. It doesn't set a timer, but only clears it. It might still be subject to a data race over the same data.

Either we also use the mutex in `no_more_data` etc. or the `RwLock<InventoryManager>` that the `InventoryAgent` uses must not be dropped in-between reading and modifying the data.

Anyway. Since this is not very critical. I'd suggest we merge the already fixed timer issue and open a seperate issue for this.

PR #124, opened 2020-01-24
Bump bitvec to 0.17, remove bit_vec

riptl:
- Remove `nimiq_utils::bit_vec`
 - Remove all usages of crate `bit_vec`
 - Bump `bitvec` to 0.17


paberr:
TODO: Check whether there are unit tests.

riptl:
While looking for usages, I noticed we are using two different crates: `bitvec` and `bit-vec`.
The latter has 10x more usages on crates.io but I personally like the former a lot more, it's really well documented and also has more features.

Usages:
 * Undocumented util iterators over crate `bit-vec` https://github.com/nimiq/core-rs-albatross/blob/albatross/utils/src/bit_vec.rs (EDIT: Turns out they iterate over the underlying storage bytes of a bitvec)
 * Mnemonic: `bit-vec`
 * beserial: De/serialize impls for `bitvec::BitVec`
 * slot: reward and slot allocation via `bitvec::BitVec`

riptl:
@paberr I replaced all `bit_vec` usages with `bitvec`
 * eliminating a few copies in `nimiq_mnemonic` since `bitvec` supports slicing
 * removing the `utils::bit_vec` chunk iterators since `bitvec` supports iterating over the underlying `BitStore` fields via `chunks()`

Slicing comes with a downside: It allows for unaligned `BitSlice`s that are annoying to serialize in `beserial`. I added a serialize fast path that copies the `BitSlice` buffer to the writer in the case it's aligned, falling back to the chunk iterator and writing single bytes if not.
This is unfortunate in the case of slices that are only low-aligned, where it would be possible to copy everything in one go except the last few remaining bits. (`[0b01010101 ... 0b10]101010`).
I don't know how often this special case happens happens but since all bitvec on-chain Albatross data structures I know of are 2^n size and not sliced (validator slots), I think it's ok.

I checked all usages for tests:
 * Usages in Albatross slots are included in the Albatross blockchain tests. Some tests specific for tests would be nice in the future.
 * Mnemonics are tested extensively.
 * Lastly, I added `bitvec` tests in `beserial`. To simplify testing, I removed the `Cursor` type paramter (equivalent to `BitOrder` in `bitvec`), since in a library literally called big-endian-serial, we don't need LE support ;) It also seems like `bitvec` goes even further than just little endian: The `Lsb0` BitOrder "traverses an element from `LSbit` to `MSbit`".

beserial/src/bitvec.rs
```diff
@@ -10,18 +10,37 @@ fn bits_to_bytes(lhs: usize) -> usize {
...
-        assert_eq!(slice.len(), bits_to_bytes(self.len()));
-        size += slice.len();
+        let base_slice = self.as_slice();
+        let target_len = bits_to_bytes(self.len());
+        if base_slice.len() == target_len {
+            // Fast path: The underlying slice is aligned with the BitSlice edges
+            writer.write_all(base_slice)?;
+        } else {
+            // Slow path: Slice is unaligned, iterate over chunks of 8 bits
+            // TODO Copy instead via to_vec()?
```
riptl, line 28: Opinions on this? (serializing an unaligned BitSlice)

jeffesquivels:
Closing this as the first part of it was merged with #137 and the second part is now #138.

PR #127, opened 2020-02-04
Staking changes

paberr:
This PR reflects the changes to the staking system by clearly separating validators and stakers.

**TL;DR:**

- explicit validator registration (distinguishing between validators and stakers in the contract)
- reward goes to an address specified by the validator
- stakes are still in the contract, but no reward address can be specified for a stake
- allow zero value transactions for signalling transactions (such as validator creation)
- no on-chain reward distribution as an off-chain distribution is already auditable on the chain (in contrast to PoW) and it would just add more complexity

**Current issues**
In our current implementation, validators are only modelled very implicitly by the stakes for these. This comes with a set of problems, when it comes to signalling transactions such as the "unparking" (stake re-activation) of a validator after it was view-changed: Right now, such transactions have to be sent by the stakers, which means that they:
Require a hot staker key
Require the staker to be online and act for the validator
Which is a bit contradicting our delegation model.

Furthermore, since each staker can set its own reward address, any staker can just stake for a validator – without its consent – and collect the rewards. The only way for a validator to prevent this is to create view-changes whenever incorrect stakes are chosen for block production. Thus, this slows down the chain.

Finally, we were discussing the feasibility and necessity of an on-chain reward distribution that doesn't require to trust the validator.

**Solutions**
We are going to stick to the centralized registry instead of having the stakes stored in a distributed fashion.
Apart from that, we solve the issue evolving around the signalling transactions and the "incorrect" stakes by separating validators and stakes inside the contract. A validator can be created, updated, and destroyed and contains at least the following information:
BLS hot key
A reward address
The staker only has to specify which validator it delegates its stake to.

Now, a staker cannot use invalid reward addresses anymore as all the reward goes to the validator. Moreover, signalling transactions do not need to be sent by stakers anymore, but are sent by the validator. For this, we will need to allow zero value transactions to the contract (which isn't a huge deal). The fees for the transaction can now be paid from any address. We will require these transactions to contain a signature by the BLS hot key in the data field as well.

We decided against an on-chain reward distribution to stakers, because of the added complexity in comparison to a relatively low benefit. In contrast to PoW, the whole reward distribution can be audited on the chain, so it is a fair assumption to trust the validator with it.


terorie:
[Refactor & finalize staking](https://trello.com/c/B3vhK7lt/35-refactor-finalize-staking)

terorie:
[Refactor & finalize staking](https://trello.com/c/B3vhK7lt/35-refactor-finalize-staking)

blockchain-albatross/src/blockchain.rs
```diff
@@ -1430,9 +1430,9 @@ impl Blockchain {
 
         // Get stake slots
```
jeffesquivels: May be it would be a good idea to update these comments in this commit too?

primitives/account/src/staking_contract/actions/staker.rs
```diff
@@ -0,0 +1,220 @@
...
+
+        if (inactive_stake.balance > value) != receipt.is_some() {
+            return Err(AccountError::InvalidForRecipient);
+        }
+
+        let retire_time = receipt.map(|r| r.retire_time).unwrap_or_default();
+        self.reactivate_sender(staker_address, value, Some(retire_time)).map(|_| ())
+    }
+
+    /// Reverts a retire transaction.
```
jeffesquivels: Is this the correct comment?

primitives/account/src/staking_contract/actions/validator.rs
```diff
@@ -0,0 +1,272 @@
...
+
+        // We first remove all stakes the validator holds and will re-add stakes afterwards
+        // when calling `retire_recipient`.
+        self.balance = Account::balance_sub(self.balance, validator.balance)?;
+
+        let mut retirement_by_address = BTreeMap::new();
+        for (staker_address, &stake) in validator.active_stake_by_address.read().iter() {
+            let receipt = self.retire_recipient(staker_address, stake, Some(inactive_validator.retire_time))?;
+            retirement_by_address.insert(staker_address.clone(), RetirementReceipt {
+                stake: stake,
```
jeffesquivels: May be replace this with just `stake,`?

primitives/account/src/staking_contract/actions/validator.rs
```diff
@@ -0,0 +1,272 @@
...
+        });
+        Ok(())
+    }
+
+    /// Revert inactivating a validator entry.
+    pub(super) fn revert_retire_validator(&mut self, validator_key: BlsPublicKey) -> Result<(), AccountError> {
+        self.reactivate_validator(validator_key).map(|_| ())
+    }
+
+    /// Revert inactivating a validator entry.
```
jeffesquivels: Is this the right comment?

primitives/account/src/staking_contract/actions/validator.rs
```diff
@@ -0,0 +1,272 @@
...
+    pub(super) fn reactivate_validator(&mut self, validator_key: BlsPublicKey) -> Result<InactiveValidatorReceipt, AccountError> {
+        // Move validator from inactive map to active map/set.
+        let inactive_validator = self.inactive_validators_by_key.remove(&validator_key)
+            .ok_or(AccountError::InvalidForRecipient)?;
+
+        // All checks passed, not allowed to fail from here on!
+        self.active_validators_sorted.insert(Arc::clone(&inactive_validator.validator));
+        self.active_validators_by_key.insert(validator_key.clone(), inactive_validator.validator);
```
jeffesquivels: May be remove the `.clone()` since `validator_key` is not used anymore in this function?

primitives/account/src/staking_contract/validator.rs
```diff
@@ -0,0 +1,237 @@
...
+        Validator {
+            balance: initial_balance,
+            reward_address,
+            validator_key,
+            active_stake_by_address: Default::default(),
+        }
+    }
+
+    pub fn update_validator(&self, new_reward_address: Option<Address>, new_validator_key: Option<BlsPublicKey>) -> Self {
+        let active_stake_by_address = mem::replace(self.active_stake_by_address.write().deref_mut(), Default::default());
```
jeffesquivels: May be use `mem::take()` instead of `mem::replace()` here?

primitives/account/src/staking_contract/validator.rs
```diff
@@ -0,0 +1,237 @@
...
+        Validator {
+            balance: self.balance,
+            reward_address: new_reward_address.unwrap_or_else(|| self.reward_address.clone()),
+            validator_key: new_validator_key.unwrap_or_else(|| self.validator_key.clone()),
+            active_stake_by_address: RwLock::new(active_stake_by_address),
+        }
+    }
+
+    fn with_balance(&self, balance: Coin) -> Self {
+        let active_stake_by_address = mem::replace(self.active_stake_by_address.write().deref_mut(), Default::default());
```
jeffesquivels: May be use mem::take() instead of mem::replace() here?

primitives/account/src/staking_contract/validator.rs
```diff
@@ -0,0 +1,237 @@
...
+                    ValidatorEntry::Inactive(_, ref mut error) => {
+                        error.replace(err);
+                    },
+                }
+            },
+        }
+    }
+}
+
+impl ValidatorEntry {
```
jeffesquivels: Is there a reason to have two `impl ValidatorEntry` blocks one inmediately after the other here?

primitives/transaction/src/account/staking_contract/mod.rs
```diff
@@ -42,24 +47,18 @@ impl AccountTransactionVerification for StakingContractVerifier {
                         return Err(TransactionError::InvalidForRecipient);
                     }
                 },
-                IncomingStakingTransactionData::CreateValidator { .. } => {
-                    // Implicitly checks that value > 0.
+                IncomingStakingTransactionData::CreateValidator { .. } => {// Implicitly checks that value > 0.
```
jeffesquivels: Was this join of these two lines on purpose?

primitives/src/policy.rs
```diff
@@ -103,10 +102,11 @@ fn compute_block_reward(current_supply: u64, block_height: u32) -> u64 {
 pub fn block_reward_at(block_height: u32) -> Coin {
     assert!(block_height >= 1, "block_height must be >= 1");
     let current_supply = supply_after(block_height - 1);
-    compute_block_reward(current_supply, block_height).try_into().unwrap()
+    compute_block_reward(current_supply, block_height)
+        .try_into()
+        .unwrap()
```
jeffesquivels, line 107: :+1:

jeffesquivels (dismissed):
LGTM, just some minor comments.

Also, please remember to rebase against albatross.

primitives/account/src/staking_contract/validator.rs
paberr: I didn't even know that! :)

primitives/account/src/staking_contract/validator.rs
paberr: No :)

primitives/transaction/src/account/staking_contract/mod.rs
paberr: No

jeffesquivels (approved):

PR #128, opened 2020-02-07
Bls12-377

brunoffranca:
## Pull request checklist

* We ported the bls library to use the curve BLS12-377 instead of the curve BLS12-381. This is necessary for the ZK proof syncing (aka Nano Syncing).
* We no longer use the "librustzcash" for the elliptic curve arithmetic. Instead, we use the "algebra" crate at the "ZEXE" repo.
* We also cleaned up the organization of the library and added some documentation.

## Update (21/05/2020):
The scope of this PR expanded _significantly_ since its creation. It now covers:

* Porting the bls crate to use the curve MNT6-753 instead of the curve BLS12-381. This is necessary for the Nano nodes syncing.
* Changing from use of the "librustzcash" for the elliptic curve arithmetic to the "algebra" and "algebra-core" crates from the "ZEXE" library.
* Creating the entire nano-sync crate which generates ZK proofs for the Nano node sync using recursive SNARKs on the MNT4-753/MNT6-753 elliptic curve cycle.
* Generally improve documentation for both crates.


bls/Cargo.toml
```diff
@@ -11,16 +11,12 @@ edition = "2018"
...
 parking_lot = { version = "0.9", optional = true }
 rand = "0.7"
 rand_chacha = "0.2"
-
-ff = { git = "https://github.com/paberr/librustzcash" }
-group = { git = "https://github.com/paberr/librustzcash" }
-pairing = { git = "https://github.com/paberr/librustzcash" }
-
+algebra = { git = "https://github.com/nimiq/zexe"}
+num-traits = { version = "0.2", default-features = false }
```
paberr: The `num_traits` dependency should be located in the external dependencies section.
We usually order the dependencies *alphabetically* and first all external ones, then the internal ones.

`algebra` is kind of a special case here, that's why I would put it in between (just like it was the case for `ff`, ...)

bls/README.md
```diff
@@ -1,60 +1,60 @@
...
+let message1 = "Some message";
 let sig1 = keypair1.sign(&message1.as_bytes());
 inputs.push((keypair1.public, message1));
-asig.aggregate(&sig1);
+agg_sig.aggregate(&sig1);
 
-let keypair2 = Keypair::<Bls12>::generate(&mut rng);
-let message2 = "Some other unique message";
+let keypair2 = Keypair::generate(&mut rng);
+let message2 = "Another message";
```
paberr: This should not work, right? We don't support different messages anymore.

bls/README.md
```diff
@@ -1,60 +1,60 @@
...
 ## Security Warnings
 
 This library does not make any guarantees about constant-time operations, memory access patterns, or resistance to side-channel attacks.
 
+**Important Note:** When aggregating signatures of the same message, please do read Section 3.2 of [this paper](https://crypto.stanford.edu/~dabo/pubs/papers/aggreg.pdf).
+An adversary that is able to choose arbitrary keys can forge signatures by publishing public keys it does not know the secret key for. This is known as a "Rogue Key Attack".
+
+Thus, one countermeasure is to require the adversary to prove knowledge of the secret key. This can be done by, for example, publishing a signature of his/her own public key.
+
+The aggregation of signatures of distinct messages is not vulnerable to the above mentioned attack. So, another countermeasure is for the signer to implicitly append his/her public key to the message, hence creating distinct messages. However, this library does not implement the aggregation of signatures for different messages.
```
paberr: But we don't support it anymore.

bls/src/tests.rs
```diff
@@ -0,0 +1,51 @@
+use super::*;
```
paberr: As you moved the tests into a separate file, I suggest actually moving them into a tests directory (instead of `src`).
See our other crates for examples.

bls/src/types/aggregate_public_key.rs
```diff
@@ -0,0 +1,77 @@
+use super::*;
+
+// An aggregate public key. Mathematically, it is equivalent to a regular public key. However, we created a new type for it in order to help differentiate between the two use cases.
```
paberr: You should use `///` for these comments, making them actual doc-strings.
That means when running `cargo doc`, it will be part of the documentation for the struct/method.

bls/src/types/aggregate_public_key.rs
```diff
@@ -0,0 +1,77 @@
...
+#[derive(Clone, Copy)]
+pub struct AggregatePublicKey(pub(crate) PublicKey);
+
+impl AggregatePublicKey {
+    // Creates a new "empty" aggregate public key. It is simply the identity element of the elliptic curve, also known as the point at infinity.
```
paberr: Same at most places. Simply add another slash in front of all doc-comments.

bls/src/types/aggregate_signature.rs
```diff
@@ -0,0 +1,59 @@
+use super::*;
+
+// An aggregate signature. Mathematically, it is equivalent to a regular signature. However, we created a new type for it in order to help differentiate between the two use cases.
```
paberr:
```suggestion
/// An aggregate signature. Mathematically, it is equivalent to a regular signature. However, we created a new type for it in order to help differentiate between the two use cases.
```
For the whole file.

bls/src/types/compressed_public_key.rs
```diff
@@ -0,0 +1,75 @@
+use super::*;
+
+// The serialized compressed form of a public key. This form consists of the x-coordinate of the point (in the affine form), one bit indicating the sign of the y-coordinate, one bit indicating if it is the "point-at-infinity" and one bit indicating that this is the compressed form.
```
paberr:
```suggestion
/// The serialized compressed form of a public key. This form consists of the x-coordinate of the point (in the affine form), one bit indicating the sign of the y-coordinate, one bit indicating if it is the "point-at-infinity" and one bit indicating that this is the compressed form.
```
Whole file.

bls/src/types/compressed_signature.rs
```diff
@@ -0,0 +1,76 @@
+use super::*;
+
+// The serialized compressed form of a signature. This form consists of the x-coordinate of the point (in the affine form), one bit indicating the sign of the y-coordinate, one bit indicating if it is the "point-at-infinity" and one bit indicating that this is the compressed form.
```
paberr:
```suggestion
/// The serialized compressed form of a signature. This form consists of the x-coordinate of the point (in the affine form), one bit indicating the sign of the y-coordinate, one bit indicating if it is the "point-at-infinity" and one bit indicating that this is the compressed form.
```
Whole file.

bls/src/types/keypair.rs
```diff
@@ -0,0 +1,54 @@
+use super::*;
+
+// Simply a struct combining the secret key and the public key types.
```
paberr:
```suggestion
/// Simply a struct combining the secret key and the public key types.
```

bls/src/types/public_key.rs
```diff
@@ -0,0 +1,81 @@
+use super::*;
+
+#[derive(Clone, Copy)]
+pub struct PublicKey {
+    // The projective form is the longer one, with three coordinates. The affine form is the shorter one, with only two coordinates. Calculation is faster with the projective form.
```
paberr:
```suggestion
    /// The projective form is the longer one, with three coordinates. The affine form is the shorter one, with only two coordinates. Calculation is faster with the projective form.
```
Whole file.

bls/src/types/secret_key.rs
```diff
@@ -0,0 +1,52 @@
+use super::*;
+
+#[derive(Clone, Copy)]
+pub struct SecretKey {
+    // This is simply a number in the finite field Fr.
```
paberr:
```suggestion
    /// This is simply a number in the finite field Fr.
```
Whole file.

network-primitives/src/genesis/unit-albatross.toml
```diff
@@ -10,7 +10,7 @@ staker_address = "NQ57 UC15 80L7 LHCK DBTB 709R M91Q PRG5 DL00"
 reward_address = "NQ57 UC15 80L7 LHCK DBTB 709R M91Q PRG5 DL00"
 balance = 100000
 # secret_key = "49ea68eb6b8afdf4ca4d4c0a0b295c76ca85225293693bc30e755476492b707f"
```
paberr: Is the secret key still correct? If not update it.

paberr: I guess it should be this one now:
```suggestion
# secret_key = "d52d4d68b4e605d5a59c2c0e4f5d4e92ed57ddb65044d85dc6d7b77c819c3610"
```

tools/src/bls/main.rs
```diff
@@ -1,8 +1,8 @@
 extern crate nimiq_bls as bls;
 
 use beserial::Serialize;
-use bls::{SecureGenerate};
-use bls::bls12_381::{PublicKey, SecretKey};
+use bls::SecureGenerate;
```
paberr: I would actually merge the two lines now, since both use the `bls` crate.

paberr (changes requested):
Looks good to me. Only minor changes requested.

bls/README.md
brunoffranca: Well, it says that we don't support it. I still believe it's important to list all of the possible countermeasures.

brunoffranca:
Finished the rebase. It builds and all tests pass.

bls/Cargo.toml
```diff
@@ -1,33 +1,30 @@
 [package]
+name = "nimiq-bls"
+version = "0.1.0"
 authors = ["Jack Grigg <str4d@i2pmail.org>", "The Nimiq Core Development Team <info@nimiq.com>"]
 description = "BLS signatures"
-documentation = "https://github.com/nimiq/bls"
-homepage = "https://github.com/nimiq/bls"
 license = "MIT/Apache-2.0"
-name = "nimiq-bls"
 repository = "https://github.com/nimiq/bls"
```
paberr: The repository URL is wrong.

bls/Cargo.toml
```diff
@@ -1,33 +1,30 @@
 [package]
+name = "nimiq-bls"
+version = "0.1.0"
 authors = ["Jack Grigg <str4d@i2pmail.org>", "The Nimiq Core Development Team <info@nimiq.com>"]
 description = "BLS signatures"
-documentation = "https://github.com/nimiq/bls"
-homepage = "https://github.com/nimiq/bls"
```
paberr, line 5: To be consistent, we should keep the homepage and set it to:
`homepage = "https://nimiq.com"`

bls/Cargo.toml
```diff
@@ -1,33 +1,30 @@
...
-rand_chacha = "0.2"
 
-ff = { git = "https://github.com/paberr/librustzcash" }
-group = { git = "https://github.com/paberr/librustzcash" }
-pairing = { git = "https://github.com/paberr/librustzcash" }
+algebra = { git = "https://github.com/nimiq/zexe", branch = "nimiq", features=["mnt6_753"]}
+algebra-core = { git = "https://github.com/nimiq/zexe", branch = "nimiq"}
 
 beserial = { path = "../beserial", version = "0.1", optional = true }
+nano-sync = { path = "../nano-sync", version = "0.1" }
```
paberr: Should we name it `nimiq-nano-sync`?

bls/src/types/public_key.rs
```diff
@@ -0,0 +1,102 @@
...
+        let rhs = MNT6_753::pairing(hash_curve, self.public_key);
+        lhs == rhs
+    }
+
+    /// Transforms a public key into a serialized compressed form.
+    /// This form consists of the x-coordinate of the point (in the affine form),
+    /// one bit indicating the sign of the y-coordinate
+    /// and one bit indicating if it is the "point-at-infinity".
+    pub fn compress(&self) -> CompressedPublicKey {
+        let mut buffer = [0u8; 288];
```
paberr: It would be nicer to have all of these sizes as constants, but this works too.

bls/src/types/signature.rs
```diff
@@ -0,0 +1,65 @@
...
+        let bits = bytes_to_bits(hash.as_bytes());
+
+        // Get the generators for the Pedersen hash.
+        let generators = pedersen_generators(256);
+
+        // Calculate the Pedersen hash.
+        let point = pedersen_hash(bits, generators, sum_generator_g1_mnt6());
+
+        point
+    }
```
paberr: As we discussed, I agree that this should be fine intuitively.
However, we should double check again.

nano-sync/.gitignore
```diff
@@ -0,0 +1,11 @@
...
+/target
+**/*.rs.bk
+/.idea
+*.iml
+.DS_Store
+/client/db
+/client/peer_key.dat
+Cargo.lock
+/fixed-unsigned/target
+/devnet-environments/*/build/
```
paberr, line 11: This file should be removed (as it's not the root of git anymore).

nano-sync/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "nano-sync"
+version = "0.1.0"
+authors = ["The Nimiq Core Development Team <info@nimiq.com>"]
+description = "Recursive SNARKs for Nimiq's Nano Nodes"
+license = "MIT/Apache-2.0"
+repository = "https://github.com/nimiq/nano-sync"
```
paberr: Fix URL and add `homepage`.

nano-sync/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+algebra-core = { git = "https://github.com/nimiq/zexe.git", branch = "nimiq"}
+crypto-primitives = { git = "https://github.com/nimiq/zexe.git", branch = "nimiq", features = ["r1cs", "groth16"] }
+groth16 = { git = "https://github.com/nimiq/zexe.git", branch = "nimiq", features = ["parallel"] }
+r1cs-core = { git = "https://github.com/nimiq/zexe.git", branch = "nimiq" }
+r1cs-std = { git = "https://github.com/nimiq/zexe.git", branch = "nimiq", features = ["mnt4_753", "mnt6_753"] }
+
+[features]
+default = []
+cost-analysis = ["colored"]
+print-trace = ["groth16/print-trace"]
```
paberr: Missing newline at end of file.

nano-sync/src/compression.rs
```diff
@@ -0,0 +1,405 @@
+use std::io::{Error, ErrorKind};
```
paberr, line 1: This file is nearly the same as in the bls crate.
(I think this one here is the "newer" version of it.)

We should think about a way to not have this code duplication.

paberr: One way we could do it is:
Moving this version to the bls crate and importing it from there.

nano-sync/src/circuits/mod.rs
```diff
@@ -0,0 +1,5 @@
+//! This module contains the zk-SNARK circuits that form the nano sync program. Each circuit produces
```
paberr, line 1: I'll create a separate review for the circuits once I'm through the rest of the code.

nano-sync/src/constants.rs
```diff
@@ -0,0 +1,382 @@
...
+
+/// This is the depth of the PKTree circuit.
+pub const PK_TREE_DEPTH: usize = 5;
+
+/// This is the number of leaves in the PKTree circuit. It must be equal to 2**PK_TREE_DEPTH.
+pub const PK_TREE_BREADTH: usize = 32;
+
+/// This is the length of one epoch in Albatross. Basically, the difference in the block numbers of
+/// two consecutive macro blocks.
+pub const EPOCH_LENGTH: u32 = 300;
```
paberr, line 31: We should import these protocol constants from the `nimiq-primitives` crate using the `policy` feature instead of having them duplicate.

nano-sync/src/gadgets/mnt4/check_sig.rs
```diff
@@ -0,0 +1,91 @@
...
+
+impl CheckSigGadget {
+    /// This function checks the correctness of an aggregated signature for n different messages
+    /// from n different public keys. It implements the signature aggregation scheme from:
+    /// https://crypto.stanford.edu/%7Edabo/pubs/papers/aggreg.pdf
+    /// It implements the following verification formula:
+    /// e(sig, gen) = e(hash_1, pk_1) * e(hash_2, pk_2) * ... * e(hash_n, pk_n)
+    pub fn check_signatures<CS: r1cs_core::ConstraintSystem<MNT4Fr>>(
+        mut cs: CS,
+        public_keys: Vec<&G2Gadget>,
```
paberr: Accept `&[&G2Gadget]` instead of `Vec`.

nano-sync/src/gadgets/mnt4/check_sig.rs
```diff
@@ -0,0 +1,91 @@
...
+        hash_points: Vec<&G1Gadget>,
```
paberr: See above.

nano-sync/src/gadgets/mnt4/check_sig.rs
```diff
@@ -0,0 +1,91 @@
...
+                PairingGadget::prepare_g2(cs.ns(|| format!("pub_key_p {}", i)), &public_key)?;
+            pub_key_p_vars.push(pub_key_p_var);
+        }
+
+        // Prepare all the hash elliptic curve points.
+        let mut hash_p_vars = vec![];
+
+        for (i, hash_point) in hash_points.iter().enumerate() {
+            let hash_p_var: G1PreparedGadget =
+                PairingGadget::prepare_g1(cs.ns(|| format!("hash_p {}", i)), &hash_point)?;
```
paberr:
```suggestion
                PairingGadget::prepare_g1(cs.ns(|| format!("hash_p {}", i)), hash_point)?;
```
`hash_point` should already be a reference.

nano-sync/src/gadgets/mnt4/check_sig.rs
```diff
@@ -0,0 +1,91 @@
...
+        );
+
+        assert!(hash_points.len() > 1, "Min one message is required");
+
+        // Prepare all the public key elliptic curve points.
+        let mut pub_key_p_vars = vec![];
+
+        for (i, public_key) in public_keys.iter().enumerate() {
+            let pub_key_p_var: G2PreparedGadget =
+                PairingGadget::prepare_g2(cs.ns(|| format!("pub_key_p {}", i)), &public_key)?;
```
paberr:
```suggestion
                PairingGadget::prepare_g2(cs.ns(|| format!("pub_key_p {}", i)), public_key)?;
```
`public_key` should already be a reference.

nano-sync/src/gadgets/mod.rs
```diff
@@ -0,0 +1,7 @@
+//! This module contains all of the "gadgets" for the zk-SNARK circuits. These are smaller, modular pieces
```
paberr, line 1: I will check gadgets separately with the circuits.

nano-sync/src/primitives/macro_block.rs
```diff
@@ -0,0 +1,168 @@
...
+
+        bytes.extend_from_slice(&block_number.to_be_bytes());
+
+        bytes.extend_from_slice(&self.header_hash);
+
+        bytes.extend(&pks_commitment);
+
+        let bits = bytes_to_bits(&bytes);
+
+        //Calculate the Pedersen generators and the sum generator. The formula used for the ceiling
```
paberr:
```suggestion
        // Calculate the Pedersen generators and the sum generator. The formula used for the ceiling
```

nano-sync/src/primitives/merkle_tree.rs
```diff
@@ -0,0 +1,239 @@
+use std::cmp;
```
paberr, line 1: Zexe has an implementation for merle trees (that also works with Pedersen hashes).
Shouldn't we use that one? Is there anything that prevents us from using their implementation?

nano-sync/src/primitives/pedersen.rs
```diff
@@ -0,0 +1,185 @@
...
+            depth: 0,
+            leaf_length: 32,
+            node_offset: i as u32,
+            xof_digest_length: 65535,
+            node_depth: 0,
+            inner_length: 32,
+            salt: [1; 8],
+            // This needs to be set to an unique value, since we want a different random stream for
+            // each generator series that we create.
+            personalization: [0; 8],
```
paberr: A lot of zeros don't look much unique to me. :)

nano-sync/src/primitives/merkle_tree.rs
paberr: Ah, was it the issue that theirs starts with a 0 and that is problematic due to the incomplete addition?

paberr: If so, we could probably just implement a version of Pedersen that works and plug it in into their merkle tree without any issues.

nano-sync/src/primitives/pedersen.rs
```diff
@@ -0,0 +1,185 @@
...
+
+/// Calculates the Pedersen hash. Given a vector of generators G_i and a vector of bits b_i, the
+/// hash is calculated like so:
+/// H = G_0 + b_1 * G_1 + b_2 * G_2 + ... + b_n * G_n
+/// where G_0 is a sum generator that is used to avoid that the sum starts at zero (which is
+/// problematic because the circuit can't handle addition with zero).
+/// The Pedersen hash guarantees that the exponent of the resulting point H is not known, the same
+/// can't be said of the Pedersen commitment. Also, note that the Pedersen hash is collision-resistant
+/// but it is not pseudo-random.
+pub fn pedersen_hash(
```
paberr: See comments for merkle tree.

nano-sync/src/primitives/state_commitment.rs
```diff
@@ -0,0 +1,65 @@
...
+    // Serialize the block number and the Merkle tree root into bits.
+    let mut bytes: Vec<u8> = vec![];
+
+    bytes.extend_from_slice(&block_number.to_be_bytes());
+
+    bytes.extend(&root);
+
+    let bits = bytes_to_bits(&bytes);
+
+    //Calculate the Pedersen generators and the sum generator. The formula used for the ceiling
```
paberr:
```suggestion
    // Calculate the Pedersen generators and the sum generator. The formula used for the ceiling
```

nano-sync/src/primitives/vk_commitment.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    bytes.extend_from_slice(serialize_g2_mnt6(vk.delta_g2.into_projective()).as_ref());
+
+    for i in 0..vk.gamma_abc_g1.len() {
+        bytes.extend_from_slice(serialize_g1_mnt6(vk.gamma_abc_g1[i].into_projective()).as_ref());
+    }
+
+    let bits = bytes_to_bits(&bytes);
+
+    //Calculate the Pedersen generators and the sum generator. The formula used for the ceiling
```
paberr:
```suggestion
    // Calculate the Pedersen generators and the sum generator. The formula used for the ceiling
```

nano-sync/src/utils.rs
```diff
@@ -0,0 +1,203 @@
...
+/// Serializes a G2 point in the MNT6-753 curve.
+pub fn serialize_g2_mnt6(point: MNT6G2Projective) -> [u8; 285] {
+    let mut buffer = [0u8; 285];
+    BeSerialize::serialize(&point.into_affine(), &mut &mut buffer[..]).unwrap();
+    buffer
+}
+
+/// Creates a random G1 point in the MNT4-753 curve.
+pub fn gen_rand_g1_mnt4() -> MNT4G1Projective {
+    let rng = &mut thread_rng();
```
paberr, line 45: We should comment that these are not cryptographically "random" as thread_rng is not a CSPRNG.
As far as I can see, we're only using it for the parameter generation, right?

nano-sync/src/utils.rs
```diff
@@ -0,0 +1,203 @@
...
+pub fn gen_rand_g2_mnt6() -> MNT6G2Projective {
+    let rng = &mut thread_rng();
+    let mut bytes = [0u8; 96];
+    let mut x = None;
+    while x.is_none() {
+        rng.fill_bytes(&mut bytes[2..]);
+        x = MNT6Fr::from_random_bytes(&bytes);
+    }
+    MNT6G2Projective::prime_subgroup_generator().mul(x.unwrap())
+}
```
paberr, line 89: Comment for all of these that these are not generated using a CSPRNG.

primitives/src/coin.rs
```diff
@@ -28,7 +28,7 @@ impl Coin {
     }
 
     // NOTE: We implement a trait that does this, but we don't always want to have to import
-    // a whole create to check if a coin value is zero.
+    // a whole trait to check if a coin value is zero.
```
paberr:
```suggestion
    // a whole crate to check if a coin value is zero.
```

tools/src/bls/main.rs
```diff
@@ -16,8 +16,4 @@ fn main() {
     println!("# Secret Key:");
     println!();
     println!("{}", hex::encode(secret_key.serialize_to_vec()));
-    println!();
-    println!("# Proof Of Knowledge:");
-    println!();
-    println!("{}", hex::encode(&secret_key.sign(&public_key).compress()));
-}
```
paberr: Why did you remove the proof of knowledge?

paberr (commented):
This is the review for everything except gadgets and circuits.
The rest will follow.

bls/Cargo.toml
brunoffranca: Sure, why not?

bls/src/types/public_key.rs
brunoffranca: They already appear as constants in CompressedPublicKey, CompressedSignature and SecretKey. But I did miss this place, it should be CompressedPublicKey::SIZE.

bls/src/types/signature.rs
brunoffranca: This was already checked a bunch of times. And the Celo guys also use this arrangement. It is fine.
But I added a disclaimer to that function saying that the input hash must already be pseudo-random.

nano-sync/src/compression.rs
brunoffranca: Won't work. I already import the nano-sync crate in the bls to access the pedersen hash function. Can't have a circular dependency.
Also, I think the one in bls is just a subset of the one in nano-sync.

brunoffranca: I can try to import it from nano-sync into bls, though.

nano-sync/src/primitives/merkle_tree.rs
brunoffranca: To be honest, their version of the Merkle tree is a bit confusing. And ya, we would need to change our Pedersen hash to make it work with Zexe's types and then plug it into their Merkle tree.

brunoffranca: I just don't see the need. We already have our code, it works and is stable.

nano-sync/src/primitives/pedersen.rs
brunoffranca: Well, unique in the sense that we want it to be different from the other generators. They all use the same seed and the same Blake2X, so we need the personalization field to be different.
But you know what? I'll set it to all fives. XD

nano-sync/src/utils.rs
brunoffranca: Hmmm. But thread_rng is a CSPRNG.
[https://docs.rs/rand/0.6.1/rand/rngs/struct.ThreadRng.html](https://docs.rs/rand/0.6.1/rand/rngs/struct.ThreadRng.html)

nano-sync/src/compression.rs
brunoffranca: Done!

nano-sync/src/constants.rs
brunoffranca: Not gonna happen. Nimiq-primitives depend on nimiq-bls which depend on nimiq-nano-sync. I can't have nimiq-nano-sync depend on nimiq-primitives.
Any ideas?

tools/src/bls/main.rs
brunoffranca: That's a good question. I don't remember doing that.
I'll add it back.

nano-sync/src/utils.rs
paberr: Interesting. Perhaps it was not always the case, at least I remember it was not. But ok, then this resolves it.

tools/src/bls/main.rs
brunoffranca: And done!

nano-sync/src/constants.rs
paberr: Only idea is moving out the policy portion into a separate crate (which doesn't need any dependencies I guess).

nano-sync/src/gadgets/pedersen.rs
```diff
@@ -0,0 +1,140 @@
...
+        // Initiate the result with the sum generator. We can't initiate it with the neutral element
+        // because that would result in an error. The addition function for EC points is incomplete
+        // and can't handle the neutral element (aka zero, point-at-infinity).
+        // This weird initialization (instead of just doing let mut result = sum_generator) is to
+        // appease Rust's borrow checker. Best to leave it as is!
+        let mut result = AffineGadget::<P, ConstraintF, F>::new(
+            sum_generator.x.clone(),
+            sum_generator.y.clone(),
+            sum_generator.infinity,
+        );
```
paberr: What about this:
```suggestion
        let mut result = sum_generator.clone();
```

nano-sync/src/gadgets/pedersen.rs
```diff
@@ -0,0 +1,140 @@
...
+{
+    /// Calculates the Pedersen hash. Given a vector of generators G_i and a vector of bits b_i, the
+    /// hash is calculated like so:
+    /// H = G_0 + b_1 * G_1 + b_2 * G_2 + ... + b_n * G_n
+    /// where G_0 is a sum generator that is used to avoid that the sum starts at zero (which is
+    /// problematic because the circuit can't handle addition with zero).
+    pub fn evaluate<CS: r1cs_core::ConstraintSystem<ConstraintF>>(
+        mut cs: CS,
+        input: &Vec<Boolean>,
+        generators: &Vec<AffineGadget<P, ConstraintF, F>>,
```
paberr: Note to myself: Check again how the `generators` are chosen.
If I see it correctly, we're not using any sub-windows here that rely on the powers of a generator, i.e.,
`H = G_0 + b_1 * G_1 + b_2 * G_1^2 + ... + b_i * G_2 + b_{i+1} * G_2^2`, right?
Instead we need n distinct generators for n bits.

Any reasons for this?

nano-sync/src/gadgets/pedersen.rs
```diff
@@ -0,0 +1,140 @@
...
+        // Initiate the result with the sum generator. We can't initiate it with the neutral element
+        // because that would result in an error. The addition function for EC points is incomplete
+        // and can't handle the neutral element (aka zero, point-at-infinity).
+        // This weird initialization (instead of just doing let mut result = sum_generator) is to
+        // appease Rust's borrow checker. Best to leave it as is!
+        let mut result = AffineGadget::<P, ConstraintF, F>::new(
+            sum_generator.x.clone(),
+            sum_generator.y.clone(),
+            sum_generator.infinity,
+        );
```
paberr: Does this work?
```suggestion
        let mut result = sum_generator.clone();
```

nano-sync/src/gadgets/pedersen.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// where G_0 is a sum generator that is used to avoid that the sum starts at zero (which is
+    /// problematic because the circuit can't handle addition with zero).
+    pub fn evaluate<CS: r1cs_core::ConstraintSystem<ConstraintF>>(
+        mut cs: CS,
+        input: &Vec<Boolean>,
+        generators: &Vec<AffineGadget<P, ConstraintF, F>>,
+        sum_generator: &AffineGadget<P, ConstraintF, F>,
+    ) -> Result<AffineGadget<P, ConstraintF, F>, SynthesisError> {
+        // This is simply the number of bits that each generator can store.
+        let capacity = 752;
```
paberr: We could declare this as a constant.

nano-sync/src/gadgets/mnt4/state_commitment.rs
```diff
@@ -0,0 +1,75 @@
+use algebra::mnt4_753::Fr as MNT4Fr;
+use r1cs_core::SynthesisError;
+use r1cs_std::bits::{boolean::Boolean, uint32::UInt32, uint8::UInt8};
+use r1cs_std::mnt6_753::G1Gadget;
+
+use crate::gadgets::mnt4::{PedersenCommitmentGadget, SerializeGadget};
+use crate::utils::reverse_inner_byte_order;
+
+/// This gadget is meant to calculate the "state commitment" in-circuit, which is simply a commitment,
```
paberr, line 9: Note to myself: Check where the state commitment is used.
Is the block number sufficient or does it also need the view-number?

nano-sync/src/gadgets/mnt4/vk_commitment.rs
```diff
@@ -0,0 +1,88 @@
...
+            &vk.delta_g2,
+        )?);
+
+        // Gamma ABC G1
+        for i in 0..vk.gamma_abc_g1.len() {
+            bits.extend(SerializeGadget::serialize_g1(
+                cs.ns(|| format!("serialize gamma abc g1: point {}", i)),
+                &vk.gamma_abc_g1[i],
+            )?);
+        }
```
paberr, line 60: Do we also need to serialize the length?

paberr: Ok, we have to be careful here.
From [the Zcash protocol specification](https://github.com/zcash/zips/blob/master/protocol/protocol.pdf) 5.4.1.7:
> PedersenHash and PedersenHashToPoint are required to be collision-resistant between inputs of fixed length, for a given personalization input D.

In principle this means that there might be a possibility to find collisions by using verifying keys with different lengths of `gamma_abc_g1`. So, the question is what happens in the recursive setting if I were to use a VK with a `gamma` length larger than expected.

paberr: ~Perhaps we should give the function the `gamma` length as parameter and check the `gamma.len()` to be equal to this constant on circuit.~ It doesn't make sense to get `gamma.len()`, since this is also just a constant on the circuit. What could work is getting the expected length as a parameter and serializing only up to this length.

nano-sync/src/gadgets/mnt6/y_to_bit.rs
```diff
@@ -0,0 +1,242 @@
...
+use r1cs_std::mnt4_753::{FqGadget, G1Gadget, G2Gadget};
+use r1cs_std::prelude::{AllocGadget, FieldGadget};
+use r1cs_std::{Assignment, ToBitsGadget};
+
+/// A gadget that takes an elliptic curve point as input and outputs a single bit representing the
+/// "sign" of the y-coordinate. It is meant to aid with serialization.
+/// It was originally part of the Celo light client library. (https://github.com/celo-org/bls-zexe)
+pub struct YToBitGadget;
+
+impl YToBitGadget {
```
paberr, line 15: I think the celo guys actually did a good job in cleaning up the code a bit:
https://github.com/celo-org/celo-bls-snark-rs/blob/master/crates/bls-gadgets/src/y_to_bit.rs

Specifically, we could also make use of a `normalize` method to avoid code duplication.

nano-sync/src/gadgets/mnt4/y_to_bit.rs
```diff
@@ -0,0 +1,345 @@
...
+use r1cs_std::mnt6_753::{FqGadget, G1Gadget, G2Gadget};
+use r1cs_std::prelude::{AllocGadget, FieldGadget};
+use r1cs_std::{Assignment, ToBitsGadget};
+
+/// A gadget that takes an elliptic curve point as input and outputs a single bit representing the
+/// "sign" of the y-coordinate. It is meant to aid with serialization.
+/// It was originally part of the Celo light client library. (https://github.com/celo-org/bls-zexe)
+pub struct YToBitGadget;
+
+impl YToBitGadget {
```
paberr, line 15: See comments for other YToBitsGadget.

nano-sync/src/gadgets/mnt4/merkle_tree.rs
```diff
@@ -0,0 +1,203 @@
...
+        // Checking that the number of leaves is a power of two.
+        assert_eq!((inputs.len() & (inputs.len() - 1)), 0);
+
+        // Calculate the Pedersen commitments for the leaves.
+        let mut nodes = Vec::new();
+
+        for i in 0..inputs.len() {
+            let input = &inputs[i];
+
+            let pedersen_commitment = PedersenCommitmentGadget::evaluate(
```
paberr: Since we want the merkle tree to be collision resistant, all Pedersen commitments need fixed-length inputs.
We can probably ensure this by padding.

paberr (commented):
I reviewed the gadgets.
One major thing we need to make sure we get right is the CR property of the Pedersen hashes/commitments. AFAIK it only holds for fixed-length inputs.

nano-sync/src/constants.rs
brunoffranca: Would make sense. But then I prefer to create a separate issue for it. Since it is an improvement and it touches a lot of crates.

brunoffranca:
> I reviewed the gadgets.
> One major thing we need to make sure we get right is the CR property of the Pedersen hashes/commitments. AFAIK it only holds for fixed-length inputs.

I'm absolutely certain that they are collision-resistant. They are basically Pedersen commitments, so it just derives from the discrete log assumption. We're ok there. :+1:

nano-sync/src/gadgets/pedersen.rs
brunoffranca: I'll try it. I can't remember why I did it in that weird way.

nano-sync/src/gadgets/pedersen.rs
brunoffranca: Yes, it is only one bit per generator.
Well, we need Pedersen hashes because we require not only collision-resistance but also that the exponent of the output EC point is unknown. Otherwise we could just use a Pedersen commitment.
See that the number of constraints for calculating a Pedersen hash or a Pedersen commitment is very similar. Because when we do a scalar multiplication we are using the square-and-add method, which is very similar to the way we calculate the Pedersen hash.
The performance could be improved a bit by using a Bowe-Hopwood hash but I don't think the extra complexity is worth the trouble.

brunoffranca: That reminds that I think I used Pedersen commitments for the Merkle tree, but I'm not sure if it is safe or if we need to use Pedersen hashes...

nano-sync/src/gadgets/mnt4/vk_commitment.rs
brunoffranca: There is already a maximum length implicit here because we provide the Pedersen generators. If the input is bigger than the capacity of those generators, the function will panic.

nano-sync/src/gadgets/mnt6/y_to_bit.rs
brunoffranca: They did improve it, yes. I hadn't seen that. Documentation is better also.

brunoffranca: I'll see what I can take from there.

nano-sync/src/gadgets/mnt4/merkle_tree.rs
brunoffranca: Again, the maximum length is implicit when we provide the Pedersen generators. My main worry is if the Pedersen commitments is enough or we need Pedersen hashes.

nano-sync/src/constants.rs
brunoffranca: Created issue. Done!

nano-sync/src/gadgets/pedersen.rs
brunoffranca: It does work. In the comments I said it was needed to "appease" the Rust borrow checker, but it's compiling without a problem.
Interestingly, IntelliJ complains about this.

nano-sync/src/gadgets/pedersen.rs
brunoffranca: After discussing this on the ZEXE Telegram channel, there is in fact no need to have different functions for this. We can let the Pedersen hash have each generator use all of its capacity and we can use it in the Merkle Trees.
In addition, to have an unknown exponent we just need to add a point with an unknown exponent to the otherwise "normal" Pedersen hash. Thankfully, we already do this because of the incomplete addition.

nano-sync/src/gadgets/pedersen.rs
brunoffranca: Done!

nano-sync/src/gadgets/pedersen.rs
brunoffranca: Done!

nano-sync/src/gadgets/mnt4/vk_commitment.rs
brunoffranca: What I can do here is check that the number of generators provided is exactly the number of generators needed. Right now I only check that number of generators provided is equal or more than the number of generators needed ([https://github.com/nimiq/core-rs-albatross/blob/bls12-377/nano-sync/src/gadgets/pedersen.rs#L105](https://github.com/nimiq/core-rs-albatross/blob/bls12-377/nano-sync/src/gadgets/pedersen.rs#L105)), so there might be some place where I actually provide more generators than needed and that might be exploitable.
Changing that assertion should be sufficient since then the circuit creation does the rest of the job.

nano-sync/src/gadgets/mnt4/merkle_tree.rs
brunoffranca: See the comments for the vk_commitment regarding this.

nano-sync/src/gadgets/mnt4/vk_commitment.rs
paberr: That should work.

brunoffranca: We are thinking about this whole thing wrong. The circuit is already fixed-size by definition. All the inputs, private and public, are set to a fixed length when we generate the parameters. There's no need to alter any of this.

nano-sync/src/gadgets/mnt6/y_to_bit.rs
brunoffranca: Done. I refactored the entire gadget. I have a curve-agnostic gadget (like we have with Pedersen hash) that has two methods: is_greater_half and is_equal_zero. I also removed some unnecessary code. Just like celo did.
Then I created the gadgets specific to each curve, and these gadgets have the y_to_bit_g1 and  y_to_bit_g2 methods that mostly just call the is_greater_half and is_equal_zero methods, so they are a lot simpler now.

nano-sync/src/circuits/mnt4/merger.rs
```diff
@@ -0,0 +1,235 @@
...
+// Renaming some types for convenience. We can change the circuit and elliptic curve of the input
+// proof to the merger circuit just by editing these types.
+type FirstProofSystem = Groth16<MNT6_753, MergerWrapperCircuit, Fr>;
+type SecondProofSystem = Groth16<MNT6_753, MacroBlockWrapperCircuit, Fr>;
+type TheProofGadget = ProofGadget<MNT6_753, Fq, PairingGadget>;
+type TheVkGadget = VerifyingKeyGadget<MNT6_753, Fq, PairingGadget>;
+type TheVerifierGadget = Groth16VerifierGadget<MNT6_753, Fq, PairingGadget>;
+
+/// This is the merger circuit. It takes as inputs an initial state commitment, a final state commitment
+/// and a verifying key and it produces a proof that there exists two valid SNARK proofs that transforms
```
paberr:
```suggestion
/// and a verifying key and it produces a proof that there exist two valid SNARK proofs that transform
```

nano-sync/src/circuits/mnt4/pk_tree_1.rs
```diff
@@ -0,0 +1,436 @@
+use std::fs::File;
```
paberr: It seems like tree_1 - tree_4 are all the same except the filename and the type of the `TheProofSystem` and the curves.
Would't it be much better to have one circuit for all of these instead of duplicate code?

I.e., `PKTreeIntermediateCircuit` could get some type parameters `<CurveA, CurveB, SubCircuit>` or so and receive the filename for the vk as a parameter in `new()`.

nano-sync/src/circuits/mnt4/pk_tree_5.rs
```diff
@@ -0,0 +1,314 @@
...
+
+        let commit_agg_pk_commitment_var = UInt8::alloc_input_vec(
+            cs.ns(|| "alloc commit aggregate pk commitment"),
+            self.commit_agg_pk_commitment.as_ref(),
+        )?;
+
+        let position_var =
+            UInt8::alloc_input_vec(cs.ns(|| "alloc position"), self.position.as_ref())?
+                .pop()
+                .unwrap();
```
paberr: I don't get why we allocate a whole vector for each of the tree_i circuits when we discard all but one element.

nano-sync/src/circuits/mnt4/pk_tree_5.rs
```diff
@@ -0,0 +1,314 @@
...
+
+        // Process public inputs.
+        next_cost_analysis!(cs, cost, || { "Process public inputs" });
+
+        let chunk_start = VALIDATOR_SLOTS / PK_TREE_BREADTH * (self.position[0] as usize);
```
paberr: Guessing from `position_var` and the `self.position[0]` here, `self.position` has exactly one element, right?
Why does it have to be a `Vec` then?

nano-sync/src/circuits/mnt4/pk_tree_5.rs
```diff
@@ -0,0 +1,314 @@
...
+        let prepare_signer_bitmap_var = prepare_signer_bitmap_var[chunk_start..chunk_end].to_vec();
+
+        let commit_signer_bitmap_var =
+            commit_signer_bitmap_var.to_bits(cs.ns(|| "commit signer bitmap to bits"))?;
+
+        let commit_signer_bitmap_var = commit_signer_bitmap_var[chunk_start..chunk_end].to_vec();
+
+        let mut path_var = position_var.into_bits_le();
+
+        path_var.truncate(PK_TREE_DEPTH);
```
paberr: It's non-obvious what's happening here. I would add a comment.

nano-sync/src/circuits/mnt4/macro_block.rs
```diff
@@ -0,0 +1,425 @@
...
+            }
+
+            commit_agg_pk_chunks_commitments.push(commitment);
+        }
+
+        // Preparing the inputs for the SNARK proof verification. All the inputs need to be Vec<UInt8>,
+        // so they need to be converted to such.
+        next_cost_analysis!(cs, cost, || { "Prepare inputs for SNARK" });
+
+        let position = vec![UInt8::constant(0)];
```
paberr, line 347: If the position for the root is always [0], why do we need to pass it?
We don't need a `Vec<Uint8>` at least.

nano-sync/src/circuits/mnt6/pk_tree_0.rs
```diff
@@ -0,0 +1,238 @@
...
+        //    L = 2 * P
+        //    R = 2 * P + 1
+        // For efficiency reasons, we calculate the positions using bit manipulation.
+        next_cost_analysis!(cs, cost, || { "Calculate positions" });
+
+        let mut bits = position_var.into_bits_le();
+
+        bits.pop();
+        bits.insert(0, Boolean::Constant(false));
+        let left_position = vec![UInt8::from_bits_le(&bits)];
```
paberr: I would add a comment to make clear what you're doing here (additionally to the high level description above).
```suggestion
        // Calculate `P << 1`.
        bits.pop();
        bits.insert(0, Boolean::Constant(false));
        let left_position = vec![UInt8::from_bits_le(&bits)];
```

nano-sync/src/circuits/mnt6/pk_tree_0.rs
```diff
@@ -0,0 +1,238 @@
...
+
+        bits.remove(0);
+        bits.insert(0, Boolean::Constant(true));
+        let right_position = vec![UInt8::from_bits_le(&bits)];
```
paberr:
```suggestion
        // `bits` is currently `P << 1 = L`, so calculate `L | 1`, which is equivalent to `L + 1`.
        bits.remove(0);
        bits.insert(0, Boolean::Constant(true));
        let right_position = vec![UInt8::from_bits_le(&bits)];
```

nano-sync/src/circuits/mnt4/pk_tree_5.rs
```diff
@@ -0,0 +1,314 @@
...
+
+        MerkleTreeGadget::verify(
+            cs.ns(|| "verify Merkle proof for pks"),
+            &bits,
+            &pks_nodes_var,
+            &path_var,
+            &pk_commitment_var,
+            &pedersen_generators_var,
+            &sum_generator_g1_var,
+        )?;
```
paberr: I would add a comment here to remind people what the merkle tree is composed of.
As far as I understood it contains the concatenated public keys for each chunk as its leaves, right?

nano-sync/src/circuits/mnt4/pk_tree_5.rs
```diff
@@ -0,0 +1,314 @@
...
+            &pedersen_commitment,
+        )?;
+
+        let pedersen_bits = reverse_inner_byte_order(&pedersen_bits[..]);
+
+        let mut reference_commitment = Vec::new();
+
+        for i in 0..pedersen_bits.len() / 8 {
+            reference_commitment.push(UInt8::from_bits_le(&pedersen_bits[i * 8..(i + 1) * 8]));
+        }
```
paberr: Add comment what you're doing here.
If I'm not mistaken you convert the `pedersen_bits` into a `Vec<UInt8>`.

nano-sync/src/circuits/mnt4/pk_tree_5.rs
```diff
@@ -0,0 +1,314 @@
...
```
paberr: See above.

nano-sync/src/circuits/mnt4/macro_block.rs
```diff
@@ -0,0 +1,425 @@
+use std::fs::File;
```
paberr, line 1: This file is still on my TODO list for tomorrow.

paberr (commented):
Review of most circuits except the macro block circuit.

nano-sync/src/circuits/mnt4/merger.rs
brunoffranca: Done.

nano-sync/src/circuits/mnt4/pk_tree_5.rs
brunoffranca: We don't discard all but one elements, the vector only has one element. Since the input gadget only works with Vec<UInt8> I'm forced to use a 1-element vector here.

brunoffranca: Actually I was having a bug because of it and it was you who found the solution. Maybe I should write some comments explaining it.

nano-sync/src/circuits/mnt4/pk_tree_5.rs
brunoffranca: Yes. See answer above.

nano-sync/src/circuits/mnt4/pk_tree_5.rs
brunoffranca: Agree, I'll add some explanation.

nano-sync/src/circuits/mnt4/macro_block.rs
brunoffranca: We pass it because I didn't want to hardcode it into pk_tree_0. Anyway, if I'm going to merge pk_tree 0/2/4 to avoid code duplication, then I need it to have the position as an input.

nano-sync/src/circuits/mnt4/pk_tree_5.rs
brunoffranca: You're correct. I'll add a comment.

nano-sync/src/circuits/mnt4/pk_tree_1.rs
brunoffranca: Not quite. pk_tree 1 and 3 are indeed equal. And pk_tree 0, 2 and 4 are also equal. I think I had them all different at the time because I thought I was going to hardcode the vk in them, not feeding them through a file.
But I'm not sure how to have the filename for the vk as a parameter. I can't have it as a parameter to `new()`, that function only returns a PKTreeCircuit struct and all the fields of that struct must be public/private inputs to the circuit.
One option might be adding it as a parameter to the `generate_constraints` function?

paberr: Ah, I compared a few files in my editor and probably compared 1 with 3 and 2 with 4. 👍 
> I can't have it as a parameter to `new()`, that function only returns a PKTreeCircuit struct and all the fields of that struct must be public/private inputs to the circuit.

I don't get why you cannot have it as a parameter to `new()`.

nano-sync/src/circuits/mnt4/pk_tree_5.rs
paberr: I would remove all these `Vec` occurrences and only pass a single `Uint8`.
The input gadget problem can be easily solved by calling:
```rust
        proof_inputs.append(&mut RecursiveInputGadget::to_field_elements::<Fr>(
            &[right_position],
        )?);
```
Note that I replaced `&right_position` by `&[right_position]`.

This results in a much cleaner signature for the functions then. :)

paberr (commented):
The macro block circuit looks good to me. No further comments there.

nano-sync/src/circuits/mnt4/pk_tree_1.rs
brunoffranca: Because whatever parameter I add to 'new()' ZEXE will assume that it is an input to the ZK circuit.

brunoffranca: Done!

nano-sync/src/circuits/mnt4/pk_tree_5.rs
brunoffranca: Done.

nano-sync/src/circuits/mnt6/pk_tree_0.rs
brunoffranca: Done.

nano-sync/src/circuits/mnt6/pk_tree_0.rs
brunoffranca: Done.

nano-sync/src/circuits/mnt4/pk_tree_5.rs
brunoffranca: Seems a bit unnecessary, but I added it.

nano-sync/src/circuits/mnt4/pk_tree_5.rs
brunoffranca: Done.

nano-sync/src/circuits/mnt4/pk_tree_5.rs
brunoffranca: Done!

nano-sync/src/circuits/mnt4/pk_tree_5.rs
brunoffranca: Done!

nano-sync/src/circuits/mnt4/pk_tree_5.rs
brunoffranca: Done!

nano-sync/src/circuits/mnt4/pk_tree_5.rs
brunoffranca: Update: This started to give problems. Whenever it verified a SNARK proof inside a circuit it would complain that the length of the verifying key is different from the length of the proof.
I think this was the original problem we had and that just ended with us inputting everything as a Vec UInt8.
So I just reverted back to that. It works fine if we pass it as a Vec UInt8 and I really don't feel like doing a lot of debugging on this one. :p

paberr: That's because of `Uint8::alloc_input` vs. `Uint8::alloc_input_vec`. The first one allocates one field element per bit, the second allocates one field element for the whole Uint8.
I just made the changes myself to speed things up and will merge the code soon.

paberr (approved):

PR #129, opened 2020-02-08
Port of some important fixes from devnet branch

paberr:
These fixes mainly concern view-changes and rebranching behaviour.

### DIFF
--- a/blockchain-albatross/src/blockchain.rs
+++ b/blockchain-albatross/src/blockchain.rs
@@ -93,7 +93,7 @@ pub struct Blockchain {
     pub notifier: RwLock<Notifier<'static, BlockchainEvent>>,
     pub chain_store: Arc<ChainStore>,
     pub(crate) state: RwLock<BlockchainState>,
-    pub push_lock: Mutex<()>, // TODO: Not very nice to have this public
+    push_lock: Mutex<()>,
 
     #[cfg(feature = "metrics")]
     metrics: BlockchainMetrics,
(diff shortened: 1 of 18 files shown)


blockchain-albatross/src/blockchain.rs
```diff
@@ -423,6 +434,82 @@ impl Blockchain {
...
+    fn order_chains(&self, block: &Block, prev_info: &ChainInfo, txn_option: Option<&Transaction>) -> ChainOrdering {
+        let mut chain_order = ChainOrdering::Unknown;
+        if block.parent_hash() == &self.head_hash() {
+            chain_order = ChainOrdering::Extend;
+        } else {
+            // To compare two blocks, we need to compare the view number at the intersection.
+            //   [2] - [2] - [3] - [4]
+            //      \- [3] - [3] - [3]
+            // The example above illustrates that you actually want to choose the lower chain,
+            // since it's view change happened way earlier.
```
riptl:
```suggestion
            // since its view change happened way earlier.
```
:)

terorie:
```suggestion
            // since its view change happened way earlier.
```

riptl (dismissed):
I wish every PR was as clean as this.
Maybe small TODO for the future is an explanation what the BitSets in BlockDescriptor express.
It might not immediately be clear to the reader that they mark slot availability (correct term?) and also whether set membership marks a slashed slot or an active one.

blockchain-albatross/src/blockchain.rs
paberr: Done

paberr:
> I wish every PR was as clean as this.

Thanks :)
> Maybe small TODO for the future is an explanation what the BitSets in BlockDescriptor express.
> It might not immediately be clear to the reader that they mark slot availability (correct term?) and also whether set membership marks a slashed slot or an active one.

True

riptl (approved):

jeffesquivels (approved):
LGTM

PR #131, opened 2020-02-11
Add Travis CI & deps.rs badges

riptl:
Adds a deps.rs badge so we can keep an overview about how outdated the deps are.
Let's maybe add a Travis CI badge too?


paberr:
Sure, go ahead. Travis CI badge makes sense in my opinion.

paberr (dismissed):

paberr (approved):

PR #135, opened 2020-02-17
nimiq-bls: Print proof of knowledge

riptl:
Only usage of `nimiq-bls` I know of is in scripts/devnet_create.py but it's not affected by the changes.


riptl:
Seems like tests fail because of improved clippy lints

brunoffranca (approved):
Seems good to me. Not sure why my review was requested...

PR #137, opened 2020-03-31
Bump Bitvec to 0.17

jeffesquivels:
It seems that older-than-current versions of Bitvec had a bug and they were yanked from crates.io (see https://github.com/myrrlyn/bitvec/issues/59).


paberr (dismissed):

riptl:
related https://github.com/nimiq/core-rs-albatross/pull/124

paberr (approved):

PR #142, opened 2020-04-17

vlvrd:
This PR enables validators to recognize when they are parked and, if they are, allows them to send an unpark signalling transaction.

### DIFF
--- a/lib/src/client.rs
+++ b/lib/src/client.rs
@@ -82,7 +82,12 @@ impl TryFrom<ClientConfig> for ClientInner {
         // Initialize peer key
         config.storage.init_key_store(&mut network_config)?;
 
-        // Load validator key (before we give away ownership of the storage config
+        // Load transaction key (before we give away ownership of the storage config)
+        #[cfg(feature="validator")]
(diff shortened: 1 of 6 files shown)


lib/src/client.rs
```diff
@@ -82,7 +82,11 @@ impl TryFrom<ClientConfig> for ClientInner {
         // Initialize peer key
         config.storage.init_key_store(&mut network_config)?;
 
-        // Load validator key (before we give away ownership of the storage config
+        // Load network key (before we move the network config)
+        #[cfg(feature="validator")]
+        let network_key = network_config.key_pair().clone();
```
paberr: The network key is usually only used to sign network messages, so that other nodes can check the authenticity.
It would be better to add a new config option, perhaps called `validator_wallet_key`.

The new config option should be placed at the same place as the `validator_key`, so you can check out the code related to that.

paberr (changes requested):
Looks great!
There's only one point remaining, which should be addressed. :)

jeffesquivels:
Superseded by #146

PR #145, opened 2020-05-13
Backports from devnet-rc3 branch

jeffesquivels:


paberr (approved):
LGTM
A lot of the network code will change, but back porting this for now is still good.

PR #146, opened 2020-05-14

vlvrd:
Reopened the PR to use my local branch. These changes enable validators to recognize when they are parked and, if they are, allows them to send an unpark signalling transaction.

### DIFF
--- a/lib/src/config/config.rs
+++ b/lib/src/config/config.rs
@@ -2,6 +2,7 @@ use std::convert::TryFrom;
 use std::fmt::Display;
 use std::path::{Path, PathBuf};
 use std::net::IpAddr;
+use std::str::FromStr;
 
 use derive_builder::Builder;
 use enum_display_derive::Display;
@@ -120,6 +121,15 @@ pub enum ProtocolConfig {
 #[derive(Debug, Clone)]
 pub struct ValidatorConfig {
     // TODO
+    /// Validator wallet key
+    validator_wallet_key: Option<keys::KeyPair>,
+}
+
+impl ValidatorConfig {
+    #[cfg(feature="validator")]
+    pub fn validator_wallet_key(self) -> Result<Option<keys::KeyPair>, Error> {
+        Ok(self.validator_wallet_key)
+    }
 }
 
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
@@ -693,7 +703,9 @@ impl ClientConfigBuilder {
     /// yet, this will just enable the validator.
     #[cfg(feature="validator")]
     pub fn validator(&mut self) -> &mut Self {
-        self.validator = Some(Some(ValidatorConfig {}));
+        self.validator = Some(Some(ValidatorConfig {
+            validator_wallet_key: None
+        }));
         self
     }
 
@@ -847,7 +859,20 @@ impl ClientConfigBuilder {
         // Configure validator
         #[cfg(feature="validator")] {
             if config_file.validator.is_some() {
-                self.validator = Some(Some(ValidatorConfig {}));
+                let wallet_private_key = config_file.validator.as_ref()
+                    .map(|settings| settings.clone().wallet_private_key)
+                    .expect("Failed to load validator settings");
+
+                let wallet_key = wallet_private_key.map(|config_private_key| {
+                    Some(keys::KeyPair::from(
+                        keys::PrivateKey::from_str(&config_private_key)
+                            .map_err(|e| Error::config_error(format!("Invalid wallet private key: {:?}: {}", &config_private_key, e))).ok()?
+                    ))
+                }).unwrap_or_else(|| None);
+
+                self.validator = Some(Some(ValidatorConfig {
+                    validator_wallet_key: wallet_key
+                }));
             }
         }
 
--- a/lib/src/config/config_file/mod.rs
+++ b/lib/src/config/config_file/mod.rs
@@ -534,4 +534,5 @@ impl From<MempoolFilterSettings> for MempoolRules {
 #[serde(deny_unknown_fields)]
 pub struct ValidatorSettings {
     pub key_file: Option<String>,
+    pub wallet_private_key: Option<String>,
 }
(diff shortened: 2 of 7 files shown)


lib/src/config/config_file/mod.rs
```diff
@@ -530,6 +532,12 @@ impl From<MempoolFilterSettings> for MempoolRules {
     }
 }
 
+#[derive(Clone, Debug, Deserialize, Default)]
+#[serde(deny_unknown_fields)]
+pub struct ValidatorWalletSettings {
+    pub key_file: Option<String>,
```
paberr: You can move this to the `ValidatorSettings` and call it `wallet_key_file`.
That way we don't need a separate section in the config.

paberr (commented):
Generally looks good to me.
The only change I would still suggest is moving the wallet key file option into the `ValidatorSettings`.

Also, I thought a bit about whether we should make the wallet key completely optional.
I.e., make `fn validator_wallet_key(&self)` return `Result<Option<keys::KeyPair>, Error>` and `Validator` accept an option as well.
In case it is `None`, the automatic unparking is disabled.
But it's just an idea. What do you think?

vlvrd:
@paberr moved the wallet key to ValidatorSettings 👍 Regarding the `validator_wallet_key` rework: I'm wondering in which cases would that function should return `None`. Should I consider that `validator_wallet_key.dat` won't always exist?

paberr:
> @paberr moved the wallet key to ValidatorSettings 👍 

Thanks, looks good now.
> Regarding the `validator_wallet_key` rework: I'm wondering in which cases would that function should return `None`. Should I consider that `validator_wallet_key.dat` won't always exist?

Good question. Probably we would need a new boolean config option to enable/disable it. But I think, for now, the current version is ok. :)

paberr (dismissed):

lib/src/config/config.rs
```diff
@@ -847,7 +859,24 @@ impl ClientConfigBuilder {
...
+                let wallet_key: Option<keys::KeyPair>;
+                if wallet_private_key.is_some() {
+                    let config_private_key = wallet_private_key.unwrap();
+                    wallet_key = Some(keys::KeyPair::from(
+                        keys::PrivateKey::from_str(&config_private_key)
+                            .map_err(|e| Error::config_error(format!("Invalid wallet private key: {:?}: {}", &config_private_key, e)))?
+                    ));
+                } else {
+                    wallet_key = None;
+                }
```
paberr: Here's the more Rusty way to do this.
```suggestion
                let wallet_key = wallet_private_key.map(|config_private_key| {
                    let config_private_key = wallet_private_key.unwrap();
                    Some(keys::KeyPair::from(
                        keys::PrivateKey::from_str(&config_private_key)
                            .map_err(|e| Error::config_error(format!("Invalid wallet private key: {:?}: {}", &config_private_key, e)))?
                    ))
                });
```

validator/src/validator.rs
```diff
@@ -257,6 +263,37 @@ impl Validator {
...
+            .with_validity_start_height(validity_start_height)
+            .with_recipient(recipient.generate().unwrap());
+
+        let mut proof_builder = tx_builder.generate().unwrap().unwrap_signalling();
+        proof_builder.sign_with_validator_key_pair(&self.validator_key);
+        let mut proof_builder = proof_builder.generate().unwrap().unwrap_basic();
+        proof_builder.sign_with_key_pair(&self.validator_wallet_key.clone().unwrap());
+        let transaction = proof_builder.generate().unwrap();
+
+        self.consensus.mempool.push_transaction(transaction.clone());
```
paberr: Minor suggestion here: If we add something to this function in the future, we have to be careful to notice the early return. Thus, I'd replace the return by putting the whole thing into an if. Also, then we can use the more Rusty way – a `if let` statement.
```suggestion
        if let Some(ref validator_wallet_key) = self.validator_wallet_key {
            if self.is_parked(validator_registry) {
                // If we're a parked validator with a wallet key, we signal to unpark automatically.
                let validator_registry = NetworkInfo::from_network_id(self.blockchain.network_id).validator_registry_address().expect("Albatross consensus always has the address set.");

                let mut recipient = Recipient::new_staking_builder(validator_registry.clone());
                recipient.unpark_validator(&self.validator_key.public);

                // Send the unpark transaction with a fixed height each epoch in case there's already a pushed unpark transaction.
                // Note: If the validity window is ever less than an epoch's length, using the last macro block's height would require
                // handling the case where the last macro block is more blocks away than the validity window inverval, since otherwise
                // invalid transactions would be created.
                let validity_start_height = policy::macro_block_before(self.consensus.mempool.current_height());

                let mut tx_builder = TransactionBuilder::new();
                tx_builder.with_sender(validator_registry.clone())
                    .with_value(Coin::ZERO)
                    .with_network_id(self.blockchain.network_id)
                    .with_validity_start_height(validity_start_height)
                    .with_recipient(recipient.generate().unwrap());

                let mut proof_builder = tx_builder.generate().unwrap().unwrap_signalling();
                proof_builder.sign_with_validator_key_pair(&self.validator_key);
                let mut proof_builder = proof_builder.generate().unwrap().unwrap_basic();
                proof_builder.sign_with_key_pair(validator_wallet_key);
                let transaction = proof_builder.generate().unwrap();

                self.consensus.mempool.push_transaction(transaction.clone());
            }
        }
```

paberr (commented):
I added two minor comments to make the code look a bit more idiomatic to Rust. :)
Otherwise it's fine.

lib/src/config/config.rs
```diff
@@ -847,7 +859,20 @@ impl ClientConfigBuilder {
...
+                let wallet_private_key = config_file.validator.as_ref()
+                    .map(|settings| settings.clone().wallet_private_key)
+                    .expect("Failed to load validator settings");
+
+                let wallet_key = wallet_private_key.map(|config_private_key| {
+                    Some(keys::KeyPair::from(
+                        keys::PrivateKey::from_str(&config_private_key)
+                            .map_err(|e| Error::config_error(format!("Invalid wallet private key: {:?}: {}", &config_private_key, e))).ok()?
+                    ))
+                }).unwrap_or_else(|| None);
```
paberr, line 871: Just FYI, you don't need this. `wallet_private_key` is already an `Option` and `unwrap_or_else` is called when this is `None`, replacing it with a `None`. So, basically a no-op. :)

paberr (approved):

PR #149, opened 2020-05-19
Implement fork proofs

vlvrd:


blockchain-albatross/src/blockchain.rs
```diff
@@ -592,6 +599,72 @@ impl Blockchain {
...
             }
 
+            // Check if there are two blocks in the same slot. Since we already
+            // verified the validator for the current slot, this is enough to
+            // check for fork proofs. A fork proof doesn't apply to the genesis
+            // block, so skip block 0.
+            if block.block_number() != 0 {
+                // Count the microblocks after the last macroblock
+                let count = block.block_number() - policy::macro_block_before(block.block_number());
+                let microblocks: Vec<Block> = self.chain_store.get_blocks_backward(block.parent_hash(), count - 1, false, Some(&read_txn));
```
paberr: Fork blocks may only occur at the same height and view-number.
More specifically, a fork proof consists of two blocks:
- that are different (i.e., the content and thus the hash is different)
- at the same height
- with the same view number
- from the same slot (i.e., signed by the same validator)

So, we only need to get all micro blocks at the height `block.block_number()`, check if they have the same view number and are signed by the same person.

blockchain-albatross/src/blockchain.rs
```diff
@@ -592,6 +599,72 @@ impl Blockchain {
...
+
+                // Get the microheader from the block
+                let header = block.header();
+                let microheader1;
+
+                if let BlockHeader::Micro(microheader) = header {
```
paberr: Don't we have a `unwrap_micro` method?

blockchain-albatross/src/blockchain.rs
```diff
@@ -592,6 +599,72 @@ impl Blockchain {
...
+                // Get the justification for the block. We asume that the
+                // validator's signature is valid.
+                let justification1;
+                if let Block::Micro(ref micro_block) = block {
+                    justification1 = micro_block.justification.signature;
+                } else {
+                    unreachable!();
+                };
+
+                for microblock in microblocks.iter() {
```
paberr: If we fetch all micro blocks at the given height, we still need the iteration here. :+1:

blockchain-albatross/src/blockchain.rs
```diff
@@ -592,6 +599,72 @@ impl Blockchain {
...
+                    // Cannot fork macroblocks
+                    if let Block::Macro(_) = microblock {
```
paberr: We don't need this.

blockchain-albatross/src/blockchain.rs
```diff
@@ -592,6 +599,72 @@ impl Blockchain {
...
+                        continue;
+                    }
+
+                    let (microblock_slot, _) = self.get_slot_at(microblock.block_number(), microblock.view_number(), Some(&read_txn)).unwrap();
+
+                    // If there's another microblock set to this slot, we
+                    // notify the fork event.
+                    if slot == microblock_slot {
```
paberr: We need to check here:
- whether the view number is the same
*and*
- whether the the slot is the same

blockchain-albatross/src/blockchain.rs
```diff
@@ -592,6 +599,72 @@ impl Blockchain {
...
+                        let header = microblock.header();
+                        let microheader2;
+
+                        if let BlockHeader::Micro(microheader) = header {
```
paberr: There are `unwrap_micro_ref` and `unwrap_micro`, which simplify this.

paberr (changes requested):
The changes that need to be made are commented inline.
Feel free to ask in case of any questions. :)

blockchain-albatross/src/chain_store.rs
```diff
@@ -159,6 +159,41 @@ impl ChainStore {
...
+        let txn = match txn_option {
+            Some(txn) => txn,
+            None => {
+                read_txn = ReadTransaction::new(&self.env);
+                &read_txn
+            }
+        };
+
+        // Seek to the first block at the given height.
+        let mut blocks= Vec::new();
```
paberr:
```suggestion
        let mut blocks = Vec::new();
```

blockchain-albatross/src/blockchain.rs
```diff
@@ -592,6 +599,44 @@ impl Blockchain {
                 return Err(PushError::InvalidBlock(BlockError::InvalidJustification));
             }
 
+            // Check if there are two blocks in the same slot. Since we already
+            // verified the validator for the current slot, this is enough to
+            // check for fork proofs. A fork proof doesn't apply to the genesis
+            // block, so skip block 0.
+            if block.block_number() != 0 {
```
paberr: I don't think we need this check as the block number 0 is a macro block anyway.

blockchain-albatross/src/blockchain.rs
```diff
@@ -592,6 +599,44 @@ impl Blockchain {
...
+
+                // Get the microheader from the block
+                let microheader1 = block.clone().unwrap_micro().header;
+
+                // Get the justification for the block. We asume that the
+                // validator's signature is valid.
+                let justification1 = block.clone().unwrap_micro().justification.signature;
+
+                // Get the view number from the block
+                let view_number = block.clone().view_number();
```
paberr: Cloning is a heavy operation that we should avoid as much as possible.
```suggestion
                let microheader1 = &micro_block.header;

                // Get the justification for the block. We asume that the
                // validator's signature is valid.
                let justification1 = &micro_block.justification.signature;

                // Get the view number from the block
                let view_number = micro_block.header.view_number;
```

blockchain-albatross/src/blockchain.rs
```diff
@@ -592,6 +599,44 @@ impl Blockchain {
...
+
+                for microblock in microblocks.iter() {
+                    let (microblock_slot, _) = self.get_slot_at(microblock.block_number(), microblock.view_number(), Some(&read_txn)).unwrap();
+
+                    // If there's another microblock set to this slot, we
+                    // notify the fork event.
+                    if slot == microblock_slot && view_number == microblock.view_number() {
```
paberr: If the `view_number` is the same, the slot is the same, too.
```suggestion
                    // If there's another microblock set to this slot, we
                    // notify the fork event.
                    if view_number == microblock.view_number() {
```

blockchain-albatross/src/blockchain.rs
```diff
@@ -592,6 +599,44 @@ impl Blockchain {
...
+                        let microheader2 = microblock.clone().unwrap_micro().header;
+                        let justification2 = block.clone().unwrap_micro().justification.signature;
+
+                        let proof = ForkProof {
+                            header1:        microheader1.clone(),
+                            header2:        microheader2.clone(),
+                            justification1: justification1,
+                            justification2: justification2};
```
paberr: Avoid the cloning here.
```suggestion
                        let micro_block2 = microblock.unwrap_micro();
                        let microheader2 = micro_block2.header;
                        let justification2 = micro_block2.justification.signature;

                        let proof = ForkProof {
                            header1:        microheader1.clone(), // Previously, we only had a reference.
                            header2:        microheader2,
                            justification1: justification1.clone(), // Previously, we only had a reference.
                            justification2: justification2
                        };
```

paberr (changes requested):
Okay, looks quite good now.
I pointed out a few improvements to be made.
Also, I would rename `microheader` and `microblocks` etc to match our usual naming using `_`.

blockchain-albatross/src/blockchain.rs
```diff
@@ -592,6 +599,39 @@ impl Blockchain {
...
+            // Get the view number from the block
+            let view_number = block.view_number();
+
+            for micro_block in micro_blocks.iter() {
+                // If there's another microblock set to this view number, we
+                // notify the fork event.
+                if view_number == micro_block.view_number() {
+                    let test = micro_block.unwrap_micro();
+                    let micro_header2 = micro_block.clone().unwrap_micro().header;
+                    let justification2 = micro_block.clone().unwrap_micro().justification.signature;
```
paberr: Ah, I think my previous suggestion was probably not working as `micro_block` is a reference.
To get rid of cloning the whole block twice here, you can do:
```suggestion
                    let micro_header2 = micro_block.unwrap_micro_ref().header.clone();
                    let justification2 = micro_block.unwrap_micro_ref().justification.signature.clone();
```
Or alternatively:
```suggestion
                    let micro_block = micro_block.clone().unwrap_micro();
                    let micro_header2 = micro_block.header;
                    let justification2 = micro_block.justification.signature;
```

In the first case, you're only cloning the parts you need, in the second you clone the whole block once.

And to have another alternative: The whole cloning is required, since we do `for micro_block in micro_blocks.iter() {` and `.iter()` gives us an iterator over references `&Block`.
However, since we don't need the `micro_blocks` for anything else afterwards, we could do:
`for micro_block in micro_blocks {` or even `for micro_block in micro_blocks.drain(..).map(|block| block.unwrap_micro()) {`.
The first will consume `micro_blocks` and iterate over `Block`. The second requires `micro_blocks` to be declared as `mut` as it empties the vector (plus I also did the conversion to `MicroBlock` in the `map()` here).

paberr (changes requested):
Ok, final comment :)

blockchain-albatross/src/blockchain.rs
paberr: I would actually choose one of the latter two possibilities as they avoid the cloning altogether. :)

paberr (approved):

PR #150, opened 2020-05-25
Add beserial derivation for enums

paberr:
This PR adds new derive functionality that will make a lot of our enums easier.
Currently, beserial can only derive `Serialize`, `Deserialize` for enums that have **no data** attached.

This lead to the common pattern of having an enum `XType` for every data carrying enum `X`.
```rust
enum X {
    OptionA(SomeTypeA),
    OptionB(SomeTypeB),
}
#[derive(Serialize, Deserialize)]
enum XType {
    OptionA = 0,
    OptionB = 1,
}
```
Then, we use this type enum to implement a (de-)serialization for `X` manually.

While for some enums this might still makes sense, this is a huge overhead for others.

This PR adds functionality to automatically derive (de-)serialization for these types of enums, too. So, the example from above would then look like:
```rust
#[derive(Serialize, Deserialize)]
enum X {
    #[beserial(discriminant = 0)]
    OptionA(SomeTypeA),
    #[beserial(discriminant = 1)] // The attributes are optional.
    OptionB(SomeTypeB),
}
```

This PR includes tests for this functionality and fixes a bug in the previous derivation if no discriminants were given.

### DIFF
--- a/beserial/beserial_derive/src/lib.rs
+++ b/beserial/beserial_derive/src/lib.rs
@@ -1,5 +1,82 @@
 #![recursion_limit = "128"]
 
+//! This crate allows to automatically derive Serialize and Deserialize implementations for beserial.
+//!
+//! While the derivation for basic structs works simply by adding `#[derive(Serialize, Deserialize)]`,
+//! there are multiple advanced options possible.
+//!
+//! ## (De-)serializing structs
+//! If the struct contains fields that have an unknown size (such as `Vec`) and these implement
+//! `SerializeWithLength` and `DeserializeWithLength`, automatic derivation is still possible.
+//! You can add the `#[beserial(len_type(X))]` attribute to the field and specify how the length
+//! should be encoded (`u8`, `u16`, or `u32`).
+//!
+//! Other field options are:
+//! - `#[beserial(skip)]` skips this field during (de-)serialization and relies on `Default` instead
+//! - `#[beserial(skip(default = X))]` same as `skip` but allows to give a custom default value literal
+//! - `#[beserial(len_type(X))]` see above, allows to derive (de-)serialization for fields
+//!   implementing `(De-)SerializeWithLength`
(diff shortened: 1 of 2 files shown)


beserial/beserial_derive/src/lib.rs
```diff
@@ -213,47 +450,114 @@ fn impl_deserialize(ast: &syn::DeriveInput) -> TokenStream {
...
 
-            let ty= if uvar {
-                enum_type.unwrap_or_else(|| Ident::new("u64", syn::export::Span::call_site()))
+            let ty = if uvar {
+                enum_type.unwrap_or_else(|| Ident::new("u64", Span::call_site()))
             } else {
-                enum_type.unwrap_or_else(||panic!("Deserialize can not be derived for enum {} without repr(u*) or repr(i*)", name))
+                enum_type.unwrap_or_else(|| {
+                    panic!(
+                        "Serialize can not be derived for enum {} without repr(u*) or repr(i*)",
```
jeffesquivels: Shouldn't this still be "Deserialize" instead of "Serialize"?

jeffesquivels (dismissed):
LGTM, just one minor comment about the text of a panic!()

beserial/beserial_derive/src/lib.rs
paberr: Copy & paste error :P

paberr:
@jeffesquivels Ok, panic is fixed as well as a small bug in the code generation.
Previously, I generated the code as `Serialize::serialize(&#ident, writer)`. However, as `#ident` is already a reference by the `ref` in the `match` rules, this leads to a double reference. While `Serialize` can still handle this, `SerializeWithLength` will run into problems.

I adapted the test to reflect this. :)

paberr:
I added some documentation for `beserial_derive` (as it was not documented at all) and made the limit feature available during derivation.

jeffesquivels (approved):

PR #151, opened 2020-06-09
Update the seed node for the public devnet

jeffesquivels:


paberr (approved):
If the address is correct, this LGTM.

PR #152, opened 2020-06-09
Use fee option for fee when creating tx via RPC

sisou:
Somehow, instead of the option parameter `fee`, `value` was used for the fee. Someone even managed to put a `FIXME` comment next to it :smile:


jeffesquivels (approved):
LGTM

PR #154, opened 2020-06-10
Add RPC validator signalling methods

sisou:
There are lots of validator & staking related methods still missing in the RPC server. Most of them are for signalling transactions, which require a signature of the tx by the validator private key in their transaction data.

To create this signature, the mempool RPC handler needs to get a reference of the client's validator object.

This code here works and I was able to reactivate my validator with it (on first try!) :blush: 

**~~I have done this here without regarding that the validator is an `Option` (because the validator can be disabled in the config). Currently the `Option` is simply `unwrap`ped, which I understand will panic when the option is actually empty. Since I have no idea how to handle Options properly in Rust, I leave that part to the experts~~ :wink:**

### TODO
- [ ] Add comments to explain why we first add an empty signature, maybe refactor variable names?
- [x] Properly handle the `validator` argument as an `Option` (not simply `unwrap()`)
- [ ] Add test(s)?

### DIFF
--- a/lib/src/extras/rpc_server.rs
+++ b/lib/src/extras/rpc_server.rs
@@ -63,7 +63,7 @@ pub fn initialize_rpc_server(client: &Client, config: RpcServerConfig) -> Result
     let wallet_manager = Arc::clone(&wallet_handler.unlocked_wallets);
     handler.add_module(wallet_handler);
 
-    let mempool_handler = MempoolAlbatrossHandler::new(client.mempool(), Some(wallet_manager));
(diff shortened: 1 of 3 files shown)


paberr:
Generally, it looks good. :)
However, I would have preferred to use the [Transaction Builder](https://github.com/nimiq/core-rs-albatross/tree/albatross/transaction-builder) crate. That makes the transactions a bit more readable.

paberr (dismissed):
Looks good to me!

paberr (approved):

PR #155, opened 2020-06-13
View change on macro block observed on the alpha testnet

jeffesquivels:
While running the alpha testnet we ran into an issue where the network would partition because of a fork and wasn't able to produce blocks anymore.

While reviewing the logs, it became clear that there were some nodes that were not able to rebranch to the correct chain because somehow they thought the right block was invalid, they all had messages similar to this on their logs:

```
2020-06-10 05:50:13 INFO  blockchain           | New block is on Better chain with fork at #167295 (current #167296.0, new block #167296.1)
 2020-06-10 05:50:13 WARN  blockchain           | Rejecting block - invalid seed (Forged)
 2020-06-10 05:50:13 WARN  blockchain           | Rejecting block - Bad header / justification
```

While looking into it, the problem seemed to be related to the fact that block number 167296 is a macro block, and after further inspection, I think the problem was caused because `get_slot_at()` was using `state.current_slots` to get the information used to validate the slot for 167296.1, but since 167296.0 was already accepted, `state.current_slots` was already updated at that point and no longer had the information about the right slot for 167296.1.

This same scenario would happen whenever the current head of the chain is at a macro block and we are trying to get the slot for a block in that same epoch (i.e. that has this same block number or lower), so the fix in this PR deals with it by detecting when this is happening and using `state.previous_slots` to get the correct slot for the block.

### DIFF
--- a/blockchain-albatross/src/blockchain.rs
+++ b/blockchain-albatross/src/blockchain.rs
@@ -1327,11 +1327,19 @@ impl Blockchain {
         };
 
         // Gets slots collection from either the cached ones, or from the macro block.
+        // Note: We need to handle the case where `state.block_number()` is at a macro block (so `state.current_slots`
+        // was already updated by it, pushing this epoch's slots to `state.previous_slots` and deleting previous'
+        // epoch's slots).
         let slots_owned;
         let slots = if policy::epoch_at(state.block_number()) == policy::epoch_at(block_number) {
-            state.current_slots.as_ref().expect("Missing current epoch's slots")
+            if policy::is_macro_block_at(state.block_number()) {
+                state.previous_slots.as_ref()
+                    .unwrap_or_else(|| panic!("Missing epoch's slots for block {}.{}", block_number, view_number))
+            } else {
+                state.current_slots.as_ref().expect("Missing current epoch's slots")
+            }
         }
-        else if policy::epoch_at(state.block_number()) == policy::epoch_at(block_number) + 1 {
+        else if !policy::is_macro_block_at(state.block_number()) && policy::epoch_at(state.block_number()) == policy::epoch_at(block_number) + 1 {
             state.previous_slots.as_ref()
                 .unwrap_or_else(|| panic!("Missing previous epoch's slots for block {}.{}", block_number, view_number))
         }


sisou:
So this happened for view changes for a macro block! Great debugging, Jeff! :+1:

paberr:
That's interesting!
I'm wondering how that happens and it would probably make sense if you also involve @brunoffranca into the discussion.
If a validator accepted the macro block, it should have sufficient signatures for everyone to accept it and the validators who accepted it should not vote for a view change.
So, in theory, view changing and accepting the previous block should be mutually exclusive due to the 2/3 majority required for both.

Also, the fix probably won't help as we don't have code for rebranching across macro blocks.

jeffesquivels:
>If a validator accepted the macro block, it should have sufficient signatures for everyone to accept it and the validators who accepted it should not vote for a view change.
>So, in theory, view changing and accepting the previous block should be mutually exclusive due to the 2/3 majority required for both.

Hmm, yes, this makes a lot of sense. So it's indeed very interesting that what I observed in the logs is happening at all. I can only think of three possible causes:

1. There's a bug in the code that accepts macro blocks that is allowing a block without enough signatures to be accepted
2. There's a bug where validators will still sign a new macro block (with a higher view change number) even if they already signed a previous macro block (with a lower view change number).
3. Since we're restarting the alpha testnet sometimes, it may be that block #167296.0 came from a previous "session" of the testnet, but I think it would be very difficult for that to happen at that height (since I would have expected the correct chain from the current session to fork from the correct chain from the previous session way earlier than after 167296 blocks)>

Are there any more possibilities I may be missing?

Btw, independently of whether this PR fixes this specific problem or not, do you think it still makes sense to merge it? Basically, the problem is still there that after accepting a macro block, the right slots for any previous micro block (in the same epoch) are no longer in `state.current_slots` but in `state.previous_slots` (what I don't know is if, for some external reason, the scenario where we are at a macro block and then ask for the slot of a previous micro block would never happen).

brunoffranca:
Yes, the fix still makes sense. There are (rare) situations where a macro block can get enough signatures and still get view-changed, so it's important that the nodes can see what is the right branch.
That being said, there are only two situations (that I can think of) where this would happen:
1) There are more than 1/3 malicious validators and they withheld their commit signatures until the block got view-changed.
2) There was some network partition immediately before the commit phase. In this case, everyone already had the block proposal and the prepare signatures. Then they would send the commit signatures but they would not reach everyone. Nodes would then timeout and start a view change. Now you have a full set of commit signatures and a full set of view change messages that will be delivered when the network comes back together.
@jeffesquivels is any of these cases possible to have happened in the testnet? Because if not, it would be important to find how this happened.

paberr:
As I don't think anyone modified the code to some malicious behaviour, option 1 is nearly impossible.
Option 2 also seems quite unlikely, what do you think Jeff?

But I agree it still makes sense to merge this fix although it's not fixing the bug we are seeing here.

jeffesquivels:
Hmm, I agree with @paberr that it is unlikely that we had intentionally malicious validators, so the first scenario seems unlikely. The second one is more probable, but still unlikely.

I will try to unravel from the logs what happened before this view change on the macro block happened to see if we can get a better hypothesis (although that may be difficult because the logs don't capture the whole state of the network at the point of failure).


Btw, I just moved the code from this PR into #160 so that we can separate both things (this discussion and the code change).

jeffesquivels:
Closing this PR since we're replacing this whole part of the code.

PR #163, opened 2020-06-19
Constrain fork proofs to relevant epochs.

paberr:
Currently, we can let validators produce invalid fork proofs simply by sending them a fork for an epoch other than the current or previous one.

This PR mitigates the problem in two ways:
1. After each epoch, it removes fork proofs that are invalid from there on.
2. The blockchain only checks for forks in the relevant epochs.

However, this PR does not address that the same fork proof, currently, might appear in multiple blocks (which is just redundant information).

### DIFF
--- a/validator/src/slash.rs
+++ b/validator/src/slash.rs
@@ -1,6 +1,6 @@
-use block_albatross::{ForkProof, Block, MicroBlock};
-use std::collections::HashSet;
 use beserial::Serialize;
+use block_albatross::{Block, ForkProof, MacroBlock, MacroHeader, MicroBlock};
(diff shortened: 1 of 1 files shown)


sisou:
@paberr It seems you applied a formatter to `blockchain.rs`, which leads to a huge change set and merge conflicts with the other open PRs. Was that on purpose?

paberr:
Yes, we agreed to format our files gradually.

brunoffranca (dismissed):

brunoffranca (dismissed):

brunoffranca (approved):

PR #164, opened 2020-06-19
Remove assertion because it's not valid for all cases

### LINKED ISSUE #158: Fork contains macro block panic
After restarting the alpha testnet, we got panics on several nodes with the following pattern:

```
2020-06-12 03:34:36 INFO  blockchain           | New block is on Unknown chain with fork at #252 (current #255.0, new block #255.0)
2020-06-12 03:34:36 DEBUG blockchain           | Creating/extending fork with block 7ac9154a37e81741d01c68897547b7cc07438858b5f502f33b0f270ee15e1d3d, block number #255, view number 0
2020-06-12 03:34:36 INFO  blockchain           | New block is on Better chain with fork at #252 (current #255.0, new block #256.0)
2020-06-12 03:34:36 DEBUG blockchain           | Rebranching to fork 0cd121eb034ce1fa5647a48e6dc28697f165d3025649c392eacdaf46a6d7d926, height #256, view number 0
2020-06-12 03:34:42 ERROR panic                | thread 'tokio-runtime-worker-2' panicked at 'assertion failed: `(left == right)`
  left: `Macro`,
 right: `Micro`: Fork contains macro block': blockchain-albatross/src/blockchain.rs:777
```

It seems the panic is happening under this scenario:

1. The node is currently on the last micro block of the epoch (i.e. preceding the macro block)
2. Another block for that same height and view change number is received but the node can't tell which chain is better at this point
3. The macro block that succeeds this micro block arrives and it is part of the the fork (i.e. not the main chain) which (correctly) causes a rebranch

Now, that means that [`rebranch()`](https://github.com/nimiq/core-rs-albatross/blob/f0e9454c8c17fd4da368d93584dbe90b81bedd08/blockchain-albatross/src/blockchain.rs#L764) will be called with the `chain_info` parameter set to the `ChainInfo` of the macro block that was just received (see [this line](https://github.com/nimiq/core-rs-albatross/blob/f0e9454c8c17fd4da368d93584dbe90b81bedd08/blockchain-albatross/src/blockchain.rs#L640) and then [this line](https://github.com/nimiq/core-rs-albatross/blob/f0e9454c8c17fd4da368d93584dbe90b81bedd08/blockchain-albatross/src/blockchain.rs#L650) in the code) which means [this assert](https://github.com/nimiq/core-rs-albatross/blob/f0e9454c8c17fd4da368d93584dbe90b81bedd08/blockchain-albatross/src/blockchain.rs#L777) will be false as `current.1` is the `ChainInfo` from the macro block (as assigned on [this line](https://github.com/nimiq/core-rs-albatross/blob/f0e9454c8c17fd4da368d93584dbe90b81bedd08/blockchain-albatross/src/blockchain.rs#L773)).

I think in this case this happened because after restarting the alpha testnet, there were external nodes that were still sharing the blocks from the previous "session", but it seems this scenario would also be possible under normal operation of the network if an attacker gets the chance to be the validator of the last micro block of an epoch, creating two valid blocks and broadcasting them (after this only one macro block will be issued, but nodes on the chain where the macro block was not built on will need to rebranch, hitting this same `assert` and panicking).

@paberr @brunoffranca wdyt? should this be addressed somehow or is there any other reason why this scenario could never happen?

jeffesquivels:
This PR is meant to fix #158

### DIFF
--- a/blockchain-albatross/src/blockchain.rs
+++ b/blockchain-albatross/src/blockchain.rs
@@ -813,10 +813,6 @@ impl Blockchain {
         let mut fork_chain: Vec<(Blake2bHash, ChainInfo)> = vec![];
         let mut current: (Blake2bHash, ChainInfo) = (block_hash, chain_info);
         while !current.1.on_main_chain {
-            // A fork can't contain a macro block. We already received that macro block, thus it must be on our
-            // main chain.
-            assert_eq!(current.1.head.ty(), BlockType::Micro, "Fork contains macro block");
-
             let prev_hash = current.1.head.parent_hash().clone();
             let prev_info = self.chain_store
                 .get_chain_info(&prev_hash, true, Some(&read_txn))


brunoffranca:
If we are going to implement Tendermint, this PR and corresponding issue no longer makes sense.

PR #166, opened 2020-07-04
New reward

### LINKED ISSUE #89: Reward distribution
Right now, we distribute the reward equally between all non-misbehaving validators.
While this incentives reporting of forks, it also incentivises validators attacking each other to maximise their reward.

The current specification thus says:
> His reward is confiscated and burned\footnote{The reward is not divided among the other validators so as to not incentivize them to attack each other (ex: by doing a denial-of-service).}.

brunoffranca:
This PR handles several things related to the way rewards are calculated and distributed:

- Rewards are now divided among all validators. But the rewards for misbehaving validators are burned. This fixes #89.
- The reward remainder is given to a single random validator. Previously it was randomly divided among all validators but this is unnecessary since the remainder is always very small.
- Added the new supply curve to the policy file.
- Changed RewardPot to use the new supply curve and also simplified the way rewards are calculated. Now micro block rewards consist only of the transaction fees and macro block rewards consist only of the coinbase.
- Added the field `extra_data` to MacroBlock. This was necessary in order to encode the genesis supply into the genesis block, but it might also be useful for other purposes.
- Added some documentation all-around.

Checklist:

### DIFF
--- a/blockchain-albatross/src/blockchain.rs
+++ b/blockchain-albatross/src/blockchain.rs
@@ -12,7 +12,6 @@ use parking_lot::{
 
 use account::inherent::AccountInherentInteraction;
 use account::{Account, Inherent, InherentType};
-use account::{Account, Inherent, InherentType};
 use accounts::Accounts;
 use beserial::Serialize;
 use block::{
@@ -38,7 +37,7 @@ use tree_primitives::accounts_proof::AccountsProof;
 use tree_primitives::accounts_tree_chunk::AccountsTreeChunk;
 use utils::merkle;
 use utils::observer::{Listener, ListenerHandle, Notifier};
-use vrf::{AliasMethod, VrfSeed, VrfUseCase};
+use vrf::VrfSeed;
 
 use crate::chain_info::ChainInfo;
 use crate::chain_store::ChainStore;
@@ -132,7 +131,7 @@ pub struct BlockchainState {
     macro_head_hash: Blake2bHash,
 
     // TODO: Instead of Option, we could use a Cell here and use replace on it. That way we know
-    // at compile-time that there is always a valid value in there.
+    //       at compile-time that there is always a valid value in there.
     current_slots: Option<Slots>,
     previous_slots: Option<Slots>,
 }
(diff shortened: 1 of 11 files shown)


blockchain-albatross/src/blockchain.rs
```diff
@@ -1928,65 +1936,75 @@ impl Blockchain {
...
         let mut slashed_set_iter = slashed_set.iter().peekable();
 
         // All accepted inherents.
         let mut inherents = Vec::new();
-        // Remember the number of eligible slots a stake had (that was able to accept the inherent)
+
+        // Remember the number of eligible slots that a validator had (that was able to accept the inherent)
         let mut num_eligible_slots_for_accepted_inherent = Vec::new();
 
+        // Remember the total amount of reward must be burned. The reward for a slot is burned
```
jeffesquivels: I think a "that" is missing here: `that must be burned.`

blockchain-albatross/src/reward_registry/reward_pot.rs
```diff
@@ -1,135 +1,240 @@
...
 
-        reward
+        // Set the current reward to the newly calculated reward.
+        txn.put(&self.reward_pot, Self::CURRENT_REWARD_KEY, &current_reward);
     }
 
-    fn reward_for_macro_block(block: &MacroBlock) -> Coin {
-        policy::block_reward_at(block.header.block_number)
-    }
+    /// Rollbacks the RewardPot database for a micro block. It takes the whole micro block as input
```
jeffesquivels: s/Rollbacks/Rollsback/ ? :thinking:

jeffesquivels (dismissed):

blockchain-albatross/src/blockchain.rs
brunoffranca: Done!

blockchain-albatross/src/reward_registry/reward_pot.rs
brunoffranca: Rolls back seems to be the most correct. But I'm not getting a lot of results from Google.

jeffesquivels (approved):

PR #168, opened 2020-07-15
Reformatted the entire repo with rustfmt.

brunoffranca:
Reformatted the entire repo using rustfmt (through InteliiJ IDEA). Should conform to standard Rust code now. No changes to content were made.


paberr (dismissed):
I haven't checked each individual file but only scrolled over some of them.

paberr (approved):
Rebase

PR #171, opened 2020-07-17
Block timestamps checks

brunoffranca:
Note: this requires PR #166 to be merged before.

## Pull request checklist

1.  It adds checks that a block's timestamp is valid. Mainly that it is not before the previous block timestamp and that it is not too far in the future.
2. Reorganized the blockchain-albatross crate.
3. Added some documentation.


blockchain-albatross/src/blockchain/blockchain.rs
```diff
@@ -430,9 +309,33 @@ impl Blockchain {
...
+
+        // Check that the current block timestamp less the node's system time is less than or equal
+        // to the allowed maximum drift. Basically, we check that the block isn't from the future.
+        // This is necessary so that validators cannot create blocks far in the future, which would
+        // give them bigger rewards.
+        // Both times are given in Unix time standard.
+        let now = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .expect("System time is before Unix epoch")
+            .as_millis() as u64;
```
paberr, line 328: We usually want to rely on the NetworkTime instead of our local SystemTime.

paberr (changes requested):
As you also pointed out in one TODO, we should use the network time instead.

blockchain-albatross/src/blockchain/blockchain.rs
brunoffranca: Agree with you there. But NetworkTime isn't implemented yet. We can merge this one now and then I'll do a PR for the NetworkTime later (which will be a lot bigger anyway).

brunoffranca:
Since #182 has a big refactor of the blockchain-albatross crate, it's best to redo this PR based on that one (and ignore the reorganization done by this PR).

brunoffranca:
Deprecated. Solved by commit 8b68de3dbecf1f706fd048be9285ea01340a3aea.

PR #175, opened 2020-07-29
Intermediate macro blocks

nibhar:
## Pull request checklist

This pull request adds intermediate macro blocks as the third block type. They work the same as macro blocks do, with the exception that the validator set does not change.


paberr:
While refactoring the whole slash registry/reward registry thing, I realised that we *do* need to include both slashed sets in the election blocks.

Previous to this PR, election blocks only contained the slashed set of the previous epoch (not the current one that ends with the block at hand). This was and is required for syncing, so that rewards can be calculated correctly.

This PR now added the current slashed set to macro blocks that are no election blocks. However, consider the following scenario: I synced up to the latest election block and start producing blocks. The election block didn't include the "current epoch" slashed set (which now after beginning the next epoch is the "previous epoch" slashed set). So, upon arriving at the first checkpoint block, this information is missing.

I'll address this issue in my follow up PR that refactors said code.

blockchain/src/blockchain/mod.rs
```diff
@@ -1004,12 +1002,20 @@ impl AbstractBlockchain for Blockchain {
...
     }
 
+    fn get_batch_transactions(
+        &self,
+        batch: u32,
+        txn_option: Option<&Transaction>,
+    ) -> Option<Vec<BlockchainTransaction>> {
+        self.get_epoch_transactions(batch, txn_option)
+    }
+
```
brunoffranca, line 1012: What does this actually do? It seems like it returns the transactions for epoch x when you ask for the transactions for batch x.

nibhar: This is part of the 1.0 chain object. Therefore there is no concept of epochs or batches and the functions threat the input as the actual block height, effectively making both functions exactly the same.

It is a remnant of the single repo for both versions of the protocol.

brunoffranca: Ok. Seems good to me then.

brunoffranca:
We only need to solve the issue of failing when we provide any extra data to the macro blocks and then it can be merged.

PR #204, opened 2020-11-03
Fix weird compiler errors in tests

jgraef:
## Pull request checklist

In `collections/linked_list.rs` line 1413 compilation fails, because the compiler needs more type annotations. Similar in other locations where something like `assert_eq(..., vec![])` is used.

Compiler version: `rustc 1.48.0-nightly (ef663a8a4 2020-09-30)`

Also fixes an issue in `nimiq-collections` when testing without `bitset` feature. I moved the tests into the `bitset` module like it is intended for unit tests.


jgraef:
Included it with other minor fixes: #205

PR #216, opened 2021-01-04

### LINKED ISSUE #344: Unpark validators automatically
Validators should be able to recognize if they have been parked in the staking contract so that they can send an unpark transaction and prevent being automatically inactivated.

vlvrd:


validator/src/validator.rs
```diff
@@ -127,9 +132,13 @@ impl<TNetwork: Network, TValidatorNetwork: ValidatorNetwork>
         let network1 = Arc::clone(&network);
         let (proposal_sender, proposal_receiver) = ProposalBuffer::new();
 
+        // FIXME: Placeholder, update this line :point_down: as soon as the staking contract changes are merged
+        let id = ValidatorId::default();
```
sisou: The staking contract changes are merged, so this should be possible to do correctly now?

vlvrd: It is, this branch is outdated (it was updated before the staking contract changes went into _albatross_)

client.toml
```diff
@@ -14,6 +14,7 @@ port = 8648
 level = "debug"
 
 [validator]
-#wallet_account = ""
-#wallet_password = ""
+validator_address = "4204 2042 0420 4204 2042 0420 4204 2042 0420"
```
brunoffranca: This doesn't seem like a valid address.

validator/src/validator.rs
```diff
@@ -62,12 +67,15 @@ struct ProduceMicroBlockState {
...
-    // TODO: Also have the validator ID here.
+
     signing_key: bls::KeyPair,
-    wallet_key: Option<keys::KeyPair>,
     database: Database,
     env: Environment,
 
+    validator_address: Address,
+    cold_key: KeyPair,
+    warm_key: KeyPair,
```
brunoffranca, line 77: I'm going to need methods to fetch these. Just like the signing_key() method.

PR #221, opened 2021-02-11
Introduce validator ID

brunoffranca:

It adds the validator ID field, which is a field supposed to uniquely and statically identify validators in the staking contract


brunoffranca (commented):
Seems good to me.

PR #238, opened 2021-03-20
Implement getTransactionByHash RPC method

sisou:
This PR implements the RPC `getTransactionByHash` method. However, to be able to include block information in the result, such as block_number, confirmations, block_time, the RPC method does not use the exposed `history_store.get_transaction`, as that function loses the block information from the `ExtendedTransaction` struct and only returns the transaction itself.

@brunoffranca Please let me know if this is approach is correct, or if the block information can be retrieved differently.

### TODO
- Find the `transaction_index` from the block, to include in the result
- Should this method also return mempool transactions?

### DIFF
--- a/blockchain-albatross/src/history_store/history_store.rs
+++ b/blockchain-albatross/src/history_store/history_store.rs
@@ -9,7 +9,6 @@ use merkle_mountain_range::store::memory::MemoryStore;
 
 use nimiq_database::{Database, Environment, ReadTransaction, Transaction, WriteTransaction};
 use nimiq_hash::{Blake2bHash, Hash};
-use nimiq_transaction::Transaction as BlockchainTransaction;
 
 use crate::history_store::mmr_store::MMRStore;
 use crate::history_store::{
@@ -189,11 +188,11 @@ impl HistoryStore {
     }
 
     /// Gets a basic transaction (no inherents!) given its hash.
-    pub fn get_transaction(
+    pub fn get_extended_transaction_by_transaction_hash(
         &self,
-        hash: Blake2bHash,
+        hash: &Blake2bHash,
         txn_option: Option<&Transaction>,
-    ) -> Option<BlockchainTransaction> {
+    ) -> Option<ExtendedTransaction> {
         let read_txn: ReadTransaction;
         let txn = match txn_option {
             Some(txn) => txn,
@@ -204,12 +203,10 @@ impl HistoryStore {
         };
 
         // Get leaf hash.
-        let leaf_hash = self.get_leaf_hash(&hash, Some(txn))?;
+        let leaf_hash = self.get_leaf_hash(hash, Some(txn))?;
 
         // Get extended transaction.
-        let ext_tx = self.get_extended_tx(&leaf_hash, Some(txn))?;
-
-        Some(ext_tx.unwrap_basic().clone())
+        self.get_extended_tx(&leaf_hash, Some(txn))
     }
 
     /// Gets all extended transactions for a given epoch.
@@ -408,7 +405,7 @@ impl HistoryStore {
 
     /// Gets an extended transaction by its hash. Note that this hash is the leaf hash (see MMRHash)
     /// of the transaction, not a simple Blake2b hash of the transaction.
-    pub fn get_extended_tx(
+    fn get_extended_tx(
         &self,
         hash: &Blake2bHash,
         txn_option: Option<&Transaction>,
@@ -426,7 +423,7 @@ impl HistoryStore {
     }
 
     /// Gets a leaf hash from the hash of its transaction (only basic, no inherents!).
-    pub fn get_leaf_hash(
+    fn get_leaf_hash(
         &self,
         hash: &Blake2bHash,
         txn_option: Option<&Transaction>,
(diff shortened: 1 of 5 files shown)


brunoffranca:
I can make history_store.get_transaction return the ExtendedTransaction, it's easy. I just thought you only wanted the Transaction.

sisou:
Yes, that would work. It would then be called `get_extended_transaction`. I think a method to get the wrapped transaction is still useful, so both?

sisou:
Thinking about it, the transaction can simply be unwrapped from the extended_tx obect, now that you added that getter. So the `get_transaction` method would be unnecessary.

sisou:
@jgraef I've been looking at the serialization of a `Transaction` via the `Transaction::from_blockchain` method defined in the rpc_interface `types.rs`. It is (currently) only used to serialize MicroBlock transactions, when a micro block is requested via RPC.

The serialization of the transaction data currently includes the `block_hash` and the index of the tx in the block.
The `block_hash` is unnecessary in the transaction object, as it is already part of the block header, also returned in a block RPC request.

If we deem the transaction_index either optional or not important enough to warrant including it here, and thus also exclude it from the transaction serialization, we can get rid of getting the block from the blockchain in the 'get_transaction_by_hash` method implemented here, as the remaining block information (block_number and timestamp) are included in the extended transaction object already.

As the one who implemented this, what do you think?

brunoffranca (approved):

brunoffranca:
> The block_hash is unnecessary in the transaction object, as it is already part of the block header, also returned in a block RPC request.

Not sure I understand that part. If you fetch a transaction from the HistoryStore, you get the block number and timestamp of the block when the transaction was included, but you can't get the hash. The history tree keeps all transactions during an epoch, so you can get a transaction proof for a transaction that was included in a block that you don't have.

I don't think we need the block hash though. It makes sense in PoW where you can have arbitrarily long forks. In Albatross a fork can only last until the next macro block at the end of the batch. And even within batches, forks are expected to be rare. The block number is probably enough information.

brunoffranca:
Now that I think about it, having the block hash seems like a very bad idea. Even history nodes aren't expected to keep all the blocks, they are only supposed to keep the election blocks and the history trees. So, if you try to get the block hash for a block that happened some time ago, there's a good chance you won't get it.

sisou:
> > The block_hash is unnecessary in the transaction object, as it is already part of the block header, also returned in a block RPC request.
> 
> Not sure I understand that part.

What I meant here is that in a micro block response, which includes the list of transactions, the block hash is already available in the block data itself, and doesn't need to be repeated in every tx object in the transactions list.

I wanted to find a reason to exclude the block hash from the transaction object. 

> The history tree keeps all transactions during an epoch, so you can get a transaction proof for a transaction that was included in a block that you don't have.

~So what does that mean? Do we need to prove transactions before returning them in the get_transaction_by_hash method, as otherwise we might be returning transactions that are actually not part of the chain?~
I think I misunderstood. You are talking about pruned blocks due to age, not fork blocks. So the history store only stores transactions that are in the chain? So when I can get a tx from the store, it's definitely in the chain? Or must we (or the user) `prove` it also?

sisou:
Following that logic to remove the block hash from the transaction result, the transaction index can also be removed, as it's not useful when the block was pruned, is that right?

brunoffranca:
> I think I misunderstood. You are talking about pruned blocks due to age, not fork blocks. So the history store only stores transactions that are in the chain? So when I can get a tx from the store, it's definitely in the chain? Or must we (or the user) prove it also?

Yes, I'm talking about pruned blocks due to age. You can't assume that nodes have past micro blocks, but you can assume that they have the history trees.
The history store stores transactions that are on the main chain of the node. Of course when the node rebranches the history store also changes. You do know that it's definitively on the chain if at least one macro block has been produced after the transaction was included. But you can easily see that with the block number.

> Following that logic to remove the block hash from the transaction result, the transaction index can also be removed, as it's not useful when the block was pruned, is that right?

I think so, yes.

jgraef:
Isn't there a use-case for a node that keeps all blocks (and thus transactions)? E.g. a blockchain explorer would need to find a transaction by hash and also in which block it was included. Using block height might be good most of the time, but not always. A explorer also could want to display transactions that were only included in forks. The transaction index is not needed, since you can find the transaction in the block by matching the txid.

The (`block_hash`, `tx_index`) tuple as a way to find a transaction was used previously, since transactions were only stored in the block store.

Anyway, I agree that this information is mostly not needed. In case for a blockchain explorer we would also need a store that maps txid -> block_hash. I'm not sure if we still have this. I'm also not familiar with the history store, so I think @brunoffranca should make the call. And we can change the RPC response however we want, the current struct is mostly taken from pre-Albatross.

But after wall, we should consider the use case where we want to keep everything and find the exact block for a transaction.

brunoffranca:
Well, for each block height there's only one block always (except during the current batch). So there's little use for the block hash, the block number is normally enough. We can create a tx_id to block hash database, but it should be cleared at the end of every batch.
About keeping all blocks. It is not that there isn't some use case, it's more that isn't feasible at all. Just a micro block header is 273 bytes. At one block per second, that's 23.5 MB per day, 8GB per year. Just on headers. If you want to keep the bodies and justifications it can quickly become too big.

brunoffranca:
But I _really really_ prefer if we use the History Tree to fetch transactions, instead of getting a block and fetching there. A lot of Albatross was designed with the idea that we would prune old blocks (except election macro blocks). If there's some specific information that you need to be included with the transactions, then it's better to add it to the `ExtendedTransaction` than to get that information from the micro blocks.

PR #246, opened 2021-03-30
Fix error message on client.toml absence

syvb:
## Pull request checklist

This fixes the error message that occurs when `client.toml` isn't found. Before, it incorrectly referred to the location where the config file was searched for as where an example was. This fixes it so it correctly labels the expected config location, and includes a reference to the actual example config.

### DIFF
--- a/lib/src/config/config_file/mod.rs
+++ b/lib/src/config/config_file/mod.rs
@@ -100,7 +100,7 @@ impl ConfigFile {
         let path = paths::home().join("client.toml");
         if !path.exists() {
             let msg = format!(
-                "Config file not found. Please create one. An example config file can be found at: {}",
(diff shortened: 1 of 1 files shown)


fiaxh:
I'd suggest to leave a recommended action in the message to give the user a clear guidance on what to do. The full version of a recommended action could be along the lines of "Please create one at {} or specify a location using `-c path/to/config.toml`". You can use the `path_example` variable to refer to the location of the example config.

syvb:
@fiaxh I've updated the error message to be like `Config file not found. Please create one at /home/smit/.nimiq/client.toml or specify a location using -c path/to/config.toml, see the example config file at /home/smit/.nimiq/client.toml.example`

PR #248, opened 2021-04-05
RPC: get extended transactions/inherent by batch

Eligioo:

New RPC method that sends the extended transactions for a given batch.

### DIFF
--- a/blockchain-albatross/src/history_store/history_store.rs
+++ b/blockchain-albatross/src/history_store/history_store.rs
@@ -255,6 +255,47 @@ impl HistoryStore {
         ext_txs
     }
 
+    /// Gets all extended transactions for a given batch.
+    pub fn get_batch_transactions(
+        &self,
+        batch_number: u32,
+        epoch_number: u32,
+        txn_option: Option<&Transaction>,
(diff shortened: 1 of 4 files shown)


brunoffranca (changes requested):
Fetching the entire epoch tree is unnecessary work. The proper way to do this would be to: 1) calculate the block numbers for that batch, there are functions for that in primitives/policy and then 2) fetch the transactions for each block number using the get_block_transactions method on HistoryStore.
But, if you only want this function to get the rewards paid to the validators, then it is unnecessary. The rewards for the validators are only paid on the macro blocks, so it is enough to get the transactions for those specific blocks. Just calculate the desired block number from the batch number and use it.

Eligioo:
@brunoffranca as I was reading your comment I was like: you are right and I know all this already but then why did I still do it this way. Feel bit ashamed haha. Anyway, I'll update the PR and I think `get_inherents_by_batch` as the RPC method name would be better than! Since it is indeed meant for the validator rewards.

Eligioo:
After discussing with @sisou how I would implement `get_inherents_by_batch`, he implemented it as `get_batch_inherents` afterwards in #249  👍

PR #249, opened 2021-04-06
Get transactions and inherents by block and batch

sisou:
## Pull request checklist

(I cancelled the pr-checks, as they are the same as the push-checks, as this branch is on top of `albatross`.)

Two new methods: `getTransactionsByBlockNumber` to get all blockchain transactions and inherents for a block, and a `getBatchInherents` method that takes the batch number and returns only the inherents of its macro block.

Two questions @brunoffranca:
1. Are inherents only included in macro blocks? E.g. the slashing inherents, are they only in the macro blocks at the end of batches?
2. Since the second method `getBatchInherents` returns the same content as `getTransactionsByBlockNumber` for macro blocks (if question 1 is answered with "yes"), do you think it is unnecessary? If one knows the batch number to give to the batch-inherents method, one would also know the macro block number and can use the txs-by-block-number method.

### DIFF
--- a/rpc-interface/src/blockchain.rs
+++ b/rpc-interface/src/blockchain.rs
@@ -5,7 +5,9 @@ use nimiq_account::Account;
 use nimiq_hash::Blake2bHash;
 use nimiq_keys::Address;
 
-use crate::types::{Block, OrLatest, SlashedSlots, Slot, Stakes, Transaction};
+use crate::types::{
+    Block, ExtendedTransactions, OrLatest, SlashedSlots, Slot, Stakes, Transaction,
+};
 
 #[cfg_attr(
     feature = "proxy",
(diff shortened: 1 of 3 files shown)


brunoffranca:
>     1. Are inherents only included in macro blocks? E.g. the slashing inherents, are they only in the macro blocks at the end of batches?
No, micro blocks can contain slash inherents. The other inherents though (reward, finalize batch, finalize epoch) only appear on macro blocks.

>     2. Since the second method `getBatchInherents` returns the same content as `getTransactionsByBlockNumber` for macro blocks (if question 1 is answered with "yes"), do you think it is unnecessary? If one knows the batch number to give to the batch-inherents method, one would also know the macro block number and can use the txs-by-block-number method.
You'd miss the slash inherents for that batch. If was I running a pool, I only care about the reward inherents, so I would probably just fetch the transactions for the macro block. But this `getBatchInherents` function might be useful for someone, I would keep it.

sisou:
> No, micro blocks can contain slash inherents. The other inherents though (reward, finalize batch, finalize epoch) only appear on macro blocks.

Hmm... Then the `getBatchInherents` function right now doesn't work correctly, it only fetches the inherents of the batch's macro block. 

How would you suggest we find every inherent in the batch's micro blocks?

The slash inherents are required to calculate a validator's trust score.

brunoffranca (approved):

PR #254, opened 2021-04-15
History Store - tx hash by address

brunoffranca:
## Pull request checklist

1. Adds a method to History Store to fetch transaction hashes by the sender/recipient address.
2. Merges two of History Store's databases into one.


blockchain-albatross/src/history_store/history_store.rs
```diff
@@ -597,32 +629,33 @@ impl HistoryStore {
...
-    /// Returns a vector containing all leaf hashes corresponding to the given block number.
-    fn get_leaf_hash_by_block(
+    /// Returns a vector containing all transaction (no inherents) hashes corresponding to the given
+    /// address.
+    pub fn get_tx_hashes_by_address(
         &self,
-        block_number: u32,
+        address: &Address,
         txn_option: Option<&Transaction>,
     ) -> Vec<Blake2bHash> {
```
sisou: The vec returned by this method can get veeeeery long for addresses with loads of transactions, as for example pools will, or games like Sunset Cyberspace.
I think it makes sense that this method gets a parameter `limit: u16` (or `max` or similar), so that it stops seeking after that amount of hashes has been found.

Will hashes be returned in the reverse order they have been inserted, so in reverse-historical order?

brunoffranca: I don't know in what order they get fetched from the database, that's a good question. Let me try to find out.

blockchain-albatross/src/history_store/history_store.rs
```diff
@@ -634,25 +742,59 @@ impl HistoryStore {
...
                 None => break,
             };
         }
 
-        leaf_hashes
+        tx_hashes
+    }
+
+    /// Returns the number of transactions (no inherents) associated to the given address.
+    pub fn get_num_txs_by_address(
```
sisou: Naming: this function does not necessarily return the number of stored txs. If an entry was deleted that was not the most recent tx, there's a gap in the indices and the number of txs is less than the last index + 1. Instead, I suggest to call this method `get_next_tx_index_for_address` or `get_last_tx_index_for_address` (and remove the +1) or something in that direction.

brunoffranca: You're right. Good catch. In that case I'll also make the method private, since then it doesn't have any use outside the History Store.

sisou (approved):

PR #255, opened 2021-04-20
Make rewards explicit transactions

brunoffranca:
## Pull request checklist

Creates transactions that match the reward inherents. The main objective is to make the rewards easier to fetch and handle for block explorers and such.

### DIFF
--- a/block-production-albatross/src/lib.rs
+++ b/block-production-albatross/src/lib.rs
@@ -210,17 +210,11 @@ impl BlockProducer {
         // Get the state.
         let state = self.blockchain.state();
 
-        // Initialize the inherents vector.
-        let mut inherents: Vec<Inherent> = vec![];
+        let inherents: Vec<Inherent> = self
+            .blockchain
+            .create_macro_block_inherents(&state, &header);
 
-        // All macro blocks are the end of a batch, so finalize the batch.
-        inherents.append(&mut self.blockchain.finalize_previous_batch(&state, &header));
-
-        // If this is an election macro block, then it also is the end of an epoch. So finalize the
-        // epoch.
-        if policy::is_election_block_at(block_number) {
-            inherents.push(self.blockchain.finalize_previous_epoch());
-        }
+        let transactions = self.blockchain.create_txs_from_inherents(&inherents);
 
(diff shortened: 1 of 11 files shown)


sisou:
@brunoffranca If I'm seeing this right rewards are now stored in the history store twice: once as an inherent and once as a transaction. Is that on purpose?

sisou:
Oh, actually not. The reward _transactions_ are not stored in the history store, only the reward inherents. Still, is that on purpose?

brunoffranca:
Actually, they should be both stored in the History Store. So, it is very much not on purpose. Where did you see that they were not stored?

sisou:
Here the transactions are not included in the extended txs that get added to the history store, instead an empty array is passed in:
https://github.com/nimiq/core-rs-albatross/blob/c04516edbc93da6fc7454633149e91a3998d9c1d/block-production-albatross/src/lib.rs#L226

brunoffranca:
Good catch. I guess Nata forgot to put the transactions from line 217 in there (and I didn't see that during review). Easy fix though, I'll do it tonight.

brunoffranca:
I ended taking 8 days off, so this got somewhat delayed. Furthermore, we couldn't have both the transactions and the inherents into the history tree since full nodes use the history tree when syncing. The solution is a bit more involved, it is in PR #260 . Anyway, the history tree now only contains the reward _transactions_, you can get them by simply asking for a macro block transactions.
P.S. The history tree now only stores slash inherents. The reward, finalize batch, and finalize epoch inherents no longer appear in the history tree.

sisou:
@brunoffranca Just tested this out for the first time. The reward transactions have a `validityStartHeight` of macro-block-height - 1. So for macroblock # 128 the validity start height is 127. I think it makes more sense to have the macro block height as the VSH. Do you think so, too, or is this not relevant?

brunoffranca:
What do you mean, the VSH?

sisou:
ValidityStartHeight. Too lazy to type it out.

brunoffranca:
You can always copy-paste. xD
It's nitpicking, but sure, here you go. c222764

PR #271, opened 2021-05-27
Add rotating_trace_log configuration to docker_config.sh

redmaner:
This change adds the option to enable and configure rotating trace log in docker using environment variables


sisou (approved):
I could imagine the variables being called `ROTATING_TRACE_LOG_*`, but I agree that that is getting long.

redmaner:
> I could imagine the variables being called `ROTATING_TRACE_LOG_*`, but I agree that that is getting long.

Since the variables are probably going to be rare use cases (let's hope so) it might be a good idea to just use the full name `ROTATING_TRACE_LOG_*` for clarity, regardless of the length. I'm down for either route

PR #272, opened 2021-07-28
Subscribe to and log consensus events in client

sisou:
Instead of (ab)using trace/debug logs in the consensus itself (that I changed to info logs in 0cc70d5a7ef7e718e77f30c5268d9b57e937ebdd), use the client to log consensus state changes. This will also be helpful as an example/extension-point for developers that are extending the client to add their own functionality.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -1,7 +1,8 @@
+use futures::StreamExt;
 use std::time::Duration;
 
 pub use nimiq::{
-    client::{Client, Consensus},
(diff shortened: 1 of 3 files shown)


sisou:
Can you explain why this was closed? Was it replaced by anything?

brunoffranca:
We were discussing it on the core call and we felt it was unnecessary. You can talk to Sebastian or Philipp. Personally I'm agnostic about this PR.

PR #273, opened 2021-07-28
Add Dockerfile for an Alpine-based image

sisou:
This Dockerfile has the advantage that it does not pull down the latest nightly build every day. It only updates when rust-stable updates (and downloads the latest nightly at that point in the build part of the Dockerfile, which is cached).

### DIFF
--- a/build_alpine.Dockerfile
+++ b/build_alpine.Dockerfile
@@ -1,0 +1,60 @@
+# syntax = docker/dockerfile:1.2
+
+# Builds a light albatross image from sources.
+# Requires Docker buildkit.
+
+ARG RUST_IMAGE=rust:1-alpine
+FROM $RUST_IMAGE AS builder
+
+# Switch to Rust nightly
+RUN rustup update nightly; rustup default nightly;
+
+# Fetch dependencies.
+RUN apk update && apk add musl-dev libressl-dev protoc
+
+# Copy sources.
+COPY . /build
+WORKDIR /build
+
+# Build.
+RUN \
+  --mount=type=cache,target=/build/target \
+  # See https://doc.rust-lang.org/cargo/guide/cargo-home.html
+  --mount=type=cache,target=/usr/local/cargo/registry \
+  --mount=type=cache,target=/usr/local/cargo/git \
+  cargo build --bin nimiq-client && cp /build/target/debug/nimiq-client /build/
+
+# This is where the final image is created
+FROM alpine
+
+# Bash is required to run the start scripts
+RUN apk add --no-cache bash
+
+# Run as unprivileged user.
+RUN adduser --disabled-password --home /home/nimiq --shell /bin/bash --uid 1001 nimiq
+USER nimiq
+
+WORKDIR /home/nimiq
+
+# Create .nimiq, so that it has the right permissions. docker-compose might want to create this
+# because we want to use this as a volume, but we don't want this directory to be owned by root.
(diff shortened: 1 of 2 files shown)


build_alpine.Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+COPY . /build
+WORKDIR /build
+
+# Build.
+RUN \
+  --mount=type=cache,target=/build/target \
+  # See https://doc.rust-lang.org/cargo/guide/cargo-home.html
+  --mount=type=cache,target=/usr/local/cargo/registry \
+  --mount=type=cache,target=/usr/local/cargo/git \
+  cargo build --bin nimiq-client && cp /build/target/debug/nimiq-client /build/
```
riptl, line 25:
```suggestion
  cargo build --bin nimiq-client --release && cp /build/target/release/nimiq-client /build/
```
Would it be fair to create a release build?

sisou: @jeffesquivels I remember how in the now dismissed commits where you changed the dockerfiles, you had a `BUILD` env variable here to set debug or release build. Do you think this is still a good idea? If not set, it would default to debug.

riptl: @sisou That would be ideal. Using the ARG instruction allows for `docker build . --build-arg ENV=release`.
https://docs.docker.com/engine/reference/builder/#arg

Something like
```dockerfile
ARG ENV=debug
RUN cargo build --profile $ENV
```

jeffesquivels:
> This Dockerfile has the advantage that it does not pull down the latest nightly build every day. It only updates when rust-stable updates (and downloads the latest nightly at that point in the build part of the Dockerfile, which is cached).

Yes, that makes a lot of sense (although the process will only pull down the image from docker hub if `--pull` is used with the build command, like this: `docker buildx build . -f build_alpine.Dockerfile --pull -t core-rs --progress=plain`). I will migrate the Ubuntu dockerfile to use one of the debian images they provide that follows this same update policy. Thanks!

build_alpine.Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+# Requires Docker buildkit.
+
+ARG RUST_IMAGE=rust:1-alpine
+FROM $RUST_IMAGE AS builder
+
+# Switch to Rust nightly
+RUN rustup update nightly; rustup default nightly;
+
+# Fetch dependencies.
+RUN apk update && apk add musl-dev libressl-dev protoc
```
jeffesquivels: I'm running into [this bug](https://gitlab.alpinelinux.org/alpine/aports/-/issues/12763) when executing this line. It should be fixed [`soonish`](https://gitlab.alpinelinux.org/alpine/aports/-/issues/12763#note_172090) (at the alpine level, at least) but we could also try to fix it with something similar to [this](https://gitlab.alpinelinux.org/alpine/aports/-/issues/12763#note_163217) (and then change it back when the issue is resolved).

jeffesquivels: It seems changing this to `RUN apk update && apk add musl-dev libretls-dev protoc` solves the issue... is there a specific reason why we're using `libressl-dev` instead of `libretls-dev`?

sisou: Nope, no reason other than that's the one I found first.

jeffesquivels: Ok, in that case, let's do that change and rebase and then this would be ready for merging :-)

PR #282, opened 2021-08-07
Fix/seed1

maestroi:
## Pull request checklist

remove of nimiq-peer key from seed 1, this made the chain crash.
nimiq-peer key is not needed because its created on startup.
>...


sisou:
Did you try running the config with the changes from https://github.com/nimiq/core-rs-albatross/pull/280 ?

jeffesquivels:
@sisou #280 is what prompted this PR, it seems the keys we have in the `docker-compose.yml` are either not in the right hex representation or invalid altogether, this PR fixes that and it made sense to remove them because right now we're not checking for the `peer_id` when connecting to the seed nodes (this will change in the future and at that point we can add some valid ones).

PR #284, opened 2021-08-09
Fix standalone compilation for several crates

jsdanielh:
Fix standalone compilation for crates:
- lib
- network-interface
- tendermint

## Pull request checklist

### DIFF
--- a/lib/src/client.rs
+++ b/lib/src/client.rs
@@ -12,7 +12,6 @@ use nimiq_network_libp2p::{
     Config as NetworkConfig, Network,
 };
 use nimiq_utils::time::OffsetTime;
-
 #[cfg(feature = "validator")]
 use nimiq_validator::validator::Validator as AbstractValidator;
 #[cfg(feature = "validator")]
@@ -28,6 +27,7 @@ use nimiq_network_libp2p::Multiaddr;
 /// Alias for the Consensus and Validator specialized over libp2p network
 pub type Consensus = AbstractConsensus<Network>;
 pub type ConsensusProxy = AbstractConsensusProxy<Network>;
+#[cfg(feature = "validator")]
 pub type Validator = AbstractValidator<Network, ValidatorNetworkImpl<Network>>;
 
 /// Holds references to the relevant structs. This is then Arc'd in `Client` and a nice API is
@@ -55,9 +55,7 @@ pub(crate) struct ClientInner {
 }
 
 impl ClientInner {
-    async fn from_config(
-        config: ClientConfig,
-    ) -> Result<(Self, Consensus, Option<Validator>), Error> {
+    async fn from_config(config: ClientConfig) -> Result<Client, Error> {
         // Get network info (i.e. which specific blokchain we're on)
         if !config.network_id.is_albatross() {
             return Err(Error::config_error(&format!(
@@ -206,17 +204,18 @@ impl ClientInner {
         network.listen_on(config.network.listen_addresses).await;
         network.start_connecting().await;
 
-        Ok((
-            ClientInner {
+        Ok(Client {
+            inner: Arc::new(ClientInner {
                 environment,
                 network,
                 consensus: consensus.proxy(),
                 #[cfg(feature = "wallet")]
                 wallet_store,
-            },
-            consensus,
-            validator,
-        ))
+            }),
+            consensus: Some(consensus),
+            #[cfg(feature = "validator")]
+            validator: validator,
+        })
     }
 }
 
@@ -241,17 +240,13 @@ impl ClientInner {
 pub struct Client {
     inner: Arc<ClientInner>,
     consensus: Option<Consensus>,
+    #[cfg(feature = "validator")]
     validator: Option<Validator>,
 }
 
 impl Client {
     pub async fn from_config(config: ClientConfig) -> Result<Self, Error> {
-        let (inner, consensus, validator) = ClientInner::from_config(config).await?;
-        Ok(Client {
-            inner: Arc::new(inner),
-            consensus: Some(consensus),
-            validator,
-        })
+        ClientInner::from_config(config).await
     }
 
     pub fn consensus(&mut self) -> Option<Consensus> {
--- a/lib/src/config/config.rs
+++ b/lib/src/config/config.rs
@@ -7,8 +7,8 @@ use std::{
 use derive_builder::Builder;
 use strum_macros::Display;
 
-#[cfg(feature = "validator")]
 use beserial::Deserialize;
+#[cfg(feature = "validator")]
 use nimiq_bls::{KeyPair as BlsKeyPair, SecretKey as BlsSecretKey};
 use nimiq_database::{
     lmdb::{open as LmdbFlags, LmdbEnvironment},
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -58,7 +58,7 @@ nimiq-network-interface = { path = "../network-interface" }
 nimiq-peer-address = { path = "../peer-address" }
 nimiq-primitives = { path = "../primitives", features = ["networks"] }
 nimiq-rpc-server = { path = "../rpc-server", optional = true }
-nimiq-utils = { path = "../utils", features = ["time"] }
+nimiq-utils = { path = "../utils", features = ["time", "key-store"] }
 nimiq-validator = { path = "../validator", optional = true }
 nimiq-validator-network = { path = "../validator-network", optional = true }
 nimiq-wallet = { path = "../wallet", optional = true }
--- a/lib/src/error.rs
+++ b/lib/src/error.rs
@@ -1,7 +1,7 @@
 use thiserror::Error;
 
 use nimiq_database::volatile::VolatileDatabaseError;
-#[cfg(feature = "rpc-server")]
+
 // #[cfg(feature = "validator")]
 // use validator::error::Error as ValidatorError;
 #[derive(Error, Debug)]
--- a/network-interface/Cargo.toml
+++ b/network-interface/Cargo.toml
@@ -23,7 +23,7 @@ futures = "0.3"
 parking_lot = "0.11"
 thiserror = "1.0"
 tokio = { version = "1.9", features = ["macros", "rt", "sync", "time"] }
-tokio-stream = "0.1"
+tokio-stream = { version = "0.1", features = ["default", "sync"] }
 
 # Used during debugging
 pretty-hex = "0.2"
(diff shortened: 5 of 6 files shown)


jeffesquivels:
It worked for me for `network-interface` and `tendermint`, but `lib` still failed to compile standalone even after I `cargo clean`ed... is it working for you? May be someone else can try it to discard that the issue I'm seeing is related to my environment?

sisou:
I'm seeing the same. `nimiq-lib` does not compile.

<details>
<summary>Build log</summary>

```
   Compiling nimiq-lib v0.1.0 (/home/soeren/nimiq/core-rs-albatross/lib)
error[E0432]: unresolved import `nimiq_bls`
  --> lib/src/config/config.rs:12:5
   |
12 | use nimiq_bls::{KeyPair as BlsKeyPair, SecretKey as BlsSecretKey};
   |     ^^^^^^^^^ use of undeclared crate or module `nimiq_bls`

error[E0412]: cannot find type `AbstractValidator` in this scope
  --> lib/src/client.rs:31:22
   |
31 | pub type Validator = AbstractValidator<Network, ValidatorNetworkImpl<Network>>;
   |                      ^^^^^^^^^^^^^^^^^ not found in this scope

error[E0412]: cannot find type `ValidatorNetworkImpl` in this scope
  --> lib/src/client.rs:31:49
   |
31 | pub type Validator = AbstractValidator<Network, ValidatorNetworkImpl<Network>>;
   |                                                 ^^^^^^^^^^^^^^^^^^^^ not found in this scope

error[E0425]: cannot find value `validator` in this scope
   --> lib/src/client.rs:218:13
    |
218 |             validator,
    |             ^^^^^^^^^ not found in this scope

error[E0573]: expected type, found derive macro `Error`
  --> lib/src/error.rs:44:6
   |
44 | impl Error {
   |      ^^^^^ not a type
   |
help: consider importing one of these items instead
   |
1  | use core::fmt::Error;
   |
1  | use crate::config::config_file::serialization::Error;
   |
1  | use log::kv::Error;
   |
1  | use nimiq_consensus::Error;
   |
     and 16 other candidates

error[E0573]: expected type, found derive macro `Error`
  --> lib/src/error.rs:56:38
   |
56 | impl From<VolatileDatabaseError> for Error {
   |                                      ^^^^^ not a type
   |
help: consider importing one of these items instead
   |
1  | use core::fmt::Error;
   |
1  | use crate::config::config_file::serialization::Error;
   |
1  | use log::kv::Error;
   |
1  | use nimiq_consensus::Error;
   |
     and 16 other candidates

error[E0603]: derive macro import `Error` is private
   --> lib/src/client.rs:24:19
    |
24  | use crate::error::Error;
    |                   ^^^^^ private derive macro import
    |
note: the derive macro import `Error` is defined here...
   --> lib/src/error.rs:1:5
    |
1   | use thiserror::Error;
    |     ^^^^^^^^^^^^^^^^
note: ...and refers to the derive macro `Error` which is defined here
   --> /home/soeren/.cargo/registry/src/github.com-1ecc6299db9ec823/thiserror-1.0.26/src/lib.rs:197:9
    |
197 | pub use thiserror_impl::*;
    |         ^^^^^^^^^^^^^^^^^ consider importing it directly

error[E0603]: derive macro import `Error` is private
   --> lib/src/config/config.rs:33:12
    |
33  |     error::Error,
    |            ^^^^^ private derive macro import
    |
note: the derive macro import `Error` is defined here...
   --> lib/src/error.rs:1:5
    |
1   | use thiserror::Error;
    |     ^^^^^^^^^^^^^^^^
note: ...and refers to the derive macro `Error` which is defined here
   --> /home/soeren/.cargo/registry/src/github.com-1ecc6299db9ec823/thiserror-1.0.26/src/lib.rs:197:9
    |
197 | pub use thiserror_impl::*;
    |         ^^^^^^^^^^^^^^^^^ consider importing it directly

error[E0603]: derive macro import `Error` is private
   --> lib/src/config/config_file/mod.rs:20:12
    |
20  |     error::Error,
    |            ^^^^^ private derive macro import
    |
note: the derive macro import `Error` is defined here...
   --> lib/src/error.rs:1:5
    |
1   | use thiserror::Error;
    |     ^^^^^^^^^^^^^^^^
note: ...and refers to the derive macro `Error` which is defined here
   --> /home/soeren/.cargo/registry/src/github.com-1ecc6299db9ec823/thiserror-1.0.26/src/lib.rs:197:9
    |
197 | pub use thiserror_impl::*;
    |         ^^^^^^^^^^^^^^^^^ consider importing it directly

error[E0603]: derive macro import `Error` is private
   --> lib/src/prelude.rs:7:12
    |
7   |     error::Error,
    |            ^^^^^ private derive macro import
    |
note: the derive macro import `Error` is defined here...
   --> lib/src/error.rs:1:5
    |
1   | use thiserror::Error;
    |     ^^^^^^^^^^^^^^^^
note: ...and refers to the derive macro `Error` which is defined here
   --> /home/soeren/.cargo/registry/src/github.com-1ecc6299db9ec823/thiserror-1.0.26/src/lib.rs:197:9
    |
197 | pub use thiserror_impl::*;
    |         ^^^^^^^^^^^^^^^^^ consider importing it directly

warning: unused import: `net::IpAddr`
 --> lib/src/config/config.rs:2:5
  |
2 |     net::IpAddr,
  |     ^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: unused import: `consts`
  --> lib/src/config/config.rs:30:9
   |
30 |         consts, paths,
   |         ^^^^^^

error[E0282]: type annotations needed
   --> lib/src/client.rs:249:45
    |
249 |         let (inner, consensus, validator) = ClientInner::from_config(config).await?;
    |                                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type

error[E0282]: type annotations needed
   --> lib/src/config/config.rs:577:9
    |
577 |         Client::from_config(self).await
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ cannot infer type

error[E0599]: no variant or associated item named `deserialize_from_vec` found for enum `nimiq_network_libp2p::Keypair` in the current scope
   --> lib/src/config/config.rs:382:46
    |
382 | ...                   IdentityKeypair::deserialize_from_vec(&hex::decode(key).unwrap())
    |                                        ^^^^^^^^^^^^^^^^^^^^ variant or associated item not found in `nimiq_network_libp2p::Keypair`
    |
    = help: items from traits can only be used if the trait is in scope
    = note: the following trait is implemented but not in scope; perhaps add a `use` for it:
            `use beserial::Deserialize;`

warning: unused import: `string::ToString`
 --> lib/src/config/config.rs:4:5
  |
4 |     string::ToString,
  |     ^^^^^^^^^^^^^^^^

Some errors have detailed explanations: E0282, E0412, E0425, E0432, E0573, E0599, E0603.
For more information about an error, try `rustc --explain E0282`.
warning: `nimiq-lib` (lib) generated 3 warnings
error: could not compile `nimiq-lib` due to 13 previous errors; 3 warnings emitted
```
</details>

jsdanielh:
Yeah, still need to work some more issues in `lib`

jsdanielh:
Thanks to @jeffesquivels we were able to fix the extra issues in `lib`.
There is an extra issue left in `rpc-interface` that we couldn't figure out but we will track it in #285.

tendermint/Cargo.toml
```diff
@@ -16,7 +16,7 @@ log = "0.4"
 
 nimiq-block = { path = "../primitives/block" }
 nimiq-hash = { path = "../hash" }
-nimiq-primitives = { path = "../primitives", features = ["policy"] }
+nimiq-primitives = { path = "../primitives", features = ["policy", "validators"] }
```
jeffesquivels: It seems to me that this could be better fixed at a lower layer:

AFAICT, the `tendermint` crate currently can't compile standalone because it [depends on the `block` crate](https://github.com/nimiq/core-rs-albatross/blob/albatross/tendermint/Cargo.toml#L17), and the `block` crate is the one that actually requires the `validators` feature to be enabled on the `primitives` crate non-conditionally (f.e. see [here](https://github.com/nimiq/core-rs-albatross/blob/albatross/primitives/block/src/macro_block.rs#L67) ), so the problem could be fixed (for both, the `block` and `tendermint` crates) by enabling the `validators` when the `block` crate depends on the `primitives` crate [here](https://github.com/nimiq/core-rs-albatross/blob/albatross/primitives/block/Cargo.toml#L34) instead of doing it here at a higher layer.

Wdyt?

lib/src/config/config.rs
```diff
@@ -574,7 +574,8 @@ impl ClientConfig {
     /// Instantiates the Nimiq client from this configuration
     ///
     pub async fn instantiate_client(self) -> Result<Client, Error> {
-        Client::from_config(self).await
+        let client: Result<Client, Error> = Client::from_config(self).await;
+        client
```
jeffesquivels: This change seems to be unnecessary on my environment, it was failing for you if you didn't specify the type for `client` explicitly, right?

lib/src/client.rs
```diff
@@ -241,17 +240,14 @@ impl ClientInner {
...
 impl Client {
     pub async fn from_config(config: ClientConfig) -> Result<Self, Error> {
-        let (inner, consensus, validator) = ClientInner::from_config(config).await?;
-        Ok(Client {
-            inner: Arc::new(inner),
-            consensus: Some(consensus),
-            validator,
-        })
+        let client: Result<Self, Error> = ClientInner::from_config(config).await;
+        client
```
jeffesquivels: This could be written more idiomatically as `ClientInner::from_config(config).await` and it works on my environment, did you had issues doing it that way?

lib/src/client.rs
```diff
@@ -55,9 +55,7 @@ pub(crate) struct ClientInner {
 }
 
 impl ClientInner {
-    async fn from_config(
-        config: ClientConfig,
-    ) -> Result<(Self, Consensus, Option<Validator>), Error> {
+    async fn from_config(config: ClientConfig) -> Result<Client, Error> {
```
jeffesquivels, line 58: I know this works, but I'm still not sure I like it semantically (i.e. a `ClientInner` method returning a `Client` which is supposed to be a wrapper around `ClientInner`). I think it would make sense to discuss this on the next core call because I'm also not sure why we're using this `Client`/`ClientInner` pattern here.

jeffesquivels (changes requested):
There's one minor change and one more semantic issue, but in general it looks good.

tendermint/Cargo.toml
jsdanielh: You're right. I already changed it.

lib/src/config/config.rs
jsdanielh: I thought so but now that I changed it, it worked

lib/src/client.rs
jsdanielh: I thought so but now that I changed it, it worked

lib/src/client.rs
jeffesquivels: For the record: we discussed it in the call and it's fine to leave it like this for the time being.

jeffesquivels (approved):
LGTM, let's merge it

PR #296, opened 2021-08-27
New block queue tests and event

viquezclaudio:
In this test multiple micro blocks are generated
and are sent, randomly, out of order.
The test verifies that all blocks are produced
properly

### DIFF
--- a/consensus/src/sync/block_queue.rs
+++ b/consensus/src/sync/block_queue.rs
@@ -50,6 +50,7 @@ pub enum BlockQueueEvent {
     AcceptedAnnouncedBlock(Blake2bHash),
     AcceptedBufferedBlock(Blake2bHash, usize),
     ReceivedMissingBlocks(Blake2bHash, usize),
+    RejectedBlock(Blake2bHash),
 }
 
 #[derive(Clone, Debug)]
@@ -432,6 +433,21 @@ impl<N: Network> Stream for Inner<N> {
                         )));
                     }
                 }
+
+                PushOpResult::Head(Err(result), hash) => {
+                    //If there was a blockchain push error, we remove the block from the pending blocks
+                    log::trace!("Head push operation failed because of {}", result);
+                    self.pending_blocks.remove(&hash);
+                    return Poll::Ready(Some(BlockQueueEvent::RejectedBlock(hash)));
+                }
+
+                PushOpResult::Buffered(Err(result), hash) => {
+                    //If there was a blockchain push error, we remove the block from the pending blocks
+                    log::trace!("Buffered push operation failed because of {}", result);
+                    self.pending_blocks.remove(&hash);
+                    return Poll::Ready(Some(BlockQueueEvent::RejectedBlock(hash)));
+                }
+
                 _ => {}
             };
         }
--- a/consensus/tests/block_queue.rs
+++ b/consensus/tests/block_queue.rs
@@ -1,3 +1,4 @@
+use rand::Rng;
 use std::marker::PhantomData;
 use std::sync::Weak;
 use std::{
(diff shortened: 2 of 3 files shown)


consensus/tests/block_queue.rs
```diff
@@ -221,6 +222,91 @@ async fn send_two_micro_blocks_out_of_order() {
...
+            .unwrap();
+
+        // run the block_queue one iteration, i.e. until it processed one block
+        block_queue.poll_next_unpin(&mut Context::from_waker(noop_waker_ref()));
+    }
+
+    // All blocks should be buffered
+    assert_eq!(blockchain1.block_number(), 0);
+
+    //Obtain the buffered blocks
```
vlvrd: Space after the slashes

consensus/tests/block_queue.rs
```diff
@@ -221,6 +222,91 @@ async fn send_two_micro_blocks_out_of_order() {
...
+    assert_eq!(buffered_blocks.len() as u64, n_Blocks - 1);
+
+    // now send block1 to fill the gap
+    tx.send((blocks[0].clone(), mock_id)).await.unwrap();
+
+    for _i in 0..n_Blocks {
+        block_queue.next().await;
+    }
+
+    //Verify all blocks except the genesis
```
vlvrd: Space after the slashes

consensus/tests/block_queue.rs
```diff
@@ -221,6 +222,91 @@ async fn send_two_micro_blocks_out_of_order() {
...
+    for i in 1..=n_Blocks {
+        assert_eq!(
+            blockchain1.get_block_at(i as u32, true, None).unwrap(),
+            ordered_blocks[(i - 1) as usize]
+        );
+    }
+
+    //No blocks buffered
```
vlvrd: Space after the slashes

vlvrd (approved):
LGTM, just a couple of cosmetic comments 👍

consensus/tests/block_queue.rs
```diff
@@ -221,6 +222,91 @@ async fn send_two_micro_blocks_out_of_order() {
...
+    assert_eq!(blockchain1.block_number(), 0);
+
+    //Obtain the buffered blocks
+    let buffered_blocks = block_queue.buffered_blocks().collect::<Vec<_>>();
+    assert_eq!(buffered_blocks.len() as u64, n_Blocks - 1);
+
+    // now send block1 to fill the gap
+    tx.send((blocks[0].clone(), mock_id)).await.unwrap();
+
+    for _i in 0..n_Blocks {
```
vlvrd: Suggestion: Since you're not planning on using the `_i` variable, just use the underscore (`_`) to achieve the same effect

consensus/src/sync/block_queue.rs
```diff
@@ -432,6 +433,18 @@ impl<N: Network> Stream for Inner<N> {
                         )));
                     }
                 }
+
+                // TODO this case should be handled
+                PushOpResult::Head(Err(result), hash) => {}
+
+                PushOpResult::Buffered(Err(result), hash) => {
+                    self.pending_blocks.remove(&hash);
```
viquezclaudio, line 447: I think this was a bug in the code, because, previously, the error cases were not handled, so if a block was rejected by the blockchain it remained inside `pending_blocks`and nobody cleared that invalid block

paberr:
One comment related to the code, but not related to your changes:
Line 309 in the `block_queue.rs` file mentions `// TODO Ban peer`.
However, this code is also triggered when the peer sends us an "orphan" block, which might simply mean we haven't synced yet.
Banning the peer is thus not always a good idea here!

EDIT: Ok, I'm not sure when exactly we're calling the push_block function here. It might be that we only do it if we expect the block to be accepted. In this case, banning is an option, of course. I suspect that this is the case as we only call it depending on the block number.
In that case, we should properly document it. :)

PR #299, opened 2021-08-31
Upgrade several dependencies

### LINKED ISSUE #274: Update arkworks dependencies to 0.3
Because of a bug in the 0.2 version, we should update all arkworks dependencies to 0.3.
[https://twitter.com/arkworks_rs/status/1412886038272352256?s=21](https://twitter.com/arkworks_rs/status/1412886038272352256?s=21)
Note: The 0.3 version uses rand-0.8 while some of our crates use rand-0.7. This might create some problems. [https://github.com/arkworks-rs/std/blob/master/Cargo.toml](https://github.com/arkworks-rs/std/blob/master/Cargo.toml)

jsdanielh:
- Update dalek cryptography dependencies:
  - Update curve25519-dalek to 3.
  - Use ed25519-zebra instead of ed25519-dalek since the latter is no
    longer actively maintained and it's the only one that solves the
    validation issues regarding batch validation of signatures.
  - Also this change is updating the sha2 dependency from 0.8 to
    0.9.

- Update the version of arkworks dependencies
  Update the version of the arkworks dependencies for all subcrates
  from 0.2 to 0.3.
  This closes #274.

- Update the version of the rand dependency
  Update the version of the rand dependency for all subcrates.
  This also includes:
   - Update of rand_chacha used in the build-tools crate.
   - Update of rand_core used in the utils crate.

## Pull request checklist


brunoffranca:
I wouldn't say that ed25519-dalek and curve25519-dalek are not being maintained. It's more that they are stable. They are crypto libraries, so it's not like they will add any new features.
However, if you really want to change, change it to ZcashFoundation/ed25519-zebra. That one is used in the Zcash blockchain so it's guaranteed to be actively maintained. And it solves some validation issues that most ed25519 libraries have.

vlvrd (dismissed):

jsdanielh:
> I wouldn't say that ed25519-dalek and curve25519-dalek are not being maintained. It's more that they are stable. They are crypto libraries, so it's not like they will add any new features.

Mmm, ed25519-dalek looks to me not maintained in the sense that there has been almost a year since the maintainer doesn't reply to issues or pull requests to the point that people have already started to fork it.
Right now, the reason why I would like to change it is to be able to use a single version of rand and rand_core across our code such that we don't get errors because of mixed versions of those dependencies. I see already [a PR](https://github.com/dalek-cryptography/ed25519-dalek/pull/160) trying to change it in the official repository but there hasn't been any response from the maintainers.

> However, if you really want to change, change it to ZcashFoundation/ed25519-zebra. That one is used in the Zcash blockchain so it's guaranteed to be actively maintained. And it solves some validation issues that most ed25519 libraries have.

Yeah, I took a look at the `ZcashFoundation/ed25519-zebra` but I found the following situation:
- `ZcashFoundation/ed25519-zebra` -> depends on `curve25519-dalek = "3"` & `rand_core = "0.6"`
- `curve25519-dalek` -> depends on `rand_core = "0.5"` (which according to [this comment](https://github.com/ZcashFoundation/ed25519-zebra/pull/44#issuecomment-810348332) it doesn't affect `ZcashFoundation/ed25519-zebra`).
- We would depend on both: `ZcashFoundation/ed25519-zebra` & `curve25519-dalek`, having us to have inner dependencies on 2 different versions of `rand_core`.

That's why I took a look at other forks of the dalek crates and found the ones maintained by _novifinancial_ where they were able to keep both `ed25519-dalek-fiat` and `curve25519-dalek-fiat` using the same set of dependencies.

Do you still think we should move to `ZcashFoundation/ed25519-zebra`?

brunoffranca:
I did some digging and apparently the dalek-cryptography Github org was taken over ([here](https://www.reddit.com/r/rust/comments/lcsg0u/ownership_and_maintenance_struggles_in/) and [here](https://twitter.com/hdevalence/status/1356778933677563905)). So it's probably safe to assume that it will become unmaintained (or at least unreliable).
Using Novi's libraries also seem problematic to me. First, they have changes that are specific to their needs. Second, Novi (previously Calibra) is the wallet of the infamous Facebook cryptocurrency Diem (previously Libra). Given the problems that that project has faced so far, I doubt they will last much longer.
Here's what we can do instead. For curve25519 use `zkcrypto/curve25519-dalek-ng`. That one is mantained by one of the original maintainers of `curve25519-dalek` and has all the necessary updates (in fact Novi's repo just cherry-picked commits from this repo). For ed25519 the best option still seems to be `ZcashFoundation/ed25519-zebra`. It's the only one that solves the validation issues (which is very much needed if we want to do batch validation of signatures) and zkcrypto doesn't have a ed25519 library.

brunoffranca:
Now that I'm thinking about it, I'm not sure if we need the change the curve25519 dependency on `ZcashFoundation/ed25519-zebra`. It might work as is. Give it a try.

brunoffranca:
This still has the wrong dalek crates.

jsdanielh:
> This still has the wrong dalek crates.

Yeah, I still need to work on this. Plan is to update to `ZcashFoundation/ed25519-zebra` but that is WIP

brunoffranca:
Seems good to me. But why didn't you use the curve25519-dalek-ng crate instead of the curve25519-dalek?

brunoffranca (approved):

jsdanielh:
> Seems good to me. But why didn't you use the curve25519-dalek-ng crate instead of the curve25519-dalek?

It was to be in sync with what `ed25519-zebra` [uses](https://github.com/ZcashFoundation/ed25519-zebra/blob/main/Cargo.toml#L20).

brunoffranca:
Fair enough.

Issue #281, opened 2021-08-07
Improve selection of slot owners


brunoffranca:
The function `get_slot_owner_number_at` could be improved. Randomly picking slots until you find a non-disabled one is certainly not the most efficient way of doing this. See code here (https://github.com/nimiq/core-rs-albatross/blob/albatross/blockchain/src/abstract_blockchain.rs#L165-L197).

PR #319, opened 2021-09-13
Improve selection of slot owners

brunoffranca:
## Pull request checklist

#### This fixes issue #281.

### DIFF
--- a/blockchain/src/abstract_blockchain.rs
+++ b/blockchain/src/abstract_blockchain.rs
@@ -183,16 +183,32 @@ pub trait AbstractBlockchain {
 
         // Check if all slots are disabled. In this case, we will accept any slot, since we want the
         // chain to progress.
-        let all_disabled = disabled_slots.len() == policy::SLOTS as usize;
+        if disabled_slots.len() == policy::SLOTS as usize {
+            // Sample a random slot number.
+            return rng.next_u64_max(policy::SLOTS as u64) as u16;
+        }
 
-        // Sample until we find a slot that is not slashed
-        loop {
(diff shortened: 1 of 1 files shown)


blockchain/src/abstract_blockchain.rs
```diff
@@ -183,16 +183,32 @@ pub trait AbstractBlockchain {
...
+        let mut r =
+            rng.next_u64_max((policy::SLOTS as usize - disabled_slots.len()) as u64) as usize;
 
-        // Sample until we find a slot that is not slashed
-        loop {
-            let slot_number = rng.next_u64_max(policy::SLOTS as u64) as u16;
+        // Now we just iterate over all the slots until we find the r-th non-disabled slot.
+        let mut slot_number = 0;
+        for is_disabled in disabled_slots.iter_bits() {
+            if is_disabled {
```
sisou, line 198: I don't understand this algorithm. Isn't _every_ `is_disabled` true? After all you are iterating over the list of disabled slots.

And if disabled slots are empty? Then it runs into `unreachable!()`, no?

sisou: Ok, I just ran it in the Docker setup, and against my expectations it works. And validators in the second epoch are still random, even if no slots where disabled.

I still don't understand how, though.

brunoffranca: BitSets work a bit differently than you think. They basically store a list of booleans (it's actually stored as a vector of u64, but you get the point). When you call len(), it gives you the number of elements that are set to true. When you call iter_bits(), you iterate over all the elements.

PR #322, opened 2021-09-15
Remove interior mutability primitives from Blockchain, use outer RwLock

### LINKED ISSUE #320: Blockchain mutabillity
The interior mutability of the Blockchain struct is problematic in some cases and in addition to that more error prone.

It is easy to forget to `blockchain.lock()` when nothing is forcing you to, as everything will just work as expected, except that it not necessarily does. The state of the data structure can change between calls to distinct methods and since there is no persistent read lock held the data is potentially inconsistent.

The internal locking of `Blockchain` should thus be removed in favour of an outside `Arc<RwLock<Blockchain>>` to make acquiring a consistent state much more explicit and less error prone.

The new Mempool (#305) should already build on this as it is going to be one of the components holding read locks on it.

nibhar:
## Pull request checklist

This PR resolves #320.

The `Blockchain` struct itself no longer uses any `RwLock` internally. The locking now happens on the outside as `Arc<Blockchain>` is changed to `Arc<RwLock<Blockchain>>` everywhere.

The only operations requiring write are subscribing a listener to the observable as well as `Blockchain::push` and `Blockchain::push_history_sync`. 
Both the `push*` functions do work on an `RwLockUpgradableReadGuard` such that the time spend with a held `RwLockWriteGuard` is minimised, while giving the option of having held the lock prior to the `push*` calls.  
The subscribing of listeners does happen mostly on startup and should not be too bad, however there are plans to remove the Observable here as we did throughout the project.

Neither the obsolete `nimiq-network` nor the currently unused and `metrics-server` were updated.

Important to note is that the `nimiq-nano-blockchain` is untouched by this as well, as I am unsure of whether or not it is actually needed.

### DIFF
--- a/blockchain/src/blockchain/blockchain.rs
+++ b/blockchain/src/blockchain/blockchain.rs
@@ -1,7 +1,5 @@
 use std::sync::Arc;
 
-use parking_lot::{Mutex, RwLock};
-
 use nimiq_account::{Account, Accounts};
 use nimiq_block::Block;
 use nimiq_database::{Environment, WriteTransaction};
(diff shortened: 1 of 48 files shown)


blockchain/src/blockchain/blockchain.rs
```diff
@@ -34,17 +32,15 @@ pub struct Blockchain {
...
-    pub chain_store: Arc<ChainStore>,
+    pub chain_store: ChainStore,
     // The history store is a database containing all of the history trees and transactions.
-    pub history_store: Arc<HistoryStore>,
+    pub history_store: HistoryStore,
     // The current state of the blockchain.
-    pub state: RwLock<BlockchainState>,
-    // A write lock for the blockchain. Guarantees that only one thread writes to it at a time.
-    pub(crate) push_lock: Mutex<()>,
+    pub state: BlockchainState,
```
paberr, line 43: Actually, the BlockchainState was only needed for the locking. So, now we could simply remove it. But I'm equally fine with keeping it as it encapsulates the state quite nicely.

paberr (approved):
Generally looks good to me. I mainly checked the push function and skipped most of the test-related code.

blockchain/src/blockchain/blockchain.rs
nibhar: I think keeping it makes sense as you can easily borrow a reference to it.

PR #345, opened 2021-09-22
RPC: Add method isAccountImported

redmaner:
This PR adds a new RPC method to the RPC interface / server "isAccountImported" which returns whether an account with the given address is loaded into the wallet store. This is a follow up to PR #327


vlvrd (commented):
Thanks a lot @redmaner, LGTM 👍 Did you test it in the dev setup?

redmaner:
@fiaxh Yes I tested it in the dev setup. Tried the function before and after importing an account. Returns `false` and then `true` as expected.

PR #350, opened 2021-09-27
RPC: Add method getBalance

redmaner:

This PR adds a new method to the RPC interface and server: `getBalance` which returns the balance of an account. Important to note is that this currently only covers accounts, not staking contract actors like stakers and validators. I also don't know if it is is preferred to have those actors in this method, although the account protocol seems to leave room for that [as seen here](https://github.com/nimiq/core-rs-albatross/blob/albatross/primitives/account/src/account.rs#L46). I tested this in the docker setup, and for normal accounts works fine.


sisou:
Is there a reason why you cannot (or don't want to) use the `getAccount` method instead?

redmaner:
> Is there a reason why you cannot (or don't want to) use the `getAccount` method instead?

Good point. I have force updated my branch by using the `getAccount` method directly. Think that is indeed a better approach. Also rebuilded and retested it to make sure these changes work.

brunoffranca:
Replaced by #351. Sorry, redmaner!

PR #352, opened 2021-09-29
RPC fixes

### LINKED ISSUE #229: RPC - Bugs, missing methods, inconsistencies
This is a collection issue for bugs, missing methods and inconsistencies I found when testing the RPC server:

- [x] Method `blockByNumber` does not accept a `number` as the first argument, only stringified numbers
- [x] Methods `blockByNumber` and `blockByHash` require a second argument (`include transactions: bool`) - I think this should be optional and default to `false`
- [x] Naming inconsistency: `blockByNumber/Hash` vs `getTransactionByHash` - make `get` prefix consistent (either with or without)
- [x] Panic when requesting `getAccount` with the staking contract address: `thread 'tokio-runtime-worker' panicked at 'Failed to serialize return value: Error("the enum variant Account::Staking cannot be serialized", line: 0, column: 0)': /usr/local/cargo/git/checkouts/jsonrpc-9fb05bf7e43ac4a6/1d17aa2/nimiq-jsonrpc-server/src/lib.rs:513`
- [x] Missing method to request current `consensus` state
- [x] Missing method to request current `peerCount`, `peerList`, `peerState <peerId>`
- [x] Missing method to set `log` level
- [x] Missing method to set `constant`s
- [x] Missing method to retrieve `validatorKey`, `validatorId`
- [x] Missing method to retrieve `minFeePerByte`
- [x] Missing method to `unsubscribe`from subscriptions
- [x] `mempool` methods throw "not implemented" error
- [x] Properties of returned objects mix `camelCase` and `snake_case` within the same object (example: `Block { blockNumber, extra_data, parentHash, parent_election_hash }`
- [x] Accounts from `getAccount` are returned with the account type as a property instead of a value: `{"Basic":{"balance":168157712802}}` - the type should rather be the value of a `type` property
- [x] Interacting with a locked wallet panics instead of returning the error correctly: `thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: UnlockedWalletNotFound(Address("8314ae0f9e42e6d872c0bc07b09e2a278b1dfbae"))': rpc-server/src/dispatchers/consensus.rs:124`
- [x] `getTransactionByHash`, `getTransactionReceipt` not implemented
- [x] All transaction-creation/sending methods require a last `validityStartHeight` argument. I think this should be optional and default to the current head height
- [x] `validityStartHeight` argument for transaction creation/sending methods does not accept numbers, only stringified numbers
- [x] Closing the websocket after subscribing to headers does not close the stream on the server and leads to continuous errors: `ERROR nimiq_jsonrpc_server | Queue error: channel closed`
- [x] Creating a validator reactivation transactions fails (on `soeren/rpc` branch)

brunoffranca:
## Pull request checklist

Cleans up, refactors, adds some methods and documents the RPC crates.


rpc-interface/src/validator.rs
```diff
@@ -1,10 +1,18 @@
...
 )]
 #[async_trait]
 pub trait ValidatorInterface {
     type Error;
+
+    async fn get_address(&mut self) -> Result<Address, Self::Error>;
+
+    async fn get_warm_key(&mut self) -> Result<String, Self::Error>;
+
+    async fn get_hot_key(&mut self) -> Result<String, Self::Error>;
```
sisou, line 17: To properly namespace these methods (which are global), they should include the word `validator`. Eg. `get_validator_address`, `get_validator_warm_key`, `get_validator_hot_key`.

rpc-server/src/dispatchers/network.rs
```diff
@@ -23,22 +23,23 @@ impl NetworkDispatcher {
...
-        // TODO: Wait for connection pool implementation
-        todo!()
+        Ok(self.network.get_peers().len())
     }
 
-    async fn get_peer_state(&mut self, _peer_id: String) -> Result<Peer, Self::Error> {
-        // TODO: Wait for connection pool implementation
-        todo!()
+    /// Returns a list with the IDs of all our peers.
+    async fn get_peer_list(&mut self) -> Result<Vec<String>, Self::Error> {
```
sisou, line 37: Can there also be a method like `get_peer_by_id` where I can query the peer struct, to e.g. get its IP, type, etc.?

rpc-server/src/dispatchers/consensus.rs
```diff
@@ -36,6 +37,7 @@ impl ConsensusDispatcher {
         }
     }
 
+    /// Pushes the given transaction to the network.
     async fn push_transaction(&self, tx: Transaction) -> Result<Blake2bHash, Error> {
```
sisou: How is the transaction send as JSON? How does the object have to be structured to be unserializable?
For HEX transactions we have the method `send_raw_transaction`.

rpc-server/src/dispatchers/consensus.rs
```diff
@@ -54,13 +57,16 @@ impl ConsensusDispatcher {
             .get(address)
             .ok_or_else(|| Error::UnlockedWalletNotFound(address.clone()))?
             .key_pair
-            .clone()) // TODO: Avoid cloning
+            .clone())
     }
 
+    /// Returns the network ID for our current blockchain.
     fn network_id(&self) -> NetworkId {
```
sisou: All other methods to get data are prefixed with `get_`. Is there a reason this one isn't?

rpc-server/src/dispatchers/consensus.rs
```diff
@@ -93,7 +122,7 @@ impl ConsensusInterface for ConsensusDispatcher {
         validity_start_height: ValidityStartHeight,
     ) -> Result<String, Error> {
         let transaction = TransactionBuilder::new_simple(
-            &self.get_wallet_keypair(&wallet)?,
+            &&self.get_wallet_keypair(&wallet)?,
```
sisou: Double `&&` by accident? Clippy also complains about this.

rpc-server/src/dispatchers/consensus.rs
```diff
@@ -173,31 +218,50 @@ impl ConsensusInterface for ConsensusDispatcher {
...
+                staker_address,
+                value,
+                fee,
+                validity_start_height,
+            )
             .await?;
         self.send_raw_transaction(raw_tx).await
     }
 
+    /// Returns a serialized `update_staker` transaction. You can the transaction fee from a basic
```
sisou:
```suggestion
    /// Returns a serialized `update_staker` transaction. You can pay the transaction fee from a basic
```

sisou: This same word is missing on multiple comments.

rpc-server/src/dispatchers/consensus.rs
```diff
@@ -303,34 +427,38 @@ impl ConsensusInterface for ConsensusDispatcher {
...
         let raw_tx = self
-            .create_unstake_transaction(wallet, recipient, value, fee, validity_start_height)
+            .create_unstake_transaction(staker_wallet, recipient, value, fee, validity_start_height)
             .await?;
         self.send_raw_transaction(raw_tx).await
     }
 
+    /// Returns a serialized `new_validator` transaction. You need to provide the address of a basic
+    /// account (the sender wallet) to pay the transaction fee.
     async fn create_new_validator_transaction(
```
sisou: From which address is the 10k NIM deposit paid? Also from the sender, or from the `validator_wallet`? Can that be included in the comment?

rpc-server/src/dispatchers/consensus.rs
```diff
@@ -383,17 +515,20 @@ impl ConsensusInterface for ConsensusDispatcher {
         self.send_raw_transaction(raw_tx).await
     }
 
+    /// Returns a serialized `update_validator` transaction. You need to provide the address of a basic
+    /// account (the sender wallet) to pay the transaction fee.
     async fn create_update_validator_transaction(
```
sisou: This comment should also explain what JSON value to send for the `signal_data` to have the desired of the three effects.
I believe sending `null` means `None`, so no change. An empty string means "change the data to `None`". And a string obviously means "set the data to this".

rpc-server/src/dispatchers/consensus.rs
```diff
@@ -303,34 +427,38 @@ impl ConsensusInterface for ConsensusDispatcher {
...
     }
 
+    /// Returns a serialized `new_validator` transaction. You need to provide the address of a basic
+    /// account (the sender wallet) to pay the transaction fee.
     async fn create_new_validator_transaction(
         &mut self,
-        wallet: Address,
+        sender_wallet: Address,
+        validator_wallet: Address,
         warm_key: Address,
```
sisou: In the `update_validator` method, `warm_key` is named `new_warm_address`. I think `warm_address` fits better here, as we only need the address, not the private key.

rpc-server/src/dispatchers/blockchain.rs
```diff
@@ -291,12 +432,61 @@ impl BlockchainInterface for BlockchainDispatcher {
         Ok(active_validators)
     }
 
-    async fn get_validator(
+    /// Returns information about the currently slashed slots. This includes slots that lost rewards
+    /// and that were disabled.
+    async fn get_current_slashed_sets(&mut self) -> Result<SlashedSlots, Self::Error> {
```
sisou: Function name suggestion: `get_current_slashed_slots` or `get_current_slashed_set`, not `sets`.

Seeing that it returns `SlashedSlots`, that would speak for the first suggestion. But the next method is called `get_parked_set`, which speaks for the second suggestion. Your call :smile:

rpc-interface/src/validator.rs
sisou: Regarding the warm key: Is this method returning the warm _address_, or is the warm private key actually accessible in the node? If the latter, why do I need to pass the `warm_secret_key` into the validator transaction creation methods then?

brunoffranca: The method is actually returning the warm private key. Well, you can create validator transactions for any validator for which you have the warm private key. You may want to have the validator running on one computer and use another computer to manage the validator and send these transactions.

rpc-server/src/dispatchers/network.rs
brunoffranca: Nope, the network doesn't provide that information. I can only fetch the PeerId. https://github.com/nimiq/core-rs-albatross/blob/albatross/network-libp2p/src/message/peer.rs#L24-L31

rpc-server/src/dispatchers/consensus.rs
brunoffranca: That is just an internal method, you're not supposed to use it directly.

rpc-server/src/dispatchers/consensus.rs
brunoffranca: Also from the sender wallet. Will do.

rpc-server/src/dispatchers/consensus.rs
sisou: This method is only used in one other method. Maybe the code can then be embedded into the other method, so this one is not on the public interface.

rpc-server/src/dispatchers/network.rs
sisou: https://github.com/nimiq/core-rs-albatross/issues/357

rpc-server/src/dispatchers/consensus.rs
brunoffranca: Done.

rpc-server/src/dispatchers/consensus.rs
brunoffranca: Done.

rpc-server/src/dispatchers/consensus.rs
brunoffranca: Done.

rpc-server/src/dispatchers/consensus.rs
brunoffranca: Done.

rpc-server/src/dispatchers/consensus.rs
brunoffranca: Done.

rpc-server/src/dispatchers/consensus.rs
brunoffranca: Done.

rpc-server/src/dispatchers/consensus.rs
brunoffranca: Done.

rpc-server/src/dispatchers/blockchain.rs
brunoffranca: Done.

sisou:
I added a commit on this PR for backward compatibility with the 1.0 structs.

And I also pushed another, more extensive one, on a branch on top of yours, which you can fast-forward to if you agree with the changes. The changes are all to keep compatibility with the old structs, and to then stay consistent with these changes across the new types also: https://github.com/nimiq/core-rs-albatross/commit/5fda1246945d0b73e5f1c945abaa7677258c79a7

sisou:
The 1.0 block object included a `size` field, that stated the block's size in bytes. This information is especially helpful for chain analysis and block explorers. Do you think that could be made available in the RPC API, @brunoffranca?

brunoffranca:
I think so, shouldn't be a problem.

redmaner:
@brunoffranca I was planning to check the RPC changes, but ran into a build error when building this branch.
```
 => ERROR [builder 5/5] RUN   --mount=type=cache,target=/build/target   --mount=type=cache,target=/usr/local/cargo/registry   --mount=type=cache,target=/usr/local/cargo/git   cargo build --bin nim  1.2s
------                                                                                                                                                                                                     
 > [builder 5/5] RUN   --mount=type=cache,target=/build/target   --mount=type=cache,target=/usr/local/cargo/registry   --mount=type=cache,target=/usr/local/cargo/git   cargo build --bin nimiq-client && cp /build/target/debug/nimiq-client /build/:
#17 0.289     Updating crates.io index
#17 0.988     Updating git repository `https://github.com/nimiq/jsonrpc.git`
#17 1.160 error: failed to select a version for the requirement `ed25519-zebra = "^3.0.0"`
#17 1.160 candidate versions found which didn't match: 2.2.0, 2.1.2, 2.1.1, ...
#17 1.160 location searched: crates.io index
#17 1.160 required by package `nimiq-keys v0.1.0 (/build/keys)`
------
executor failed running [/bin/sh -c cargo build --bin nimiq-client && cp /build/target/debug/nimiq-client /build/]: exit code: 101
```
Seems to also happen on the albatross branch. Maybe related to this commit https://github.com/nimiq/core-rs-albatross/commit/61ef7fcba1f083205dbcf0b9346e83bd7f98c74d. Pretty strange since `ed25519-zebra` is on crates.io version 3.0.0

brunoffranca:
This branch is not building currently. But albatross is (at least it is on my machine). Do `cargo update` and then try building again.

redmaner:
Thanks you are right. Albatross branch builds fine after a cargo update. Can you leave a comment when this branch builds again, then I'll do my testing once the WIP is done.

brunoffranca:
You're welcome. If you don't need the mempool methods, you can just discard the last commit. Then it should work fine. Otherwise, this PR will probably be finished in the next couple of days. I'll warn you then.

sisou:
@brunoffranca The `ValidatorDispatcher` and `MempoolDispatcher` are not registered yet in `lib/src/extras/rpc_server.rs`, so their methods cannot be used yet.

brunoffranca:
@redmaner You can test now.

lib/src/extras/rpc_server.rs
```diff
@@ -59,8 +50,10 @@ pub fn initialize_rpc_server(
         client.consensus_proxy(),
         Some(unlocked_wallets),
     ));
-    dispatcher.add(wallet_dispatcher);
+    dispatcher.add(MempoolDispatcher::new(client.mempool()));
     dispatcher.add(NetworkDispatcher::new(client.network()));
+    dispatcher.add(ValidatorDispatcher::new(client.validator_proxy()));
```
sisou, line 55: Since the methods to retrieve the validator and mempool from the client are behind compiler feature flags, I think these must also be behind these checks? Wouldn't it otherwise lead to compiler errors when the validator feature is disabled during compilation?

rpc-interface/src/blockchain.rs
```diff
@@ -76,18 +81,24 @@ pub trait BlockchainInterface {
         max: Option<u16>,
     ) -> Result<Vec<Transaction>, Self::Error>;
 
-    async fn list_stakes(&mut self) -> Result<HashMap<Address, Coin>, Self::Error>;
+    async fn get_account(&mut self, address: Address) -> Result<Account, Self::Error>;
```
redmaner: I would consider renaming this method to `getAccountByAddress` to have consistent naming with `getValidatorByAddress` and `getStakerByAddress`

rpc-interface/src/blockchain.rs
```diff
@@ -18,52 +21,54 @@ pub trait BlockchainInterface {
...
+        block_number: u32,
+    ) -> Result<Vec<Transaction>, Self::Error>;
 
-    async fn get_transactions_by_block_number(
+    async fn get_inherents_by_block_number(
         &mut self,
         block_number: u32,
+    ) -> Result<Vec<Inherent>, Self::Error>;
+
+    async fn get_batch_transactions(
```
redmaner: I would consider renaming:
* `getBatchTransactions` to `getTransactionsByBatchNumber`
* `getBatchInherents` to `getInherentsByBatchNumber`

For consistency with naming which seems to follow `get some resource by some parameter`. These above methods deviate from the other methods.

rpc-server/src/dispatchers/blockchain.rs
```diff
@@ -108,96 +125,130 @@ impl BlockchainInterface for BlockchainDispatcher {
...
-    async fn get_raw_transaction_info(&mut self, _raw_tx: String) -> Result<(), Error> {
-        Err(Error::NotImplemented)
+        Ok(transactions)
     }
 
-    async fn get_transaction_by_hash(&mut self, hash: Blake2bHash) -> Result<Transaction, Error> {
-        // TODO: Check mempool for the transaction, too
+    /// Returns all the inherents (including reward inherents) for the given block number. Note
+    /// that this only considers blocks in the main chain.
+    async fn get_inherents_by_block_number(
```
redmaner, line 171: Inherents have a `ty` field, which I assume is an abbreviation for `type`? If so, is it possible to call this field just `type` because that would be a lot clearer. Furthermore the value is now an integer, maybe they can be represented as strings for extra clarity?

rpc-interface/src/consensus.rs
```diff
@@ -16,6 +16,11 @@ pub trait ConsensusInterface {
 
     async fn is_established(&mut self) -> Result<bool, Self::Error>;
```
redmaner: `isEstablished` refers to the consensus being established right? I think this method was called this way in 1.0 too. But maybe it is worth considering to clarify this a bit more that it refers to the consensus. A few suggestions: `hasConsensus` or `isConsensusEstablished`

rpc-interface/src/consensus.rs
```diff
@@ -16,6 +16,11 @@ pub trait ConsensusInterface {
...
 
+    async fn get_raw_transaction_info(
+        &mut self,
+        raw_tx: String,
+    ) -> Result<Transaction, Self::Error>;
+
     async fn send_raw_transaction(&mut self, raw_tx: String) -> Result<Blake2bHash, Self::Error>;
 
     async fn create_basic_transaction(
```
redmaner: The current RPC interface doesn't allow / provide options to set the `data` field in a transaction. Although this field is present in transactions if I have seen it correctly. Is it possible to add support for setting the `data` field via RPC methods, or will this be added later?

redmaner (commented):
Did a few checks, mainly on the responses side of things. Looks really nice, great work! 
I have a few comments relating to consistency / clarity of the RPC interface. I'm not a rust developer, so I leave the actual code review to others :+1:

lib/src/extras/rpc_server.rs
brunoffranca: Nope. The whole `initialize_rpc_server` is already conditional on the "rpc-server" compilation flag. And the "rpc-server" flag includes the "validator" flag. So no problems here.

rpc-interface/src/blockchain.rs
brunoffranca: Done.

rpc-interface/src/blockchain.rs
brunoffranca: Done.

rpc-server/src/dispatchers/blockchain.rs
brunoffranca: Nope, not possible. `type` is a reserved word in Rust, which is actually very annoying. That's why we use `ty`.

rpc-interface/src/consensus.rs
brunoffranca: Apparently it was just called `consensus` in 1.0 (https://github.com/nimiq/core-js/wiki/JSON-RPC-API#consensus). @sisou do you want to weigh in?

rpc-interface/src/consensus.rs
brunoffranca: Do you mean adding arbitrary data to the `data` field? I guess I could add that option to the basic transactions, but I'm not sure we want to encourage that behavior...

redmaner: Well I refer to the `data` field which is also present in 1.0, which is hex encoded with a max size of bytes if remember correctly? Reason I would like to use it is to add some description to the transaction. As a pool for example it would be nice to refer to which batch / batches a payout belongs to. But I understand that such a field can be misused, so I do understand your restraint to expose via the RPC interface. I can live without it as well, at least we now have some discussion about it that can be referred to if others ask in the future

rpc-server/src/dispatchers/blockchain.rs
redmaner: That's odd, because the JSON response of `getAccount` has a `type` field representing the type of the account. That's also the reason why I made this comment here, since `type` is preferable over `ty`

lib/src/extras/rpc_server.rs
sisou: Those will likely be separated in the future though, so that RPC servers can be run without being a validator. But if the compiler will find it then when that change is made, it's fine.

rpc-server/src/dispatchers/blockchain.rs
sisou: Serde can be told to rename a field when serializing to JSON, which is what we did with the other structs. I'll look into it on Monday, I probably overlooked the inherent when I did the last cleanup/flattening.

rpc-interface/src/consensus.rs
sisou: In 1.0 I believe it returned a string ('established', 'syncing')? I'm fine with both, for compatibility, the 1.0 version is better obvs. But there's enough incompatible changes already that implementors will have to have a compat wrapper or change their integration anyway.

rpc-interface/src/consensus.rs
sisou: The tx format is the same as in 1.0, so if you can generate a valid tx, you can send it with `sendRawTransaction` and `pushTransaction`.
This method here is for `basic` txs, which don't have a data field. That needs an extended tx, which then also should allow setting flags, sender and recepient types.
So either this method is adapted to accept all of those and constructs the correct tx based on inputs, or a separate method for extended txs.

redmaner: I think the latter, as in adding separate method for extended txs, makes most sense then

rpc-interface/src/consensus.rs
brunoffranca: Done.

rpc-server/src/dispatchers/blockchain.rs
brunoffranca: No need, @sisou. Found it.

rpc-interface/src/consensus.rs
brunoffranca: I don't see the point of having a method for extended transactions if all the cases are already covered by the other methods and we have a method to submit raw transactions. If we want to "support" basic transactions with data then I would prefer to just add the data field to the basic transaction method.

P.S. Now I'm seeing that we are missing methods to handle HTLC and Vesting contracts. I need to add these.

rpc-interface/src/blockchain.rs
```diff
@@ -76,18 +81,24 @@ pub trait BlockchainInterface {
...
+    async fn get_account_by_address(&mut self, address: Address) -> Result<Account, Self::Error>;
+
+    async fn get_active_validators(&mut self) -> Result<HashMap<Address, Coin>, Self::Error>;
+
+    async fn get_current_slashed_slots(&mut self) -> Result<SlashedSlots, Self::Error>;
 
-    async fn get_validator(
+    async fn get_previous_slashed_slots(&mut self) -> Result<SlashedSlots, Self::Error>;
+
+    async fn get_parked_set(&mut self) -> Result<ParkedSet, Self::Error>;
```
redmaner: Just looking at this now, what is a parked set exactly? Are these parked validators?

brunoffranca: Yes, they are the parked validators for the current epoch.

redmaner: does it make sense for clarity to reference validator in the method name? For example `get_parked_validator_set` or `get_parked_validators`? That would be more clear and complements well with `get_active_validators` method

brunoffranca: Sure.

primitives/trie/src/key_nibbles.rs
```diff
@@ -1,6 +1,8 @@
 use std::borrow::Cow;
 use std::cmp;
 use std::fmt;
+
```
styppo: Fixed in 0d3c6a30386845beb1387eb34f48fcfa76ecdeb3, this adds the unused import again

brunoffranca: Fixed.

sisou:
This PR was reverted by force-pushing the `albatross` branch to e1954e8f6cd2641e588278eac4e5e02fe873bfeb.

brunoffranca:
PR was re-merged.

PR #364, opened 2021-10-14
Streamline Mempool

brunoffranca:
## Pull request checklist

1. Improves readability and reduces code duplication in the mempool crate.
2. Adds methods to mempool that are necessary for the RPC API.
3. Simplifies the ordering used in the mempool.

### DIFF
--- a/mempool/src/filter.rs
+++ b/mempool/src/filter.rs
@@ -6,8 +6,8 @@ use nimiq_transaction::{Transaction, TransactionFlags};
 /// Struct defining a Mempool filter
 #[derive(Debug)]
 pub struct MempoolFilter {
-    blacklist: LimitHashSet<Blake2bHash>,
-    rules: MempoolRules,
+    pub blacklist: LimitHashSet<Blake2bHash>,
+    pub rules: MempoolRules,
 }
 
 impl MempoolFilter {
--- a/mempool/src/filter.rs
+++ b/mempool/src/filter.rs
@@ -34,12 +34,12 @@ impl MempoolFilter {
         self
     }
 
-    /// Checks wether a transaction is blacklisted
+    /// Checks whether a transaction is blacklisted
     pub fn blacklisted(&self, hash: &Blake2bHash) -> bool {
         self.blacklist.contains(hash)
     }
 
-    /// Checks wether a transaction is accepted according to the general Mempool filter rules
+    /// Checks whether a transaction is accepted according to the general Mempool filter rules
     ///
     /// The following rules are checked in this function:
     /// - tx_fee
--- a/mempool/src/filter.rs
+++ b/mempool/src/filter.rs
@@ -6,8 +6,8 @@ use nimiq_transaction::{Transaction, TransactionFlags};
 /// Struct defining a Mempool filter
 #[derive(Debug)]
 pub struct MempoolFilter {
-    pub blacklist: LimitHashSet<Blake2bHash>,
-    pub rules: MempoolRules,
+    pub(crate) blacklist: LimitHashSet<Blake2bHash>,
+    pub(crate) rules: MempoolRules,
 }
 
 impl MempoolFilter {
--- a/mempool/src/lib.rs
+++ b/mempool/src/lib.rs
@@ -1,4 +1,3 @@
-#![deny(missing_docs)]
 #![feature(hash_drain_filter)]
 
 //! Mempool implementation
@@ -10,6 +9,8 @@
 extern crate log;
 /// Mempool config module
 pub mod config;
+/// Mempool executor module
+pub mod executor;
 /// Mempool filter module
 pub mod filter;
 /// Main mempool module
--- a/mempool/src/lib.rs
+++ b/mempool/src/lib.rs
@@ -1,3 +1,4 @@
+#![deny(missing_docs)]
 #![feature(total_cmp)]
 
 //! Mempool implementation
--- a/mempool/src/lib.rs
+++ b/mempool/src/lib.rs
@@ -1,4 +1,4 @@
-#![feature(hash_drain_filter)]
+#![feature(total_cmp)]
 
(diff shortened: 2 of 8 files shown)


mempool/src/mempool.rs
```diff
@@ -254,556 +156,340 @@ impl Mempool {
...
-    // Transactions ordered by fee (higher fee transactions pop first)
-    transactions_by_fee: KeyedPriorityQueue<Transaction, TransactionFeePerByteOrdering>,
-
-    // Transactions ordered by age (older transactions pop first)
-    // TODO: This could be improved to store a reference to the transaction and not the transaction itself
-    transactions_by_age: KeyedPriorityQueue<Transaction, TransactionAgeOrdering>,
+    /// Returns a vector with accepted transactions from the mempool.
+    ///
+    /// Returns the highest fee per byte up to max_bytes transactions and removes them from the mempool
+    pub fn fill_block(&self, max_bytes: usize) -> Result<Vec<Transaction>, Error> {
```
jsdanielh: I actually liked more the `get_transactions` name since we are actually returning a vector of `transactions` but I see the reason of the change being having another function that returns all transactions below :thinking:
Maybe `get_transactions_batch`?

primitives/transaction/src/lib.rs
```diff
@@ -556,7 +523,17 @@ impl PartialOrd for Transaction {
 
 impl Ord for Transaction {
     fn cmp(&self, other: &Self) -> Ordering {
-        self.cmp_mempool_order(other)
+        Ordering::Equal
+            .then_with(|| self.recipient.cmp(&other.recipient))
+            .then_with(|| self.validity_start_height.cmp(&other.validity_start_height))
+            .then_with(|| other.fee.cmp(&self.fee))
+            .then_with(|| other.value.cmp(&self.value))
```
jsdanielh, line 530: For consistency I would use:
`self.value.cmp(&other.value)`
(and the same for fee)

mempool/src/mempool.rs
```diff
@@ -254,556 +156,340 @@ impl Mempool {
...
+//       transactions over large size transactions (assuming they have the same fee per byte). Or
+//       we might prefer basic transactions over staking contract transactions, etc, etc.
+#[derive(PartialEq, PartialOrd)]
+pub struct FeeWrapper(f64);
+
+impl Eq for FeeWrapper {}
+
+impl Ord for FeeWrapper {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.0.total_cmp(&other.0)
```
jsdanielh, line 499: Clippy is saying that we must either implement `Ord` or `PartialOrd` (which we are deriving)

mempool/src/lib.rs
```diff
@@ -1,5 +1,4 @@
-#![deny(missing_docs)]
```
jsdanielh, line 1: Mmm, I like this since it actually force us to document public facing components

mempool/src/mempool.rs
```diff
@@ -254,556 +156,340 @@ impl Mempool {
...
+    /// Gets all transactions in the mempool.
+    pub fn get_transactions(&self) -> Vec<Transaction> {
+        self.state.read().transactions.values().cloned().collect()
+    }
+}
 
-            // Spawn the transaction verification task
-            tokio::task::spawn(async move {
-                log::debug!("Starting execution of new verif task");
+pub struct MempoolState {
```
jsdanielh: Mmm, maybe `pub(crate)`?

mempool/src/filter.rs
```diff
@@ -6,8 +6,8 @@ use nimiq_transaction::{Transaction, TransactionFlags};
 /// Struct defining a Mempool filter
 #[derive(Debug)]
 pub struct MempoolFilter {
-    blacklist: LimitHashSet<Blake2bHash>,
-    rules: MempoolRules,
+    pub blacklist: LimitHashSet<Blake2bHash>,
```
jsdanielh: Add: `/// Mempool filter blacklist` such that the missing documentation check passes

mempool/src/filter.rs
```diff
@@ -6,8 +6,8 @@ use nimiq_transaction::{Transaction, TransactionFlags};
...
+    pub rules: MempoolRules,
```
jsdanielh: Add `/// Mempool filter rules` such that the missing documentation check passes

jsdanielh (commented):
Looks good to me. Just some minor comments

mempool/src/mempool.rs
brunoffranca: We already use batch for the period between two macro blocks, it would be confusing. `get_transactions_block` is probably fine.

primitives/transaction/src/lib.rs
brunoffranca: That reverses the direction of the ordering. I didn't write this part, so I don't know if there's a reason for this particular ordering. I would leave it as is.

mempool/src/mempool.rs
brunoffranca: Clippy can be very annoying sometimes. Solved.

mempool/src/lib.rs
brunoffranca: But it's very annoying when you are working in the crate. I'll re-add it to the crate.

mempool/src/mempool.rs
brunoffranca: Done.

mempool/src/filter.rs
brunoffranca: I just made it pub(crate). Fixed.

mempool/src/filter.rs
brunoffranca: Same as above.

jsdanielh (approved):

viquezclaudio (approved):
Refactoring looks good to me

PR #368, opened 2021-10-19
Manually add transaction to Mempool

brunoffranca:
## Pull request checklist

1. Separated the transaction verification logic into its own function and simplified it.
2. Added a method to the Mempool to add a transaction manually.

### DIFF
--- a/mempool/src/executor.rs
+++ b/mempool/src/executor.rs
@@ -4,21 +4,16 @@ use std::sync::Arc;
 
 use futures::task::{Context, Poll};
 use futures::{stream::BoxStream, Future, StreamExt};
-
 use parking_lot::RwLock;
 
-use nimiq_account::{Account, BasicAccount};
-
-use nimiq_blockchain::{AbstractBlockchain, Blockchain};
-use nimiq_hash::Hash;
-
+use nimiq_blockchain::Blockchain;
 use nimiq_network_interface::network::{Network, Topic};
-use nimiq_primitives::coin::Coin;
-use nimiq_primitives::networks::NetworkId;
+
 use nimiq_transaction::Transaction;
 
 use crate::filter::MempoolFilter;
 use crate::mempool::MempoolState;
+use crate::verify::{verify_tx, ReturnCode};
 
 const CONCURRENT_VERIF_TASKS: u32 = 1000;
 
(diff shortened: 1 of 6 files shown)


mempool/src/executor.rs
```diff
@@ -123,169 +94,26 @@ impl<N: Network> Future for MempoolExecutor<N> {
...
-                    ) {
-                        log::debug!("Transaction filtered: Not accepting transaction due to recipient balance");
-                        return ReturnCode::Filtered;
-                    }
-                }
-
-                if sender_in_fly_balance <= blockchain_sender_balance {
-                    // 6. Add transaction to the mempool
-                    log::debug!(" Accepting new transaction");
+                let rc = verify_tx(&tx, blockchain, mempool_state.clone(), filter);
```
jsdanielh: I actually prefer `Arc::clone(&mempool_state)`

mempool/src/mempool.rs
```diff
@@ -360,6 +353,23 @@ impl Mempool {
...
 
+    /// Adds a transaction to the Mempool.
+    pub fn add_transaction(&self, transaction: Transaction) -> ReturnCode {
+        let blockchain = Arc::clone(&self.blockchain);
+        let mempool_state = Arc::clone(&self.state);
+        let filter = Arc::clone(&self.filter);
+
+        log::debug!("Verifying manually added transaction.");
+
+        let rc = verify_tx(&transaction, blockchain, mempool_state.clone(), filter);
```
jsdanielh: Same here, I actually prefer `Arc::clone(&mempool_state)`

mempool/src/verify.rs
```diff
@@ -0,0 +1,144 @@
...
+    transaction: &Transaction,
+    blockchain: Arc<RwLock<Blockchain>>,
+    mempool_state: Arc<RwLock<MempoolState>>,
+    filter: Arc<RwLock<MempoolFilter>>,
+) -> ReturnCode {
+    // Check if we already know the transaction
+    let mempool_state = mempool_state.read();
+
+    if mempool_state.contains(&transaction.hash()) {
+        //We already know this transaction, no need to process
```
jsdanielh: Add a space after `//`

mempool/src/verify.rs
```diff
@@ -0,0 +1,144 @@
...
+        log::debug!("Transaction invalid at block {}", block_height);
+        return ReturnCode::Invalid;
+    }
+
+    if blockchain.contains_tx_in_validity_window(&transaction.hash()) {
+        log::debug!("Transaction has already been mined");
+        return ReturnCode::Invalid;
+    }
+
+    // 4. Sequentialize per Sender to Check Balances and acquire the upgradable from the blockchain),
```
jsdanielh: Looks like the final parenthesis is not needed

mempool/src/verify.rs
```diff
@@ -0,0 +1,144 @@
...
+        log::debug!("Transaction filtered: Not accepting transaction due to recipient balance");
+        return ReturnCode::Filtered;
+    }
+
+    if sender_in_fly_balance > blockchain_sender_balance {
+        log::debug!("Dropped because sum of txs in mempool is larger than the account balance");
+        return ReturnCode::NotEnoughFunds;
+    }
+
+    log::debug!(" Accepting new transaction");
```
jsdanielh: Looks like the space after the quotes could be removed

jsdanielh (commented):
Looks good to me. Only few minor comments

mempool/src/executor.rs
brunoffranca: The two forms are equivalent (https://doc.rust-lang.org/std/sync/struct.Arc.html#cloning-references). Do you still prefer  `Arc::clone(&mempool_state)`?

jsdanielh: Yeah, they are equivalent but as stated [here](https://github.com/rust-lang/rust/pull/42137):
> The `Arc::clone(&from)` syntax is the most idiomatic because it conveys more explicitly
> the meaning of the code. In the example above, this syntax makes it easier to see that
> this code is creating a new reference rather than copying the whole content of foo.

brunoffranca: Fair enough!

PR #372, opened 2021-10-21
Fix locking issues for the mempool

jsdanielh:
This PR contains two commits.
The first commit originally tried to increase the mempool performance by:
- Having the `mempool_state` and `blockchain` locks less time taken.
- Spawning the transaction verification task.

By implementing this optimization and in conversations with @nibhar we realized there were
locking issues with the current implementation and a second commit was added.
This commit fixes possible issue with the `mempool_state` locking. Before this change the
read lock of the `mempool_state` was dropped before taking a write lock to it to push a
transaction. This may lead to issues were a transaction could change the `mempool_state` in
between.
This change re-adds the upgradable RW lock of the `mempool_state` which is now returned by
the `verify_tx` function such that the caller can upgrade the lock and push a transaction.

## Pull request checklist

### DIFF
--- a/mempool/src/verify.rs
+++ b/mempool/src/verify.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use parking_lot::RwLock;
+use parking_lot::{RwLock, RwLockUpgradableReadGuard};
 
 use nimiq_account::{Account, BasicAccount};
 use nimiq_blockchain::{AbstractBlockchain, Blockchain};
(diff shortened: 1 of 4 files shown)


mempool/src/verify.rs
```diff
@@ -115,9 +139,11 @@ pub(crate) fn verify_tx(
     let sender_in_fly_balance = transaction.total_value().unwrap() + sender_current_balance;
     let recipient_in_fly_balance = transaction.total_value().unwrap() + recipient_current_balance;
 
+    let filter = filter.read();
+
     // Check the balance against filters
     if !filter.accepts_sender_balance(
-        transaction,
+        &transaction,
```
viquezclaudio: there are some clippy warnings because of this, otherwise looks good to me

jsdanielh: Just fixed them

brunoffranca (approved):

brunoffranca:
For me it is slower though. I get ~800 tps on `albatross` and only get ~600 tps on this branch. (on debug)

P.S. - On release mode I get the exact same value for both. 3200 tps.

viquezclaudio (approved):

Issue #373, opened 2021-10-21
After killing/restarting validators multiple times there are no more new blocks


viquezclaudio:
In a 4 validators setup, I have been able to consistently get to a state where the network does not longer make any progress (it just stalls)
In order to reproduce this issue, I have 4 validators  (A,B,C,D) running and producing blocks, and then:

- Kill Validator A
- Delete all of the state from Validator A
- Restart Validator A
- Wait for A to be in sync and produce some blocks
- Kill Validator B
- Delete all of the state from Validator B
- Restart Validator B
- Wait for B to be in sync
- **Validators no longer make any further progress**

As part of the debug process, it has been determined that:

- If instead of deleting all of the state of one validator, I just remove the _devalbatross-history-consensus_ folder for that validator, the issue is no longer reproduced.
- If I modify the following lines of code, the issue is no longer reproduced:

```
diff --git a/validator-network/src/network_impl.rs b/validator-network/src/network_impl.rs
index c56bc7eb..d2c8bfcc 100644
--- a/validator-network/src/network_impl.rs
+++ b/validator-network/src/network_impl.rs
@@ -138,13 +138,13 @@ where
             .ok_or(NetworkError::UnknownValidator(validator_id))?
             .clone();
 
-        let entry = state.validator_peer_id_cache.entry(public_key.clone());
+        // let entry = state.validator_peer_id_cache.entry(public_key.clone());
 
-        match entry {
-            Entry::Occupied(occupied) => Ok(occupied.get().clone()),
-            Entry::Vacant(vacant) => {
+        // match entry {
+        //     Entry::Occupied(occupied) => Ok(occupied.get().clone()),
+        //     Entry::Vacant(vacant) => {
                 if let Some(peer_id) = Self::resolve_peer_id(&self.network, &public_key).await? {
-                    Ok(vacant.insert(peer_id).clone())
+                    Ok(peer_id)
                 } else {
                     log::error!(
                         "Could not find peer ID for validator in DHT: public_key = {:?}",
@@ -152,8 +152,8 @@ where
                     );
                     Err(NetworkError::UnknownValidator(validator_id))
                 }
-            }
-        }
+        //     }
+        // }
     }
 }
```

The logs from the 4 validators are attached.
2021-10-21 10:43:31.604 is where validator B got consensus
2021-10-21 10:43:32.707 proposal for round 0 times out
and then no more progress

[...]

PR #374, opened 2021-10-22
Fetch validators peer_keys if a lookup fails

nibhar:
## Pull request checklist

Whenever a validator disconnects and reconnects with a new peer_key the validator network can not handle it. This PR remedies that situation by re fetching the peer_key if the peer can not be found.


viquezclaudio:
With this change I cannot reproduce [373](https://github.com/nimiq/core-rs-albatross/issues/373) anymore

PR #381, opened 2021-11-01
Bash script to test 4 validators scenario

viquezclaudio:
Simple bash script to test 4 validators connecting/disconnecting
It also generates logs for each validator and searches for panics


viquezclaudio:
This PR is missing a github file in order to continuously execute this test

PR #382, opened 2021-11-02
Only start processing macro proposals for current height

### LINKED ISSUE #385: current_timestamp >= previous_timestamp
In the 4xValidators reconnecting scenario, after multiple re connection/restarts, it is sometimes observed that some validator panics because of the following assertion:
` thread 'tokio-runtime-worker' panicked at 'assertion failed: current_timestamp >= previous_timestamp': blockchain/src/reward.rs:47`

This has been difficult to reproduce, however after further debugging and via the attached logs, it was observed that the block 
proposal triggering the panic is for an old block.
Based on this logs, it seems that in order to hit this condition, we need to receive a previous proposal, while syncing, that comes from the same validator as the one that we are expecting the current proposal.

[20211101_141943-timestamp-logs.zip](https://github.com/nimiq/core-rs-albatross/files/7462685/20211101_141943-timestamp-logs.zip)

nibhar:
## Pull request checklist

This PR adds an additional check to received proposals to have their block height match the expected block height.

Prior to this PR there is a rare case where a proposal for a previous block from the same proposer as the current proposal could be received which would result in an assertion failing stating that a block must have an equal or higher timestamp than its predecessor.

### DIFF
--- a/validator/src/tendermint.rs
+++ b/validator/src/tendermint.rs
@@ -211,11 +211,13 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> TendermintOutsideDeps
         &mut self,
         round: u32,
     ) -> Result<ProposalResult<Self::ProposalTy>, TendermintError> {
-        let (timeout, validator_id, validator_key) = {
+        let (timeout, validator_id, validator_key, expected_height) = {
             let blockchain = self.blockchain.read();
+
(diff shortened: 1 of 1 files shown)


viquezclaudio (approved):

viquezclaudio:
This is the proposed fix for issue
[385](https://github.com/nimiq/core-rs-albatross/issues/385)

PR #386, opened 2021-11-02
Bash script to test 4 validators scenario

viquezclaudio:
Simple bash script to test 4 validators connecting/disconnecting
It also generates logs for each validator and searches for panics
A Github file is included so we can automatically execute this
scenario as part of our CIT

## Pull request checklist

- [x ] All tests pass. Demo project builds and runs.
- [x ] I have resolved any merge conflicts.

### DIFF
--- a/.github/workflows/reconnecting_validators.yml
+++ b/.github/workflows/reconnecting_validators.yml
@@ -1,0 +1,78 @@
+name: Four validators scenario
+
+on:
+  schedule:
+    - cron: '0 * * * *'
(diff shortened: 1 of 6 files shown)


viquezclaudio:
This is an example of a CIT run that actually hit a known issue /bug:
https://github.com/viquezclaudio/core-rs-albatross/runs/4084404500?check_suite_focus=true

jeffesquivels (approved):

PR #390, opened 2021-11-03
Deregister listeners when a notifier stream receiver is gone

### LINKED ISSUE #383: Deregister listeners created as streams
When a listener is registered as a stream, there isn't a proper mechanism to deregister it. As such, because the listener is still going to get notified of new events even if the receiver is no longer there, the sender will keep attempting to send it the new values and this error pops up:

`ERROR observer | Failed to send event to channel: channel closed`

Related to #352, nimiq/jsonrpc#2

vlvrd:
This closes #383


sisou:
I tested this branch and can confirm that it works, the observer error is now only logged once after the socket (and thus the stream) was closed, no longer on every following block. :+1:

PR #400, opened 2021-11-05
Remove signature verification for gossipsub messages

jsdanielh:
Remove signature verification for libp2p gossipsib messages.
This also required to change the privacy settings of gossipsub
messages to only contain the peer ID instead of the sender public
key.
This change also improves the performance of the network as found
after profiling the code and seeing that a lot of CPU was being
spent in signature verifications for gossipsub messages.

## Pull request checklist


vlvrd (approved):
LGTM

PR #420, opened 2021-11-15
Increase the initial DB map size

### LINKED ISSUE #419: Problem when trying to resize the LMDB database
In a 4 validators + spammer + seed node scenario, running each one in a docker container, one of the validators halted (it didn't crashed), stopped producing blocks and stopped printing any log messages after a LMDB resize operation took place:

```
seed1_1    |  2021-11-13 01:06:54.525 DEBUG behaviour            | Removing stale sender from global message_receivers: TYPE_ID: 121
seed1_1    |  2021-11-13 01:06:55.321 DEBUG block_queue          | Received block #8544 via gossipsub
seed1_1    |  2021-11-13 01:06:55.322 DEBUG block_queue          | Received block #8544 via gossipsub
seed1_1    |  2021-11-13 01:06:55.388 DEBUG block_queue          | Received block #8544 via gossipsub
seed1_1    |  2021-11-13 01:06:55.401 INFO  lmdb                 | DB resize (percent-based)
seed1_1    |  2021-11-13 01:06:55.401 INFO  lmdb                 | DB map size: 6442450944
seed1_1    |  2021-11-13 01:06:55.401 INFO  lmdb                 | Space used: 3891486720
seed1_1    |  2021-11-13 01:06:55.401 INFO  lmdb                 | Space remaining: 2550964224
seed1_1    |  2021-11-13 01:06:55.401 INFO  lmdb                 | Percent used: 0.60
seed1_1    |  2021-11-13 01:06:55.587 INFO  lmdb                 | LMDB Mapsize increased. Old: 6144 MiB, New: 7168 MiB
```

jsdanielh:
Increase the initial DB map size from 1G to 1T.
Also increase the percentage threshold that defines when to resize
the DB to 100% such that the DB never gets resized.
This is to address issues resizing LMDB which is not well supported
and the recommendation is to use a large initial mapping size.
This will affect 32-bit systems and systems without sparse file
system support that are also written in the system requirements
section in the `README.md` file.

This closes #419.

## Pull request checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,11 @@ The Nimiq Rust client comes without wallet and can currently not be used to send
 - [Nimiq Developer Reference](https://nimiq-network.github.io/developer-reference/): Details of the protocol architecture.
 
 
+## System requirements
+- 32-bit architectures are not supported.
+- File systems with sparse file support (most modern files systems support it).
+
+
 ## Install
 
 Besides [Rust nightly](https://www.rust-lang.org/learn/get-started#installing-rust) itself, the following packages are required to be able to compile this source code:
--- a/database/src/lmdb.rs
+++ b/database/src/lmdb.rs
@@ -7,7 +7,6 @@ use std::sync::Arc;
 pub use lmdb_zero::open;
 use lmdb_zero::traits::LmdbResultExt;
 pub use lmdb_zero::Error as LmdbError;
-use rand::{thread_rng, Rng};
 
 use crate::cursor::{RawReadCursor, ReadCursor, WriteCursor as WriteCursorTrait};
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -27,6 +27,11 @@ The Nimiq Rust client comes without wallet and can currently not be used to send
 - [Nimiq Developer Reference](https://nimiq-network.github.io/developer-reference/): Details of the protocol architecture.
 
 
+## System requirements
+- 32-bit architectures are not supported.
+- File systems with sparse file support (most modern files systems support it).
```
sisou, line 32: Do these requirements only apply when the size of 1TB is used? If a config file specifies a smaller file size, can 32-bit systems without sparse file support still work?

sisou (approved):
Seems the changes do what the description says they should do. 

However, does this mean a 1TB file will be created whenever I start a node? :astonished: This will make testing much harder?

But it seems the size is also configurable with the config file. Is the example config file currently maintained? Can this PR add an example/default entry to the example config file?

README.md
jsdanielh: Yes, but the limit would be 4G which we think we can quickly reach according to the initial data we have collected.

jsdanielh:
> However, does this mean a 1TB file will be created whenever I start a node? This will make testing much harder?

It means that's the maximum size the database can grow to when running on systems that use sparse file systems. Since we want to avoid resizing the DB, we are setting an initial value that we preview wouldn't need resizing.

PR #421, opened 2021-11-15
Re-implement KeyNibbles using arrays instead of vectors

jsdanielh:
- Re-implement KeyNibbles using arrays
  - Re-implement KeyNibbles using arrays instead of vectors.
    This is to improve the performance of KeyNibbles since the profiler
    showed that with vectors there is plenty of CPU cycles spent
    in vector heap allocation and handling.

- Fix some typos in the database crate

## Pull request checklist

### DIFF
--- a/database/src/lib.rs
+++ b/database/src/lib.rs
@@ -185,7 +185,7 @@ impl<'env> WriteTransaction<'env> {
     }
 
     /// Puts a key/value pair into the database by copying it into a reserved space in the database.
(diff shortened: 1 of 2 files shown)


sisou:
Why? Why are arrays better than vectors?

jsdanielh:
> Why? Why are arrays better than vectors?

Just edited the PR description to add a little more of context :smiley:

brunoffranca (approved):

viquezclaudio (approved):

PR #456, opened 2021-11-26
Add network fixes

jsdanielh:
Decrease the DHT record TTL and related settings:
- Reduce record TTL from 36 h to 6 min.
- Reduce the publication interval from 24h to 1 min.
- Reduce the replication interval from 60 min to 20s.
    
Remove the peer from the network observable peer map and let the
peer's dispatcher that the connections is going to be closed.
This fixes issues related to stale dispatchers and peers when a
peer closes and then rejoins.

## Pull request checklist

### DIFF
--- a/network-libp2p/src/connection_pool/behaviour.rs
+++ b/network-libp2p/src/connection_pool/behaviour.rs
@@ -543,7 +543,7 @@ impl NetworkBehaviour for ConnectionPoolBehaviour {
 
     fn inject_connection_closed(
         &mut self,
-        _peer_id: &PeerId,
+        peer_id: &PeerId,
         _conn: &ConnectionId,
         endpoint: &ConnectedPoint,
         _handler: <Self::ProtocolsHandler as IntoProtocolsHandler>::Handler,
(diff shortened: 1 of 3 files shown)


network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -573,6 +573,19 @@ impl NetworkBehaviour for ConnectionPoolBehaviour {
...
+        // Notify handler about the connection is going to be shut down
+        self.actions
+            .push_back(NetworkBehaviourAction::NotifyHandler {
+                peer_id: *peer_id,
+                handler: NotifyHandler::Any,
+                event: HandlerInEvent::Close {
+                    reason: CloseReason::RemoteClosed,
+                },
+            });
+        // Remove peer from the observable peer map
```
vlvrd: Is this ever triggered? The peer object should be removed when the `ConnectionClosed` Swarm event is fired and the `PeerLeft` event is sent

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -573,6 +573,19 @@ impl NetworkBehaviour for ConnectionPoolBehaviour {
         };
 
         self.addresses.mark_closed(address.clone());
+        // Notify handler about the connection is going to be shut down
+        self.actions
+            .push_back(NetworkBehaviourAction::NotifyHandler {
+                peer_id: *peer_id,
+                handler: NotifyHandler::Any,
```
vlvrd, line 580: We can run into race conditions if we don't address the specific connection here with `NotifyHandler::One`. This is probably best changed in line 522

vlvrd (changes requested):

vlvrd (approved):

vlvrd (approved):

PR #497, opened 2021-12-06
Removes unstaking delays

brunoffranca:
## Pull request checklist

This includes changes in the primitives/account, primitives/transaction, transaction-builder, rpc-interface, rpc-client and rpc-server crates.

- Deletes the retire, reactivate and deduct fees transactions from the staker account. They are no longer needed. 
- Modifies the unstake transaction to have immediate effect. 
- Renames the "retire" and "drop" validator transactions to "inactivate" and "delete".

### DIFF
--- a/primitives/account/src/account.rs
+++ b/primitives/account/src/account.rs
@@ -53,7 +53,7 @@ impl Account {
             Account::StakingValidatorsStaker(_) => {
                 unimplemented!()
             }
-            Account::StakingStaker(ref account) => account.active_stake + account.inactive_stake,
+            Account::StakingStaker(ref account) => account.balance,
         }
     }
 
--- a/primitives/account/src/staking_contract/receipts.rs
+++ b/primitives/account/src/staking_contract/receipts.rs
@@ -43,7 +43,7 @@ pub struct UnparkValidatorReceipt {
 }
 
 #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
-pub struct DropValidatorReceipt {
+pub struct DeleteValidatorReceipt {
     pub signing_key: SchnorrPublicKey,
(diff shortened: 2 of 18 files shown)


maestroi:
does this feature mean someone can unstake their nim mid epoch?

brunoffranca:
Yes.

maestroi:
but the stake is finalized in the election block right, so during the epoch the amount is set?

brunoffranca:
Not sure I'm understanding your question. Validators are chosen based on whatever stake they have delegated to them in the election block. During the epoch the stakes can change and it is reflected immediately in the staking contract. But of course the amount of stake you have only really matters on election blocks.

Eligioo:
And this is not a problem because Albatross doesn't slash misbehaving nodes. Only refrain from creating more blocks for a while right?

maestroi:
That means people can just stake just before election block and after they can retrackt right ?

redmaner:
What I'm wondering though. From a pool perspective we payout rewards as stake. But what if in the meantime someone mid epoch decides to completely unstake their balance. Is it then still possible to payout as stake if there is no stake anymore? Is paying out rewards as stake still the best way?

sisou:
> What I'm wondering though. From a pool perspective we payout rewards as stake. But what if in the meantime someone mid epoch decides to completely unstake their balance. Is it then still possible to payout as stake if there is no stake anymore? Is paying out rewards as stake still the best way?

This is a good point, but as stakers are identified by their address, the pool could fall back to pay out as a regular tx to the staker's address in case the staker is not available anymore.

brunoffranca:
@Eligioo That's right, we don't slash stakes only rewards, so unstaking mid-epoch is not a problem.

@maestroi If they want to, yes. All that matters is what stake a validator has in the election block.

PR #538, opened 2022-01-06
Remove unnecesary blockchain read from mempool

viquezclaudio:
- The network_id is basically an static parameter, so it is not necessary to get the blockchain read lock to get this value each time a transaction is verified
- Fixed a bug that was not limiting the number of concurrent transaction verification tasks

### DIFF
--- a/mempool/src/executor.rs
+++ b/mempool/src/executor.rs
@@ -10,6 +10,7 @@ use parking_lot::{RwLock, RwLockUpgradableReadGuard};
 use nimiq_blockchain::Blockchain;
 use nimiq_network_interface::network::Network;
 use nimiq_network_interface::prelude::MsgAcceptance;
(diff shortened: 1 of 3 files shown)


mempool/src/executor.rs
```diff
@@ -34,6 +35,9 @@ pub(crate) struct MempoolExecutor<N: Network> {
     // Reference to the network, to alow for message validation
     network: Arc<N>,
 
+    // Network ID, used for tx verification
+    network_id: Arc<NetworkId>,
```
jsdanielh, line 39: Mmm... `NetworkId` is an `enum`. I don't think we need to arc it. A copy should work instead

viquezclaudio: That is  because we need to send it to multiple tx verify tasks that are running in parallel...

PR #542, opened 2022-01-07
Change header for typed messages

jsdanielh:
Change header for typed messages such that the header no longer
contains fields with variable length.

## Pull request checklist


vlvrd (approved):
👍

Issue #356, opened 2021-10-07
Possible performance optimizations


brunoffranca:
Here are some possible performance optimizations to Albatross. They might or might not be practical, and might or might not be implemented in the future.

1. **Changing the signature scheme used in the micro blocks.** Micro blocks contain two BLS signatures in the MNT6-753 elliptic curve (one for signing the block and the other for the VRF). These are very expensive to create and verify. We can change both to use curve 25519 and drastically improve performance. For signing the block there is the ed25519 scheme and for the VRF there is the VXEdDSA scheme (see https://github.com/w3f/schnorrkel, https://www.signal.org/docs/specifications/xeddsa/#vxeddsa and https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vrf-06). The simplest way to do this is probably to use the validators warm key.
2. **Getting rid of the checkpoint macro blocks.** Checkpoint blocks are mostly used for syncing and Albatross can work without them. They do introduce small delays every couple of minutes which hurts the performance a bit. We could maybe use a finality gadget (like GRANDPA, see https://arxiv.org/abs/2007.01560) to helps us with the syncing and get rid of the checkpoint blocks. This has the advantage of reducing the number of micro blocks that a node needs to download when syncing, since the finality gadget will most likely always be only a few micro blocks behind the head of the chain (after all, most validators can probably agree on a block that is 10 confirmations deep, for example). The main problem is the added complexity. We would also need to find a solution for the longer duration between rewards. This was a problem before the introduction of the checkpoint blocks, if a validator misbehaved even a single time it would lose the reward for the entire epoch. [A possible solution is for rewards to be slashed gradually. For example, for each time you get view-changed, you lose 1/10 of your reward.]
[...]

PR #569, opened 2022-01-23
Switch BLS groups

brunoffranca:
## Pull request checklist

This implements proposal 3 in issue #356. Turns out the performance improvement for the nano zkp proofs is even bigger than anticipated since allocating a G2 point in the circuit is orders of magnitude more expensive than allocating a G1 point.


block-production/tests/mod.rs
```diff
@@ -286,7 +286,7 @@ fn it_can_revert_unpark_transactions() {
 }
 
 #[test]
-fn it_can_revert_create_stacker_transaction() {
+fn it_can_revert_create_staker_transaction() {
```
paberr, line 289: Good catch 😆

bls/src/types/compressed_public_key.rs
```diff
@@ -19,15 +19,15 @@ use crate::PublicKey;
 /// and one bit indicating if it is the "point-at-infinity".
 #[derive(Clone)]
 pub struct CompressedPublicKey {
-    pub public_key: [u8; 285],
+    pub public_key: [u8; 95],
 }
 
 impl CompressedPublicKey {
-    pub const SIZE: usize = 285;
+    pub const SIZE: usize = 95;
```
paberr, line 26: I haven't checked this number.

nano-zkp/src/gadgets/mnt4/hash_to_curve.rs
```diff
@@ -78,69 +78,112 @@ impl HashToCurve {
...
+            c0_bits.iter().map(|i| i.value().unwrap()).collect(),
+            c1_bits.iter().map(|i| i.value().unwrap()).collect(),
+            c2_bits.iter().map(|i| i.value().unwrap()).collect(),
             y_bit.value()?,
         );
 
-        // Allocate the nonce bits and convert to a field element.
+        // Allocate the nonce bits.
         let nonce_bits_var = Vec::<Boolean<MNT4Fr>>::new_witness(cs.clone(), || Ok(nonce_bits))?;
-        let nonce = Boolean::le_bits_to_fp_var(&nonce_bits_var)?;
```
paberr, line 92: Why can't we use this anymore?

nano-zkp/src/gadgets/mnt4/hash_to_curve.rs
```diff
@@ -78,69 +78,112 @@ impl HashToCurve {
...
+        x_coordinate.enforce_equal(&x)?;
         y_coordinate.enforce_equal(y_bit)?;
-        coordinates[2].enforce_equal(&FpVar::zero())?;
+        inf_coordinate.enforce_equal(&Boolean::constant(false))?;
 
-        // We don't need to scale by the cofactor since MNT6-753 has a cofactor of one.
+        // TODO! The hash point still needs to be scaled by the cofactor of the curve. Otherwise the
+        //  signature won't verify.
+        // We now scale by the cofactor.
+        //let scaled_hash_var = hash_var_affine.fixed_scalar_mul_le()?;
```
paberr, line 145: Why didn't you implement this part?

paberr (commented):
A few comments already. You also did mention the spammer. Does it require changes?

paberr:
> A few comments already. You also did mention the spammer. Does it require changes?

Okay, I see. The spammer is not compiling (but it is on the main branch).

brunoffranca:
> > A few comments already. You also did mention the spammer. Does it require changes?
> 
> Okay, I see. The spammer is not compiling (but it is on the main branch).

It wasn't building on the main chain also, I think. I need to look into this.

bls/src/types/compressed_public_key.rs
brunoffranca: It is correct. A single field element is 95 bytes, three field elements are 285 bytes.

nano-zkp/src/gadgets/mnt4/hash_to_curve.rs
brunoffranca: Because it is no longer a Fp, it is a Fp3. And I couldn't find any equivalent method for Fp3Var.

nano-zkp/src/gadgets/mnt4/hash_to_curve.rs
brunoffranca: Honestly, because it is more complicated than it seems and is not really needed now. I prefer to merge this PR first and then I can implement this part when I'm doing the Poseidon hash PR.

PR #583, opened 2022-01-28
Include unittest to create chain with txns

viquezclaudio:
Created a new unittest that can be used to exercise a chain
with transactions indefinitely

### DIFF
--- a/block-production/Cargo.toml
+++ b/block-production/Cargo.toml
@@ -42,7 +42,8 @@ nimiq-utils = { path = "../utils"}
 nimiq-test-utils = { path = "../test-utils" }
 nimiq-transaction-builder = { path = "../transaction-builder" }
 nimiq-keys = { path = "../keys" }
-
+tempdir = "0.3"
+rand = "0.8"
 [features]
 default = []
 test-utils = []
--- a/block-production/tests/mod.rs
+++ b/block-production/tests/mod.rs
@@ -1,20 +1,24 @@
+use parking_lot::RwLock;
 use std::convert::TryInto;
 use std::sync::Arc;
-
-use parking_lot::RwLock;
+use tempdir::TempDir;
 
 use beserial::Deserialize;
 use nimiq_block::{Block, BlockError, ForkProof};
 use nimiq_block_production::BlockProducer;
 use nimiq_blockchain::{AbstractBlockchain, Blockchain, PushError, PushResult};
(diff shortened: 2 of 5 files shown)


block-production/Cargo.toml
```diff
@@ -20,6 +20,7 @@ maintenance = { status = "experimental" }
 log = "0.4"
 hex = "0.4"
 parking_lot = { git = "https://github.com/styppo/parking_lot.git" }
+rand = "0.8"
```
jsdanielh: Maybe a `dev` dependency instad?

block-production/tests/mod.rs
```diff
@@ -1,24 +1,28 @@
...
 use nimiq_test_utils::blockchain::{
-    fill_micro_blocks, sign_macro_block, sign_view_change, signing_key, voting_key,
+    fill_micro_blocks, fill_micro_blocks_with_txns, sign_macro_block, sign_view_change,
+    signing_key, voting_key,
 };
 use nimiq_transaction_builder::TransactionBuilder;
 use nimiq_utils::time::OffsetTime;
 use nimiq_vrf::VrfSeed;
+use parking_lot::RwLock;
+use tempdir::TempDir;
```
jsdanielh: We normally separate internal vs external dependencies... I would put them above as they were

test-utils/src/blockchain.rs
```diff
@@ -1,5 +1,8 @@
+use std::str::FromStr;
 use std::sync::Arc;
+use std::time::Instant;
 
+use nimiq_primitives::coin::Coin;
```
jsdanielh: I would place them below with nimiq dependencies

block-production/tests/mod.rs
```diff
@@ -27,6 +31,8 @@ pub const ACCOUNT_SECRET_KEY: &str =
 
 const STAKER_ADDRESS: &str = "NQ20TSB0DFSMUH9C15GQGAGJTTE4D3MA859E";
 
+const VOLATILE_ENV: bool = false;
```
jsdanielh: I would consider leaving the default in `true` since most of our tests run in a volatile environment

jsdanielh (changes requested):

viquezclaudio:
changes applied @jsdanielh

jsdanielh (approved):

PR #594, opened 2022-02-07
mempool: Reject messages for certain transaction errors

jsdanielh:
Report a rejection to the `Gossipsub` validation result when the
transaction verification fails and the error is one of these two
cases:
- Transaction has an invalid signature.
- Transaction is not valid for the current validation window.

This would make that upon receiving these type of messages, the
sending peer score would be penalized.

## Pull request checklist

### DIFF
--- a/mempool/src/verify.rs
+++ b/mempool/src/verify.rs
@@ -33,8 +33,14 @@ pub enum SignVerifReturnCode {
 pub enum VerifyErr {
     /// Sender doesn't have enough funds.
     NotEnoughFunds,
-    /// Transaction is invalid
-    Invalid,
+    /// Transaction signature is invalid
+    InvalidSignature,
+    /// Transaction not valid for the validation window
+    InvalidTxWindow,
+    /// Transaction not valid for the current block height
+    InvalidBlockHeight,
+    /// Transaction sender doesn't exist
+    InvalidSender,
     /// Transaction is already known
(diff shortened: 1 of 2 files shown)


viquezclaudio (approved):

mempool/src/verify.rs
```diff
@@ -123,12 +138,12 @@ pub(crate) async fn verify_tx<'a>(
...
     if !transaction.is_valid_at(block_height) {
         log::debug!("Transaction invalid at block {}", block_height);
-        return Err(VerifyErr::Invalid);
+        return Err(VerifyErr::InvalidBlockHeight);
     }
 
     if blockchain.contains_tx_in_validity_window(&transaction.hash(), None) {
         log::debug!("Transaction has already been mined");
-        return Err(VerifyErr::Invalid);
+        return Err(VerifyErr::InvalidTxWindow);
```
sisou, line 146: When I first saw the `InvalidTxWindow` error, I thought it would be the error that the transaction is invalid because of block height (the window already passed, or hasn't started). I see now that that would be the `InvalidBlockHeight` error.

The naming might be confusing, because an invalid block height is also an invalid tx window at the same time.

How about renaming `VerifyErr::InvalidTxWindow` => `VerifyErr::AlreadyMined`, or `AlreadyIncluded`?

PR #598, opened 2022-02-10
Add prometheus metrics to the spammer

jsdanielh:
Added some prometheus metrics to the spammer to be exported in a
fixed port: 9501. For now only these metrics are exported:
- Block time
- Transactions per second
- Block number

This is under a feature called `metrics` in the spammer that can be
optionally enabled with the `-m` or `--metrics` argument in the
`devnet.sh` script.

## Pull request checklist

### DIFF
--- a/spammer/src/main.rs
+++ b/spammer/src/main.rs
@@ -3,8 +3,14 @@ use std::path::PathBuf;
 use std::str::FromStr;
 
 use futures::StreamExt;
+#[cfg(feature = "metrics")]
+use lazy_static::lazy_static;
+#[cfg(feature = "metrics")]
+use prometheus::{IntGauge, Registry};
 use rand::{thread_rng, RngCore};
 use structopt::StructOpt;
+#[cfg(feature = "metrics")]
+use warp::{Filter, Rejection, Reply};
 
 use nimiq::client::ConsensusProxy;
 pub use nimiq::{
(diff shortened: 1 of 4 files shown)


spammer/src/main.rs
```diff
@@ -70,6 +87,63 @@ struct StatsExert {
...
+    let metrics_route = warp::path!("metrics").and_then(metrics_handler);
+
+    warp::serve(metrics_route).run(([0, 0, 0, 0], 9501)).await;
+}
+
+#[cfg(feature = "metrics")]
+async fn update_metric_counters(block_time: i64, spammer_tps: i64) {
+    TPS.set(spammer_tps);
+    BLOCK_TIME.set(block_time);
+    BLOCK_NUMBER.inc();
```
viquezclaudio: Can we change this to a gauge, just as block time or block number?
and also obtain the block number from the blockchain itself, because as it is now, if you kill/restart the spammer, the block number will restart at 0 and not from the current block number

viquezclaudio (changes requested):

spammer/src/main.rs
jsdanielh: Done @viquezclaudio

spammer/src/main.rs
```diff
@@ -70,6 +87,63 @@ struct StatsExert {
...
+#[cfg(feature = "metrics")]
+async fn register_custom_metrics() {
+    REGISTRY
+        .register(Box::new(TPS.clone()))
+        .expect("collector can be registered");
+
+    REGISTRY
+        .register(Box::new(BLOCK_TIME.clone()))
+        .expect("collector can be registered");
+
```
viquezclaudio, line 136: you are missing the block number registration

viquezclaudio (changes requested):

spammer/src/main.rs
jsdanielh: Yeah, you're right. Changes made @viquezclaudio

viquezclaudio (approved):

PR #630, opened 2022-03-04
Change blake2b hash function to blake3

hrxi:
See https://blake3.io/. This should improve performance a little, as
blake3 is substantially faster than blake2b. In a tiny
microbenchmark for a payload size of 32 bytes, it was 30% faster than
blake2b.

TPS in the devnet test script varied so much across runs that no
definite improvement could be determined in the global context.

---

Also add benchmarks for hashes, allow to compute nimiq addresses from private keys with the nimiq-address tool.

---

This includes #629 because separating it would unfortunately cause a merge conflict in `genesis/src/networks.rs`.


hrxi:
I'm not entirely sure whether changing blake2b for blake3 is still worth it if we can't even drop the dependency on the blake2-rfc crate tbh.

hrxi:
As discussed with @styppo, we're probably not going to do it given that we don't see benefits.

PR #656, opened 2022-03-10
Albagen integration to devnet scripts

viquezclaudio:
Add albagen as one possible service for the docker-compose file that is created by the
devnet_create script


sisou:
What is albagen? I can't find anything with Google (except a drug) and Docker hub doesn't know it either. No results on Github.

PR #684, opened 2022-03-14
Some style fixes, fix/ignore clippy warnings, failure detection in devnet script

hrxi:
Exit on errors in the devnet script
---

This e.g. stops the script if the Rust compiler returns an error
compiling the project.

Generally, having
```sh
set -o errexit
set -o nounset
set -o pipefail
```
on top of bash scripts is a good idea.

`errexit` exits the script with a non-zero exit status if any called
command returns a non-zero exit status (and is itself not inside an `if`
or similar). A few commands return a non-zero exit status on non-error
conditions, e.g. `grep`, which then might make stuff like `|| true` or
`|| test $? = 1` necessary to ignore all or specific non-zero exit
statuses.

`nounset` exits the script if an unset variable is used.

`pipefail` extends the `errexit` check to commands that are not the
final part of a pipeline.

Fix or disable all remaining clippy warnings
---

Warnings in general and also clippy warnings only have a useful impact
if they're not routinely ignored. This fixes most sensible clippy
warnings.

The followings warnings were disabled globally:

`clippy::large_enum_variant` is a warning that fires when one of an
enum's variants is 200 bytes larger than the next one.

`clippy::too_many_arguments` is a warning that warns for functions with
more than 7 parameters.

`clippy::type_complexity` warns about "complex types".

These three lints don't show obvious problems that *need* to be fixed
and were thus mostly ignored.


hrxi:
I only grouped these changes together because they're small and don't change functionality. Obviously, if only some of them are desired, or if some need further discussion, I can easily move them to a different pull request.

PR #694, opened 2022-03-17
Spammer enhancements part 1

viquezclaudio:
- The spammer can use a generation profile to control different tx functionality
- It now includes Bursty Basic Transactions and the possibility to send txns from accounts that are created
- By default, the behaviour of the spammer is the same as it was before the enhancements.
- Vesting contracts are also generated, along with redeem transactions accordingly.

### DIFF
--- a/spammer/src/spammer_profile.toml
+++ b/spammer/src/spammer_profile.toml
@@ -1,0 +1,10 @@
+#[generation]
+# Generation weights in terms of: BaseBasicTransactions, BurstyBasicTransactions, Vesting Contracts
+weights = [ 7, 2, 1]
+
+# Probability of sending from existing accounts to other accounts
+many_to_many = 0.5
+
+# The Base of BasicTransactions that are generated per block. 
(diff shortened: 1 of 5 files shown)


spammer/src/spammer_profile.toml
```diff
@@ -0,0 +1,10 @@
+#[generation]
+# Generation weights in terms of: BaseBasicTransactions, BurstyBasicTransactions, Vesting Contracts
+weights = [ 7, 2, 1]
+
+# Probability of sending from existing accounts to other accounts
+many_to_many = 0.5
+
+# The Base of BasicTransactions that are generated per block. 
+# Bursty transactions are generated in the range [tpb * 10, tpb * 20]
+tpb = 50
```
jsdanielh, line 10: Looks like this could go to our config file without needing to create another configuration source

viquezclaudio: Since the spammer is some testing-only tool, I decided to not mix our config file with spammer stuff..

PR #707, opened 2022-03-21
Use the matrix feature of Github Actions to reduce duplication

hrxi:


hrxi:
I tested that the devnet scenarios Github Action still actually works.

viquezclaudio (approved):

PR #718, opened 2022-03-24
Remove most `#![allow(dead_code)]`

hrxi:
One that remains is on `nano_zkp::poseidon` added in
c8381852c2233bd48ccd3d49257baf8767c85b69 by Pascal that is planned to be
used later.

The `NetworkAction::Validate` was made obsolete by commit
e2de15e213e804e8388ce73fe368059ad4864990 but wasn't noticed back then
because of the `#![allow(dead_code)]`.


nano-zkp/src/lib.rs
```diff
@@ -1,5 +1,3 @@
-#![allow(dead_code)]
```
jsdanielh, line 1: Looks like this generated warnings of unused elements in `nano-zkp/src/gadgets/mnt4/merkle_tree.rs`, `nano-zkp/src/gadgets/mnt6/serialize.rs` and  `nano-zkp/src/gadgets/mnt6/y_to_bit.rs`. Could you check them?

jsdanielh (changes requested):

nano-zkp/src/lib.rs
hrxi: Fixed by marking those modules as `#[allow(dead_code)]` since I'd guess that we want to use them at some point.

PR #719, opened 2022-03-24
Split the `connections_are_properly_closed` test into two

### LINKED ISSUE #708: `network-libp2p`'s `connections_are_properly_closed` sometimes fails
On current master (139045070f00a4931c9c0f42ede33c26361e3779), you can reproduce it like this:

```
$ cargo test -p nimiq-network-libp2p --test network --all-features -- none
$ while target/debug/deps/network-31361f49325afbf3 connections_are_properly_closed; do :; done
[…]
running 1 test
test connections_are_properly_closed ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 7 filtered out; finished in 0.03s


running 1 test
test connections_are_properly_closed ... FAILED

failures:

---- connections_are_properly_closed stdout ----
thread 'connections_are_properly_closed' panicked at 'assertion failed: `(left == right)`
  left: `1`,
 right: `0`', network-libp2p/tests/network.rs:415:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    connections_are_properly_closed

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 7 filtered out; finished in 0.05s

```

It also failed on CI once: https://github.com/nimiq/core-rs-albatross/runs/5633337430?check_suite_focus=true.

hrxi:
The problem with the old one was that the peers re-discovered each other
after being disconnected, hence the
`assert_eq!(net.get_peers().len(), 0)` failed sometimes.

Fix this by splitting it up into two tests, one where we drop the
assertions and one where we disconnect the networks so that the peers
can't find each other again.

Fixes #708.

### DIFF
--- a/network-libp2p/tests/network.rs
+++ b/network-libp2p/tests/network.rs
@@ -393,7 +393,7 @@ async fn both_peers_can_talk_with_each_other() {
 }
 
 #[test(tokio::test)]
-async fn connections_are_properly_closed() {
+async fn connections_are_properly_closed_events() {
     let (net1, net2) = create_connected_networks().await;
 
     let peer1 = net2.get_peer(*net1.local_peer_id()).unwrap();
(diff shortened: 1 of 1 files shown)


network-libp2p/tests/network.rs
```diff
@@ -405,15 +405,33 @@ async fn connections_are_properly_closed() {
...
+    let (net1, net2) = create_connected_networks().await;
+
+    let peer1 = net2.get_peer(*net1.local_peer_id()).unwrap();
+
+    let mut events1 = net1.subscribe_events();
+
+    let net2_peer_id = *net2.local_peer_id();
+    drop(net2);
+
+    peer1.close(CloseReason::Other);
```
jsdanielh, line 427: I'd first close `peer1` connection and then drop `net2` since `peer1` comes from `net2`.

hrxi: I misidentified whose network `peer1` is, I thought it was a peer of `net1`. I now swapped the roles of `net1` and `net2` and it should be good now?

PR #720, opened 2022-03-24
libp2p: Increase the maximum response size

jsdanielh:
Increase the maximum response size from 2 kB to 5 kB since it was
seen in the devnet that some request easily go over 2 kB causing
a connection to be closed.


hrxi:
I guess there's some motivation? ;) Could you write it into the commit description?

jsdanielh:
> I guess there's some motivation? ;) Could you write it into the commit description?

Sure, just edited the commit and PR description.

PR #724, opened 2022-03-25
Add Loki integration to the devnet script, ship CI logs directly to our instance

hrxi:


hrxi:
Ask @jeffesquivels if you need the NIMIQ_LOKI_URL secret.

PR #738, opened 2022-03-28
Include delegation address check for mempool verif

### LINKED ISSUE #736: Mempool accepts create stakes transaction with non existent delegation address
The mempool was accepting create_staker transactions that include a delegation address that does not exist. This causes problems during the block production phase, in particular, a panic with the _Failed to compute accounts hash during block production: NonExistentAddress_ signature

viquezclaudio:
- The mempool was accepting a create staker transaction with a non existant delegation adddress,
which caused a panic during the block production phase
- Fixes #736

### DIFF
--- a/mempool/src/verify.rs
+++ b/mempool/src/verify.rs
@@ -45,6 +45,8 @@ pub enum VerifyErr {
     Known,
     /// Transaction is filtered
     Filtered,
+    /// Transaction cannot succeed
+    CannotSucceed,
 }
 
 impl Display for VerifyErr {
@@ -70,6 +72,9 @@ impl Display for VerifyErr {
             }
             VerifyErr::Filtered => {
                 write!(f, "Filtered")
+            }
+            VerifyErr::CannotSucceed => {
+                write!(f, "Cannot succeed")
             }
         }
     }
@@ -241,8 +246,8 @@ pub(crate) async fn verify_tx<'a>(
         // If the recipient is not already in the mempool, then we need to check if the transaction
         // can succeed.
         if !StakingContract::can_create(accounts_tree, &db_txn, data) {
-            log::debug!("Outgoing staking transaction cannot pay fee.");
-            return Err(VerifyErr::NotEnoughFunds);
+            log::debug!("Outgoing staking transaction cannot succeed");
+            return Err(VerifyErr::CannotSucceed);
         }
     }
 
--- a/mempool/tests/mod.rs
+++ b/mempool/tests/mod.rs
@@ -1355,6 +1355,75 @@ async fn mempool_update_create_staker_twice() {
     );
 }
 
+#[test(tokio::test(flavor = "multi_thread", worker_threads = 10))]
+async fn mempool_update_create_staker_non_existant_delegation_addr() {
+    let mut rng = StdRng::seed_from_u64(0);
+    let time = Arc::new(OffsetTime::new());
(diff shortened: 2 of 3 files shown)


primitives/account/src/staking_contract/mod.rs
```diff
@@ -411,6 +412,17 @@ impl StakingContract {
...
+
+                //Verify we have a valid delgation address (if present)
+                if let Some(delegation) = delegation {
+                    let key = StakingContract::get_key_validator(&delegation);
+                    match accounts_tree.get(db_txn, &key) {
+                        None => {
+                            return false;
+                        }
+                        _ => {}
+                    }
```
jsdanielh, line 421: Take a look at the clippy warnings generated

primitives/account/src/staking_contract/mod.rs
```diff
@@ -411,6 +412,17 @@ impl StakingContract {
                     warn!("Cannot create staker because staker already exists.");
                     return false;
                 };
+
+                //Verify we have a valid delgation address (if present)
```
jsdanielh:
```suggestion
                // Verify we have a valid delegation address (if present)
```

mempool/tests/mod.rs
```diff
@@ -1355,6 +1355,75 @@ async fn mempool_update_create_staker_twice() {
...
+            env.clone(),
+            time,
+            NetworkId::UnitAlbatross,
+            genesis_info.block,
+            genesis_info.accounts,
+        )
+        .unwrap(),
+    ));
+
+    // Create mempool and subscribe with a custom txn stream.
```
jsdanielh:
```suggestion
    // Create mempool and subscribe with a custom txn stream
```

mempool/src/verify.rs
```diff
@@ -241,7 +241,7 @@ pub(crate) async fn verify_tx<'a>(
         // If the recipient is not already in the mempool, then we need to check if the transaction
         // can succeed.
         if !StakingContract::can_create(accounts_tree, &db_txn, data) {
-            log::debug!("Outgoing staking transaction cannot pay fee.");
+            log::debug!("Outgoing staking transaction cannot succeed");
             return Err(VerifyErr::NotEnoughFunds);
```
jsdanielh: Mmm, the error indicates that this will happen because we don't have enough funds. Wondering if we should have a different error or if the `can_create` function would fail only because we can't pay the fee

jsdanielh (changes requested):

viquezclaudio:
@jsdanielh changes applied

jsdanielh (approved):

mempool/src/verify.rs
```diff
@@ -45,6 +45,8 @@ pub enum VerifyErr {
     Known,
     /// Transaction is filtered
     Filtered,
+    ///Transaction cannot succeed
```
sisou: Missing space again between forward slashes and first word.

mempool/src/verify.rs
```diff
@@ -71,6 +73,9 @@ impl Display for VerifyErr {
             VerifyErr::Filtered => {
                 write!(f, "Filtered")
             }
+            VerifyErr::CantSucceed => {
+                write!(f, "Cant succeed")
```
sisou: Either `Cannot` or `Can't`. I'm personally in favor of `Cannot` and would also change the error to be called `CannotSucceed`.

hrxi:
LGTM.

PR #743, opened 2022-03-29
Move to `tokio` channels, remove `Peer` structs and traits

hrxi:
Move to `tokio` channels all around
---

This makes it easier to integrate everything, no longer having `futures`
and `tokio` channels around.

Remove `Peer`, `MockPeer` struct and `Peer`
---

They didn't contain a lot of functionality anymore, but created a lot of
noise at the call sites. Use `PeerId`s instead, which are copyable and
can be easily stored.

The only remaining functionality of `Peer` was the `close` function
which got replaced by the `disconnect_peer` function on the `Network`
trait.


hrxi:
Contains all the commits from #717.

PR #799, opened 2022-04-26
To keep macro_state around .clone() it instead of .take() it.

nibhar:
After a restart there is a chance that a node will initialize Tendermint at least twice without calling `poll_macro()` in the Validator. The first initialization takes the `macro_state` and leaves the second initialization without one. 
This leads to faulty behaviour and with this PR the state is cloned instead of taken.

### DIFF
--- a/validator/src/validator.rs
+++ b/validator/src/validator.rs
@@ -342,7 +342,9 @@ impl<TNetwork: Network, TValidatorNetwork: ValidatorNetwork>
                     head.seed().clone(),
                     next_block_number,
                     next_view_number,
-                    self.macro_state.take(),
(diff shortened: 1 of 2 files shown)


validator/src/validator.rs
```diff
@@ -342,7 +342,9 @@ impl<TNetwork: Network, TValidatorNetwork: ValidatorNetwork>
                     head.seed().clone(),
                     next_block_number,
                     next_view_number,
-                    self.macro_state.take(),
+                    // This cannot be .take() as there is a chance init_epoch is called multiple times without
+                    // poll_maceo being called creating a new macro_state that is Some(...).
```
jsdanielh:
```suggestion
                    // poll_macro being called creating a new macro_state that is Some(...).
```

PR #825, opened 2022-05-09
Fix bug in devnet script

viquezclaudio:
The devnet script was not properly detecting the best chain and
was not taking into consideration the batch size when performing that operation.

### DIFF
--- a/scripts/devnet/devnet.sh
+++ b/scripts/devnet/devnet.sh
@@ -28,6 +28,7 @@ devnet_create="scripts/devnet/python/devnet_create.py"
 tpb=150
 vkill=1
 down_time=10
+batch_size=32
 trap cleanup_exit INT
 
 function cleanup_exit() {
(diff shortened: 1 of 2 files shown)


scripts/devnet/devnet.sh
```diff
@@ -384,7 +395,16 @@ do
     for bn in "${bns[@]}" ; do
         block_number=$(echo $bn | cut -d "." -f 1)
         view_number=$(echo $bn | cut -d "." -f 2)
-        if [ $block_number -gt $new_block_number ] || [ $view_number -gt $new_view_number ] ; then
+
+        #First we need to detect if the blocks being compared correspond to the same batch
```
jsdanielh:
```suggestion
        # First we need to detect if the blocks being compared correspond to the same batch
```

scripts/devnet/devnet.sh
```diff
@@ -384,7 +395,16 @@ do
...
+        batch_number1=$((block_number/$batch_size))
+        batch_number2=$((new_block_number/$batch_size))
+
+        #If they are in the same batch, the greatest view number wins
```
jsdanielh:
```suggestion
        # If they are in the same batch, the greatest view number wins
```

viquezclaudio:
Changes applied @jsdanielh

jsdanielh (approved):

PR #833, opened 2022-05-11
Adjust transaction validity check

viquezclaudio:
A transaction is valid if:
- The validity start height minus a batch is less than the block height
  (this is to give some time to all validators to be in sync and catch up,
   otherwise the tx can be rejected by some of them if the one sending the tx
   is at the tip of the chain and others are a bit behind)
- The transaction its within the transaction validity window


viquezclaudio:
This fixes #834

PR #839, opened 2022-05-11
Exclude dependabot pushes from clippy runs

hrxi:
They always fail due to permission problems.


hrxi:
I'm just going to try it out, I guess.

PR #840, opened 2022-05-11
Fix oldest transaction ordering in mempool

### LINKED ISSUE #838: Mempool pruning of oldest transactions is not working properly
During a mempool execution, it was noticed that old transactions were not properly pruned, and remained in the mempool

viquezclaudio:
Transactios in mempool's oldest_transactions structure were being ordered from
oldest to newest, so the peak function was actually returning the newest one,
and what we neeeded is to obtain the oldest one to properly prune old transactions.

Fixes #838


hrxi:
LGTM.

PR #844, opened 2022-05-16
Include mempool test for aged txns

viquezclaudio:
Create a unittest that verifies that aged txns (the ones that expire because of validity window)
are porperly pruned/removed from mempool.
Test is currently disabled because it takes some time to produce a chain that triggers this condition
However, the functionality was tested by manually compiling and testing a client with a lower
validity window limit

### DIFF
--- a/mempool/tests/mod.rs
+++ b/mempool/tests/mod.rs
@@ -21,10 +21,10 @@ use nimiq_keys::{
 use nimiq_mempool::config::MempoolConfig;
 use nimiq_mempool::mempool::Mempool;
 use nimiq_network_mock::{MockHub, MockId, MockNetwork, MockPeerId};
-use nimiq_primitives::networks::NetworkId;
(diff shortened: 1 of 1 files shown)


mempool/tests/mod.rs
```diff
@@ -968,6 +968,125 @@ async fn mempool_update() {
...
+    assert_eq!(
+        mempool.num_transactions(),
+        0,
+        "Number of txns in the mempools is not what is expected"
+    );
+
+    // Get txns from mempool
+    let updated_txns = mempool.get_transactions_for_block(10_000);
+
+    // Should obtain 0 txns, as they are no longer valid due to aging)
```
jsdanielh:
```suggestion
    // Should obtain 0 txns, as they are no longer valid due to aging
```

jsdanielh (approved):

PR #845, opened 2022-05-16
network-libp2p: Do not use reserved word for logging

jsdanielh:
Change a couple of log messages to avoid using `message` as a
field in log entries since it is a reserved word in tracing.


hrxi:
Thanks for the fix. :)

PR #851, opened 2022-05-17
RPC: Move get_transaction_by_hash function from mempool into blockchain dispatcher

sisou:
Seed nodes do not have a mempool, because they do not run a validator, so this change makes the method accessible on the devnet seed node. It loses the ability to query the mempool, but transactions are confirmed (included into block) near-instantly, so mempool querying should not be necessary.

## Pull request checklist

Tested and verified with a local node.


jeffesquivels (approved):
LGTM.

PR #855, opened 2022-05-18
Add tokio metrics and remove old spammer metrics

jsdanielh:
Remove old spammer metrics and add tokio metrics like:
- Tokio runtime metrics.
- Tokio task metrics for:
  - Consensus
  - Mempool
  - Validator

Also added changes to instrument the above tasks.

## Pull request checklist

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -2,6 +2,7 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use log::info;
+
 pub use nimiq::{
     client::{Client, Consensus},
     config::command_line::CommandLine,
@@ -11,6 +12,7 @@ pub use nimiq::{
     extras::{
         deadlock::initialize_deadlock_detection,
         logging::{initialize_logging, log_error_cause_chain},
+        metrics_server::NimiqTaskMonitor,
         panic::initialize_panic_reporting,
         signal_handling::initialize_signal_handler,
     },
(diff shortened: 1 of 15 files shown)


metrics-server/src/lib.rs
```diff
@@ -82,5 +98,51 @@ pub fn start_metrics_server<TNetwork: Network>(
...
+    #[cfg(tokio_unstable)]
+    let handle = tokio::runtime::Handle::current();
+    #[cfg(tokio_unstable)]
+    let tokio_rt_monitor = RuntimeMonitor::new(&handle);
+    #[cfg(tokio_unstable)]
+    let mut tokio_rt_metrics = TokioRuntimeMetrics::new();
+    #[cfg(tokio_unstable)]
+    tokio_rt_metrics.register(nimiq_registry);
+    #[cfg(tokio_unstable)]
+    let tokio_rt_metrics = Arc::new(RwLock::new(tokio_rt_metrics));
```
fiaxh: You could put the code in a condition-block instead of writing the condition in front of every line.
```rust
#[cfg(tokio_unstable)] {
   [code]
}
```

client/src/main.rs
```diff
@@ -64,12 +66,38 @@ async fn main_inner() -> Result<(), Error> {
...
-    tokio::spawn(consensus);
+    let con_metrics_monitor = tokio_metrics::TaskMonitor::new();
+    let instr_con = con_metrics_monitor.instrument(consensus);
+    tokio::spawn(instr_con);
     let consensus = client.consensus_proxy();
+    nimiq_task_metric.push(NimiqTaskMonitor {
+        name: "consensus".to_string(),
+        monitor: con_metrics_monitor,
+    });
+
```
fiaxh, line 89: Having all this rather specific logic in the client/main.rs feels rather unfortunate to me, since the code here is otherwise fairly high-level. However, I also can't think of a nice solution. Maybe only do the instrumenting if ~the `metrics-server` feature is enabled~ `metrics_config.is_some()`, and maybe move construction of the `nimiq_task_metric` down to the metrics-server code.

viquezclaudio (approved):

PR #857, opened 2022-05-18
Cache accounts number so it can be queried cheaply

hrxi:
Make the root node of the trie special so it can store some data about
the whole trie.


hrxi:
Also added a counter for the number of non-leaf nodes.

hrxi:
Superseded by #1115.

PR #859, opened 2022-05-18
Add accounts tree perf test

viquezclaudio:
Add a couple of accounts tree performance test to simulate what we do during history sync.
The goal of the test is to measure is how long it takes to push batches, in two different configurations:
- Single sender, multiple random recipients
- Multiple sender, multiple recipients all existing

### DIFF
--- a/primitives/account/tests/accounts.rs
+++ b/primitives/account/tests/accounts.rs
@@ -12,6 +12,7 @@ use nimiq_genesis_builder::GenesisBuilder;
 use nimiq_keys::{Address, KeyPair, PublicKey, SecureGenerate};
 use nimiq_primitives::coin::Coin;
 use nimiq_primitives::networks::NetworkId;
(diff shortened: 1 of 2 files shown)


primitives/account/tests/accounts.rs
```diff
@@ -525,3 +526,247 @@ fn accounts_performance() {
...
+    let start = Instant::now();
+    txn.commit();
+    let duration = start.elapsed();
+    log::debug!(
+        "Time elapsed after account init's txn commit: {} ms, Accounts per second {}",
+        duration.as_millis(),
+        num_txns as f64 / (duration.as_millis() as f64 / 1000_f64),
+    );
+
+    log::debug!("Done adding accounts to genesis ");
```
jsdanielh:
```suggestion
    log::debug!("Done adding accounts to genesis");
```

jsdanielh (approved):

PR #865, opened 2022-05-24
Serialize Signatures to HEX for display and JSON

sisou:
Currently, signatures are serialized into an array of numbers. That is not really usable via RPC or when debugging, I think.

This PR changes the Signature serialization to the same way the [PublicKey](https://github.com/nimiq/core-rs-albatross/blob/d4cf96dfd5c57fd5caff4cea58ac9a1ff763cc38/keys/src/public_key.rs) is serialized.

I don't know what `std::borrow::Cow` is and if it's really needed, I just copy-pasted it from PublicKey.

### DIFF
--- a/keys/src/signature.rs
+++ b/keys/src/signature.rs
@@ -1,4 +1,5 @@
 use std::convert::TryFrom;
+use std::fmt;
 use std::str::FromStr;
 
 use hex::FromHex;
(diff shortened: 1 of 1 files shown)


jsdanielh (approved):
Looks good to me

PR #868, opened 2022-05-25
Fix error in zkp setup

paberr:
## Summary
This pull request fixes the ZKP setup phase.
It also introduces a new optimised build profile with debug symbols, which is useful for ZKPs (as they are too slow otherwise).

## Problem being fixed
During the ZKP setup, no specific inputs are required (which is why inputs are always declared in closures).
In fact, the setup often deliberately sets these inputs to `Err(SynthesisError::AssignmentMissing)` during the setup.
However, the setup was failing with exactly this error.

The problem was that the vector allocation `Vec::new_witness` does evaluate the closure and passes on the error.
The reason is simple: it needs to have a value to know the size of the vector to be allocated on the circuit.

## Proposed fix
The fix is to return a dummy value of the correct length in this case.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,3 +76,7 @@ opt-level = 3
 
 [profile.release]
 lto = "thin"
+
+[profile.release-with-debug]
+inherits = "release"
+debug = true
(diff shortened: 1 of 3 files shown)


paberr:
Recommitted with GPG signature

jsdanielh (approved):

PR #885, opened 2022-06-14
Second (control) mempool implementation

viquezclaudio:
- Introduced a second mempool, used for control transactions
- When blocks are produced, first we try to include control txns
- If there is space left, we include regular txns
- This means that control txns have a higher priority over reg txns.

### DIFF
--- a/consensus/src/consensus/mod.rs
+++ b/consensus/src/consensus/mod.rs
@@ -6,6 +6,7 @@ use std::task::{Context, Poll};
 use std::time::{Duration, Instant};
 
 use futures::{FutureExt, StreamExt};
+use nimiq_primitives::account::AccountType;
 use parking_lot::RwLock;
 use tokio::sync::broadcast::{channel as broadcast, Sender as BroadcastSender};
 use tokio::time::Sleep;
@@ -13,7 +14,7 @@ use tokio_stream::wrappers::BroadcastStream;
 
 use nimiq_blockchain::{AbstractBlockchain, Blockchain};
 use nimiq_database::Environment;
-use nimiq_mempool::mempool::TransactionTopic;
+use nimiq_mempool::mempool::{ControlTransactionTopic, TransactionTopic};
 use nimiq_network_interface::network::Network;
 use nimiq_transaction::Transaction;
 
@@ -42,6 +43,9 @@ impl<N: Network> Clone for ConsensusProxy<N> {
 
 impl<N: Network> ConsensusProxy<N> {
     pub async fn send_transaction(&self, tx: Transaction) -> Result<(), N::Error> {
+        if tx.sender_type == AccountType::Staking || tx.recipient_type == AccountType::Staking {
+            return self.network.publish::<ControlTransactionTopic>(tx).await;
+        }
         self.network.publish::<TransactionTopic>(tx).await
     }
 
--- a/mempool/src/executor.rs
+++ b/mempool/src/executor.rs
@@ -13,7 +13,7 @@ use nimiq_primitives::networks::NetworkId;
 use nimiq_transaction::Transaction;
 
 use crate::filter::MempoolFilter;
-use crate::mempool::{MempoolState, TransactionTopic};
+use crate::mempool::{ControlTransactionTopic, MempoolState, TransactionTopic};
 use crate::verify::{verify_tx, VerifyErr};
 
 const CONCURRENT_VERIF_TASKS: u32 = 1000;
@@ -31,7 +31,7 @@ pub(crate) struct MempoolExecutor<N: Network> {
     // Ongoing verification tasks counter
     verification_tasks: Arc<AtomicU32>,
 
-    // Reference to the network, to alow for message validation
+    // Reference to the network, to allow for message validation
     network: Arc<N>,
 
     // Network ID, used for tx verification
@@ -104,6 +104,103 @@ impl<N: Network> Future for MempoolExecutor<N> {
                 };
 
                 network.validate_message::<TransactionTopic>(pubsub_id, acceptance);
+
+                tasks_count.fetch_sub(1, AtomicOrdering::SeqCst);
+            });
+        }
+
+        // We have exited the loop, so poll_next() must have returned Poll::Ready(None).
+        // Thus, we terminate the executor future.
+        Poll::Ready(())
+    }
+}
+
+pub(crate) struct ControlMempoolExecutor<N: Network> {
+    // Blockchain reference
+    blockchain: Arc<RwLock<Blockchain>>,
+
+    // The mempool state: the data structure where the transactions are stored
+    state: Arc<RwLock<MempoolState>>,
+
+    // Mempool filter
+    filter: Arc<RwLock<MempoolFilter>>,
+
+    // Ongoing verification tasks counter
+    verification_tasks: Arc<AtomicU32>,
+
+    // Reference to the network, to allow for message validation
+    network: Arc<N>,
+
+    // Network ID, used for tx verification
+    network_id: Arc<NetworkId>,
+
+    // Transaction stream that is used to listen to transactions from the network
+    txn_stream: BoxStream<'static, (Transaction, <N as Network>::PubsubId)>,
+}
+
+impl<N: Network> ControlMempoolExecutor<N> {
+    pub fn new(
+        blockchain: Arc<RwLock<Blockchain>>,
+        state: Arc<RwLock<MempoolState>>,
+        filter: Arc<RwLock<MempoolFilter>>,
+        network: Arc<N>,
+        txn_stream: BoxStream<'static, (Transaction, <N as Network>::PubsubId)>,
+    ) -> Self {
+        Self {
+            blockchain: blockchain.clone(),
+            state,
+            filter,
+            network,
+            network_id: Arc::new(blockchain.read().network_id),
+            verification_tasks: Arc::new(AtomicU32::new(0)),
+            txn_stream,
+        }
+    }
+}
+
+impl<N: Network> Future for ControlMempoolExecutor<N> {
+    type Output = ();
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        while let Some((tx, pubsub_id)) = ready!(self.txn_stream.as_mut().poll_next_unpin(cx)) {
+            if self.verification_tasks.fetch_add(0, AtomicOrdering::SeqCst)
+                >= CONCURRENT_VERIF_TASKS
+            {
+                log::debug!("Reached the max number of verification tasks");
+                continue;
+            }
+
+            let blockchain = Arc::clone(&self.blockchain);
+            let mempool_state = Arc::clone(&self.state);
+            let filter = Arc::clone(&self.filter);
+            let tasks_count = Arc::clone(&self.verification_tasks);
+            let network_id = Arc::clone(&self.network_id);
+            let network = Arc::clone(&self.network);
+
+            // Spawn the transaction verification task
+            tokio::task::spawn(async move {
+                tasks_count.fetch_add(1, AtomicOrdering::SeqCst);
+
+                // Verifying and pushing the TX in a separate scope to drop the lock that is returned by
+                // the verify_tx function immediately
+                let acceptance = {
+                    let verify_tx_ret =
+                        verify_tx(&tx, blockchain, network_id, &mempool_state, filter).await;
+
+                    match verify_tx_ret {
+                        Ok(mempool_state_lock) => {
+                            RwLockUpgradableReadGuard::upgrade(mempool_state_lock).put(&tx);
+                            MsgAcceptance::Accept
+                        }
+                        // Reject the message if signature verification fails or transaction is invalid
+                        // for current validation window
+                        Err(VerifyErr::InvalidSignature) => MsgAcceptance::Reject,
+                        Err(VerifyErr::InvalidTxWindow) => MsgAcceptance::Reject,
+                        Err(_) => MsgAcceptance::Ignore,
+                    }
+                };
+
+                network.validate_message::<ControlTransactionTopic>(pubsub_id, acceptance);
 
                 tasks_count.fetch_sub(1, AtomicOrdering::SeqCst);
             });
--- a/mempool/src/executor.rs
+++ b/mempool/src/executor.rs
@@ -1,4 +1,5 @@
 use std::future::Future;
+use std::marker::PhantomData;
 use std::pin::Pin;
 use std::sync::atomic::{AtomicU32, Ordering as AtomicOrdering};
 use std::sync::Arc;
@@ -8,17 +9,17 @@ use futures::{ready, stream::BoxStream, StreamExt};
 use parking_lot::{RwLock, RwLockUpgradableReadGuard};
 
 use nimiq_blockchain::Blockchain;
-use nimiq_network_interface::network::{MsgAcceptance, Network};
+use nimiq_network_interface::network::{MsgAcceptance, Network, Topic};
 use nimiq_primitives::networks::NetworkId;
 use nimiq_transaction::Transaction;
 
 use crate::filter::MempoolFilter;
-use crate::mempool::{ControlTransactionTopic, MempoolState, TransactionTopic};
+use crate::mempool_state::MempoolState;
 use crate::verify::{verify_tx, VerifyErr};
 
 const CONCURRENT_VERIF_TASKS: u32 = 1000;
 
-pub(crate) struct MempoolExecutor<N: Network> {
+pub(crate) struct MempoolExecutor<N: Network, T: Topic + Unpin + Sync> {
     // Blockchain reference
     blockchain: Arc<RwLock<Blockchain>>,
 
@@ -39,9 +40,12 @@ pub(crate) struct MempoolExecutor<N: Network> {
 
     // Transaction stream that is used to listen to transactions from the network
     txn_stream: BoxStream<'static, (Transaction, <N as Network>::PubsubId)>,
+
+    // Phantom data for the unused type T
+    _phantom: PhantomData<T>,
 }
 
-impl<N: Network> MempoolExecutor<N> {
+impl<N: Network, T: Topic + Unpin + Sync> MempoolExecutor<N, T> {
     pub fn new(
         blockchain: Arc<RwLock<Blockchain>>,
         state: Arc<RwLock<MempoolState>>,
@@ -57,11 +61,12 @@ impl<N: Network> MempoolExecutor<N> {
             network_id: Arc::new(blockchain.read().network_id),
             verification_tasks: Arc::new(AtomicU32::new(0)),
             txn_stream,
+            _phantom: PhantomData,
         }
     }
 }
 
-impl<N: Network> Future for MempoolExecutor<N> {
+impl<N: Network, T: Topic + Unpin + Sync> Future for MempoolExecutor<N, T> {
     type Output = ();
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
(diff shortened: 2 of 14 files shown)


consensus/src/consensus/mod.rs
```diff
@@ -42,6 +43,9 @@ impl<N: Network> Clone for ConsensusProxy<N> {
 
 impl<N: Network> ConsensusProxy<N> {
     pub async fn send_transaction(&self, tx: Transaction) -> Result<(), N::Error> {
+        if tx.sender_type == AccountType::Staking || tx.recipient_type == AccountType::Staking {
+            return self.network.publish::<ControlTransactionTopic>(tx).await;
+        }
         self.network.publish::<TransactionTopic>(tx).await
```
sisou, line 49: So control transactions are relayed on both the control-tx topic and also still the regular tx topic?

mempool/src/executor.rs
```diff
@@ -114,3 +114,100 @@ impl<N: Network> Future for MempoolExecutor<N> {
...
+    // The mempool state: the data structure where the transactions are stored
+    state: Arc<RwLock<MempoolState>>,
+
+    // Mempool filter
+    filter: Arc<RwLock<MempoolFilter>>,
+
+    // Ongoing verification tasks counter
+    verification_tasks: Arc<AtomicU32>,
+
+    // Reference to the network, to alow for message validation
```
sisou:
```suggestion
    // Reference to the network, to allow for message validation
```

mempool/src/mempool.rs
```diff
@@ -89,17 +99,24 @@ impl Mempool {
...
         }
     }
 
-    /// Starts the mempool executor
+    /// Starts the mempool executors
     ///
     /// Once this function is called, the mempool executor is spawned.
     /// The executor will subscribe to the transaction topic from the the network.
-    pub async fn start_executor<N: Network>(&self, network: Arc<N>, monitor: Option<TaskMonitor>) {
+    pub async fn start_executor<N: Network>(
```
sisou: Should this function (and its descriptive comments) not be renamed to executor**s** (plural)?

mempool/src/mempool.rs
```diff
@@ -89,17 +99,24 @@ impl Mempool {
...
+        &self,
+        network: Arc<N>,
+        monitor: Option<TaskMonitor>,
+        ctrl_monitor: Option<TaskMonitor>,
```
sisou: Suggestion: Do not abbreviate `control`  to `ctrl` here, as you are not doing it anywhere else (so far). Should be consistently named `control_monitor`.

mempool/src/mempool.rs
```diff
@@ -127,6 +144,34 @@ impl Mempool {
...
+        let future = Abortable::new(control_mempool_executor, abort_registration);
+
+        if let Some(monitor) = ctrl_monitor {
+            tokio::spawn(monitor.instrument(future));
+        } else {
+            tokio::spawn(future);
+        }
+
+        // Set the control mempool executor handle
+        *control_executor_handle = Some(abort_handle);
```
sisou: This code is the exact same as in the `start_control_executor_with_txn_stream` function below. Could you call that function here instead?

mempool/src/mempool.rs
```diff
@@ -162,14 +207,48 @@ impl Mempool {
...
+    ///
+    /// Once this function is called, the mempool executor is spawned.
+    /// The executor won't subscribe to the transaction topic from the network but will use the provided transaction
+    /// stream instead.
+    pub async fn start_control_executor_with_txn_stream<N: Network>(
+        &self,
+        txn_stream: BoxStream<'static, (Transaction, <N as Network>::PubsubId)>,
+        network: Arc<N>,
+    ) {
+        let mut executor_handle = self.control_executor_handle.lock().await;
```
sisou: Suggestion to rename. I know it's not necessary, but it will prevent confusion when scanning the function body, which of the two functions (executor vs control_executor) I'm currently in.
```suggestion
        let mut control_executor_handle = self.control_executor_handle.lock().await;
```

sisou: On a similar note: I see a lot of code duplication. Is it maybe possible to generalize the code so that the logic is only written once but called with different params for the regular executor and control_executor?

mempool/src/mempool.rs
```diff
@@ -162,14 +207,48 @@ impl Mempool {
...
+        tokio::spawn(Abortable::new(mempool_executor, abort_registration));
+
+        // Set the executor handle
+        *executor_handle = Some(abort_handle);
+    }
+
+    /// Stops the mempool executors
     ///
     /// This functions should only be called only after one of the functions to start the executor is called.
     pub async fn stop_executor<N: Network>(&self, network: Arc<N>) {
```
sisou:
```suggestion
    pub async fn stop_executors<N: Network>(&self, network: Arc<N>) {
```

mempool/src/mempool.rs
```diff
@@ -89,17 +99,24 @@ impl Mempool {
...
+        ctrl_monitor: Option<TaskMonitor>,
+    ) {
         let mut executor_handle = self.executor_handle.lock().await;
+        let mut control_executor_handle = self.control_executor_handle.lock().await;
 
-        if executor_handle.is_some() {
+        if executor_handle.is_some() && control_executor_handle.is_some() {
             // If we already have an executor running, don't do anything
             return;
         }
```
sisou: This now only returns early, if **both** of the executors are running. If only one is running, both are recreated. Or is that something that 100% cannot happen? Or is recreating an existing executor not a problem?

sisou: And since the control transactions are sent on both topics, do all node types have both mempools? Or is the control mempool only relevant for validator nodes? If so, should the control mempool then be decoupled, so that it only runs in the validator configuration?

mempool/src/mempool.rs
```diff
@@ -178,6 +257,18 @@ impl Mempool {
...
+        // Unsubscribe to the network control TX topic before killing the executor
+        network
+            .unsubscribe::<ControlTransactionTopic>()
+            .await
+            .unwrap();
+
+        // Stop the control executor
+        control_executor_handle
+            .take()
+            .expect("Expected an control executor handle")
```
sisou:
```suggestion
            .expect("Expected a control executor handle")
```

mempool/src/mempool.rs
```diff
@@ -198,6 +289,27 @@ impl Mempool {
...
+
+        if handle.is_none() {
+            // If there isn't any executor running we return
+            return;
+        }
+
+        // Stop the executor
+        handle
+            .take()
+            .expect("Expected an control executor handle")
```
sisou:
```suggestion
            .expect("Expected a control executor handle")
```

mempool/src/mempool.rs
```diff
@@ -511,11 +616,77 @@ impl Mempool {
...
+            log::debug!("Requesting control txns and there are no txns in the mempool ");
+            return (tx_vec, 0_usize);
+        }
+
+        let mut size = 0_usize;
+
+        let mut mempool_state_upgraded = RwLockUpgradableReadGuard::upgrade(state);
+
+        loop {
+            // Get the hash of the highest paying regular transaction.
```
sisou:
```suggestion
            // Get the hash of the highest paying control transaction.
```

mempool/src/mempool.rs
```diff
@@ -558,17 +729,35 @@ impl Mempool {
...
-        self.state.read().transactions.keys().cloned().collect()
+        let state = self.state.read();
+
+        state
+            .regular_transactions
+            .transactions
+            .keys()
+            .cloned()
+            .chain(state.control_transactions.transactions.keys().cloned())
+            .collect()
```
sisou, line 708: This returns regular transactions before control transactions. But control transactions are supposed to be higher importance. Should they not be listed first?

mempool/src/mempool.rs
```diff
@@ -558,17 +729,35 @@ impl Mempool {
...
-        self.state.read().transactions.values().cloned().collect()
+        let state = self.state.read();
+
+        state
+            .regular_transactions
+            .transactions
+            .values()
+            .cloned()
+            .chain(state.control_transactions.transactions.values().cloned())
+            .collect()
```
sisou: This returns regular transactions before control transactions. But control transactions are supposed to be higher importance. Should they not be listed first?

mempool/src/mempool.rs
```diff
@@ -627,21 +804,62 @@ pub(crate) struct MempoolState {
...
+
+    pub fn contains_key(&self, hash: &Blake2bHash) -> bool {
         self.transactions.contains_key(hash)
     }
 
+    pub fn is_empty(&self) -> bool {
+        self.transactions.is_empty()
+    }
+
+    //This function is used to remove the transactions that are no longer valid at a given block height
```
sisou:
```suggestion
    // This function is used to remove the transactions that are no longer valid at a given block height
```

mempool/src/mempool.rs
```diff
@@ -664,11 +882,89 @@ impl MempoolState {
                 insertion_order: self.tx_counter,
             },
         );
+
         self.tx_counter += 1;
 
         self.oldest_transactions
             .push(tx_hash.clone(), Reverse(tx.validity_start_height));
```
sisou: Clippy warning here: redundant clone

mempool/src/mempool.rs
```diff
@@ -664,11 +882,89 @@ impl MempoolState {
...
 
+        // Update total tx size and remove the cheapest ones if we have too many txs.
+        self.total_size += tx.serialized_size();
+
+        true
+    }
```
sisou: I don't understand how this comment describes what's happening. The `total_size` is increased, but I do not see any removal of a cheapest one, or a check for too many txs?

mempool/src/mempool.rs
```diff
@@ -664,11 +882,89 @@ impl MempoolState {
...
+}
+
+impl MempoolState {
+    pub fn contains(&self, hash: &Blake2bHash) -> bool {
+        self.regular_transactions.contains_key(hash) || self.control_transactions.contains_key(hash)
+    }
+
+    pub fn get(&self, hash: &Blake2bHash) -> Option<&Transaction> {
+        if let Some(transaction) = self.regular_transactions.get(hash) {
+            return Some(transaction);
```
sisou: Clippy warnings here: unneeded `return` statements

mempool/src/mempool.rs
```diff
@@ -664,11 +882,89 @@ impl MempoolState {
...
+    pub(crate) fn put(&mut self, tx: &Transaction) -> bool {
+        let tx_hash = tx.hash();
+
+        if self.regular_transactions.contains_key(&tx_hash)
+            || self.control_transactions.contains_key(&tx_hash)
+        {
+            return false;
+        }
+
+        // If we are adding a stacking transaction we insert it into the control txns container
```
sisou: This comment describes **what** the code does. Which I can see in the code. It should instead describe **why** it does it. It should explain that staking txns _are_ control txns (which is not explained anywhere else yet).
```suggestion
        // Staking txns are control txns
```

mempool/src/mempool.rs
```diff
@@ -664,11 +882,89 @@ impl MempoolState {
...
+        if tx.sender_type == AccountType::Staking || tx.recipient_type == AccountType::Staking {
+            self.control_transactions.insert(&tx);
```
sisou: Clippy warning about creating an immediately dereferenced reference

mempool/src/mempool.rs
```diff
@@ -664,11 +882,89 @@ impl MempoolState {
...
+        } else {
+            self.regular_transactions.insert(&tx);
+        }
+
+        //Update the per sender state
```
sisou:
```suggestion
        // Update the per sender state
```

mempool/src/mempool.rs
```diff
@@ -754,39 +1053,52 @@ impl MempoolState {
...
+            self.remove_from_staking_state(&tx);
 
-        if sender_state.txns.is_empty() {
-            self.state_by_sender.remove(&tx.sender);
-        }
+            #[cfg(feature = "metrics")]
+            self.metrics.note_evicted(reason);
 
-        self.remove_from_staking_state(&tx);
+            return Some(tx);
```
sisou: Clippy warning: unneeded `return` statements

mempool/src/mempool.rs
```diff
@@ -754,39 +1053,52 @@ impl MempoolState {
...
+                self.state_by_sender.remove(&tx.sender);
+            }
 
-        Some(tx)
+            self.remove_from_staking_state(&tx);
+
+            #[cfg(feature = "metrics")]
+            self.metrics.note_evicted(reason);
+
+            return Some(tx);
```
sisou: It looks to me like the code that gets run when a transaction is removed is the **same** for both regular and control transactions, because the `sender_state` is global across both types?

In that case, the code should just be declared once and run if the deleted tx existed in either tx-type bucket, e.g. by combining the condition:
```rust
if let Some(tx) = self.regular_transactions.delete(tx_hash) || self.control_transactions.delete(tx_hash) {
```
Would that be possible?

sisou: Also, Clippy warning about unneeded `return` statement

mempool/src/mempool.rs
```diff
@@ -754,39 +1053,52 @@ impl MempoolState {
...
-                self.worst_transactions.remove(tx_hash);
-                self.oldest_transactions.remove(tx_hash);
-                if let Some(tx) = self.transactions.remove(tx_hash) {
+                if let Some(tx) = self.regular_transactions.delete(tx_hash) {
+                    self.remove_from_staking_state(&tx);
+                }
+
+                if let Some(tx) = self.control_transactions.delete(tx_hash) {
                     self.remove_from_staking_state(&tx);
                 }
```
sisou: Similar to above, can this be combined?
```rust
if let Some(tx) = self.regular_transactions.delete(tx_hash) || self.control_transactions.delete(tx_hash) {
    self.remove_from_staking_state(&tx);
}
```

validator/src/micro.rs
```diff
@@ -240,9 +240,20 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
             systemtime_to_timestamp(SystemTime::now()),
         );
 
-        let transactions = self
+        //First we try to fill the block with control transactions
```
sisou:
```suggestion
        // First we try to fill the block with control transactions
```

validator/src/validator.rs
```diff
@@ -211,6 +213,8 @@ impl<TNetwork: Network, TValidatorNetwork: ValidatorNetwork>
             mempool_state,
             #[cfg(feature = "metrics")]
             mempool_monitor: TaskMonitor::new(),
+            #[cfg(feature = "metrics")]
+            ctrl_mempool_monitor: TaskMonitor::new(),
```
sisou: Suggestion to rename, to be consistent across the codebase.
```suggestion
            control_mempool_monitor: TaskMonitor::new(),
```

sisou (changes requested):

consensus/src/consensus/mod.rs
viquezclaudio: Nop, because there is a return statement in the line where we publish to the ControlTransactionTopic

sisou: Ah, right! 👍

mempool/src/mempool.rs
viquezclaudio: I don't get this, is not exactly the same. As in normal operation, we call the _start_executors_ which would subscribe to the topics and manage the txn topics themselves, whereas the _*_with_txn_stream_ functions are used only for testing purposes and it is the callers responsibility to manage the txn topics..

mempool/src/mempool.rs
viquezclaudio: Yes I thought about that, but since the control/regular executor are really distinct entities, then even if we only have a single function that receives a parameter to decide which executor to create, we would still need to instantiate both executors, both handles, etc. so there are not much gains I think

mempool/src/mempool.rs
viquezclaudio: Both executors are created/terminated in tandem. If for some reason one of them crashes we don't have a way to recreate it (this applies to the current mempool as well, if the executor dies we are not monitoring its health and this would indicate a major problem in our code)
If a node has a mempool, it will have both a control and regular executors.

mempool/src/mempool.rs
viquezclaudio: Well I think that prioritization matters only during block production, since we want to include control transaction first.
Do you think it also matters for this function? Basically to get the control transactions hashes first as well for RPC purposes

mempool/src/mempool.rs
viquezclaudio: Same as before. Does this matter for RPC purposes?

mempool/src/mempool.rs
viquezclaudio: outdated comment, removal of cheapest ones is done outside this function

mempool/src/mempool.rs
viquezclaudio: Yes, for now we decided to consider staking transactions = control transactions. 
But I think the is an ongoing discussion..

mempool/src/mempool.rs
viquezclaudio: I wanted to do that, but that is not permitted in the language.
They are planning to introduce some chain "let" assignments:
https://github.com/rust-lang/rust/issues/53667

But it is still forbidden with the || operator:
https://github.com/rust-lang/rust/pull/94754

mempool/src/mempool.rs
viquezclaudio: Same as above, forbidden. 
Probably the reason is: what happens if both regular_transactions and control_transactions return an object?
if they both return an object, tx would be assigned to what? to the object returned by the regular_transactions or the one returned by control_transactions?

sisou: In Javascript, the functions are run after each other. If the first call returns a truthy value, the second function is not called anymore. That's what I would expect. And in this case it makes sense, because only one tx bucket can contain the tx_hash in this implementation.

viquezclaudio:
@sisou any more comments?

mempool/src/mempool.rs
sisou: It would be consistent. And would also show RPC clients that those transactions are higher priority. If some where to write a fee estimator, or a confirmation estimation, that can make sense, as the control txs are included first, and the client would not need to sort them again.

But maybe that is just premature optimization :shrug:

sisou (approved):

mempool/src/mempool.rs
nibhar: Just saw this.

`||` does not work, but `or_else(...)` should.

```suggestion
                if let Some(tx) = self.regular_transactions.delete(tx_hash)
                    .or_else(|| self.control_transactions.delete(tx_hash))
                { 
                    self.remove_from_staking_state(&tx);
                }
```

mempool/src/executor.rs
```diff
@@ -114,3 +114,100 @@ impl<N: Network> Future for MempoolExecutor<N> {
...
+        network: Arc<N>,
+        txn_stream: BoxStream<'static, (Transaction, <N as Network>::PubsubId)>,
+    ) -> Self {
+        Self {
+            blockchain: blockchain.clone(),
+            state,
+            filter,
+            network,
+            network_id: Arc::new(blockchain.read().network_id),
+            verification_tasks: Arc::new(AtomicU32::new(0)),
```
nibhar: More of a question than an actual comment:

So each mempool has their ow transaction verification task limit. That doubles the amount of verification tasks from what it was previously. Is that intended?

mempool/src/mempool.rs
```diff
@@ -54,27 +66,28 @@ pub struct Mempool {
...
     /// Default total size limit of transactions in the mempool (bytes)
-    pub const DEFAULT_SIZE_LIMIT: usize = 12000000;
+    pub const DEFAULT_SIZE_LIMIT: usize = 12_000_000;
+
+    /// Default total size limit of control transactions in the mempool (bytes)
+    pub const DEFAULT_CONTROL_SIZE_LIMIT: usize = 6_000_000;
 
     /// Creates a new mempool
     pub fn new(blockchain: Arc<RwLock<Blockchain>>, config: MempoolConfig) -> Self {
         let state = MempoolState {
```
nibhar: Maybe add a `MempoolState::new(size_limit: usize) -> MempoolState { ... }`

mempool/src/mempool.rs
```diff
@@ -511,11 +619,77 @@ impl Mempool {
...
         );
 
-        tx_vec
+        (tx_vec, size)
+    }
+
+    /// Returns a vector with accepted control transactions from the mempool.
+    ///
+    /// Returns the highest fee per byte up to max_bytes transactions and removes them from the mempool
+    pub fn get_control_transactions_for_block(
```
nibhar, line 600: I moved the logic for the expired transactions into the `MempoolState`. I did not do it for `get_transactions_for_block` because I am unsure wether or not more fine grained control over the byte size in particular is required.
However I believe even then a function like:
```
impl MempoolState {
    pub fn get_transactions_for_block(&self, max_bytes: (usize, usize)) -> Vec<Transactions>) {
        ...
    }
}
```
might be the way to go here. One thing in particular that I see is that for both the regular and the control transactions a `self.state.upgradable_read()` and a subsequent upgrade of the lock are required. Moving the logic to the `MempoolState` would remove that. The code is also pretty much identical, aside from which field to access. Abstracting what transactions shall be retrieved might also be a good idea.

If that is not desirable, I probably would still move both these functions in the more lightweight `MempoolState` and just call them from here after taking the lock.

nibhar (commented):
I created #929 to some smaller refactoring and to move some code pieces to their own files. Feel free to look through it and cherry pick if necessary.

Other than that I have little to add to what @sisou already provided. I do agree that control transactions should always come first, as there is no reason to do it the other way around and it does not introduce any complexity as far as I can tell.

mempool/src/executor.rs
viquezclaudio: Yes, I think we never really hit that limit under normal operation, should I set it to half of it, so we maintain the same total amount of verification tasks as before?

mempool/src/mempool.rs
viquezclaudio: if we move to the `pub fn get_transactions_for_block(&self, max_bytes: (usize, usize))` format, how would you specify what we are currently doing? i.e.: we know how much available bytes we have in the block, and we want to fill the block with control txns and if there is space left, fill the remaining with regular txns?

viquezclaudio: I still agree about the other benefits, the code its pretty much identical and it would be a good idea to move it to the `MempoolState`

PR #888, opened 2022-06-16
history_sync: Use chunks for pushing history

jsdanielh:
Change the `push_history_sync` functionality in the `blockchain` to
receive `HistoryTreeChunks` progressively instead of a full epoch.
This includes the following changes:
- Remove `push_history_sync` method.
- Add `start_history_sync` method that only takes a `MacroBlock` and
  returns a `HistoryPusher` instance.
- Add `HistoryPusher` struct with associated functions to:
  - Add history chunks in the form of `HistoryTreeChunk`s.
  - Commit changes when no more history chunks are expected.
Also changed the consensus `cluster` and `sync_stream` functionality
to use the new APIs and call the appropriate functions as soon as new
history chunks are received.

## Pull request checklist

### DIFF
--- a/blockchain/src/blockchain/history_sync.rs
+++ b/blockchain/src/blockchain/history_sync.rs
@@ -2,85 +2,230 @@ use std::error::Error;
 
 use parking_lot::{RwLockUpgradableReadGuard, RwLockWriteGuard};
 
-use nimiq_block::{Block, BlockError, TendermintProof};
+use nimiq_block::{Block, BlockError, MacroBlock, TendermintProof};
 use nimiq_database::WriteTransaction;
 use nimiq_hash::{Blake2bHash, Hash};
 use nimiq_primitives::coin::Coin;
 use nimiq_primitives::policy;
 
 use crate::chain_info::ChainInfo;
-use crate::history::{ExtTxData, ExtendedTransaction, HistoryStore};
-use crate::{AbstractBlockchain, Blockchain, BlockchainEvent, PushError, PushResult};
+use crate::history::{ExtTxData, ExtendedTransaction};
+use crate::{
+    AbstractBlockchain, Blockchain, BlockchainEvent, HistoryTreeChunk, PushError, PushResult,
+};
 use nimiq_account::{Inherent, InherentType};
 
-/// Implements methods to push macro blocks into the chain when an history node is syncing. This
-/// type of syncing is called history syncing. It works by having the node get all the election
-/// macro blocks since genesis plus the last macro block (most likely it will be a checkpoint block,
-/// but it might be an election block). For these macro blocks the node must also get the
-/// corresponding history tree. When the macro blocks are synced, then the node gets all the micro
-/// blocks in the current batch and pushes them normally.
-/// Note that, when pushing the macro blocks, we rely on the assumption that they were produced by
-/// honest validator sets (defined as having less than 1/3 malicious validators). Because of that
-/// we don't actually check the validity of the blocks, we just perform the minimal amount of checks
-/// necessary to verify that the given block is a successor of our current chain so far and that the
-/// corresponding history tree is actually part of the block.
-impl Blockchain {
-    /// Pushes a macro block (election or checkpoint) into the chain using the history sync method.
-    /// You can push election blocks after checkpoint blocks and vice-versa. You can also push macro
-    /// blocks even after you pushed micro blocks.
-    /// You just cannot push micro blocks with this method.
-    pub fn push_history_sync(
-        this: RwLockUpgradableReadGuard<Self>,
-        block: Block,
-        history: &[ExtendedTransaction],
+/// Struct produced after starting history sync.
+/// The object provides methods for adding history chunks and finally commit
+/// it to the history store.
+pub struct HistoryPusher {
+    /// Macro block for which the history chunks will be added
+    pub block: MacroBlock,
+    /// Number of transactions added to this HistoryPusher instance
+    pub tx_count: u64,
+    /// Cumulative transaction fees for the transactions added to this
+    /// HistoryPusher instance
+    pub cum_tx_fees: Coin,
+}
+
+impl HistoryPusher {
+    /// Adds transactions to the history pusher.
+    /// Transactions are only received in chunks using the `HistoryTreeChunk`
+    /// struct. This means that the chunks must come with a proof that relates
+    /// them with this object macro block's history root.
+    /// The proof is checked against the provided macro block's history root.
+    /// The blockchain state may be changed if a need to reverse some micro
+    /// blocks is detected in cases such as a partially known epoch or
+    /// different adopted history.
+    pub fn add_history_chunk(
+        &mut self,
+        blockchain: RwLockUpgradableReadGuard<Blockchain>,
+        history_chunk: HistoryTreeChunk,
+        chunk_index: usize,
+        chunk_size: usize,
     ) -> Result<PushResult, PushError> {
-        // Check that it is a macro block. We can't push micro blocks with this function.
-        assert!(
-            block.is_macro(),
-            "You can't push micro blocks with history sync!"
+        let macro_block = self.block.clone();
+        let staking_contract_address = blockchain.staking_contract_address();
+
+        // Check the history chunk proof against the macro block history root
+        if !history_chunk
+            .verify(macro_block.header.history_root, chunk_index * chunk_size)
+            .unwrap_or(false)
+        {
+            log::warn!(
+                "History Chunk failed to verify (chunk {} of epoch {})",
+                chunk_index,
+                self.block.epoch_number()
+            );
+            return Err(PushError::InvalidHistoryChunk);
+        }
+
+        // Create a new database write transaction.
+        let mut txn = blockchain.write_transaction();
+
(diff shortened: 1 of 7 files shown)


blockchain/src/blockchain/history_sync.rs
```diff
@@ -307,28 +149,149 @@ impl Blockchain {
...
 
                 txn.abort();
                 #[cfg(feature = "metrics")]
-                this.metrics.note_invalid_block();
+                blockchain.metrics.note_invalid_block();
                 return Err(PushError::AccountsError(e));
             }
         }
-        this.state.accounts.finalize_batch(&mut txn);
+        blockchain.state.accounts.finalize_batch(&mut txn);
```
viquezclaudio, line 169: I think this is a confusing name, because we are not finalizing a batch.
"update root" or something similar would be a better name

blockchain/src/blockchain/history_sync.rs
```diff
@@ -307,28 +149,149 @@ impl Blockchain {
...
+                    block_no = macro_block.header.block_number,
+                    previous_block_no = macro_head.block_number(),
+                    "Rejecting block",
+                );
+                return Err(PushError::AlreadyKnown);
+            }
+        }
+
+        // Check the justification.
+        if !TendermintProof::verify(&macro_block, &blockchain.current_validators().unwrap()) {
```
viquezclaudio: clippy warning here

blockchain/tests/history_sync.rs
```diff
@@ -174,37 +252,73 @@ fn history_sync_works_with_micro_blocks() {
...
+            checkpoint_block_2_1.block_number(),
+            CHUNK_SIZE,
+            0,
+            None,
+        )
+        .unwrap();
 
-        checkpoint_txs_2_1.push(ext_tx.clone());
-    }
+    //for ext_tx in &election_txs_2.history {
```
viquezclaudio: unused/commented lines

blockchain/src/blockchain/history_sync.rs
jsdanielh: Mmm, this is an accounts batch. I don't have the full context but there is a family of functions in the accounts tree talking about batches but they doesn't look related to the concept of a block batch

blockchain/src/blockchain/history_sync.rs
```diff
@@ -2,244 +2,86 @@ use std::error::Error;
...
+        let stacking_contract_address = blockchain.staking_contract_address();
 
-        // Check if we already know this block.
-        if this
-            .chain_store
-            .get_chain_info(&macro_block.hash(), false, Some(&read_txn))
-            .is_some()
+        // Check the history chunk proof against the macro block history root
+        if !history_chunk
+            .verify(macro_block.header.history_root, chunk_index * CHUNK_SIZE)
```
paberr: Does verification of consecutive chunks work like that?
As far as I know, the chunks only contain the leaves and the rightmost path to the root.
What's missing on the left must be taken from the previous proof.

Some of the verification functions have this code:
```
        if self.assume_previous {
            return Err(Error::IncompleteProof);
        }
```
And I'm basically wondering whether we'll run into this error or not.

blockchain/src/blockchain/history_sync.rs
```diff
@@ -2,244 +2,86 @@ use std::error::Error;
...
 
         // Calculate the cumulative transaction fees for the given batch. This is necessary to
         // create the chain info for the block.
-        let mut cum_tx_fees = Coin::ZERO;
-        let current_batch = policy::batch_at(block.block_number());
-        for i in (0..history.len()).rev() {
-            if policy::batch_at(history[i].block_number) != current_batch {
+        for i in (0..history_chunk.history.len()).rev() {
+            if policy::batch_at(history_chunk.history[i].block_number) != current_batch {
                 break;
```
paberr, line 76: What happens with blocks that are not in the current batch?

blockchain/src/blockchain/history_sync.rs
```diff
@@ -2,244 +2,86 @@ use std::error::Error;
...
-
-        // Set the head of the chain store to the current block.
-        this.chain_store.set_head(&mut txn, &block_hash);
-
         // We might already know the given epoch partially.
         // Revert our chain to a common ancestor state in case we have adopted a different history.
         // Also skip over any transactions that we already know.
-        let first_new_ext_tx = this.revert_to_common_state(&block, history, &mut txn);
+        let (current_chain_info, first_new_ext_tx_idx) =
+            self.revert_to_common_state(&blockchain, &history_chunk.history, &mut txn);
```
paberr, line 87: We only want to revert for the first history chunk we push. Subsequent chunks build on top of the previous one.
TODO: check that the function works as expected.

blockchain/src/blockchain/history_sync.rs
```diff
@@ -273,7 +115,7 @@ impl Blockchain {
             if policy::is_macro_block_at(*block_number) {
                 let finalize_batch = Inherent {
                     ty: InherentType::FinalizeBatch,
-                    target: this.staking_contract_address(),
+                    target: stacking_contract_address.clone(),
```
paberr:
```suggestion
                    target: staking_contract_address.clone(),
```

consensus/src/sync/history/cluster.rs
```diff
@@ -194,57 +169,65 @@ impl<TNetwork: Network + 'static> SyncCluster<TNetwork> {
...
+            };
 
-        // Queue history chunks for the given epoch for download.
-        let history_chunk_ids = (start_index
-            ..((epoch.history_len as usize).ceiling_div(CHUNK_SIZE)))
-            .map(|i| (epoch_number, pending_batch_set.block.header.block_number, i))
-            .collect();
-        self.history_queue.add_ids(history_chunk_ids);
+            // If the block is in the same epoch, add already known history.
+            // Why? It looks like it is already assuming this is a new epoch?
```
paberr: Is it?

consensus/src/sync/history/cluster.rs
```diff
@@ -254,38 +237,21 @@ impl<TNetwork: Network + 'static> SyncCluster<TNetwork> {
...
-            .unwrap_or(false)
-        {
-            log::warn!(
-                "History Chunk failed to verify (chunk {} of epoch {})",
-                chunk_index,
-                epoch_number
-            );
-            return Err(SyncClusterResult::Error);
-        }
-
```
paberr, line 273: Ok, here we use the same code as above. We should still test the code with more than 1 history chunk (tests seem to always only add 1).

consensus/src/sync/history/cluster.rs
jsdanielh: You're right. Doesn't make sense so I removed it

blockchain/src/blockchain/history_sync.rs
jsdanielh: To build the chunks we're basically using [this function](https://github.com/nimiq/core-rs-albatross/blob/albatross/blockchain/src/history/history_store.rs#L675) from `Blockchain`, which is not setting `assume_previous` such that we can verify chunks independently.
Not sure if we want to change it though

consensus/src/sync/history/cluster.rs
jsdanielh: Added a new test: `history_sync_works_with_multiple_chunks`

blockchain/src/blockchain/history_sync.rs
jsdanielh: It should since we're only updating the `blockchain` history at `commit`

blockchain/src/blockchain/history_sync.rs
jsdanielh: Added a comment to clarify it but essentially they are ignored since we only need them for the same batch the macro block is at to later update the `ChainInfo` in the `Blockchain`

blockchain/src/blockchain/history_sync.rs
paberr: Okay, yes. In that case, I think we still need to check that the proof fits to what we currently have stored.
We partially do this by reverting to the common state. However, what happens if we push a later chunk first?

Example: a batch consists of 3 chunks.
We start pushing the first chunk. We then push the third chunk (it's a valid chunk that can be verified independently).
This should fail (but not sure whether it currently does).

blockchain/src/blockchain/history_sync.rs
paberr: Ah yes, you're right. I first thought the `current_batch` refers to the batch the blockchain is currently in.
However, it seems to be the macro block we're syncing to (which is good).
It might make sense to rename the variable then.

blockchain/src/blockchain/history_sync.rs
paberr: There's still a test called `it_can_revert_create_stacker_transaction` (which I think is not from your changeset, but while you're at it we can also rename that 😄 )

blockchain/src/blockchain/history_sync.rs
paberr: Discussed in core call, all good.

jsdanielh:
Comments addressed @paberr

PR #906, opened 2022-06-28
Iicruz/rpc

ii-cruz:
- Introduces logs events to the blockchain. The Logs are aggregated by block in a struct called BlockLog.  Logs from the transactions are included as TransactionLogs, thus aggregated by transaction (TransactionLog struct). Inherent logs are included in the blocklog separately from the remaining logs.
- Adds the event stream subscription for BlockLogs on the rpc server. The logs can be subscribed to and filtered by addresses or log types.
- Adds block height and hash to the head_hash_subscribe, staker_by_address, and get_validator_by_address
- Adds the option to subscribe to full blocks (head_subscribe)

## Pull request checklist


rpc-server/src/dispatchers/blockchain.rs
```diff
@@ -468,55 +525,57 @@ impl BlockchainInterface for BlockchainDispatcher {
...
+        include_transactions: Option<bool>,
+    ) -> Result<BoxStream<'static, Result<Block, Blake2bHash>>, Error> {
+        let blockchain = Arc::clone(&self.blockchain);
+        let stream = self.head_hash_subscribe().await?;
+
+        Ok(stream
+            .map(move |hash| {
+                let blockchain_rg = blockchain.read();
+                get_block_by_hash(blockchain_rg.deref(), &hash, include_transactions)
+                    .map_err(|_| hash)
```
sisou: Does this, and the result type `Result<Block, Blake2bHash>` mean that this stream can _both_ yield a block object _and_ a hash string?

rpc-server/src/dispatchers/blockchain.rs
```diff
@@ -529,4 +588,152 @@ impl BlockchainInterface for BlockchainDispatcher {
...
+                        block_number,
+                        mut tx_logs,
+                    } => {
+                        let mut block_log = None;
+                        inherent_logs.retain(|log| is_log_related_to_addresses(&log, &addresses));
+                        tx_logs.retain(|tx_log| {
+                            is_tx_logs_related_to_any_addresses(&tx_log, &addresses)
+                        });
+                        if !inherent_logs.is_empty() || !tx_logs.is_empty() {
+                            block_log = Some(BlockLog::AppliedBlockLog {
```
sisou: This is in the `BlockLog::RevertBlockLog` matching block, so I think:
```suggestion
                            block_log = Some(BlockLog::RevertBlockLog {
```

rpc-server/src/dispatchers/blockchain.rs
```diff
@@ -529,4 +588,152 @@ impl BlockchainInterface for BlockchainDispatcher {
...
+                        mut inherent_logs,
+                        block_hash,
+                        block_number,
+                        mut tx_logs,
+                    } => {
+                        let mut block_log = None;
+                        inherent_logs.retain(|log| contains_log_type(&log, &log_types));
+                        tx_logs.retain(|tx_log| tx_logs_contains_any_log_type(&tx_log, &log_types));
+                        if !inherent_logs.is_empty() || !tx_logs.is_empty() {
+                            block_log = Some(BlockLog::AppliedBlockLog {
```
sisou: Same here, this is in the `RevertBlockLog` match block:
```suggestion
                            block_log = Some(BlockLog::RevertBlockLog {
```

rpc-server/src/dispatchers/blockchain.rs
ii-cruz: You are 100% correct. Thank you!

rpc-server/src/dispatchers/blockchain.rs
ii-cruz: You are correct. So this function uses the head_hash_subscribe to have a stream of new block hashs and then fetches the block using that hash. In case there is a new block committed, the new hash is received over the stream. If the block was reverted in the meantime, the get_block_by_hash fails. In that case, we are sending the hash as a return value. 

The reason for this behavior is that serde is not able to serialize something like Result<Block, Error>. The alternative would be to filter out the event altogether. That could work too, I'm not seeing any issues with it.

sisou:
@ii-cruz I just noticed another problem: In the block subscription stream, all results pushed over the websocket are wrapped in a `Ok {}` result, e.g.

```
{
  Ok: {
    hash: '57fa19973988afa678beba2d48798fe26f54264888ae7a60f9a16b0ba3f23b39',
    size: 315,
    batch: 9,
    epoch: 3,
    version: 1,
    number: 283,
    view: 0,
    timestamp: 1656434106296,
    parentHash: '884a89736fb6518a8f3f8ba2ae2865ac0f3b852f98a56ea410f8dd1bb5f0f187',
    seed: { signature: [Array] },
    extraData: '',
    stateHash: '39ea52d1c50e6768916f54460c24df39d18bc7a260d20a98d19c5dcafef4e3d1',
    bodyHash: '11da6d1f761ddf9bdb4c9d6e5303ebd41f61858d0a5647a1a7bfe089bf921be9',
    historyHash: '81e47a19e6b29b0a65b9591762ce5143ed30d0261e5d24a3201752506b20f15c',
    type: 'micro',
    producer: {
      slotNumber: 143,
      validator: 'NQ69 MH90 3M5H DFNU 8E84 53FX JSY6 XYQS N2CE',
      publicKey: '000b17927b6293cd141488086073077392386c752c8d990340313ff7730e24497938a088c3694c827501e6474b3c288a075f919cc5eebde20a812cf2b475da77dc961eedcc0216bf53010e2df22dd19a20fe62f0416323a966a0b78ceb465d009700ed01a91f339a087580c86bf6b68e1c71d7bbebd65c5b5fca0b36c6f9bfcbb03cb4746d2164738dcbc9f825ac2edb6178a2b43665e615ad38fe9516ae671cc5c31eab5102c92a111ca309d3a9fc88f64a0b56959db881d663bc085fc400668e719eb7a8ef674add0520f16f28b31864953c7899fb458d491486aea8e6d30ddb107e522e9409531e7d5aa877e6ef9a80ce88c176f6abc690daed3e07af5a88700bebb6c3833a198f13121f2ebf426560cbe933857f0ee522c8e430ec'
    },
    forkProofs: [],
    justification: {
      signature: '5d875768ffd066ec41522cfe15cb5f8ba4e5c1d049e6256f93d3d8a7323a1f52fb010b4fe1f109fe48dd46d24044b64a7a7e2248f10316f4950bb32e20cbe90f'
    }
  }
}
```

Please unwrap the data before sending it over the wire.

sisou:
@paberr The logs are sent as JSON with snake_cased names. Can we get the fields of enum structs to also be serialized as camelCase?

primitives/account/src/accounts.rs
```diff
@@ -172,43 +191,48 @@ impl Accounts {
...
             txn,
             inherents,
             block_height,
             timestamp,
             pre_tx_inherent_receipts,
-        )?;
+        )?);
 
-        Ok(())
+        //TODO build batchinfo with results of previous reverts
```
nibhar: Open TODO

primitives/account/src/accounts.rs
```diff
@@ -485,4 +541,35 @@ impl Accounts {
...
+
+            if let Some(tx_log) = tx_logs_create.peek_mut() {
+                if tx_log.tx_hash == tx_log_sender.tx_hash {
+                    tx_log_sender.logs.append(&mut tx_log.logs);
+                    tx_logs_create.next();
+                }
+            }
+        }
+
+        //assert tx_logs_recipients and tx_log_sender are over (?)
```
nibhar: Good idea!

primitives/account/src/staking_contract/staker.rs
```diff
@@ -505,6 +534,17 @@ impl StakingContract {
             Some(x) => x,
         };
 
+        // If it was a no-op, we end right here.
```
nibhar: Is there a reason this moved beneath the staker exist check?

primitives/account/src/staking_contract/validator.rs
```diff
@@ -236,12 +245,7 @@ impl StakingContract {
         db_txn: &mut WriteTransaction,
         validator_address: &Address,
         receipt: UpdateValidatorReceipt,
-    ) -> Result<(), AccountError> {
-        // If it was a no-op, we end right here.
-        if receipt.no_op {
-            return Ok(());
-        }
-
+    ) -> Result<Vec<Log>, AccountError> {
```
nibhar, line 259: I think this change to the return type could be applied to `revert_create_validator` as well. Makes for a more consistent interface and cleans up `primitives/account/src/staking_contract/traits.rs` a bit.

primitives/account/src/staking_contract/validator.rs
```diff
@@ -176,21 +177,24 @@ impl StakingContract {
         new_voting_key: Option<BlsPublicKey>,
         new_reward_address: Option<Address>,
         new_signal_data: Option<Option<Blake2bHash>>,
-    ) -> Result<UpdateValidatorReceipt, AccountError> {
+    ) -> Result<OperationInfo<UpdateValidatorReceipt>, AccountError> {
```
nibhar, line 191: I think the return type for non revert operations can also be applied to `create_validator`. Same Reasoning as for the revert case.

primitives/account/src/staking_contract/validator.rs
```diff
@@ -252,6 +256,16 @@ impl StakingContract {
                     });
                 }
             };
+        // If it was a no-op, we end right here.
+        if receipt.no_op {
```
nibhar: Again I wonder why this was moved beneath the validator exists check.

primitives/account/src/staking_contract/staker.rs
```diff
@@ -203,7 +215,7 @@ impl StakingContract {
         db_txn: &mut WriteTransaction,
         staker_address: &Address,
         value: Coin,
-    ) -> Result<(), AccountError> {
+    ) -> Result<Vec<Log>, AccountError> {
```
nibhar: I think it would be more consistent to return 
```suggestion
    ) -> Result<AccountInfo, AccountError> {
```
as almost all other functions in this file do with the respective `*Info` type

primitives/account/src/staking_contract/staker.rs
```diff
@@ -35,7 +36,7 @@ impl StakingContract {
         staker_address: &Address,
         value: Coin,
         delegation: Option<Address>,
-    ) -> Result<(), AccountError> {
+    ) -> Result<Vec<Log>, AccountError> {
```
nibhar: I think this could also return 
```suggestion
    ) -> Result<AccountInfo, AccountError> {
```
Same reasoning as for `stake`.

primitives/account/src/staking_contract/traits.rs
```diff
@@ -337,36 +342,50 @@ impl AccountTransactionInteraction for StakingContract {
...
                     &staker_address,
                     transaction.total_value(),
                 )?
-                .map(|r| r.serialize_to_vec())
+                .into();
             }
         };
-
-        Ok(receipt)
+        acc_info.logs.insert(
```
nibhar, line 362: Does the order of logs matter here or can this also be
```suggestion
        acc_info.logs.push(
```
?

If the order does matter, add a comment explaining why.

primitives/account/src/staking_contract/traits.rs
```diff
@@ -660,6 +714,15 @@ impl AccountInherentInteraction for StakingContract {
                             .current_lost_rewards
                             .remove(slot.slot as usize);
                     }
+                    logs.insert(
```
nibhar, line 708: Does the order of logs matter here or can this also be
```suggestion
                    logs.push(
```
?

If the order does matter, add a comment explaining why.

primitives/account/src/staking_contract/receipts.rs
```diff
@@ -8,6 +8,9 @@ use nimiq_keys::{Address, PublicKey as SchnorrPublicKey};
 /// A collection of receipts for inherents/transactions. This is necessary to be able to revert
 /// those inherents/transactions.
 
+#[derive(Clone, Copy, Debug, Serialize, Deserialize, Eq, PartialEq)]
+pub struct DefaultReceipt;
```
nibhar: What is the purpose of this, is this used anywhere?

jsdanielh (commented):
@ii-cruz Looks like there are several clippy warnings and an error. Could you address them?
You can check them from the last run of your code [here](https://github.com/nimiq/core-rs-albatross/pull/906/checks?check_run_id=7116450257).

rpc-server/src/dispatchers/blockchain.rs
```diff
@@ -529,4 +588,127 @@ impl BlockchainInterface for BlockchainDispatcher {
             })
             .boxed())
     }
+
+    /// Subscribes to pre epoch validators events.
+    #[stream]
+    async fn election_validator_subscribe(
```
jsdanielh: Maybe `validator_election_subscribe`?

rpc-client/src/main.rs
```diff
@@ -194,22 +205,50 @@ impl Command {
...
+                    while let Some(block_hash) = stream.next().await {
                         println!("{}", block_hash);
                     }
                 }
             }
 
+            Command::FollowValidator { address } => {
+                let mut stream = client
+                    .blockchain
+                    .election_validator_subscribe(address)
```
jsdanielh: Maybe `validator_election_subscribe`?

rpc-interface/src/blockchain.rs
```diff
@@ -97,10 +99,36 @@ pub trait BlockchainInterface {
...
+        include_transactions: Option<bool>,
+    ) -> Result<BoxStream<'static, Block>, Self::Error>;
+
+    #[stream]
+    async fn head_hash_subscribe(&mut self)
+        -> Result<BoxStream<'static, Blake2bHash>, Self::Error>;
 
     #[stream]
-    async fn head_subscribe(&mut self) -> Result<BoxStream<'static, Blake2bHash>, Self::Error>;
+    async fn election_validator_subscribe(
```
jsdanielh: Maybe `validator_election_subscribe`?

primitives/account/src/logs.rs
```diff
@@ -0,0 +1,364 @@
+use crate::Receipt;
```
jsdanielh, line 1: Mmm, it looks strange to me that we have this in `primitives/account` since this is not strictly related to accounts. Maybe consider moving it to one directory up (to `primitives/src`)?

primitives/account/src/logs.rs
```diff
@@ -0,0 +1,364 @@
...
+impl TransactionLog {
+    pub fn new(tx_hash: Blake2bHash, logs: Vec<Log>) -> Self {
+        Self { tx_hash, logs }
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+#[cfg_attr(feature = "serde-derive", derive(serde::Serialize, serde::Deserialize))]
+// Renaming affects only the struct names and thus their tag, the "type" field.
+#[serde(rename_all = "kebab-case", tag = "type")]
```
jsdanielh, line 260: Shouldn't this be `camelCase`? Maybe we can change it here and avoid adding it for all of the members?

primitives/account/src/logs.rs
```diff
@@ -0,0 +1,364 @@
...
+use beserial::Serialize as BeSerialize;
+use nimiq_hash::Blake2bHash;
+use nimiq_keys::Address;
+use nimiq_primitives::coin::Coin;
+use nimiq_transaction::account::htlc_contract::{AnyHash, HashAlgorithm};
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+#[cfg_attr(feature = "serde-derive", derive(serde::Serialize, serde::Deserialize))]
+// Renaming affects only the struct names and thus their tag, the "type" field.
+#[serde(rename_all = "kebab-case", tag = "type")]
```
jsdanielh, line 11: Shouldn't this be `camelCase`? Doesn't look consistent to me that we're renaming the members to use kebab-case?

primitives/account/src/logs.rs
```diff
@@ -0,0 +1,364 @@
...
+            Log::Slash {
+                validator_address, ..
+            } => validator_address == address,
+            Log::RevertContract { contract_address } => contract_address == address,
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+#[cfg_attr(feature = "serde-derive", derive(serde::Serialize, serde::Deserialize))]
```
jsdanielh, line 243: Aren't we missing `#[serde(rename_all = "camelCase")]`?

primitives/account/src/logs.rs
```diff
@@ -0,0 +1,364 @@
...
+        AccountInfo::new(
+            op_info.receipt.map(|receipt| receipt.serialize_to_vec()),
+            op_info.logs,
+        )
+    }
+}
+
+// Batch Info is used as a return type of multiple transactions or batches applied to the blockchain.
+// It stores the transaction logs, inherent logs and receipts associated with the multiple transactions of the batch.
+#[derive(Default, Eq, PartialEq, Debug)]
```
jsdanielh, line 322: Same here, aren't we missing `#[serde(rename_all = "camelCase")]`?

primitives/account/src/logs.rs
```diff
@@ -0,0 +1,364 @@
...
+            receipts,
+            tx_logs,
+            inherent_logs,
+        }
+    }
+}
+
+// Operation Info is used as a return type of some staker and validator related transactions.
+// It stores the transaction logs generated by the transaction and the receipts. The receipts can either be represented in bytes or in any receipt type defined.
+#[derive(Debug, PartialEq, Eq)]
```
jsdanielh, line 346: Same here, aren't we missing `#[serde(rename_all = "camelCase")]`?

primitives/account/src/logs.rs
```diff
@@ -0,0 +1,364 @@
...
+        Self {
+            receipts,
+            tx_logs,
+            inherent_logs,
+        }
+    }
+}
+
+// Operation Info is used as a return type of some staker and validator related transactions.
+// It stores the transaction logs generated by the transaction and the receipts. The receipts can either be represented in bytes or in any receipt type defined.
```
jsdanielh: Maybe consider splitting this line.

primitives/account/src/logs.rs
```diff
@@ -0,0 +1,364 @@
...
+    pub fn is_revert_block_log(&self) -> bool {
+        match self {
+            BlockLog::AppliedBlock { .. } => false,
+            BlockLog::RevertedBlock { .. } => true,
+        }
+    }
+}
+
+// Account Info is used as a return type of transactions to the blockchain and stores the logs and receipt associated with the transaction
+#[derive(Debug, PartialEq, Eq)]
```
jsdanielh, line 294: Same here, aren't we missing `#[serde(rename_all = "camelCase")]`?

primitives/account/src/logs.rs
```diff
@@ -0,0 +1,364 @@
+use crate::Receipt;
+use beserial::Serialize as BeSerialize;
+use nimiq_hash::Blake2bHash;
+use nimiq_keys::Address;
+use nimiq_primitives::coin::Coin;
+use nimiq_transaction::account::htlc_contract::{AnyHash, HashAlgorithm};
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+#[cfg_attr(feature = "serde-derive", derive(serde::Serialize, serde::Deserialize))]
+// Renaming affects only the struct names and thus their tag, the "type" field.
```
jsdanielh, line 10: Mmm, it should affect all of the members according to the documentation: _Rename all the fields (if this is a struct) or variants (if this is an enum) according to the given case convention._

primitives/account/src/logs.rs
```diff
@@ -0,0 +1,364 @@
...
+
+impl TransactionLog {
+    pub fn new(tx_hash: Blake2bHash, logs: Vec<Log>) -> Self {
+        Self { tx_hash, logs }
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+#[cfg_attr(feature = "serde-derive", derive(serde::Serialize, serde::Deserialize))]
+// Renaming affects only the struct names and thus their tag, the "type" field.
```
jsdanielh, line 259: Mmm, it should affect all of the members according to the documentation: _Rename all the fields (if this is a struct) or variants (if this is an enum) according to the given case convention._

primitives/account/src/logs.rs
sisou: The name of the structs (which are renamed through this) are set as the 'type' property on the resulting objects. The string value should be kebab-cased IMO, this is on purpose.

This way it looks like this:
```
    type: 'applied-block',
```

And not like this:
```
    type: 'appliedBlock',
````

Comment (sisou) [reply primitives/account/src/logs.rs:260]: Additionally, this rule only applies to direct members of the enum, not _their_ members in turn. That's why the rename rule needs to be added to all members individually anyway.

File: primitives/account/src/logs.rs (hunk 26 above)
Comment (sisou) [reply primitives/account/src/logs.rs:10]: It only renames the names of the variants, not their fields too.

File: primitives/account/src/logs.rs (hunk 21 above)
Comment (sisou) [reply primitives/account/src/logs.rs:243]: It's unnecessary since I chose to rename the only field that would be affected with its own rename rule already.

File: primitives/account/src/logs.rs (hunk 25 above)
Comment (sisou) [reply primitives/account/src/logs.rs:294]: No field would be affected by this rule, plus this struct is never sent as JSON AFAIK.

File: primitives/account/src/logs.rs (hunk 22 above)
Comment (sisou) [reply primitives/account/src/logs.rs:322]: AFAIK this struct is never sent as JSON.

File: primitives/account/src/logs.rs (hunk 23 above)
Comment (sisou) [reply primitives/account/src/logs.rs:346]: No field would be affected by this rule, plus this struct is never sent as JSON AFAIK.

File: primitives/account/src/logs.rs (hunk 19 above)
Comment (sisou) [reply primitives/account/src/logs.rs:260]: If you find a way to apply it to all children fields as well, please tell!

File: primitives/account/src/logs.rs (hunk 27 above)
Comment (sisou) [reply primitives/account/src/logs.rs:259]: It does. It renames all _direct_ fields and variants.

Comment (jsdanielh) [reply primitives/account/src/logs.rs:259]: Ok, I think I now understand the point of the comment, maybe it should be reworded as:
_Renaming affects only the enum variant names and its tag (the "type" field)_
?

File: primitives/account/src/logs.rs (hunk 19 above)
Comment (jsdanielh) [reply primitives/account/src/logs.rs:260]: Ok, yeah I see why it needs to be in the enum variants. But still it doesn't look consistent to me that this enum's tag would follow kebab case when everything else does camel case.

File: primitives/account/src/logs.rs (hunk 21 above)
Comment (jsdanielh) [reply primitives/account/src/logs.rs:243]: I see your point but I'd still add it in case we add more members in the future and we forget to revise it

File: primitives/account/src/logs.rs (hunk 26 above)
Comment (jsdanielh) [reply primitives/account/src/logs.rs:10]: Maybe it should be reworded as:
_Renaming affects only the enum variant names and its tag (the "type" field)_
?

File: rpc-interface/src/types.rs [hunk 28]
```diff
@@ -692,6 +693,143 @@ impl Validator {
...
+#[serde(rename_all = "camelCase")]
+pub struct BlockchainState<T> {
+    pub block_number: u32,
+    pub block_hash: Blake2bHash,
+    #[serde(flatten)]
+    pub value: T,
+}
+
+impl<T> BlockchainState<T> {
+    pub fn from_value(block_number: u32, block_hash: Blake2bHash, value: T) -> Self {
```
nibhar: Small nitpick.
```suggestion
    pub fn new(block_number: u32, block_hash: Blake2bHash, value: T) -> Self {
```
But since this does not contain any logic and the fields are `pub`, you can also just use 

```rust
BlockchainState { block_number, block_hash, value }
```

nibhar (changes requested):
Looks good in general in my opinion. 

Clippy warnings and errors should for sure be fixed, and I would appreciate more documentation being added. 

Outside of that I commented mostly smaller nitpicks that I found reading this.

primitives/account/src/logs.rs
ii-cruz: I think it makes sense to place it in the same place as receipts, they are reasonably similar. Another argument for letting it stay here is to that the wrapper structs that are within logs are return types of many of the account methods.

rpc-interface/src/blockchain.rs
```diff
@@ -97,10 +99,36 @@ pub trait BlockchainInterface {
...
+        &mut self,
+        address: Address,
+    ) -> Result<BoxStream<'static, BlockchainState<Validator>>, Self::Error>;
+
+    #[stream]
+    async fn logs_subscribe(&mut self) -> Result<BoxStream<'static, BlockLog>, Self::Error>;
```
jsdanielh, line 126:
```suggestion
    async fn logs_subscribe(&mut self) -> Result<BoxStream<'static, Log>, Self::Error>;
```
One more, shouldn't this be a `Log` instead of a `BlockLog`?

rpc-interface/src/blockchain.rs
```diff
@@ -97,10 +99,36 @@ pub trait BlockchainInterface {
...
+
+    #[stream]
+    async fn logs_by_type_and_addresses_subscribe(
+        &mut self,
+        addresses: Vec<Address>,
+        log_types: Vec<LogType>,
+    ) -> Result<BoxStream<'static, BlockLog>, Self::Error>;
```
jsdanielh, line 133: Same here, shouldn't this be a `Log` instead of a `BlockLog`?

rpc-interface/src/blockchain.rs
```diff
@@ -97,10 +99,36 @@ pub trait BlockchainInterface {
...
+    ) -> Result<BlockchainState<Validator>, Self::Error>;
 
-    async fn get_staker_by_address(&mut self, address: Address) -> Result<Staker, Self::Error>;
+    async fn get_staker_by_address(
+        &mut self,
+        address: Address,
+    ) -> Result<BlockchainState<Staker>, Self::Error>;
+
+    #[stream]
+    async fn head_subscribe(
```
redmaner: From a RPC user perspective, I would like to propose some naming alternatives for the streaming methods:
* `head_subscribe` -> `subscribe_for_head_block`
* `head_hash_subscribe` -> `subscribe_for_head_block_hash`
* `election_validator_subscribe` -> `subscribe_for_validator_election_by_address` 
* `logs_subscribe` -> `subscribe_for_logs` or `subscribe_for_block_logs` or `subscribe_for_event_logs`
* `logs_by_type_and_addresses_subscribe` -> `subscribe_for_logs_by_addresses_and_types` or one of the other two variations mentioned above suffixed with `_by_addresses_and_types`

redmaner:
First of all, great work @ii-cruz! Since I requested most of these features, I took the liberty of trying out your changes. I tested it quickly for now, and it looks great so far. I thinks this really adds some valuable ways to extract information that can be used by a lot of applications to great extends in the future. 

I added one comment to maybe improve the naming of the streaming methods. I prefer descriptive naming over short naming, and tried to propose a few alternatives that are more in line with the other methods. 

I agree with @nibhar on the documentation part. Especially for the filter method on logs it would be nice to know which types you could filter on for example. I think the logs methods are probably the most interesting to have documented, since the others are quite self explanatory.

rpc-interface/src/blockchain.rs
ii-cruz: Thank you! 
The methods are renamed using those suggestions :)

sisou (approved):
Just tested the latest state, both with `arpl` and with the built-in RPC client, and both work fine. The JSON sent is also correctly formatted, for the logs I tested. :+1:
LGTM :rocket:

nibhar (approved):
I'll create an issue for adding the documentation to our own rpc-client, but also to the server such that it is easily accessible. Also there is PR #923 related to this, for more consistency.

PR #920, opened 2022-07-05
Mempool txn prioritization

viquezclaudio:
- Introduces some priorities (low, medium, high) to transactions when they are added into the mempool.
- Exposes this functionality via RPC
- The unpark transaction is included into the mempool with high priority

### DIFF
--- a/mempool/src/mempool.rs
+++ b/mempool/src/mempool.rs
@@ -25,6 +25,7 @@ use crate::filter::{MempoolFilter, MempoolRules};
 #[cfg(feature = "metrics")]
 use crate::mempool_metrics::MempoolMetrics;
 use crate::mempool_state::{EvictionReason, MempoolState};
+use crate::mempool_transactions::TxPriority;
 use crate::verify::{verify_tx, VerifyErr};
 
 /// Transaction topic for the Mempool to request transactions from the network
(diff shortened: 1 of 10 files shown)


mempool/src/mempool.rs
```diff
@@ -668,7 +673,15 @@ impl Mempool {
...
             Ok(mempool_state_lock) => {
-                RwLockUpgradableReadGuard::upgrade(mempool_state_lock).put(&transaction);
+                if let Some(priority) = tx_priority {
+                    RwLockUpgradableReadGuard::upgrade(mempool_state_lock)
+                        .put(&transaction, priority);
+                } else {
+                    //If the priority is not provided we default to medium
+                    RwLockUpgradableReadGuard::upgrade(mempool_state_lock)
+                        .put(&transaction, TxPriority::MediumPriority);
+                }
```
nibhar:
```suggestion
                RwLockUpgradableReadGuard::upgrade(mempool_state_lock)
                        .put(&transaction, tx_priority.unwrap_or(TxPriority::MediumPriority));
```

mempool/src/mempool.rs
```diff
@@ -520,7 +521,7 @@ impl Mempool {
                     let pending_balance = tx.total_value() + sender_total;
 
                     if pending_balance <= sender_balance {
-                        mempool_state.put(tx);
+                        mempool_state.put(tx, TxPriority::MediumPriority);
```
nibhar, line 525: All reverted transactions are added as `MediumPriority`. The fact they might have been added as `HighPriority` is lost of course, but potentially we could still add unpark transactions as `HighPriority`.

nibhar (approved):
Good to go! My comments just to save some code and a question for trannsactions of reverted blocks.

sisou:
When querying mempool state, e.g. via RPC `mempoolContent()` method, are the transactions ordered by priority? Is the priority part of the returned object (if I set `include_transactions` to `true`)?

PR #923, opened 2022-07-06
Output reward inherent logs as transaction logs

sisou:
Validator rewards are publicly represented as transactions with transaction hashes. They are represented as coinbase transactions in macro blocks, they are returned in the `get_transaction_by_address` RPC call and can be queried by their transaction hash with `get_transaction_by_hash`.

Therefore I found it to be consistent that rewards are also represented in the transaction logs of block event subscriptions.

This requires a transaction to be created from the inherent however, to compute the transaction hash under which it will be accessible.

### DIFF
--- a/primitives/account/src/accounts.rs
+++ b/primitives/account/src/accounts.rs
@@ -801,7 +801,8 @@ impl Accounts {
             // If the tx_logs are empty, overwrite them with the senders_logs
             batch_info.tx_logs = senders_logs
         } else if !senders_logs.is_empty() {
-            unreachable!("Block contains both regular transactions and reward inherents");
+            warn!("Block contains both regular transactions and reward inherents");
+            batch_info.tx_logs.append(&mut senders_logs);
         }
         batch_info
     }
--- a/primitives/account/src/accounts.rs
+++ b/primitives/account/src/accounts.rs
@@ -2,6 +2,9 @@ use nimiq_database::{
     Environment, ReadTransaction, Transaction as DBTransaction, WriteTransaction,
 };
 use nimiq_hash::{Blake2bHash, Hash};
+use nimiq_primitives::coin::Coin;
+use nimiq_primitives::networks::NetworkId;
+use nimiq_primitives::policy::Policy;
 use nimiq_transaction::{ExecutedTransaction, Transaction, TransactionFlags};
 use nimiq_trie::key_nibbles::KeyNibbles;
 use nimiq_trie::trie::MerkleRadixTrie;
@@ -9,7 +12,7 @@ use nimiq_trie::trie::MerkleRadixTrie;
 use crate::{
     logs::{BatchInfo, TransactionLog},
     Account, AccountError, AccountInherentInteraction, AccountTransactionInteraction, Inherent,
-    Log, Receipt, Receipts, RevertTransactionLogs, TransactionInfo,
+    InherentType, Log, Receipt, Receipts, RevertTransactionLogs, TransactionInfo,
 };
 
 /// An alias for the accounts tree.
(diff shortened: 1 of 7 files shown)


jsdanielh:
Rebased branch to correctly reflect the changes since GH auto-changed the PR base to `albatross`

primitives/account/src/accounts.rs
```diff
@@ -457,10 +491,23 @@ impl Accounts {
                 data: account_info.receipt,
             });
 
-            logs.append(&mut account_info.logs);
+            if inherent.ty == InherentType::Reward {
```
ii-cruz, line 643: It's better to comment on this behavior, to have an in-place explanation for the differentiating behavior between inherent types.

primitives/account/src/accounts.rs
```diff
@@ -470,26 +517,47 @@ impl Accounts {
...
+                    let mut logs = Account::revert_inherent(
+                        &self.tree,
+                        txn,
+                        inherent,
+                        block_height,
+                        timestamp,
+                        data.as_ref(),
+                    )?;
+
+                    if inherent.ty == InherentType::Reward {
```
ii-cruz, line 688: same suggestion as on the commit_inherents.

primitives/account/src/accounts.rs
```diff
@@ -577,7 +645,15 @@ impl Accounts {
         assert_eq!(None, tx_logs_recipients.peek(), "recipients tx_logs were not fully processed. Possible cause is a ordering difference compared to senders_logs tx_logs");
         assert_eq!(None, tx_logs_create.peek(), "create tx_logs were not fully processed. Possible cause is a ordering difference compared to senders_logs tx_logs");
 
-        batch_info.tx_logs = senders_logs.to_vec();
+        // Micro blocks can contain transactions, but no reward inherents. Macro blocks in turn have reward inherents,
+        // but no transactions. Thus either the tx_logs (reward inherents converted to transactions) or the
+        // senders_logs (regular transactions) must be empty.
+        if batch_info.tx_logs.is_empty() {
+            // If the tx_logs are empty, overwrite them with the senders_logs
+            batch_info.tx_logs = senders_logs
```
ii-cruz, line 802: missing a semicoma at the end for a clearer behavior

primitives/account/src/accounts.rs
```diff
@@ -541,7 +609,7 @@ impl Accounts {
         )
     }
 
-    /// Merges the log transactions from senders, receiptients and create_contracts and returns batch_info containing the merged tx_logs.
+    /// Merges the log transactions from senders, recipients and create_contracts and returns batch_info containing the merged tx_logs.
     /// This means that the batch_info inherent_logs and receipts are preserved and returned by this method.
     /// The batch_info provided as argument is expected to have its tx_logs vec empty, otherwise its content will get overwritten.
```
ii-cruz: Update the batch_info comment line explaining the need for a empty tx_log. Something like:
For macroblocks the vec should be empty, for micro it should contain the payout rewards. Thus if there are tx_logs in batch_info, there will be no changes from this function. Otherwise, the merged tx_logs will be returned.

ii-cruz (commented):
Look great @sisou :) 
More with the goal of learning. What's the deal with network_id? What is it needed for?

sisou:
@ii-cruz The network ID is required to create transactions and calculate their hash. Since the rewards are converted into transactions, the place where that transaction is constructed and the hash is computed requires the current network ID.

Basically here https://github.com/nimiq/core-rs-albatross/pull/923/files#diff-786bcafe3374f1aa08d803a68924ccf5092fc893053dc4444bd7fa680418e4e3R501 and here https://github.com/nimiq/core-rs-albatross/pull/923/files#diff-786bcafe3374f1aa08d803a68924ccf5092fc893053dc4444bd7fa680418e4e3R546

paberr:
Not a priority at the moment and needs more discussion.

PR #925, opened 2022-07-07
Include data into transfer logs if present and if recipient is a basi…

sisou:
This PR adds a `data: Option<Vec<u8>>` field to the `Log::Transfer` struct. This is to represent custom data that is added to a regular transaction **to a _basic_ account**.

To not repeat the recipient account-type check and data length everywhere, I opted to extract the `Log::Transfer` creation into a static method on the `Log` enum.

### DIFF
--- a/primitives/account/src/logs.rs
+++ b/primitives/account/src/logs.rs
@@ -152,7 +152,7 @@ pub enum Log {
 }
 
 impl Log {
-    pub fn transfer_from_transaction(transaction: &Transaction) -> Self {
+    pub fn transfer_log_from_transaction(transaction: &Transaction) -> Self {
         Log::Transfer {
             from: transaction.sender.clone(),
             to: transaction.recipient.clone(),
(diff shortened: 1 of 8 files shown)


primitives/account/src/logs.rs
```diff
@@ -136,6 +142,21 @@ pub enum Log {
 }
 
 impl Log {
+    pub fn transfer_from_transaction(transaction: &Transaction) -> Self {
```
ii-cruz:
```suggestion
    pub fn transfer_log_from_transaction(transaction: &Transaction) -> Self {
```
I found the name a bit more confusing at first glance. Another alternative I thought of is get_transfer_log_from_transaction.

ii-cruz (approved):
Thank you @sisou. Looks neat!

primitives/account/src/logs.rs
sisou: I think your suggestion makes sense!

PR #933, opened 2022-07-12
Allow Failing Transactions

### LINKED ISSUE #905: Allow failing transactions
Doing all checks in order to accept transactions into the mempool that would never fail is expensive, at least in performance/computational terms, since this would essentially require to simulate all the transactions to know wether they will succeed or not (which was the model used in our initial mempool implementation). 
Given our high throughput and performant requirements, this is no longer an option (doing a full commit check in the mempool) and thats why we moved away from this model. 
However there are still different cases (specially related to staking) that could fail when including a transaction into a block. 
Another possibility would be to do full checks during block production, but this would open a window for DoS attacks, which is something that we want to avoid. 
The desired solution is then to allow failing transactions, that is, if a transaction fails to be applied into the accounts state, we deduct fees and mark the transaction as a failed transaction. We anticipate this to not occur frequently, as this would happen when a user is using the staking contract incorrectly. 
To properly implement this functionality several changes need to take place:

- [ ] Create and implement a commit_failed / revert_failed APIs in the accounts interaction traits
- [ ] Create a wrapper around the Transaction struct that is included in a block to signal the success or failure or a given transaction
- [ ] Modify the accounts tree commit to operate on a per transaction atomicity basis
- [ ] Expose the result of applying individual transactions to the accounts tree
- [ ] Use the result of applying transactions as part of the block production process
- [ ] Use the result of applying transactions as part of the check/verification process when we receive a block
- [ ] Modify the history store to use the new transaction wrapper
- [ ] Create proper unit tests to test this new functionality

viquezclaudio:
This is an implementation of #905 
- Allow failing transactions
- Transactions are now wrapped in an ExecutedTransaction structure that conveys the result of applying such transaction


rpc-interface/src/types.rs
```diff
@@ -443,6 +461,33 @@ impl From<nimiq_block::ForkProof> for ForkProof {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub enum ExecutedTransaction {
+    Ok(Transaction),
+    Err(Transaction),
+}
```
sisou: This is a major API change again. With this, every transaction will be wrapped into a `Ok: { ...transaction }` object (or `Err: {...}`.

I propose to serde-flatten this enum, so it doesn't wrap. But at the same time, the execution status needs to be conveyed also through the API, so this needs to be "keyed" again, so that the execution status becomes a property on the transaction object, e.g. `executionResult: bool`.

sisou (dismissed):

rpc-interface/src/types.rs
```diff
@@ -281,12 +281,32 @@ impl Block {
...
                                     .map(|tx| {
-                                        Transaction::from_blockchain(
-                                            tx,
-                                            block_number,
-                                            timestamp,
-                                            head_height,
-                                        )
+                                        // We obtain an internal executed transaction
+                                        // We need to grab the internal transaction and map it to the RPC transaction structure
+                                        match tx {
```
sisou: I think you can replace this with the `from_blockchain` method of the `ExecutedTransaction` type? There the two states are already differentiated.

rpc-interface/src/types.rs
```diff
@@ -443,6 +463,36 @@ impl From<nimiq_block::ForkProof> for ForkProof {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ExecutedTransaction {
+    transaction: Transaction,
```
sisou, line 460: Still needs a serde flatten here for the transaction property, to not be nested.

block-production/tests/mod.rs
```diff
@@ -431,6 +436,564 @@ fn it_can_revert_create_staker_transaction() {
...
+
+    // We need to check that the account has the initial funds after reverting the blocks
+    assert_eq!(
+        accounts
+            .get(&KeyNibbles::from(&Address::from(&key_pair.public)), None)
+            .unwrap()
+            .balance(),
+        Coin::from_u64_unchecked(1000000000000)
+    );
+
```
nibhar, line 505: It would probably make sense to check that the state_root and history_root are the ones we would expect them to be. That would cover the account having the correct funds, but also that the staking contract is in the same state and so on.
Probably applies to other tests, too.

primitives/account/src/basic_account.rs
```diff
@@ -209,6 +210,105 @@ impl AccountTransactionInteraction for BasicAccount {
...
+            Some(account) => account.balance(),
+        };
+
+        let new_balance = Account::balance_add(current_balance, transaction.fee)?;
+        let logs = vec![Log::PayFee {
+            from: transaction.sender.clone(),
+            fee: transaction.fee,
+        }];
+        // If the new balance is zero, it means this account didnt exist before, so we don't need to create it.
+        if new_balance == Coin::ZERO {
```
nibhar, line 290: Same as above:  With what kind of transaction would execution reach this code?
If there are cases we might also just check in the beginning for 0 fee and skip a couple of lines including db get/put as nothing will be mutated.

Additionally:
Do logs with 0 fee make sense? Only commented on this once, but there are more occurrences.

primitives/account/src/logs.rs
```diff
@@ -333,6 +385,14 @@ pub struct BatchInfo {
 }
 
 impl BatchInfo {
+    pub fn new_empty() -> Self {
```
nibhar: `Default`?

blockchain/src/history/extended_transaction.rs
```diff
@@ -128,10 +129,17 @@ impl ExtendedTransaction {
     /// their transaction hash.
     pub fn tx_hash(&self) -> Blake2bHash {
         match &self.data {
-            ExtTxData::Basic(tx) => tx.hash(),
+            ExtTxData::Basic(tx) => match tx {
+                ExecutedTransaction::Ok(txn) => txn.hash(),
+                ExecutedTransaction::Err(txn) => txn.hash(),
+            },
```
nibhar: Probably makes sense to move some of these into the `ExecutedTransaction` struct.
```rust
impl ExecutedTransaction {
    pub fn hash(&self) -> Blake2bHash {..}
}
```

blockchain/src/history/history_store.rs
```diff
@@ -204,6 +204,10 @@ impl HistoryStore {
 
             match &ext_tx.data {
                 ExtTxData::Basic(tx) => {
+                    let tx = match tx {
+                        nimiq_transaction::ExecutedTransaction::Ok(txn) => txn,
+                        nimiq_transaction::ExecutedTransaction::Err(txn) => txn,
+                    };
```
nibhar:
```suggestion
                    let tx = tx.get_raw_transaction();
```

blockchain/src/history/history_store.rs
```diff
@@ -813,6 +817,11 @@ impl HistoryStore {
 
         match &ext_tx.data {
             ExtTxData::Basic(tx) => {
+                let tx = match tx {
+                    nimiq_transaction::ExecutedTransaction::Ok(txn) => txn,
+                    nimiq_transaction::ExecutedTransaction::Err(txn) => txn,
+                };
+
```
nibhar:
```suggestion
                let tx = tx.get_raw_transaction();
```

mempool/src/mempool.rs
```diff
@@ -389,6 +390,17 @@ impl Mempool {
...
+                                    // (Assuming that all pending txns in the mempool for this sender are included in a block)
+                                    if !sender_account.can_fee_be_paid(
+                                        old_tx,
+                                        new_total + old_tx.fee,
+                                        blockchain.timestamp(),
+                                    ) {
+                                        // If the fee cannot be paid for this transaction, we filter it
+                                        return true;
+                                    }
+
```
nibhar: Is this even necessary? The following lines check if `old_tx.total_value()` is still covered, which by construction is always >= `old_tx.fee`

mempool/src/verify.rs
```diff
@@ -274,6 +276,19 @@ pub(crate) async fn verify_tx<'a>(
...
 
+    // The sender must be able to at least pay the fee (in case the tx fails), assumming all pending txns in the mempool for this sender are included in a block
+    if !sender_account.can_fee_be_paid(
+        transaction,
+        sender_current_balance + transaction.fee,
+        blockchain.timestamp(),
+    ) {
+        // If the fee cannot be paid for this transaction, we reject it.
+        return Err(VerifyErr::CantPayFee);
+    }
```
nibhar: Same here.

primitives/account/src/account.rs
```diff
@@ -73,6 +73,16 @@ impl Account {
...
+
+    // This function is used by the mempool to check if a fee is payable for a particular transaction, according to the current balance for that account in the mempool
+    pub fn can_fee_be_paid(
+        &self,
+        transaction: &Transaction,
+        mempool_balance: Coin,
+        block_time: u64,
+    ) -> bool {
+        self.can_pay_fee(transaction, mempool_balance, block_time)
+    }
```
nibhar: can we not just use `AccountTransactionInteraction::can_pay_fee`?

primitives/account/src/accounts.rs
```diff
@@ -202,19 +470,64 @@ impl Accounts {
...
+        let mut recipient_receipt;
 
-        let recipients_logs = self.revert_recipients(
-            txn,
-            transactions,
-            block_height,
-            timestamp,
-            recipient_receipts,
-        )?;
+        for (index, transaction) in transactions.iter().enumerate() {
```
nibhar: Does this not have to be reverse order, too?

primitives/account/src/basic_account.rs
```diff
@@ -209,6 +210,105 @@ impl AccountTransactionInteraction for BasicAccount {
...
+
+        let account = accounts_tree
+            .get(db_txn, &key)
+            .or_else(|| {
+                if transaction.total_value() != Coin::ZERO {
+                    None
+                } else {
+                    Some(Account::Basic(BasicAccount {
+                        balance: Coin::ZERO,
+                    }))
```
nibhar, line 230: More out of interest: What would be a transaction leading to this code being executed?

primitives/account/src/staking_contract/traits.rs
```diff
@@ -436,6 +436,96 @@ impl AccountTransactionInteraction for StakingContract {
...
+
+    fn can_pay_fee(
+        &self,
+        _transaction: &Transaction,
+        _current_balance: Coin,
+        _block_time: u64,
+    ) -> bool {
+        // TODO Implementation missing: we need to support the ability to pay the fee from a third party for outgoing stacking transactions
+        // When that is ready, we can use this function to verify if the third party can actually pay the fee
+        // For now we just return true, to be consistant with the current implementation
```
nibhar: For `Unstake` we should be able to calculate this?

primitives/account/tests/accounts.rs
```diff
@@ -801,3 +815,124 @@ fn accounts_performance_history_sync_batches_many_to_many() {
...
+
+    let (_, executed_txns) = accounts
+        .commit(&mut db_txn, &vec![tx.clone()], &[], 1, 200)
+        .unwrap();
+
+    assert_eq!(executed_txns, vec![ExecutedTransaction::Err(tx.clone())]);
+
+    db_txn.commit();
+
+    // We should deduct the fee from the staking contract balance
```
nibhar:
```suggestion
    // We should deduct the fee from the vesting contract balance
```

primitives/account/src/vesting_contract.rs
```diff
@@ -256,6 +256,130 @@ impl AccountTransactionInteraction for VestingContract {
...
+            _ => {
+                return Err(AccountError::TypeMismatch {
+                    expected: AccountType::Vesting,
+                    got: account.account_type(),
+                })
+            }
+        };
+
+        // Note that in this type of transactions the fee is paid (deducted) from the contract balance
+        let new_balance = Account::balance_sub(account.balance(), transaction.fee)?;
```
nibhar, line 284: Do we not have to check `min_cap` here as well as the correctness of the recipients signature?
Both of these checks are performed for `commit_outgoing_transaction`.

primitives/block/src/block.rs
```diff
@@ -231,7 +231,15 @@ impl Block {
...
                 .body
                 .as_ref()
-                .map(|ex| ex.transactions.iter().map(|tx| tx.fee).sum())
+                .map(|ex| {
+                    ex.transactions
+                        .iter()
+                        .map(|tx| match tx {
+                            ExecutedTransaction::Ok(tx) => tx.fee,
+                            ExecutedTransaction::Err(tx) => tx.fee,
+                        })
```
nibhar: I'd say either
```suggestion
                        .map(|tx| tx.get_raw_transaction().fee()
```
or add `ExecutedTransaction::fee(&self)`

primitives/block/src/micro_block.rs
```diff
@@ -75,7 +76,20 @@ pub struct MicroBody {
...
+
+impl MicroBody {
+    pub fn get_raw_transactions(&self) -> Vec<Transaction> {
+        // Extract the transactions from the block
+        self.transactions
+            .iter()
+            .map(|txn| match txn {
+                ExecutedTransaction::Ok(txn) => txn.clone(),
+                ExecutedTransaction::Err(txn) => txn.clone(),
+            })
```
nibhar:
```suggestion
            .map(|txn| txn.get_raw_transaction().clone())
```

mempool/src/mempool.rs
viquezclaudio: That is true, the _total_value_ check is stronger than the fee check because _total_value_ already incudes the fee, which seems to be sufficient for basic transactions.
But is this sufficient for other transaction types? i.e: vesting and htlc contracts?

mempool/src/verify.rs
viquezclaudio: This check is designed for vesting/htlc contracts, because apart from just verifying if the fee can be paid from a balance perspective, in the case of vesting/htlc we need to make sure the fee can actually be deducted from the contract (contract already vested with enough funds, etc)

primitives/account/src/basic_account.rs
viquezclaudio: This code is inherited from _commit_outgoing_transaction_ and I think it is used for signaling transactions: i.e: if the transaction total value is 0 and the account does not exist, we just use a "mock" basic account.
It was added in commit _a5ed6b79d7ca2883fe19b28e8ca8da6aaafdb294_ and if I remember correctly it was added because there was some panic or some issue that we were experimenting, however I don't know exactly what a signaling transaction that fails means, so if that is never a possibility, we can just remove this

primitives/account/src/basic_account.rs
viquezclaudio: Same as above, signaling, but I guess this can be safely removed.
I think for consistency we should allow logs with 0 fee.

primitives/account/src/logs.rs
viquezclaudio: just realized this is dead code, is not used

primitives/account/src/staking_contract/traits.rs
viquezclaudio: yes good catch!
The only problem is that I need the _accounts_tree_  to get the staking contract in order to obtain the staker balance, but I do not have the accounts_tree in this interface, is there an alternative way to do this?

primitives/account/src/vesting_contract.rs
viquezclaudio: This is covered by _can_pay_fee_ , the mempool before accepting the transaction needs to call _can_pay_fee_ which would check that the transaction signer is contract owner.
Do you need think it is necessary to also add this check here?

Saiv46:
Do I miss something? How and why do we need to handle failed transactions? If account tries to move more funds that it has, transaction will be rejected and new transaction will be needed.

viquezclaudio:
Correct, we perform balance checks in the mempool.
This is more intended for staking contract misusage (trying to delete a validator that no longer exists, etc)

PR #937, opened 2022-07-15
Iicruz/automatic validator activation

ii-cruz:
The automatic reactivation is introduced to the validator. This configuration is provided upon creation. It may also be changed using the rpc server commands. 
When a validator intentionally wants to inactivate itself, it should first ensure the automatic reactivation configuration is disabled, otherwise, the validator will send a reactivate transaction automatically.

Other changes:
- RPC server: added a command to inactivate the validator for testing purposes and to allow the feature to be fully used.
- RPC server: Corrected the unlock return type. It was giving a JSON parsing error.

## Pull request checklist

### DIFF
--- a/rpc-server/src/dispatchers/wallet.rs
+++ b/rpc-server/src/dispatchers/wallet.rs
@@ -101,7 +101,7 @@ impl WalletInterface for WalletDispatcher {
         address: Address,
         passphrase: Option<String>,
         _duration: Option<u64>,
-    ) -> Result<(), Error> {
+    ) -> Result<bool, Error> {
         let passphrase = passphrase.unwrap_or_default();
         let account = self
             .wallet_store
@@ -114,7 +114,7 @@ impl WalletInterface for WalletDispatcher {
 
         self.unlocked_wallets.write().insert(unlocked_account);
 
-        Ok(())
+        Ok(true)
     }
 
     async fn is_account_unlocked(&mut self, address: Address) -> Result<bool, Error> {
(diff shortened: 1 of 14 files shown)


rpc-server/src/dispatchers/wallet.rs
```diff
@@ -114,7 +114,7 @@ impl WalletInterface for WalletDispatcher {
 
         self.unlocked_wallets.write().insert(unlocked_account);
 
-        Ok(())
+        Ok(true)
```
paberr, line 117: Why does it now return a bool? It's only ever set to true, isn't it?

ii-cruz: It gives a json parsing error if it returns nothing/null. My understanding is that this issue was not detected previously because only Soeren is using the rpc client and he is using the javascript one.

validator/src/validator.rs
```diff
@@ -628,12 +649,41 @@ impl<TNetwork: Network, TValidatorNetwork: ValidatorNetwork>
...
+            Coin::ZERO,
+            validity_start_height,
+            blockchain.network_id(),
+        )
+        .unwrap(); // TODO: Handle transaction creation error
+        let tx_hash = reactivate_transaction.hash();
+
+        let cn = self.consensus.clone();
+        tokio::spawn(async move {
+            if cn.send_transaction(reactivate_transaction).await.is_err() {
```
viquezclaudio: Since the mempool txn prioritization PR is already merged, you should include this transaction with high priority, similar to how it is now done for the unpark transaction

viquezclaudio (approved):

viquezclaudio:
LGTM, feel free to merge these changes

PR #961, opened 2022-07-28
limiting by peer_id and request type

### LINKED ISSUE #945: Implement network rate limiting
The objective of this task is to Implement rate limiting for the request/response layer in our client to limit the number of requests that a peer can perform per time unit.
Add rate limiting per peer & per request type.

ii-cruz:

Limit the number of requests that a peer can perform per time unit and request type. The time unit used was changed from block height to (tokio) time. This second version of the PR has the code placed on the network crate.

## Pull request checklist


consensus/src/consensus/request_response.rs
```diff
@@ -22,45 +27,109 @@ impl<N: Network> Consensus<N> {
...
         network: &Arc<N>,
         stream: BoxStream<'static, (Req, N::RequestId, N::PeerId)>,
         blockchain: &Arc<RwLock<Blockchain>>,
+        max_requests: u16,
+        block_range: u32,
     ) -> impl Future<Output = ()> {
         let blockchain = Arc::clone(blockchain);
         let network = Arc::clone(network);
+        let peers_request_limit = Arc::new(futures::lock::Mutex::new(HashMap::with_capacity(
+            network.get_peers().len() * 5,
```
ii-cruz: This capacity was set in an arbitrary manner, I also considered just leaving it to whatever is the default of the structure. Suggestions?

consensus/src/consensus/request_response.rs
```diff
@@ -22,45 +27,109 @@ impl<N: Network> Consensus<N> {
...
             stream
                 .for_each_concurrent(
                     Self::MAX_CONCURRENT_HANDLERS,
                     |(msg, request_id, peer_id)| {
-                        let request_id = request_id;
                         let network = Arc::clone(&network);
                         let blockchain = Arc::clone(&blockchain);
+                        let peers_request_limit = Arc::clone(&peers_request_limit);
                         async move {
-                            let blockchain = Arc::clone(&blockchain);
```
ii-cruz, line 59: These moves were removed since they seem to be unnecessary to me. We can just use the cloned Arcs declared just before this line (right after for_each_concurrent). Please double check this and let me know if I'm missing something.

primitives/src/policy.rs
```diff
@@ -31,6 +31,23 @@ pub const VERSION: u16 = 1;
 /// Number of available validator slots. Note that a single validator may own several validator slots.
 pub const SLOTS: u16 = 512;
 
+// The max number of request to be processed per peerID and per request type.
+
+/// The range to restrict the responses to the requests on the consensus network.
+pub const MAX_RESPONSE_REQUEST_BLOCK_RANGE: u32 = 1000;
```
ii-cruz:
```suggestion
pub const MAX_RESPONSE_REQUEST_BLOCK_RANGE: u32 = 10;
```
if we end up just merging this prior to deciding which are the limits for the request types, I would suggest having a smaller range, so that no unwanted delays are introduced.

consensus/src/consensus/mod.rs
```diff
@@ -366,3 +366,54 @@ impl<N: Network> Future for Consensus<N> {
...
+            block_range: _block_range,
+            last_reset: 0,
+            occurrences_counter: 0,
+        }
+    }
+
+    /// Updates the block_range if needed and then increments the counter of number of requests by
+    /// the specified number.
+    /// Receives the number to increment the counter and the current time measured in blocks.
+    pub fn increment_and_is_allowed(&mut self, number: u16, now: u32) -> bool {
```
nibhar:
```suggestion
    pub fn increment_and_is_allowed(&mut self, request_count: u16, current_block: u32) -> bool {
```

consensus/src/consensus/mod.rs
```diff
@@ -366,3 +366,54 @@ impl<N: Network> Future for Consensus<N> {
...
+
+impl RateLimit {
+    pub fn new(allowed_occurrences: u16, _block_range: u32) -> Self {
+        RateLimit {
+            allowed_occurrences,
+            block_range: _block_range,
+            last_reset: 0,
+            occurrences_counter: 0,
+        }
+    }
```
nibhar:
```suggestion
    pub fn new(allowed_occurrences: u16, block_range: u32) -> Self {
        RateLimit {
            allowed_occurrences,
            block_range,
            last_reset: 0,
            occurrences_counter: 0,
        }
    }
```

consensus/src/consensus/request_response.rs
nibhar: Where does the `* 5` come from?

consensus/src/consensus/request_response.rs
```diff
@@ -22,45 +27,109 @@ impl<N: Network> Consensus<N> {
...
-                            let network = Arc::clone(&network);
-                            let request_id = request_id;
                             tokio::spawn(async move {
-                                trace!("[{:?}] {:?} {:#?}", request_id, peer_id, msg);
+                               trace!("[{:?}] {:?} {:#?}", request_id, peer_id, msg);
```
nibhar:
```suggestion
                                trace!("[{:?}] {:?} {:#?}", request_id, peer_id, msg);
```

consensus/src/messages/handlers.rs
```diff
@@ -1,6 +1,7 @@
 use std::collections::HashSet;
```
nibhar: I believe all changes in this file are unnecessary.

consensus/src/consensus/request_response.rs
```diff
@@ -84,3 +153,89 @@ impl<N: Network> Consensus<N> {
...
+
+        match response {
+            Ok(_) => {}
+            Err(e) => {
+                panic!(
+                    "First request should be the only one working, however it failed - {:?}",
+                    e
+                )
+            }
+        }
```
nibhar:
```suggestion
        // Make sure that the first request resolves as expected as it must not exceed the rate limit.
        assert!(response.is_ok(), "First request must return Ok Result");
```

consensus/src/consensus/request_response.rs
```diff
@@ -84,3 +153,89 @@ impl<N: Network> Consensus<N> {
...
+
+            match response {
+                Err(RequestError::InboundRequest(InboundRequestError::Timeout)) => {}
+                Ok(_) => {
+                    panic!("Expected to timeout due to exceeding the max rate limit, returned Ok instead.");
+                }
+                Err(e) => {
+                    panic!("Expected to timeout due to exceeding the max rate limit, failed with a different error message - {:?}", e);
+                }
+            }
```
nibhar:
```suggestion
            // Make sure the request timed out and did not return any other Error
            assert_eq!(
                response, 
                Err(RequestError::InboundRequest(InboundRequestError::Timeout)), 
                "Subsequent requests must return Timeouts",
            );
```

consensus/src/consensus/request_response.rs
```diff
@@ -22,45 +27,109 @@ impl<N: Network> Consensus<N> {
...
                             tokio::spawn(async move {
-                                trace!("[{:?}] {:?} {:#?}", request_id, peer_id, msg);
+                               trace!("[{:?}] {:?} {:#?}", request_id, peer_id, msg);
+
+                                // Gets lock of peer requests limits read and write on it
+                                let mut peers_request_limit = peers_request_limit.lock().await;
+                                // If the peer has never sent a request of this type, creates a new entry
+                                let requests_limit = peers_request_limit
+                                    .entry(peer_id)
+                                    .or_insert_with(|| RateLimit::new(max_requests, block_range));
```
nibhar: These entries never get evicted, letting this Map grow forever.

We need to introduce means to remove peers from that map once we are no longer connected to them. Likely it would also be necessary to wait with the removal until the `last_reset + block_range` block was pushed as one could otherwise bypass the rate limiting by disconnecting and reconnecting.

consensus/src/consensus/request_response.rs
```diff
@@ -22,45 +27,109 @@ impl<N: Network> Consensus<N> {
...
+            network,
+            stream,
+            blockchain,
+            policy::MAX_RESPONSE_REQUEST_HEAD,
+            policy::MAX_RESPONSE_REQUEST_BLOCK_RANGE,
+        ));
     }
 
-    pub(crate) fn request_handler<Req: Handle<Req::Response> + Request>(
+    pub(crate) fn request_handler<Req: Handle<Req::Response, N> + Request>(
```
nibhar: the `, N` should be unnecessary as it was already specified on the `impl<N: Network>` in Line 22.
```suggestion
    pub(crate) fn request_handler<Req: Handle<Req::Response> + Request>(
```

nibhar (dismissed):

consensus/src/consensus/request_response.rs
nibhar: It did not show me this comment while reviewing.

The default size should be fine. `network.get_peers().len()` should also be fine, as peers will likely send requests at some point while connected and there will not be many (if any) when this executes. 

Also note that these get initialized very early, so I would probably just go with default size. Another option might be the peer count floor a node would try to maintain, as eventually that number of peers should be reached at least.

primitives/src/policy.rs
```diff
@@ -31,6 +31,23 @@ pub const VERSION: u16 = 1;
...
+/// The The max number of MacroChain requests per peer.
```
viquezclaudio: All these comments have an extra "The"

primitives/src/policy.rs
```diff
@@ -31,6 +31,23 @@ pub const VERSION: u16 = 1;
...
+pub const MAX_RESPONSE_REQUEST_MACRO_CHAIN: u16 = 1000;
```
viquezclaudio: This is just a minor naming suggestion, instead of "MAX_RESPONSE_REQUEST_*" rename it to "MAX_REQUEST_RESPONSE_*" because "request response" is how we usually refer to this component and also to match the file name as well: request_response.rs

consensus/src/consensus/request_response.rs
ii-cruz: This will be complicated to implement I believe. Basically, the rate-limiting is being implemented where the processing of the requests is at, which is out of the network.  The network is where the disconnect peer function exists. 

I think the easiest thing is to include the hashmap of the rate counters inside the network implementation and introduce some functions on the network interface rather than creating it on the request_response. Can you think of something else?

consensus/src/messages/mod.rs
```diff
@@ -58,6 +60,9 @@ impl RequestCommon for RequestMacroChain {
     type Kind = RequestMarker;
     const TYPE_ID: u16 = 200;
     type Response = MacroChain;
+
```
paberr: You can remove the empty line for consistency with the other implementations.

consensus/src/messages/mod.rs
```diff
@@ -58,6 +60,9 @@ impl RequestCommon for RequestMacroChain {
...
+    const MAX_REQUESTS: u32 = policy::MAX_REQUEST_RESPONSE_MACRO_CHAIN;
+    const TIME_WINDOW: Duration = policy::MAX_REQUEST_RESPONSE_TIME_WINDOW;
```
paberr: Since this seems to be the same for most implementations, you could make it default and only override if necessary.

In Rust you can do something like:
```Rust
trait Test {
    const A: u32 = 1;
}

struct B;

impl Test for B {
    const A: u32 = 2;
}
```

consensus/src/sync/history/sync_stream.rs
```diff
@@ -207,6 +207,7 @@ impl<TNetwork: Network> Stream for HistorySync<TNetwork> {
 
 #[cfg(test)]
 mod tests {
+    // use std::collections::HashMap;
```
paberr: Remove this.

handel/src/update.rs
```diff
@@ -83,4 +85,7 @@ impl<
     // messages upon receiving them.
     const TYPE_ID: u16 = C::TYPE_ID;
     type Response = ();
+
+    const MAX_REQUESTS: u32 = policy::MAX_REQUEST_RESPONSE_LEVEL_UPDATE_MESSAGE;
+    const TIME_WINDOW: Duration = policy::MAX_REQUEST_RESPONSE_TIME_WINDOW;
```
paberr: See comment about default.

network-interface/src/request/mod.rs
```diff
@@ -121,6 +124,8 @@ pub trait RequestCommon:
     type Kind: RequestKind;
     const TYPE_ID: u16;
     type Response: Deserialize + Serialize + Send;
+    const MAX_REQUESTS: u32;
+    const TIME_WINDOW: Duration;
```
paberr: Here's where you can add the default.

network-libp2p/src/rate_limiting.rs
```diff
@@ -0,0 +1,47 @@
...
+            time_window,
+            last_reset,
+            occurrences_counter: 0,
+        }
+    }
+
+    /// Updates the last_reset if needed and then increments the counter of number of requests by
+    /// the specified number.
+    /// Receives the number to increment the counter and the current time measured in seconds.
+    pub fn increment_and_is_allowed(&mut self, request_count: u32, current_time: Instant) -> bool {
```
paberr: I was curious to know why the `current_time` is passed as an argument and not calculated inside the function.
Was there a specific reason for it?

network-libp2p/tests/request_response.rs
```diff
@@ -394,3 +541,297 @@ async fn test_valid_request_no_response_no_receiver() {
         ),
     };
 }
+
+use nimiq_network_interface::peer::CloseReason;
+
+async fn disconnect_succefully(net1: &Arc<Network>, net2: &Arc<Network>) {
```
paberr:
```suggestion
async fn disconnect_successfully(net1: &Arc<Network>, net2: &Arc<Network>) {
```

network-libp2p/tests/request_response.rs
```diff
@@ -394,3 +541,297 @@ async fn test_valid_request_no_response_no_receiver() {
...
+    let event1 = events1.next().await.unwrap().unwrap();
+    log::trace!(event = ?event1, "Event 1");
+    assert_peer_left(&event1, &net2.get_local_peer_id());
+
+    let event2 = events2.next().await.unwrap().unwrap();
+    log::trace!(event = ?event2, "Event 2");
+    assert_peer_left(&event2, &net1.get_local_peer_id());
+}
+
+async fn reconnect_succefully(net1: &Arc<Network>, addr1: Multiaddr, net2: &Arc<Network>) {
```
paberr:
```suggestion
async fn reconnect_successfully(net1: &Arc<Network>, addr1: Multiaddr, net2: &Arc<Network>) {
```

network-libp2p/tests/request_response.rs
```diff
@@ -394,3 +541,297 @@ async fn test_valid_request_no_response_no_receiver() {
...
+        assert!(net2.has_peer(net1.get_local_peer_id()));
+        assert!(net1.has_peer(net2.get_local_peer_id()));
+
+        log::info!(
+            "{:?} sends request {:?}",
+            net1.get_local_peer_id(),
+            test_request.clone(),
+        );
+
+        // Send the request and get future for the response
```
paberr:
```suggestion
        // Send the request and get future for the response.
```

network-libp2p/tests/request_response.rs
```diff
@@ -394,3 +541,297 @@ async fn test_valid_request_no_response_no_receiver() {
...
+        let response = net2.request::<TestRequest4>(test_request.clone(), net1.get_local_peer_id());
+
+        log::info!("Succeed {:?}, {:?}", i, Instant::now());
+
+        // Check the received response
```
paberr:
```suggestion
        // Check the received response.
```

network-libp2p/tests/request_response.rs
```diff
@@ -394,3 +541,297 @@ async fn test_valid_request_no_response_no_receiver() {
...
+        log::info!("Fail {:?}, {:?}", i, Instant::now());
+
+        log::debug!(
+            "{:?} sends request {:?}",
+            net1.get_local_peer_id(),
+            test_request.clone(),
+        );
+        let response = net2.request::<TestRequest4>(test_request.clone(), net1.get_local_peer_id());
+
+        // Check the received response
```
paberr:
```suggestion
        // Check the received response.
```

network-libp2p/tests/request_response.rs
```diff
@@ -394,3 +541,297 @@ async fn test_valid_request_no_response_no_receiver() {
...
+
+#[test(tokio::test)]
+async fn it_can_limit_requests_rate() {
+    let (net1, net2) = TestNetwork::create_connected_networks().await;
+    let net1 = Arc::new(net1);
+    let net2 = Arc::new(net2);
+
+    let test_response = TestResponse4 { response: 43 };
+
+    // Subscribe for receiving requests
```
paberr:
```suggestion
    // Subscribe for receiving requests.
```

network-libp2p/tests/request_response.rs
```diff
@@ -394,3 +541,297 @@ async fn test_valid_request_no_response_no_receiver() {
...
+            let test_response = test_response.clone();
+            let network1 = Arc::clone(&network1);
+            async move {
+                let _result = network1
+                    .respond::<TestRequest4>(request_id, test_response.clone())
+                    .await;
+            }
+        });
+
+    // Spawn the request listener future
```
paberr:
```suggestion
    // Spawn the request listener future.
```

network-libp2p/tests/request_response.rs
```diff
@@ -394,3 +541,297 @@ async fn test_valid_request_no_response_no_receiver() {
...
+    let ((net1, addr1), (net2, _), (net3, _), (net4, _)) =
+        TestNetwork::create_4_connected_networks().await;
+    let net1 = Arc::new(net1);
+    let net2 = Arc::new(net2);
+    let net3 = Arc::new(net3);
+    let net4 = Arc::new(net4);
+
+    let test_response = TestResponse4 { response: 43 };
+
+    // Subscribe for receiving requests
```
paberr:
```suggestion
    // Subscribe for receiving requests.
```

network-libp2p/tests/request_response.rs
```diff
@@ -394,3 +541,297 @@ async fn test_valid_request_no_response_no_receiver() {
...
+#[test(tokio::test)]
+async fn it_can_reset_requests_rate_with_reconnections() {
+    let ((net1, addr1), (net2, _), (net3, _), _) = TestNetwork::create_4_connected_networks().await;
+    let net1 = Arc::new(net1);
+    let net2 = Arc::new(net2);
+    let net3 = Arc::new(net3);
+
+    let test_response = TestResponse4 { response: 43 };
+
+    // Subscribe for receiving requests
```
paberr:
```suggestion
    // Subscribe for receiving requests.
```

network-libp2p/tests/request_response.rs
```diff
@@ -394,3 +541,297 @@ async fn test_valid_request_no_response_no_receiver() {
...
+            let test_response = test_response.clone();
+            let network1 = Arc::clone(&network1);
+            async move {
+                let _result = network1
+                    .respond::<TestRequest4>(request_id, test_response.clone())
+                    .await;
+            }
+        });
+
+    // Spawn the request listener future
```
paberr:
```suggestion
    // Spawn the request listener future.
```

network-libp2p/src/network.rs
```diff
@@ -1298,6 +1352,149 @@ impl Network {
...
+        let mut rate_limits_pending_deletion_l = rate_limits_pending_deletion.lock();
+        let mut peer_request_limits_l = peer_request_limits.lock();
+
+        // Go through all existing request types of the given peer and deletes the limit counters if possible or marks it for deletion.
+        if let Some(request_limits) = peer_request_limits_l.get_mut(&peer_id) {
+            request_limits.retain(|req_type, rate_limit| {
+                // Gets the requests limit and deletes it if no counter info would be lost, otherwise places it as pending deletion.
+                if !rate_limit.can_delete(TokioInstant::now()) {
+                    rate_limits_pending_deletion_l
+                        .push_back(((peer_id, *req_type), rate_limit.next_reset_time()));
```
paberr, line 1409: If I'm not mistaken, this could lead to adding the same thing multiple times, does it?

paberr: And if so, is it a problem?

network-libp2p/src/network.rs
```diff
@@ -1298,6 +1352,149 @@ impl Network {
...
+            let current_timestamp = TokioInstant::now();
+            if block_expiration <= &current_timestamp {
+                let mut peer_request_limits_l = peer_request_limits.lock();
+
+                if let Some(peer_req_limits) =
+                    peer_request_limits_l
+                        .get_mut(peer_id)
+                        .and_then(|peer_req_limits| {
+                            if let Some(rate_limit) = peer_req_limits.get(req_type) {
+                                // If the peer has reconnected the rate limit may be inforcing a new limit. In this case we only remove the pending deletion.
```
paberr:
```suggestion
                                // If the peer has reconnected the rate limit may be enforcing a new limit. In this case we only remove the pending deletion.
```

network-libp2p/src/network.rs
```diff
@@ -1298,6 +1352,149 @@ impl Network {
...
+
+        // Iterates from the oldest to the most recent object of the queue, deletes the entries that have expired.
+        // The queue is ordered from the oldest to the most recent object added, meaning that its loosely ordered by expiration date.
+        // Thus, the iteration stops upon the first object that is hasn't yet expired. In the case where the queue is out of order,
+        // the clean up will reach it after a few more blocks (max block_range).
+        while let Some(((peer_id, req_type), block_expiration)) =
+            rate_limits_pending_deletion_l.front()
+        {
+            let current_timestamp = TokioInstant::now();
+            if block_expiration <= &current_timestamp {
```
paberr: Is it still correct to call it `block_expiration` as it's not based on blocks anymore?

network-libp2p/src/network.rs
```diff
@@ -1298,6 +1352,149 @@ impl Network {
...
+    fn clean_up(
+        peer_request_limits: Arc<Mutex<HashMap<PeerId, HashMap<u16, RateLimit>>>>,
+        rate_limits_pending_deletion: Arc<Mutex<VecDeque<((PeerId, u16), TokioInstant)>>>,
+    ) {
+        let mut rate_limits_pending_deletion_l = rate_limits_pending_deletion.lock();
+
+        // Iterates from the oldest to the most recent object of the queue, deletes the entries that have expired.
+        // The queue is ordered from the oldest to the most recent object added, meaning that its loosely ordered by expiration date.
+        // Thus, the iteration stops upon the first object that is hasn't yet expired. In the case where the queue is out of order,
+        // the clean up will reach it after a few more blocks (max block_range).
```
paberr: Is it still correct to talk about blocks as it's not based on blocks anymore?

paberr (changes requested):
Generally looks very good. 👍 Just a few style issues.

network-libp2p/src/network.rs
ii-cruz: nope. Good catch

network-libp2p/src/network.rs
ii-cruz: If we receive two connection closed messages we can try to remove the rate limits of the said peer again and place the limits again on the queue for pending deletion queue. Those will be cleared out similarly at the same time when the expiration date has passed. 

So duplicates are possible in case of duplicate connection closed messages but they would be cleared out all at the same time eventually (max time_window defined).

paberr: That sounds good to me then.

paberr (commented):
It looks like some comments are still unresolved.

paberr (approved):
Looks alright now 🥳

PR #962, opened 2022-07-29
Add a new type of blocks: skip blocks.

### LINKED ISSUE #867: View Changes only come into effect after block is published
## Current Behaviour and Problem
View changes currently only change the validator state when being attached to a block. This can lead to double-spending and re-organisation attacks:

1. Validator A is the current block producer (and will be the next block producer after a view change).
2. It does not produce a block and waits until the view change completes.
3. **Problem:** Despite the view change being known by other validators, it won't have any effect until it is published in a block on the new view. This block would have to be produced by validator A, though.
4. Validator A instead publishes the original block, which is accepted by other validators due to this issue.
5. At some point later in time (before a macro block), validator A can publish the view-changed block and cause a reorg.

## Proposed Solution
The problem occurs because the completed view-change itself does not trigger any permanent state change in the validator.

The proper fix is: view changes should be sent around not only as part of the block but as individual messages to make sure all validators receive it. Completed view changes should have an immediate effect and validators should not accept blocks on the old view once the view change is complete.

jsdanielh:
Add skip block as a new type of block. This also required changes in
several places in the code mainly in `blockchain` and `validator`
crates such that this type of block can be produced and accepted by
the blockchain.

Also remove the concept of view change from the code and rename the
view change delay in favor of the block producer timeout. This
removes the `view_number` field in either macro and micro blocks.
For micro blocks it also implies that there are no view change
proofs as justification.

This fixes #867.

## Pull request checklist


nibhar:
I started looking into this PR. I noticed you introduced a new `struct SkipBlock { ... }`.

What is the reasoning behind that? I believe we could've just changed `MicroJustification` to 
```rust
pub enum MicroJustification {
    Micro(Signature),
    Skip(SkipBlockProof),
}
```

That I think would reduce the change set quite drastically and less code to maintain, as almost everything we need to do for skip blocks we also need to do for micro blocks. The necessary steps that change, like inherents, could still be done since we can distinguish them by their proof, making the changes much more local.

jsdanielh:
> I started looking into this PR. I noticed you introduced a new `struct SkipBlock { ... }`.
> 
> What is the reasoning behind that? I believe we could've just changed `MicroJustification` to
> 
> ```rust
> pub enum MicroJustification {
>     Micro(Signature),
>     Skip(SkipBlockProof),
> }
> ```
> 
> That I think would reduce the change set quite drastically and less code to maintain, as almost everything we need to do for skip blocks we also need to do for micro blocks. The necessary steps that change, like inherents, could still be done since we can distinguish them by their proof, making the changes much more local.

The reasoning was based on these thoughts:
- Semantics: It's a new block type and I thought it would be more clear if instead of recycling the `micro` block concept, we totally differentiate it with a new type. Also the change to using an `enum` in the block justification I think wouldn't be enough since there are other changes from `micro` blocks such as no `body_root` (since it doesn't have a body) which also allows us to actually `assert` in the `Block` `primitives` if someone actually tries to get the body of a skip block. Lastly, identification of a skip block becomes really easy (it's just a matter of going to the `Block` primitives and call `is_skip`) instead of going to the `Justification` `enum` and check if this is a skip block.
- Maintainability: Although I agree that less code to maintain is better, `Rust`'s `match` pattern applied to a `Block` type, force the developer to actually handle a `SkipBlock` which may not always be something like "_do the same as in micros_" but could imply different handling.
- Size optimizations: Although the savings are not that significant, using a different type for `SkipBlocks` saves a couple of bytes in each `micro` block namely:
  - The byte for the `enum` identification in the `Justification`.
  - The byte for the `Option` identification that we would have to add to `body_root` as explained before.

jsdanielh:
> I started looking into this PR. I noticed you introduced a new `struct SkipBlock { ... }`.
> 
> What is the reasoning behind that? I believe we could've just changed `MicroJustification` to
> 
> ```rust
> pub enum MicroJustification {
>     Micro(Signature),
>     Skip(SkipBlockProof),
> }
> ```
> 
> That I think would reduce the change set quite drastically and less code to maintain, as almost everything we need to do for skip blocks we also need to do for micro blocks. The necessary steps that change, like inherents, could still be done since we can distinguish them by their proof, making the changes much more local.

@nibhar Done. Ready for a another review

jsdanielh:
@nibhar Added the suggested extra checks in the last commit

PR #939, opened 2022-07-15
RPC validator by address change stakers model

Eligioo:
## Pull request checklist

- [x] All tests pass. Demo project builds and runs.
- [x] I have resolved any merge conflicts.

## What's in this pull request?

Currently the `get_validator_by_address` RPC method returns a list of stakers something like this:

```
{
    ...
    stakers: {
        'NQ68 SN2E JS5Y 81K0 LFCD PDQM 1Q3U CFQC M3EB': 5019227844,
        'NQ84 9YK6 GGAD QDJG G36V DSC0 T5CE 70J0 35FA': 5345850189,
        'NQ26 9VB9 BYG6 P2A8 L9VX N5UD JNNN VDK2 3F1J': 2500452602,
        'NQ79 JA7K CKTA NA1Y 132A 27DR 8DD9 KSGK YPL1': 4280474570,
        'NQ06 LN8U XASN DULU 7E0E S5J0 X98Y 9XX3 2T01': 5279178854,
        'NQ37 M3DJ 4H9M J1G0 M8FT 8JRC PMR2 3KF8 BLVH': 2501716832,
        'NQ82 QAGH NMDY 05QN 6PEU A8Y1 295A 8EYS KVVL': 1026045298,
        'NQ56 XBC7 YNRV 426X RYJA 1D5Q A6BL J0NX 7C3N': 5447595387,
        'NQ77 2EHU RFRN PYMR NT2A DE7B QSQH R274 CVT6': 2076636785,
        'NQ95 BDRT AXGU YKLF DXPA SY49 XM61 RDYQ Q3CH': 5259736823,
        'NQ56 M3J9 N18Y 5KE4 PT1J YTAE Q0G8 31JE FFCB': 150280652,
        'NQ47 QHA5 04RC A2GF BSRL DVA8 NR65 3VNA ATL0': 31101308333,
        'NQ91 YJNG P54M J335 JRVQ QDLY 4CYJ X5K9 R8RT': 1000000,
        'NQ15 447K N4UD EFGG T7DE SX90 XLXH 87D9 SBNV': 15767555561,
        'NQ75 YK89 T1CP PP91 BT30 S4TA 9SAK R9C3 SX0V': 5291847757,
        'NQ53 CFRG QYYC HNFT XLKR RH6X H381 K1K2 K985': 4949844002,
        'NQ27 F703 GJUQ YBQ6 Y5B5 B3XS A56Q CDM6 S6NQ': 5271704262
    }
    ...
}
```

[...]

Issue #861, opened 2022-05-19
[RPC][request] Improved / extended websocket functionality


redmaner:
Hi team,

Since beta community developers are now able to request access to the albatross RPC interface to provide feedback, I thought it was time to start an issue with some feature requests regarding the RPC inteface of my own. 

## 1. Improved websocket `headSubscribe` functionality
The RPC inteface supports a specific RPC method over websockets `headSubscribe` which pushes the block hash of new blocks that have been validated / added to the blockchain. This feature provides some first hooks to receive active updates from the blockchain, but is still rather limited in my opinion. 

### Problem 
Currently the `headSubscribe` method pushes block hashes to the client. It is still required for the client to retrieve the entire block via`getBlockByHash` in order to discern what type of block has been added to the blockchain and make a decision on whether to do something with it or not. The current implementation has the effect of creating a push pull mechanism where the client is still pulling all blocks, which in my view moves potential efficiency gains of events pushed to the client away.

I foresee applications that may be interested in only specific blocks, for example only macro blocks or only blocks with transactions or only an election block, but in order to filter these we have to still request all blocks from the node to filter the blocks we are interested in.

### Possible solutions 
1. One way to already have some more idea from a client side perspective if you are interested in a block would be to receive more information than only the hash. This could include the block hash, the block number, the block type and in case of a macro block if it is a checkpoint block or election block. This would give applications options to implement event driven design, without having to pull data for all blocks received which would make implementations more efficient. 
2. Another option would be to push the entire block that you would also receive from responses like `getBlockByHash`. Including transactions or only transaction hashes could be a parameter that is also used for `getLatestBlock`, `getBlockByHash` and `getBlockByNumber`

## 2. websocket events for validator and or staker during election block 
This request relates to adding functionality to the websocket RPC interface to receive events / updates related to the staking contract, especially or maybe evenly exclusively when an election block begins. 

### Problem 
One of the major challenges for pools in particular is to keep track of the staking contract. The staking contract always receives transactions and only has a solidified state during the last micro block of an epoch and during the election block. Before and after these blocks, the staking contract can change. It is therefore important for pools to check in this small window of two blocks to check if a validator is elected for a new epoch and receive the list of delegated stakers. Currently the only way to do this is to keep track of the blockheight and make sure to request the data within the limited time window. 

### Solution 
[...]

PR #963, opened 2022-07-31
Update return result of getActiveValidators

Eligioo:
## Pull request checklist

Update the return type of `getActiveValidators` to align it with `getValidatorByAddress` after it changed in #939.

In this PR, the return type of `getActiveValidators` has now become `Result<Vec<Staker>, Self::Error>`. Officially we are not returning a list of `Staker`s but a list of `Validator`s, hence the RPC name, but this was the best solution without defining a new struct. Also `Validator` already has been defined as a RPC type. If you want this to become something else, I'm more than happy to apply feedback.

### DIFF
--- a/rpc-server/src/dispatchers/blockchain.rs
+++ b/rpc-server/src/dispatchers/blockchain.rs
@@ -460,17 +460,16 @@ impl BlockchainInterface for BlockchainDispatcher {
     }
 
     /// Returns a map of the currently active validator's addresses and balances.
-    async fn get_active_validators(&mut self) -> Result<Vec<Staker>, Error> {
-        let staking_contract = self.blockchain.read().get_staking_contract();
+    async fn get_active_validators(&mut self) -> Result<Vec<Validator>, Self::Error> {
+        let blockchain = self.blockchain.read();
+        let staking_contract = blockchain.get_staking_contract();
 
         let mut active_validators = vec![];
 
-        for (address, balance) in staking_contract.active_validators {
-            active_validators.push(Staker {
-                address,
-                balance,
-                delegation: None,
-            });
+        for (address, _) in staking_contract.active_validators {
+            if let Ok(v) = get_validator_by_address(&blockchain, &address, None) {
+                active_validators.push(v.value);
+            }
         }
 
         Ok(active_validators)
--- a/rpc-server/src/dispatchers/blockchain.rs
+++ b/rpc-server/src/dispatchers/blockchain.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, ops::Deref, sync::Arc};
+use std::{ops::Deref, sync::Arc};
 
 use async_trait::async_trait;
 use futures::{future, stream::BoxStream, StreamExt};
@@ -8,7 +8,7 @@ use nimiq_account::{BlockLog, StakingContract, TransactionLog};
 use nimiq_blockchain::{AbstractBlockchain, Blockchain, BlockchainEvent};
 use nimiq_hash::Blake2bHash;
 use nimiq_keys::Address;
-use nimiq_primitives::{coin::Coin, policy};
+use nimiq_primitives::policy;
 use nimiq_rpc_interface::types::{
     is_of_log_type_and_related_to_addresses, BlockchainState, ParkedSet, Validator,
 };
@@ -460,13 +460,17 @@ impl BlockchainInterface for BlockchainDispatcher {
     }
 
     /// Returns a map of the currently active validator's addresses and balances.
-    async fn get_active_validators(&mut self) -> Result<HashMap<Address, Coin>, Error> {
+    async fn get_active_validators(&mut self) -> Result<Vec<Staker>, Error> {
         let staking_contract = self.blockchain.read().get_staking_contract();
 
-        let mut active_validators = HashMap::new();
+        let mut active_validators = vec![];
 
         for (address, balance) in staking_contract.active_validators {
-            active_validators.insert(address, balance);
+            active_validators.push(Staker {
+                address,
+                balance,
+                delegation: None,
+            });
         }
 
         Ok(active_validators)
--- a/rpc-server/src/dispatchers/blockchain.rs
+++ b/rpc-server/src/dispatchers/blockchain.rs
@@ -49,7 +49,7 @@ fn get_block_by_hash(
         .ok_or_else(|| Error::BlockNotFound(hash.clone().into()))
 }
 
-/// Tries to fetch a validator information given its address. It has an option to include a map
+/// Tries to fetch a validator information given its address. It has an option to include a collection
 /// containing the addresses and stakes of all the stakers that are delegating to the validator.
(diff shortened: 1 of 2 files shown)


redmaner:
What is the advantage of having an array of validators instead of a map? I actually like the map much more in this case, it is simple yet effective.

Eligioo:
@redmaner unlike an object where you can add more properties over time (less chance of breaking changes in the future when the methods are more used), a map is not extensible like this.

Eligioo:
@jsdanielh it's available for reviewing.

nibhar:
While I agree that a struct is more versatile, `Result<Vec<Staker>, Self::Error>` is not my favourite as the new return value. 

`Validator` does not seem inappropriate to me as the return type. It does yield more information which is maybe not strictly necessary as other RPC calls exist to retrieve those, but it also does not hurt. 
I have not dug in too deep, but what speaks against using i.e `Result<BlockchainState<Vec<Validator>>, Self::Error>` making it even more consistent with `get_validator_by_address`?

I would think that incorporating `BlockchainState` in it would provide value regardless, even if defining a new type would be the better approach.

Also something unrelated I noticed is that all of those `Error` really should be `Self::Error`.

Eligioo:
@nibhar 

>I have not dug in too deep, but what speaks against using i.e `Result<BlockchainState<Vec<Validator>>, Self::Error>` making it even more consistent with `get_validator_by_address`?

The generic parameter `T` in `BlockchainState<T>` is annotated with a `#[serde(flatten)]` which will cause a `panic` since you can only flatten structs and maps. `Vec<Validator>` is a sequence. [Source](https://github.com/nimiq/core-rs-albatross/blob/albatross/rpc-interface/src/types.rs#L728) I could do a `Result<Vec<Validator>, Self::Error>`, so not wrapping it into a `BlockchainState`.
 
> Also something unrelated I noticed is that all of those `Error` really should be `Self::Error`.

I can create a PR as follow-up aligning all the `Error`s with `Self::Error` if you want.

rpc-server/src/dispatchers/blockchain.rs
```diff
@@ -460,13 +460,17 @@ impl BlockchainInterface for BlockchainDispatcher {
     }
 
     /// Returns a map of the currently active validator's addresses and balances.
-    async fn get_active_validators(&mut self) -> Result<HashMap<Address, Coin>, Error> {
+    async fn get_active_validators(&mut self) -> Result<Vec<Staker>, Error> {
```
jsdanielh:
```suggestion
    async fn get_active_validators(&mut self) -> Result<Vec<Validator>, Error> {
```

jsdanielh (commented):
I agree with @nibhar, the resulting type to me should be `Validator`

nibhar:
@Eligioo 
> @nibhar
> 
> > I have not dug in too deep, but what speaks against using i.e `Result<BlockchainState<Vec<Validator>>, Self::Error>` making it even more consistent with `get_validator_by_address`?
> 
> The generic parameter `T` in `BlockchainState<T>` is annotated with a `#[serde(flatten)]` which will cause a `panic` since you can only flatten structs and maps. `Vec<Validator>` is a sequence. [Source](https://github.com/nimiq/core-rs-albatross/blob/albatross/rpc-interface/src/types.rs#L728) I could do a `Result<Vec<Validator>, Self::Error>`, so not wrapping it into a `BlockchainState`.

I see. Thank you for pointing it out! 
I do not know exactly why it is flattened and maybe it does not have to be. Maybe @sisou can join in on this discussion here, as he introduced the `#[serde(flatten)]`. 

I would prefer `Result<Vec<Validator>, Self::Error>` over `Result<Vec<Staker>, Self::Error>` in case it must be flattened. 
However I would also prefer it with the `BlockchainState` as that appears to me exactly what it is made for, but I could misunderstand its purpose.

> > Also something unrelated I noticed is that all of those `Error` really should be `Self::Error`.
> 
> I can create a PR as follow-up aligning all the `Error`s with `Self::Error` if you want.

If you like, contributions are always much appreciated!

sisou:
`BlockchainState` wraps it's target. But I don't want it wrapped when returned from the API. If I query a validator, why should the validator data be one level lower than the block height it was queried at. If anything, the block height is secondary information and should be wrapped.

So that's why the contents are flattened, to make it all one level of object.

Since vecs can't be flattened, it means creating a new type where the vec is nested under the block info. Or leaving the block info out, then the vec can be on the root level.

rpc-server/src/dispatchers/blockchain.rs
Eligioo: Solved in last commit

Eligioo:
I understand Sören in this case that's it's prettier to receive a
``` 
{
    block_number: ...,
    block_hash: ...,
    ...fields of T
}
```

rather than  a
```
{
    block_number: ...,
    block_hash: ...,
    T: {
        ...fields of T
    }
}
```
But with the current definition we can't turn a `Vec<Validator>` into a `BlockchainState<Vec<Validator>>`. I could be missing background info about this `BlockchainState` struct but why would it be relevant to also receive the `block_number` and `block_hash` in this case.


@jsdanielh I'm returning a `Result<Vec<Validator>, Self::Error>` now.

nibhar:
Because it provides the context that validator was retrieved in. That block height is not a part of the validator. Since not all returned values need that context, all those that do immediately indicate this accessing the data. Wrapping the context is certainly also an option but in case of collections leads to a bunch of duplicates. 
Personally I do not see an issue in the retrieved value being a level lower than the context it is in.

I would say that `BlockchainState<T>` should be used in a bunch more places, like `get_parked_validators`, `get_previous_slashed_slots` and the like ,as was discussed in #861. So I would not leave it out, but add it here as a start.

nibhar:
That being said, introducing all those `BlockchainState<T>` can be done in another PR after we discussed how to do it.

Adding i.e
```rust
pub struct Validators {
    pub validators: Vec<Validator>,
}
```
would work with 
```rust
async fn get_active_validators(&mut self) -> Result<BlockchainState<Validators>, Self::Error> { ... }
```

Eligioo:
It makes more sense now due to the context you've provided. Thanks.

I can easily rewrite and apply this definition underneath as a first step towards '`BlockchainState`'ing more methods. 
> ```rust
> async fn get_active_validators(&mut self) -> Result<BlockchainState<Validators>, Self::Error> { ... }
> ```

If you agree with that I'll use this PR for `getActiveValidators`, finish it up and use #970 to do the rest.

nibhar:
I would say leave it as is for this PR and let's continue the discussion about `BlockchainState<T>` in #970. Depending of the outcome we will then add the context consistently throughout the RPC calls.

ii-cruz (commented):
All seems good to me, changes look straightforward. Overall it makes the rpc interface more intuitive.  
Thank you @Eligioo

jsdanielh (approved):

jsdanielh:
Merged in 8ef1ade.

Issue #957, opened 2022-07-25
Blockchain push panic during devnet execution 


jsdanielh:
In the latest devnet execution a panic was observed in the `blockchain` `push` function:
```
Jul 25 09:22:33 validator1 nimiq-client[99778]: 2022-07-25T09:22:33.751971Z ERROR panic                | thread 'tokio-runtime-worker' panicked at 'Trying to rebranch across macro block': blockchain/src/blockchain/push.rs:405
Jul 25 09:22:33 validator1 nimiq-client[99778]:    0: <backtrace::capture::Backtrace as core::default::Default>::default
Jul 25 09:22:33 validator1 nimiq-client[99778]:    1: log_panics::Config::install_panic_hook::{{closure}}
Jul 25 09:22:33 validator1 nimiq-client[99778]:    2: std::panicking::rust_panic_with_hook
Jul 25 09:22:33 validator1 nimiq-client[99778]:              at /rustc/43347397f7c5ca9a670a3bb3890c7187e24a52ab/library/std/src/panicking.rs:702:17
Jul 25 09:22:33 validator1 nimiq-client[99778]:    3: std::panicking::begin_panic_handler::{{closure}}
Jul 25 09:22:33 validator1 nimiq-client[99778]:              at /rustc/43347397f7c5ca9a670a3bb3890c7187e24a52ab/library/std/src/panicking.rs:586:13
Jul 25 09:22:33 validator1 nimiq-client[99778]:    4: std::sys_common::backtrace::__rust_end_short_backtrace
Jul 25 09:22:33 validator1 nimiq-client[99778]:              at /rustc/43347397f7c5ca9a670a3bb3890c7187e24a52ab/library/std/src/sys_common/backtrace.rs:138:18
Jul 25 09:22:33 validator1 nimiq-client[99778]:    5: rust_begin_unwind
Jul 25 09:22:33 validator1 nimiq-client[99778]:              at /rustc/43347397f7c5ca9a670a3bb3890c7187e24a52ab/library/std/src/panicking.rs:584:5
Jul 25 09:22:33 validator1 nimiq-client[99778]:    6: core::panicking::panic_fmt
Jul 25 09:22:33 validator1 nimiq-client[99778]:              at /rustc/43347397f7c5ca9a670a3bb3890c7187e24a52ab/library/core/src/panicking.rs:142:14
[...]

PR #980, opened 2022-08-05
Fix history sync blockchain state mismatch issue

### LINKED ISSUE #979: History sync fails in devnet
History sync fails in devnet with the following error:
```
 DEBUG cluster              | Processing epoch #11290 (52 history items, chunk 2/2100)
 DEBUG cluster              | Committing epoch #11290
 WARN  history_sync         | Rejecting block block=#1445120.4:MA:6622ff11fa reason="header accounts hash doesn't match real accounts hash" state_root=f5e18d69b8366425c8870b3d13e0ae9444722c0829597dbe9b77877ea6524c8e wanted_state_root=f9370f0c3ac7b78a2a4a93d2c7d79c347756ac7c6fcf6c908a45b791ed4da191
```

jsdanielh:
Fix state mismatch after doing history sync.
The error was caused by history chunks that had transactions of
the same block number in different chunks.
Account's `commit_batch` requires all transactions of the same block
to be passed such that transactions can be applied in the expected
order: pre-inherents, senders, recipients and then post-inherents.

The solution implemented caches the last block transactions when
`add_history_chunk` is called and then merge them with the following
chunk if it happens to carry transactions of the same block. If
`commit` is called, it also checks for outstanding transactions in
the cache and applies them before doing the respective checks and
blockchain state updates.

This fixes #979.

## Pull request checklist

### DIFF
--- a/blockchain/src/blockchain/history_sync.rs
+++ b/blockchain/src/blockchain/history_sync.rs
@@ -7,6 +7,7 @@ use nimiq_database::WriteTransaction;
 use nimiq_hash::{Blake2bHash, Hash};
 use nimiq_primitives::coin::Coin;
 use nimiq_primitives::policy;
+use nimiq_transaction::Transaction;
 
 use crate::chain_info::ChainInfo;
 use crate::history::{ExtTxData, ExtendedTransaction};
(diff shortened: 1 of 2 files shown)


nibhar (approved):
Ultimately when we refactor parts of the Sync to mitigate the inconsistent state of #957 we might want to just block align chunks, making this unnecessary again. That really depends on the implementation details though and this PR should make the sync work properly again for the time being.

PR #1023, opened 2022-09-09
Introducing the RPC result with metadata

### LINKED ISSUE #970: Include `BlockchainState` in RPC responses where applicable
As was discussed in #861 RPC requests whose returned values depend on the head state of the blockchain should include the head block number and head hash the value was queried at. This provides context to the validity of those values as they change with the head changing.

Currently `BlockchainState<T>` fulfils this purpose for a select number of requests. 

Other RPC calls whose return values depend on the head state do not include these values yet.

Further discussion about the `BlockchainState<T>` can also be found in #963. The discussion is out of scope for that PR, thus this Issue is supposed to progress that conversation and also to track which RPC calls to modify.

ii-cruz:
This pull request adds a new RPC return type to all RPC methods with the data and metadata fields. As of now, All methods except the ones mentioned below return empty metadata.

The metadata includes the block hash and block number for the following RPC functions:
- get_slot_at
- get_account_by_address
- get_active_validators
- get_current_slashed_slots
- get_previous_slashed_slots
- get_parked_validators
- get_validator_by_address
- get_staker_by_address
- subscribe_for_validator_election_by_address
- subscribe_for_logs_by_addresses_and_types

This pull request removes the deprecated functions:
- create_update_transaction
- send_update_transaction

## Pull request checklist

### DIFF
--- a/rpc-client/src/subcommands/accounts_subcommands.rs
+++ b/rpc-client/src/subcommands/accounts_subcommands.rs
@@ -89,7 +89,7 @@ pub enum AccountCommand {
         #[clap(short, long)]
         public_key: PublicKey,
 
-        /// The signature returned upon signing the message. The r and s bystes should be all concatenated
+        /// The signature returned upon signing the message. The r and s bytes should be all concatenated
         /// into one continous input.
         #[clap(short, long)]
         signature: Signature,
@@ -111,7 +111,7 @@ impl HandleSubcommand for AccountCommand {
     async fn handle_subcommand(self, mut client: Client) -> Result<(), Error> {
         match self {
             AccountCommand::List { short } => {
-                let accounts = client.wallet.list_accounts().await?;
+                let accounts = client.wallet.list_accounts().await?.data;
                 for address in &accounts {
                     if short {
                         println!("{}", address.to_user_friendly_address());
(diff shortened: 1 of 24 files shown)


rpc-client/src/subcommands/accounts_subcommands.rs
```diff
@@ -111,10 +111,10 @@ impl HandleSubcommand for AccountCommand {
     async fn handle_subcommand(self, mut client: Client) -> Result<(), Error> {
         match self {
             AccountCommand::List { short } => {
-                let accounts = client.wallet.list_accounts().await?;
+                let accounts = client.wallet.list_accounts().await?.data;
                 for address in &accounts {
                     if short {
-                        println!("{}", address.to_user_friendly_address());
+                        println!("{:?}", address.to_user_friendly_address());
```
paberr: What's the reason for this change? `{}` should be perfectly fine for a string.

paberr: Same question applies to most of the other `{:?}`.

rpc-interface/Cargo.toml
```diff
@@ -24,6 +24,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_with = "2.0"
 thiserror = "1.0"
 clap = { version = "3.2", features = ["derive"] }
+parking_lot = { git = "https://github.com/styppo/parking_lot.git" }
```
paberr, line 27: Why do we need this dependency in the interface?

rpc-interface/src/consensus.rs
```diff
@@ -1,11 +1,12 @@
 use async_trait::async_trait;
 
+use crate::types::{RPCResult, Transaction, ValidityStartHeight};
 use nimiq_hash::Blake2bHash;
 use nimiq_keys::Address;
 use nimiq_primitives::coin::Coin;
 use nimiq_transaction::account::htlc_contract::{AnyHash, HashAlgorithm};
 
-use crate::types::{Transaction, ValidityStartHeight};
+type Result<T, E> = core::result::Result<RPCResult<T, ()>, E>;
```
paberr: Since you are defining this type in multiple files, it might make sense to move it to another file, make it `pub` and import it instead.
We might also think about giving it a different name but I'm terrible at names. 😅

paberr: (probably into the `types.rs` file)

paberr (commented):
Only some minor comments.

rpc-interface/Cargo.toml
ii-cruz: yep removed

rpc-interface/src/consensus.rs
ii-cruz: done

rpc-client/src/subcommands/accounts_subcommands.rs
```diff
@@ -120,32 +120,37 @@ impl HandleSubcommand for AccountCommand {
                             .blockchain
                             .get_account_by_address(address.clone())
                             .await?;
-                        println!("{}: {:#?}", address.to_user_friendly_address(), account);
```
paberr: I think you should keep the `#` option in the format strings.

rpc-client/src/subcommands/accounts_subcommands.rs
ii-cruz: The new return type RPCResult does not implement display so I replaced all {} to {:?}. This specific one is not an RPCResult and was needlessly replaced so I've put it back to {}.

paberr (approved):

paberr (approved):

PR #1027, opened 2022-09-13
Do not cancel running aggregations. Restart aggregations on startup.

nibhar:
A validator once restarted during macro block production will remember its past votes, but it will not start those aggregations to let other validators know about them. This PR fixes that by introducing an init() function to tendermint, which will restart all aggregations prior to the one in the current state if existent. The one in the current state will, (and has been) started automatically.

Furthermore pre commit aggregations were canceled up until this PR once they reach a clear result. That can in rare cases lead to a stalling situation. There is a threshold once reached where these aggregations could be terminated, but for the sake of a timely fix, in this PR aggregations are no longer terminated at all.

### DIFF
--- a/tendermint/src/outside_deps.rs
+++ b/tendermint/src/outside_deps.rs
@@ -87,6 +87,15 @@ pub trait TendermintOutsideDeps: Send + Unpin + Sized + 'static {
         proposal_hash: Option<Self::ProposalHashTy>,
     ) -> Result<(Self, AggregationResult<Self::ProposalHashTy, Self::ProofTy>), TendermintError>;
 
+    /// Rebroadcasts a vote and starts an aggregations, but does not wait for the result.
+    /// Used for restarting aggregations from state.
+    fn rebroadcast_and_aggregate(
+        &self,
+        round: u32,
+        step: Step,
+        proposal_hash: Option<Self::ProposalHashTy>,
+    );
+
     /// Returns the current aggregation for a given round and step. The returned aggregation might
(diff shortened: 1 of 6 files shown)


tendermint/src/outside_deps.rs
```diff
@@ -87,6 +87,15 @@ pub trait TendermintOutsideDeps: Send + Unpin + Sized + 'static {
         proposal_hash: Option<Self::ProposalHashTy>,
     ) -> Result<(Self, AggregationResult<Self::ProposalHashTy, Self::ProofTy>), TendermintError>;
 
+    /// Rebroadcasts avote and starts an aggregations, but does not wait forthe result.
```
jsdanielh:
```suggestion
    /// Rebroadcasts a vote and starts an aggregations, but does not wait for the result.
```

tendermint/src/tendermint.rs
```diff
@@ -67,6 +67,25 @@ impl<DepsTy: TendermintOutsideDeps> Tendermint<DepsTy> {
...
             }
+        };
+
+        this.init();
+
+        this
+    }
+
+    fn init(&mut self) {
+        // Restart all aggregations prior to our current states aggregation
```
jsdanielh:
```suggestion
        // Restart all aggregations prior to our current state aggregation
```

jsdanielh (commented):
Just some minor typos

jsdanielh (approved):

Issue #1036, opened 2022-09-17
When an epoch is adopted via history sync, the mempool is not properly updated


viquezclaudio:
During a public devnet execution, the following logs were observed:

```
Sep 15 00:22:03 validator4 nimiq-client[890919]: 2022-09-15T00:22:03.802936Z DEBUG verify               | Mempool-verify tx invalid at this block height block_height=6694273 validity_start_height=6694315
Sep 15 00:22:03 validator4 nimiq-client[890919]: 2022-09-15T00:22:03.805475Z DEBUG verify               | Mempool-verify tx invalid at this block height block_height=6694273 validity_start_height=6694315
Sep 15 00:22:03 validator4 nimiq-client[890919]: 2022-09-15T00:22:03.808163Z DEBUG verify               | Mempool-verify tx invalid at this block height block_height=6694273 validity_start_height=6694315
Sep 15 00:22:03 validator4 nimiq-client[890919]: 2022-09-15T00:22:03.809769Z DEBUG verify               | Mempool-verify tx invalid at this block height block_height=6694273 validity_start_height=6694315
Sep 15 00:22:03 validator4 nimiq-client[890919]: 2022-09-15T00:22:03.811740Z DEBUG verify               | Mempool-verify tx invalid at this block height block_height=6694273 validity_start_height=6694315
Sep 15 00:22:03 validator4 nimiq-client[890919]: 2022-09-15T00:22:03.814369Z DEBUG verify               | Mempool-verify tx invalid at this block height block_height=6694273 validity_start_height=6694315
Sep 15 00:22:03 validator4 nimiq-client[890919]: 2022-09-15T00:22:03.815508Z DEBUG verify               | Mempool-verify tx invalid at this block height block_height=6694273 validity_start_height=6694315
Sep 15 00:22:03 validator4 nimiq-client[890919]: 2022-09-15T00:22:03.818267Z DEBUG verify               | Mempool-verify tx invalid at this block height block_height=6694273 validity_start_height=6694315
Sep 15 00:22:03 validator4 nimiq-client[890919]: 2022-09-15T00:22:03.820622Z DEBUG verify               | Mempool-verify tx invalid at this block height block_height=6694273 validity_start_height=6694315
Sep 15 00:22:03 validator4 nimiq-client[890919]: 2022-09-15T00:22:03.822959Z DEBUG verify               | Mempool-verify tx invalid at this block height block_height=6694273 validity_start_height=6694315
[...]

PR #1043, opened 2022-09-20
Create a new HistoryAdoptedEvent

viquezclaudio:
Created a new event that is used to notify the mempool when a new history was adopted, in order to prune transactions from the mempool that are already included in the blockchain. Fixes issue #1036

### DIFF
--- a/mempool/src/mempool.rs
+++ b/mempool/src/mempool.rs
@@ -289,6 +289,23 @@ impl Mempool {
             .abort();
     }
 
+    /// This function prunes transactions from the mempool that were already included in the blockchain
+    /// This is generally used by history sync when it adopts epochs
+    pub fn mempool_clean_up(&self) {
+        // Acquire the mempool and blockchain locks
+        let blockchain = self.blockchain.read();
+        let mut mempool_state = self.state.write();
+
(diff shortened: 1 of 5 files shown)


mempool/src/mempool.rs
```diff
@@ -289,6 +289,23 @@ impl Mempool {
             .abort();
     }
 
+    /// This function prunes transactions from the mempool that were already included in the blockchain
+    /// This is generally used by history sync when it adopts epochs
+    pub fn mempool_clean_up(&self, _: &Blake2bHash) {
```
jsdanielh: The argument is unused. This opens up the question on why do we need it to be sent with the event?

viquezclaudio: Is used by the _get_newest_hash_ function also to be consistent with the other events in case it is needed in the future

jsdanielh: I'd propose then to remove the unused argument in the mempool and then change
`on_blockchain_history_adopted(&mut self, hash: &Blake2bHash)` for `on_blockchain_history_adopted(&mut self, _: &Blake2bHash)`

jsdanielh (approved):

PR #1103, opened 2022-11-02
Changed the DB sync mode to Durable

viquezclaudio:
Changed the MDBX sync mode to Durable, which provides the following properties:
Atomicity, Consistency, Isolation and Durability.
This comes with a performance impact, that according to unittesting (which pushes full blocks),
it is ~10% slower than UtterlyNoSync debug mode
But it is 2x slower than UtterlyNoSync and similar to NoMetaSync in release mode.
A tradeoff needs to be made between database consistency and speed, which can be offset by the
desired block times during production

### DIFF
--- a/database/src/mdbx.rs
+++ b/database/src/mdbx.rs
@@ -57,7 +57,7 @@ impl MdbxEnvironment {
         let db_flags = libmdbx::EnvironmentFlags {
             no_rdahead: true,
             mode: libmdbx::Mode::ReadWrite {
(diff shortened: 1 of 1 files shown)


hrxi (approved):

jsdanielh:
Just a small observation in the commit message I think it should say: `MetaNoSync` instead of `MetaSync`

jsdanielh (approved):

jsdanielh (approved):

PR #1105, opened 2022-11-04
Basic infrastructure for FullNodes

viquezclaudio:
- Added new sync mode options to the config file: When in 'full' mode, the complete transaction history is not stored
- Removed unused protocols flags from peer contacts
- Introduced the ability to exchange service flags and based on the flags, store the appropriate peer contacts


blockchain/src/blockchain/blockchain.rs
```diff
@@ -8,6 +8,7 @@ use nimiq_genesis::NetworkInfo;
 use nimiq_hash::Blake2bHash;
 use nimiq_primitives::coin::Coin;
 use nimiq_primitives::networks::NetworkId;
+
```
jsdanielh: This extra line looks unnecessary.

blockchain/src/blockchain/blockchain.rs
```diff
@@ -29,6 +30,8 @@ const BROADCAST_MAX_CAPACITY: usize = 256;
 pub struct Blockchain {
     // The environment of the blockchain.
     env: Environment,
+    // Blockchain configuration options
```
jsdanielh: Mmm, actually we right way is to use `///` instead. I think the rest of the member documentation are wrong

blockchain/src/blockchain/push.rs
```diff
@@ -307,10 +307,21 @@ impl Blockchain {
...
+            if this.config.keep_history {
+                // If we are a full history node, we only prune the chain store
+                this.chain_store.prune_epoch(pruned_epoch, &mut txn);
+            } else {
+                // Prune the History Store.
+                this.history_store.remove_history(&mut txn, pruned_epoch);
+
+                // Prune the Chain Store.
+                this.chain_store.prune_epoch(pruned_epoch, &mut txn);
+            }
```
jsdanielh, line 326: The only comment here is that maybe we would want to store history only when `keep_history` is set. And thus we wouldn't need to prune anything for that case here.

blockchain/src/blockchain/push.rs
```diff
@@ -307,10 +307,21 @@ impl Blockchain {
         this.chain_store.set_head(&mut txn, &block_hash);
 
         if is_election_block {
-            this.chain_store.prune_epoch(
-                Policy::epoch_at(block_number).saturating_sub(MAX_EPOCHS_STORED),
-                &mut txn,
-            );
+            let epochs_stored = cmp::max(this.config.max_epochs_stored, Policy::MIN_EPOCHS_STORED);
```
jsdanielh:
```suggestion
            let max_epochs_stored = cmp::max(this.config.max_epochs_stored, Policy::MIN_EPOCHS_STORED);
```

blockchain/src/blockchain/push.rs
```diff
@@ -307,10 +307,21 @@ impl Blockchain {
...
+
+            // Calculate the epoch to be pruned. Saturate at zero.
+            let pruned_epoch = Policy::epoch_at(block_number).saturating_sub(epochs_stored);
+
+            if this.config.keep_history {
+                // If we are a full history node, we only prune the chain store
```
jsdanielh:
```suggestion
                // If we are a full node, we only prune the chain store
```

blockchain/src/blockchain/push.rs
```diff
@@ -307,10 +307,21 @@ impl Blockchain {
...
-            this.chain_store.prune_epoch(
-                Policy::epoch_at(block_number).saturating_sub(MAX_EPOCHS_STORED),
-                &mut txn,
-            );
+            let epochs_stored = cmp::max(this.config.max_epochs_stored, Policy::MIN_EPOCHS_STORED);
+
+            // Calculate the epoch to be pruned. Saturate at zero.
+            let pruned_epoch = Policy::epoch_at(block_number).saturating_sub(epochs_stored);
+
+            if this.config.keep_history {
```
jsdanielh, line 316:
```suggestion
            if !this.config.keep_history {
```
This logic seems backwards to me because I'd expect history nodes to set `keep_history`

genesis/src/networks.rs
```diff
@@ -11,10 +10,7 @@ use nimiq_account::AccountsList;
 use nimiq_database::volatile::VolatileEnvironment;
 use nimiq_genesis_builder::{GenesisBuilder, GenesisBuilderError, GenesisInfo};
 use nimiq_hash::Blake2bHash;
-use nimiq_keys::PublicKey;
-use nimiq_peer_address::address::seed_list::SeedList;
-use nimiq_peer_address::address::{NetAddress, PeerAddress, PeerAddressType, PeerId};
-use nimiq_peer_address::services::ServiceFlags;
+
```
jsdanielh: Extra line seems unnecessary.

lib/src/client.rs
```diff
@@ -82,6 +83,38 @@ pub(crate) struct ClientInner {
     zkp_component: ZKPComponentProxy,
 }
 
+/// This function is used to generate the serices flags (provided, needed) based upon the configured sync mode
```
jsdanielh:
```suggestion
/// This function is used to generate the services flags (provided, needed) based upon the configured sync mode
```

lib/src/client.rs
```diff
@@ -82,6 +83,38 @@ pub(crate) struct ClientInner {
...
+                | Services::ACCOUNTS_CHUNKS
+                | Services::TRANSACTION_INDEX
+        }
+        // Services provided by full nodes
+        crate::config::config::SyncMode::Full => {
+            log::info!("Client configured as a full node");
+            Services::ACCOUNTS_PROOF | Services::FULL_BLOCKS | Services::ACCOUNTS_CHUNKS
+        }
+        // Services provided by nano nodes
+        crate::config::config::SyncMode::Nano => Services::empty(),
```
jsdanielh: Seems not consistent to not log anything for `Nano`

lib/src/client.rs
```diff
@@ -82,6 +83,38 @@ pub(crate) struct ClientInner {
...
+    };
+
+    let required_services = match sync_mode {
+        // Services provided by history nodes
```
jsdanielh:
```suggestion
        // Services required by history nodes
```

lib/src/client.rs
```diff
@@ -82,6 +83,38 @@ pub(crate) struct ClientInner {
...
+        crate::config::config::SyncMode::History => Services::HISTORY | Services::FULL_BLOCKS,
+        // Services provided by full nodes
```
jsdanielh:
```suggestion
        // Services required by full nodes
```

lib/src/client.rs
```diff
@@ -82,6 +83,38 @@ pub(crate) struct ClientInner {
...
+        crate::config::config::SyncMode::Full => Services::FULL_BLOCKS | Services::ACCOUNTS_CHUNKS,
+        // Services provided by nano nodes
```
jsdanielh:
```suggestion
        // Services required by light nodes
```

lib/src/client.rs
```diff
@@ -82,6 +83,38 @@ pub(crate) struct ClientInner {
...
+                | Services::ACCOUNTS_PROOF
+                | Services::ACCOUNTS_CHUNKS
+                | Services::TRANSACTION_INDEX
+        }
+        // Services provided by full nodes
+        crate::config::config::SyncMode::Full => {
+            log::info!("Client configured as a full node");
+            Services::ACCOUNTS_PROOF | Services::FULL_BLOCKS | Services::ACCOUNTS_CHUNKS
+        }
+        // Services provided by nano nodes
```
jsdanielh:
```suggestion
        // Services provided by light nodes
```

lib/src/client.rs
jsdanielh:
```suggestion
        crate::config::config::SyncMode::Light => Services::empty(),
```

lib/src/client.rs
```diff
@@ -82,6 +83,38 @@ pub(crate) struct ClientInner {
...
+        crate::config::config::SyncMode::Nano => Services::empty(),
+    };
+
+    let required_services = match sync_mode {
+        // Services provided by history nodes
+        crate::config::config::SyncMode::History => Services::HISTORY | Services::FULL_BLOCKS,
+        // Services provided by full nodes
+        crate::config::config::SyncMode::Full => Services::FULL_BLOCKS | Services::ACCOUNTS_CHUNKS,
+        // Services provided by nano nodes
+        crate::config::config::SyncMode::Nano => Services::CHAIN_PROOF | Services::ACCOUNTS_PROOF,
```
jsdanielh:
```suggestion
        crate::config::config::SyncMode::Light => Services::CHAIN_PROOF | Services::ACCOUNTS_PROOF,
```

lib/src/config/config_file/client.example.toml
```diff
@@ -59,6 +59,7 @@ seed_nodes = [
 # Possible values: "main", "test", "dev", "test-albatross", "dev-albatross"
 # Default: "dev-albatross"
 #network = "main"
+sync_mode = "full"
```
jsdanielh, line 64: Which should we set as default in the example? `full` or `history`? Also we should provide some comments as in the `network` case.

lib/src/config/config.rs
```diff
@@ -51,6 +51,8 @@ use crate::{
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash, Display)]
 pub enum SyncMode {
     History,
+    Full,
+    Nano,
```
jsdanielh:
```suggestion
    Light,
```

lib/src/config/config.rs
```diff
@@ -908,13 +855,17 @@ pub struct ZKPConfig {
 
     /// ZKP prover path for verifying and proving keys.
     pub setup_keys_path: PathBuf,
+
+    /// ZKP propagation
```
jsdanielh: I'd improve this comment a little, like: `Flag for enabling ZKP propagation`

lib/src/config/config_file/mod.rs
```diff
@@ -190,6 +164,8 @@ pub struct ConsensusSettings {
 #[serde(rename_all = "lowercase")]
 pub enum SyncMode {
     History,
+    Full,
+    Nano,
```
```suggestion
    Light,
```

lib/src/config/config_file/mod.rs
```diff
@@ -207,6 +183,8 @@ impl FromStr for SyncMode {
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         Ok(match s.to_lowercase().as_str() {
             "history" => Self::History,
+            "full" => Self::Full,
+            "nano" => Self::Nano,
```
```suggestion
            "light" => Self::Light,
```

lib/src/config/config_file/mod.rs
```diff
@@ -216,6 +194,8 @@ impl From<SyncMode> for config::SyncMode {
     fn from(sync_mode: SyncMode) -> Self {
         match sync_mode {
             SyncMode::History => Self::History,
+            SyncMode::Full => Self::Full,
+            SyncMode::Nano => Self::Nano,
```
```suggestion
            SyncMode::Light => Self::Light,
```

lib/src/config/config.rs
```diff
@@ -804,17 +753,15 @@ impl ClientConfigBuilder {
...
 
         // Configure RPC server
         #[cfg(feature = "rpc-server")]
         {
             if let Some(rpc_config) = &config_file.rpc_server {
-                let bind_to = rpc_config
-                    .bind
-                    .as_ref()
-                    .and_then(|addr| addr.into_ip_address());
+                let bind_to = rpc_config.bind.as_ref().map(|addr| addr.parse().unwrap());
```
jsdanielh: I would change the `unwrap` for a log message and an appropriate error

lib/src/config/config.rs
```diff
@@ -858,7 +805,7 @@ impl ClientConfigBuilder {
                 let ip = metrics_config
                     .bind
                     .as_ref()
-                    .and_then(|addr| addr.into_ip_address());
+                    .map(|addr| addr.parse().unwrap());
```
jsdanielh: Same here: I would change the `unwrap` for a log message and an appropriate error

network-mock/src/network.rs
```diff
@@ -566,4 +566,8 @@ impl Network for MockNetwork {
             Err(MockNetworkError::CantRespond(request_id))
         }
     }
+
+    fn peer_provides_required_services(&self, _peer_id: Self::PeerId) -> bool {
+        true
```
jsdanielh, line 571: Shouldn't we implement this method to add some testing?

network-interface/src/network.rs
```diff
@@ -52,6 +52,7 @@ pub trait Network: Send + Sync + Unpin + 'static {
 
     fn get_peers(&self) -> Vec<Self::PeerId>;
     fn has_peer(&self, peer_id: Self::PeerId) -> bool;
+    fn peer_provides_required_services(&self, peer_id: Self::PeerId) -> bool;
```
jsdanielh, line 56: I'd suggest to add some documentation to the method (I know the rest should have it as well)

scripts/devnet/devnet.sh
```diff
@@ -10,6 +10,8 @@ validators=()
 vpids=()
 #Array with seed/spammer pids
 spids=()
+#Array with history_nodes pids
```
jsdanielh, line 13: What is the difference between `vpids` and this array? Maybe state that `vpids` are meant for `full` nodes while `hpids` for `history`?

scripts/devnet/devnet.sh
```diff
@@ -297,6 +303,14 @@ spids+=($!)
 sleep 3
 echo "Done."
 
+# Launch the history node
+echo "Starting history node..."
+mkdir -p temp-state/dev/history1
```
jsdanielh, line 308: Looks like we are going to support only one history node?

scripts/devnet/python/devnet_create.py
```diff
@@ -317,6 +320,56 @@ def create_validator(path, i, min_peers, enable_metrics):
     }
 
 
+def create_history_node(path, i, min_peers, enable_metrics):
```
jsdanielh, line 323: I'd vote against duplicating this function and add a flag for `create_validator` like:
```
def create_validator(path, i, min_peers, enable_metrics, sync_mode):
```

scripts/devnet/python/devnet_create.py
```diff
@@ -328,6 +381,11 @@ def create_validator(path, i, min_peers, enable_metrics):
     print("Created validator: {}..".format(
         validator["voting_key"]["public_key"][0:16]))
 
+# Create history node
+create_history_node(
+    output / "history{:d}".format(0+1), 0, min_peers, args.metrics)
```
jsdanielh:
```suggestion
    output / "history1", 0, min_peers, args.metrics)
```

network-libp2p/tests/network.rs
```diff
@@ -203,6 +204,8 @@ async fn create_network_with_n_peers(n_peers: usize) -> Vec<Network> {
 
         log::debug!(address = %addr, peer_id = %network.get_local_peer_id(), "Network {}", peer);
 
+        peer_ids.push(network.get_local_peer_id());
```
jsdanielh: Where is it used?

network-libp2p/tests/network.rs
```diff
@@ -189,6 +189,7 @@ async fn create_network_with_n_peers(n_peers: usize) -> Vec<Network> {
     let mut addresses = Vec::new();
     let mut events = Vec::new();
     let mut rng = rand::thread_rng();
+    let mut peer_ids = Vec::new();
```
jsdanielh: Where is it used?

network-libp2p/src/discovery/behaviour.rs
```diff
@@ -247,7 +251,6 @@ impl NetworkBehaviour for DiscoveryBehaviour {
                 debug!("Doing house-keeping in peer address book");
                 let mut peer_address_book = self.peer_contact_book.write();
                 peer_address_book.update_own_contact(&self.keypair);
-                peer_address_book.house_keeping();
```
jsdanielh: Why did we remove the house keeping from the `peer_address_book`?

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -61,117 +28,44 @@ bitflags! {
...
+        /// The node provides the full transaction history.
         ///
-        /// However, if [`Services::BLOCK_HISTORY`] is not set, this service is only provided for the latest
-        /// [`nimiq_primitives::policy::NUM_BLOCKS_VERIFICATION`] blocks.
-        ///
-        const BLOCK_PROOF = 1 << 2;
+        const HISTORY = 1 << 1;
 
-        /// The node provides a chain proof for the tip of the current main chain.
+        /// The ZKP that proves the latest election block
```
```suggestion
        /// The node provides the ZKP for the latest election block
```
(proof is duplicated since `ZKP` already includes proof)

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -433,14 +314,13 @@ impl PeerContactBook {
...
-        services_filter: Services,
-    ) {
+    pub fn insert_filtered(&mut self, contact: SignedPeerContact, services_filter: Services) {
         let info = PeerContactInfo::from(contact);
-        if info.matches(protocols_filter, services_filter) {
+        if info.matches(services_filter) {
+            log::debug!(
+                " Inserting peer_id: {} into my peer contacts, because is interesting to me, it provides: {:?}",
+                info.peer_id,info.services()
+            );
```
```suggestion
            log::debug!(
                added_peer =  info.peer_id,
                services = info.services(),
                "Inserting into my peer contacts, because is interesting to me",
            );
```
Use the `tracing` way of logging

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -455,27 +335,23 @@ impl PeerContactBook {
     pub fn insert_all_filtered<I: IntoIterator<Item = SignedPeerContact>>(
         &mut self,
         contacts: I,
-        protocols_filter: Protocols,
+
```
jsdanielh: Unnecessary extra line

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -523,7 +399,7 @@ impl PeerContactBook {
                 .peer_contacts
                 .iter()
                 .filter_map(|(peer_id, peer_contact)| {
-                    if peer_contact.exceeds_age(&self.config, unix_time) {
+                    if peer_contact.exceeds_age(Duration::from_secs(60 * 15), unix_time) {
```
jsdanielh: Why did we change this for something hardwired? In any case it should be a constant instead

network-libp2p/src/network.rs
```diff
@@ -177,15 +179,28 @@ impl PubsubId<PeerId> for GossipsubId<PeerId> {
     }
 }
 pub struct Network {
+    // The local ID that is used to identify our peer
```
jsdanielh:
```suggestion
    /// The local ID that is used to identify our peer
```

network-libp2p/src/network.rs
```diff
@@ -177,15 +179,28 @@ impl PubsubId<PeerId> for GossipsubId<PeerId> {
...
     local_peer_id: PeerId,
-    connected_peers: Arc<RwLock<HashSet<PeerId>>>,
+    // This hash map mantains an association between PeerIds and PeerContact:
+    // If the peer is interesting, i.e.: it provides services that are interested to us,
+    //  we store a some value with the peer contact itself
+    // Otherwise,
+    //  we just store a None value
+    //
```
jsdanielh:
```suggestion
    /// This hash map maintains an association between `PeerId`s and `PeerContact`s:
    /// If the peer is interesting, i.e. it provides services that are interesting to us,
    /// we store a some value with the peer contact itself
    /// Otherwise, we just store a None value
    ///
```

network-libp2p/src/network.rs
```diff
@@ -177,15 +179,28 @@ impl PubsubId<PeerId> for GossipsubId<PeerId> {
...
+    connected_peers: Arc<RwLock<HashMap<PeerId, Option<PeerContact>>>>,
+    // Stream used to send event messages
```
jsdanielh:
```suggestion
    /// Stream used to send event messages
```

network-libp2p/src/network.rs
```diff
@@ -177,15 +179,28 @@ impl PubsubId<PeerId> for GossipsubId<PeerId> {
...
     events_tx: broadcast::Sender<NetworkEvent<PeerId>>,
+    // Stream used to send action messages
```
jsdanielh:
```suggestion
    /// Stream used to send action messages
```

network-libp2p/src/network.rs
```diff
@@ -177,15 +179,28 @@ impl PubsubId<PeerId> for GossipsubId<PeerId> {
...
     action_tx: mpsc::Sender<NetworkAction>,
+    // Stream used to send validation messages
```
jsdanielh:
```suggestion
    /// Stream used to send validation messages
```

network-libp2p/src/network.rs
```diff
@@ -177,15 +179,28 @@ impl PubsubId<PeerId> for GossipsubId<PeerId> {
...
     validate_tx: mpsc::UnboundedSender<ValidateMessage<PeerId>>,
+    // Rate limiting capabilities
```
jsdanielh:
```suggestion
    /// Rate limiting capabilities
```

network-libp2p/src/network.rs
```diff
@@ -177,15 +179,28 @@ impl PubsubId<PeerId> for GossipsubId<PeerId> {
...
     peer_request_limits: Arc<Mutex<HashMap<PeerId, HashMap<u16, RateLimit>>>>,
-
+    // Metrics used for data analysis
     #[cfg(feature = "metrics")]
     metrics: Arc<NetworkMetrics>,
+    // Required services from other peeers. This is defined on init, based on our client type
```
jsdanielh:
```suggestion
    /// Required services from other peers. This is defined on init, based on our client type
```

network-libp2p/tests/discovery.rs
```diff
@@ -222,66 +218,12 @@ pub async fn test_dialing_peer_from_contacts() {
...
+        peer_id,
+        peer_contact: _,
+    })) = node1.swarm.next().await
     {
         log::info!(%peer_id, "Established PEX with peer");
         assert_eq!(peer2_id, peer_id);
     }
 }
-
-#[test]
```
jsdanielh, line 233: Why did we remove this test?

blockchain/src/blockchain/push.rs
viquezclaudio: This is actually the case for history nodes: those are the nodes that would have keep_history set to true

blockchain/src/blockchain/push.rs
viquezclaudio: Same comment as above:
history nodes would have keep_history = true, and thus:

history nodes should only prune the chain store
fill nodes should prune the chain store + history store

so it is right as it currently is

blockchain/src/blockchain/push.rs
jsdanielh: Ok, I see where the confusion is: We're deleting history for the full node for `pruned_epoch` which means that we will have history for a full node for `epochs_stored` and in that case we don't want any history.

blockchain/src/blockchain/push.rs
jsdanielh: Ok, I see where the confusion is: We're deleting history for the full node for `pruned_epoch` which means that we will have history for a full node for `epochs_stored` and in that case we don't want any history.

scripts/devnet/devnet.sh
viquezclaudio: Yes, validators are expected to be run as full nodes and not history nodes

scripts/devnet/devnet.sh
viquezclaudio: Yes, one seed, one history and multiple validators (full nodes)

scripts/devnet/python/devnet_create.py
viquezclaudio: although in theory you can run validators as history nodes, the use case would be to run validators in full nodes, so the validator configuration is not necessary for history nodes

blockchain/src/blockchain/push.rs
```diff
@@ -307,10 +307,22 @@ impl Blockchain {
...
+
+            // Calculate the epoch to be pruned. Saturate at zero.
+            let pruned_epoch = Policy::epoch_at(block_number).saturating_sub(max_epochs_stored);
+
+            if this.config.keep_history {
+                // If we are a history node, we only prune the chain store
+                this.chain_store.prune_epoch(pruned_epoch, &mut txn);
+            } else {
+                // Prune the History Store.
+                this.history_store.remove_history(&mut txn, pruned_epoch);
```
jsdanielh:
```suggestion
                this.history_store.remove_history(&mut txn, Policy::epoch_at(block_number).saturating_sub(1));
```
We don't want to keep history at all for full node

lib/src/config/config.rs
```diff
@@ -814,7 +764,7 @@ impl ClientConfigBuilder {
                 let bind_to = rpc_config
                     .bind
                     .as_ref()
-                    .and_then(|addr| addr.into_ip_address());
+                    .map(|addr| addr.parse().expect(" Error parsing the RPC server address"));
```
jsdanielh: I'd prefer to return a graceful error

lib/src/config/config.rs
```diff
@@ -855,10 +805,10 @@ impl ClientConfigBuilder {
         #[cfg(feature = "metrics-server")]
         {
             if let Some(metrics_config) = &config_file.metrics_server {
-                let ip = metrics_config
-                    .bind
-                    .as_ref()
-                    .and_then(|addr| addr.into_ip_address());
+                let ip = metrics_config.bind.as_ref().map(|addr| {
+                    addr.parse()
+                        .expect("Error parsing the IP address of the metrics server")
```
jsdanielh: I'd prefer to return a graceful error

lib/src/config/config_file/mod.rs
```diff
@@ -178,10 +151,11 @@ pub struct TlsSettings {
 #[derive(Clone, Debug, Deserialize, Default)]
 #[serde(deny_unknown_fields)]
 pub struct ConsensusSettings {
-    #[serde(rename = "type")]
     #[serde(default)]
     pub sync_mode: SyncMode,
     #[serde(default)]
+    pub max_epochs_stored: usize,
```
jsdanielh, line 160: I'd suggest to add rust doc for the members

lib/src/config/config_file/mod.rs
```diff
@@ -190,6 +164,8 @@ pub struct ConsensusSettings {
 #[serde(rename_all = "lowercase")]
 pub enum SyncMode {
```
jsdanielh, line 171: I'd suggest to add rust doc for the members and the `enum`

network-interface/src/network.rs
```diff
@@ -52,6 +52,8 @@ pub trait Network: Send + Sync + Unpin + 'static {
 
     fn get_peers(&self) -> Vec<Self::PeerId>;
     fn has_peer(&self, peer_id: Self::PeerId) -> bool;
+    // This method returns true when the given peer provides the services flags that are required by us
```
jsdanielh:
```suggestion
    /// This method returns true when the given peer provides the services flags that are required by us
```

blockchain/src/blockchain/blockchain.rs
```diff
@@ -27,52 +27,81 @@ const BROADCAST_MAX_CAPACITY: usize = 256;
...
+    /// The coin supply at the genesis block. This is needed to calculate the rewards.
     pub(crate) genesis_supply: Coin,
-    // The timestamp at the genesis block. This is needed to calculate the rewards.
+    /// The timestamp at the genesis block. This is needed to calculate the rewards.
     pub(crate) genesis_timestamp: u64,
 }
 
+/// Contains various blockchain configuration knobs
+pub struct BlockchainConfig {
+    // Flag indicating if the full history should be stored
```
jsdanielh:
```suggestion
    /// Flag indicating if the full history should be stored
```

blockchain/src/blockchain/blockchain.rs
```diff
@@ -27,52 +27,81 @@ const BROADCAST_MAX_CAPACITY: usize = 256;
...
+    pub keep_history: bool,
+    // Maximum number of epochs (other than the current one) that the ChainStore will store fully.
+    // Epochs older than this number will be pruned.
```
jsdanielh:
```suggestion
    /// Maximum number of epochs (other than the current one) that the ChainStore will store fully.
    /// Epochs older than this number will be pruned.
```

lib/src/config/config.rs
```diff
@@ -66,13 +68,16 @@ pub struct ConsensusConfig {
     pub sync_mode: SyncMode,
     #[builder(default = "3")]
     pub min_peers: usize,
+    #[builder(default = "1")]
+    pub max_epochs_stored: u32,
```
jsdanielh, line 79: I'd suggest to add rust doc for the members and the `struct`

lib/src/config/config.rs
```diff
@@ -51,6 +51,8 @@ use crate::{
 #[derive(Debug, Copy, Clone, Eq, PartialEq, Hash, Display)]
 pub enum SyncMode {
     History,
+    Full,
+    Light,
```
jsdanielh, line 58: I'd suggest to add rust doc for the members and the `enum`

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -16,41 +16,12 @@ use parking_lot::RwLock;
...
-            self.max_age_websocket
-        } else if protocols.contains(Protocols::RTC) {
-            self.max_age_webrtc
-        } else {
-            self.max_age_dumb
-        }
-    }
-}
+/// If a peer's age exceeds this value in seconds, it is removed
+pub const MAX_PEER_AGE: u64 = 900;
```
```suggestion
/// If a peer's age exceeds this value in seconds, it is removed (30 min)
pub const MAX_PEER_AGE: u64 = 30 * 60;
```

jsdanielh: Also can we make it a constant of `PeerContactBook`?

network-libp2p/src/network.rs
```diff
@@ -177,15 +179,28 @@ impl PubsubId<PeerId> for GossipsubId<PeerId> {
...
 pub struct Network {
+    /// The local ID that is used to identify our peer
     local_peer_id: PeerId,
-    connected_peers: Arc<RwLock<HashSet<PeerId>>>,
+    /// This hash map mantains an association between PeerIds and PeerContact:
+    /// If the peer is interesting, i.e.: it provides services that are interested to us,
+    ///  we store a some value with the peer contact itself
+    /// Otherwise,
+    ///  we just store a None value
+    ///
```
jsdanielh:
```suggestion
    /// we store a some value with the peer contact itself
    /// Otherwise,
    /// we just store a None value
```
(extra spaces and lines)

network-libp2p/tests/network.rs
```diff
@@ -222,10 +222,13 @@ async fn create_network_with_n_peers(n_peers: usize) -> Vec<Network> {
         .take(n_peers * (n_peers - 1))
         .for_each(|event| async move {
             match event {
-                Ok(NetworkEvent::PeerJoined(_)) => {}
-                _ => panic!("Unexpected NetworkEvent: {:?}", event),
+                Ok(NetworkEvent::PeerJoined(peer_id)) => {
+                    log::info!(" Received peer joined event {}", peer_id);
```
jsdanielh:
```suggestion
                    log::info!(%peer_id, " Received peer joined event");
```

network-libp2p/tests/network.rs
```diff
@@ -222,10 +222,13 @@ async fn create_network_with_n_peers(n_peers: usize) -> Vec<Network> {
...
+                }
+                _ => log::error!("Unexpected NetworkEvent: {:?}", event),
```
jsdanielh:
```suggestion
                _ => log::error!(?event, "Unexpected NetworkEvent"),
```

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -61,117 +32,52 @@ bitflags! {
...
-            Protocol::Ws(_) => Self::WS,
-            Protocol::Wss(_) => Self::WSS,
-            #[cfg(test)]
-            Protocol::Memory(_) => Self::MEM,
-            _ => Self::empty(),
-        }
+impl fmt::Display for Services {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "BitSet(32)")?;
+        fmt::Display::fmt(self, f)?;
```
jsdanielh: Oh yeah, and beware that this is an infinite recursive call

lib/src/config/config_file/mod.rs
```diff
@@ -178,18 +151,29 @@ pub struct TlsSettings {
 #[derive(Clone, Debug, Deserialize, Default)]
 #[serde(deny_unknown_fields)]
 pub struct ConsensusSettings {
-    #[serde(rename = "type")]
     #[serde(default)]
+    /// Posible synchronization modes to reach consensus according to the client type
```
```suggestion
    /// Possible synchronization modes to reach consensus according to the client type
```

lib/src/config/config_file/mod.rs
```diff
@@ -178,18 +151,29 @@ pub struct TlsSettings {
 #[derive(Clone, Debug, Deserialize, Default)]
 #[serde(deny_unknown_fields)]
 pub struct ConsensusSettings {
```
jsdanielh, line 154: Could you add documentation for the `struct`

network-libp2p/tests/network.rs
```diff
@@ -222,10 +222,13 @@ async fn create_network_with_n_peers(n_peers: usize) -> Vec<Network> {
         .take(n_peers * (n_peers - 1))
         .for_each(|event| async move {
             match event {
-                Ok(NetworkEvent::PeerJoined(_)) => {}
-                _ => panic!("Unexpected NetworkEvent: {:?}", event),
+                Ok(NetworkEvent::PeerJoined(peer_id)) => {
+                    log::info!(%peer_id, " Received peer joined event");
```
jsdanielh:
```suggestion
                    log::info!(%peer_id, "Received peer joined event");
```

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -377,18 +261,18 @@ impl PeerContactInfo {
         self.contact.inner.timestamp.is_none()
     }
 
-    /// Returns whether the peer contact exceeds its age limit (specified in `config`).
```
jsdanielh, line 380: Can we keep the documentation and rewrite it appropiately?

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -402,17 +286,17 @@ impl PeerContactInfo {
...
-
     own_peer_contact: PeerContactInfo,
 
     peer_contacts: HashMap<PeerId, Arc<PeerContactInfo>>,
 }
 
 impl PeerContactBook {
-    pub fn new(config: PeerContactBookConfig, own_peer_contact: SignedPeerContact) -> Self {
+    /// If a peer's age exceeds this value in seconds, it is removed (30 minutes)
+    pub const MAX_PEER_AGE: u64 = 1800;
```
```suggestion
    pub const MAX_PEER_AGE: u64 = 30 * 60;
```
I'd prefer this to see easily this is 30 min in fact

jsdanielh (approved):

PR #1112, opened 2022-11-08
Add a block separation time

jsdanielh:
- Add a block separation time and change the micro block producer within the `validator` to check when a block needs to be either:
  - Immediately produced because the expected block time for a block has already passed.
  - Delayed in order to comply with the block separation time.
  The block separation time will be taken as a point for the reward function later to incentivize validators to produce blocks within this time.
- Change the maximum timestamp drift to 1 minute
    - Change the maximum timestamp drift from 10 minutes to 1 minute. This is to only accept blocks whose timestamp differs less than the expected batch time which is targeting 1s * `BLOCKS_PER_BATCH`.
      The latter will have a value of 60 after #940 gets merged.
    - Change some tests utilities to set a less aggressive timestamp offset when producing blocks.
    - Change the `README` file to add a recommendation to run a clock synchronization protocol such as `NTP`.
- Introduced a penalty for not producing batches in time
  - Each batch should be produced 'in time'. That means there is a target block time, and if validators delay the block production process, then their batch rewards would be affected proportionally to the delay.
  - Re-generate ZKP keys and proofs to make all zkp tests to pass. 
- Use a dynamic block producer timeout depending on whether the block production within the batch is on time or is delayed.
- Change skip block timestamp generation and verification such that the expected timestamp is the producer timeout plus the greater between:
  - The original expected timestamp of the block being skipped.
  - The previous block timestamp.
- Decrease the block producer timeout from 4s to 2s. This is to adjust the timeout to the expected time of producing 2 consecutive blocks.
- Create a test for delayed batch rewards.

## Pull request checklist


validator/src/micro.rs
```diff
@@ -95,26 +88,36 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
                 if !in_current_state(&blockchain.head()) {
                     break Some(None);
                 } else if self.is_our_turn(&blockchain) {
-                    if !logged {
+                    // We want to produce a block at the expected timestamp for this block in this epoch
+                    // as it is calculated by the reward function
+                    let last_election_block = blockchain.election_head();
+                    let target_block_ts = last_election_block.header.timestamp
+                        + self.block_separation_time.as_secs()
+                            * 1000
```
sisou: You can do this to avoid the `*1000` line.
```suggestion
                        + self.block_separation_time.as_millis() as u64
```

validator/src/micro.rs
```diff
@@ -141,16 +144,17 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
                             .map(move |result| ProduceMicroBlockEvent::MicroBlock(block1, result))
                             .ok();
                         break Some(event);
-                    }
+                    } else {
+                        // Set the delay for the time we expect to delay
+                        delay = Duration::from_millis(target_block_ts - now);
```
sisou: Unnecessary comment IMO :D

validator/src/micro.rs
```diff
@@ -141,16 +144,17 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
...
                     break None;
                 }
             }
-
             // We have dropped the blockchain lock.
-            // Wait a bit before trying to produce a block again.
-            delay += Self::CHECK_MEMPOOL_DELAY;
-            time::sleep(Self::CHECK_MEMPOOL_DELAY).await;
+            // Wait for the expected timestamp to arrive before actually produce the block
+            time::sleep(delay).await;
```
sisou:
```suggestion
            // Wait for the expected timestamp to arrive before actually producing the block
            time::sleep(delay).await;
```

PR #1113, opened 2022-11-08
Fix several clippy warnings

jsdanielh:
Fix several clippy warnings in the `keys`, `lib`, `nano-zpk`, `peer-address`, `tools` and `vrf` crates.
Warnings were related to needless borrows.
Also ignore the clippy warning for converting a boolean to `int` using `if` in the `beserial` crate since the solution doesn't make the code clearer.

## Pull request checklist

### DIFF
--- a/beserial/src/lib.rs
+++ b/beserial/src/lib.rs
@@ -198,6 +198,7 @@ impl Deserialize for bool {
 
 impl Serialize for bool {
     fn serialize<W: WriteBytesExt>(&self, writer: &mut W) -> Result<usize, SerializingError> {
+        #[allow(clippy::bool_to_int_with_if)]
         writer.write_u8(if *self { 1u8 } else { 0u8 })?;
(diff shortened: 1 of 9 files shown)


beserial/src/lib.rs
```diff
@@ -198,7 +198,7 @@ impl Deserialize for bool {
 
 impl Serialize for bool {
     fn serialize<W: WriteBytesExt>(&self, writer: &mut W) -> Result<usize, SerializingError> {
-        writer.write_u8(if *self { 1u8 } else { 0u8 })?;
+        writer.write_u8(u8::from(*self))?;
```
hrxi: This doesn't make the code clearer to me. Maybe ignore the warning for this line if you agree.

jsdanielh: I agree. Added an ignore to the warning

hrxi (approved):

PR #1118, opened 2022-11-10
Couple of warning fixes, debug compile time improvements

hrxi:
Fix clippy warnings, and improve test running times.

PR #1121, opened 2022-11-11
Fix clippy warnings

fiaxh:


jsdanielh:
I think this is the same as 2695d624e60bd16fcee03a167afe963ff2f92dc9 of #1118.

fiaxh:
You are right, this is a duplicate.

PR #1127, opened 2022-11-15
Unify devnet.sh and devnet_mac.sh

hrxi:
We don't want to maintain two versions of this script; they already diverged.

This gets rid of `&>>` redirection operator and replaces it with an equivalent because Apple doesn't update the bash they ship with their computers.


paberr:
This will unfortunately still cause problems on macOS due to the first few lines:
```
set -o errexit
set -o nounset
set -o pipefail
```

heinrich5991:
Can you show the errors? I think I tried that on a macOS computer in Apple's bash (version 2 or so) and it worked.

hrxi:
Can you show the errors? I think I tried that on a macOS computer in Apple's bash (version 2 or so) and it worked.

paberr:
Yes, you're partially right:
```
Starting seed node...
Done.
Starting validators...
    Starting Validator: 1
    Starting Validator: 2
    Starting Validator: 3
    Starting Validator: 4
Done.
./scripts/devnet/devnet.sh: line 340: kindexes[*]: unbound variable
```

The script will not fail in these lines.
It just makes it surface due to the nounset. 😅

jsdanielh:
This script will be gone with #1198.

PR #1129, opened 2022-11-15
Return `Result` for calculating `pk_tree_root`

jsdanielh:
Change the `MacroBlock` `pk_tree_root` function to return a `Result` and do additional checks on the provided `Validators`. This is to avoid hitting assertions when calling `pk_tree_construct` since the function assumes that the correct number of public keys are passed as arguments.

## Pull request checklist


paberr (approved):
Lgtm

PR #1131, opened 2022-11-16
Introducing blockchain proxies enum types

ii-cruz:
It introduces blockchain types by the creation of two enums - BlockchainProxy and BlockchainReadProxy. This is needed for the different sync strategies that require different blockchain implementations (eg. light and full sync).

## Pull request checklist


blockchain-proxy/Cargo.toml
```diff
@@ -0,0 +1,28 @@
...
+tokio-stream = { version = "0.1", features = ["sync"] }
+futures = { package = "futures-util", version = "0.3" }
+
+nimiq-block = { path = "../primitives/block" }
+nimiq-blockchain = { path = "../blockchain" }
+nimiq-database = { path = "../database" }
+nimiq-genesis = { path = "../genesis" }
+nimiq-hash = { path = "../hash" }
+nimiq-primitives = { path = "../primitives" }
+nimiq-transaction = { path = "../primitives/transaction" }
```
jsdanielh, line 28: Could you check all these dependencies? It looks to me like at least `nimiq-transaction` and `nimiq-database` are not used.

blockchain-proxy/src/blockchain_proxy.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+/// The `BlockchainReadProxy` implements `AbstractBlockchain` and allows to access common blockchain functions.
+/// It is a wrapper around read locked versions of our blockchain types.
+pub enum BlockchainReadProxy<'a> {
+    Full(Arc<RwLockReadGuard<'a, Blockchain>>),
+}
+
+impl<'a> AbstractBlockchain for BlockchainReadProxy<'a> {
+    fn network_id(&self) -> nimiq_genesis::NetworkId {
```
jsdanielh: I'd suggest to do `use nimiq_genesis::NetworkId` in the beginning of the file and then only having `NetworkId` here. This also applies to the rest of the return types, like `nimiq_block::Block`, etc.

blockchain-proxy/src/blockchain_proxy.rs
```diff
@@ -0,0 +1,180 @@
...
+    ($self: ident, $t: ident, $f: ident $(, $arg:expr )*) => {
+        match $self {
+            $t::Full(ref blockchain) => AbstractBlockchain::$f(&***blockchain, $( $arg ),*),
+        }
+    };
+}
+
+/// The `BlockchainProxy` is our abstraction over multiple types of blockchains.
+/// Currently, it holds the:
+/// - (Full)Blockchain, which is capable of storing the full history, transactions, and full blocks.
```
jsdanielh, line 23: I'd suggest to move this documentation to the member documentation (1 line below, just on top of `Full(...)`)

blockchain/src/blockchain/history_sync.rs
```diff
@@ -376,14 +376,31 @@ impl Blockchain {
...
-        this.notifier
-            .notify(BlockchainEvent::HistoryAdopted(block_hash.clone()));
+        if let Err(e) = this
+            .notifier
+            .send(BlockchainEvent::HistoryAdopted(block_hash.clone()))
+        {
+            log::error!(
+                "Error sending the History Adopted event to the events notifier {}",
+                e
+            );
```
jsdanielh, line 386: I'd suggest to use the tracing style, which I believe would be:
```
log::error!(
    error = e,
    "Error sending the History Adopted event to the events notifier"
);
```
And the same for the log messages in this PR

consensus/Cargo.toml
```diff
@@ -30,6 +30,7 @@ tokio-stream = { version = "0.1", features = ["sync"] }
 
 beserial = { path = "../beserial", features = ["derive"] }
 nimiq-block = { path = "../primitives/block" }
+nimiq-blockchain-proxy = { path = "../blockchain-proxy" }
 nimiq-bls = { path = "../bls" }
 nimiq-blockchain = { path = "../blockchain" }
```
jsdanielh, line 35: Is the `blockchain` still needed?

lib/src/client.rs
```diff
@@ -1,12 +1,13 @@
 use std::sync::Arc;
 
+use nimiq_blockchain_proxy::BlockchainProxy;
 use nimiq_bls::cache::PublicKeyCache;
 use nimiq_nano_zkp::NanoZKP;
 use nimiq_primitives::policy::Policy;
 use parking_lot::{Mutex, RwLock};
 
 use nimiq_block::Block;
-use nimiq_blockchain::{AbstractBlockchain, Blockchain};
```
jsdanielh, line 9: I'd suggest to keep it like this and remove the added `use` in line 52.

rpc-server/src/dispatchers/consensus.rs
```diff
@@ -1,10 +1,10 @@
 use std::sync::Arc;
 
 use async_trait::async_trait;
+use nimiq_blockchain::AbstractBlockchain;
```
jsdanielh: It looks to me like it was ok as it was. I think the separation is for first import standard library components, then external and then our own.

jsdanielh (commented):
Looks great. I only have some minor comments.

consensus/Cargo.toml
ii-cruz: It is. The blockchain crate has the abstract blockchain trait, so we need to import it in several places in the consensus to interact with whatever is inside the blockchain proxy.
We are also directly using the blockchain type for at least for now. There are some match cases for specifying the behavior if it's a full blockchain when pushing a block.

blockchain-proxy/Cargo.toml
ii-cruz: They were all needed. I changed the implementation of Abstract Blockchain to make that clearer.

jsdanielh (approved):

jsdanielh:
I'll rebase it to get it merged

PR #1147, opened 2022-11-23
Modify light blockchain to receieve 'this' lock

viquezclaudio:
Previously the light blockchain was receiving a mut reference to self, changed this to the same use model that is used in the regular blockchain where a RwLock of 'this' is used.

### DIFF
--- a/light-blockchain/src/blockchain.rs
+++ b/light-blockchain/src/blockchain.rs
@@ -1,4 +1,4 @@
-use std::sync::{Arc, RwLock};
+use std::sync::Arc;
 
 use nimiq_block::{Block, MacroBlock};
 use nimiq_blockchain::ChainInfo;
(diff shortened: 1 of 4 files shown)


light-blockchain/src/blockchain.rs
```diff
@@ -57,7 +57,7 @@ impl LightBlockchain {
             election_head: genesis_block.clone().unwrap_macro(),
             current_validators: genesis_block.validators(),
             genesis_block,
-            chain_store: RwLock::new(chain_store),
+            chain_store: chain_store,
```
jsdanielh:
```suggestion
            chain_store,
```
Clippy warning here

jsdanielh (changes requested):

jsdanielh (approved):
LGTM

PR #1154, opened 2022-11-28
Use `BlockchainProxy` for consensus handlers and move functions to `AbstractBlockchain`

jsdanielh:
- Change some of the `consensus` request handlers to use `BlockchainProxy` instead of `Blockchain`. This applies to the following requests:
  - `RequestMacroChain`
  - `RequestBlock`
  - `RequestMissingBlocks`
  - `RequestHead`
- Move `get_macro_blocks` to `AbstractBlockchain` and add its implementation for `Blockchain`, `LightBlockchain` and `BlockchainProxy`.
  Also implement `get_blocks` for `LightBlockchain`.  
- Use the `include_nody` argument for functions of the `AbstractBlockchain` implementation for `LightBlockchain` and return no blocks when set to `true`.

## Pull request checklist

### DIFF
--- a/light-blockchain/src/abstract_blockchain.rs
+++ b/light-blockchain/src/abstract_blockchain.rs
@@ -1,6 +1,6 @@
 use futures::stream::BoxStream;
 use nimiq_block::{Block, MacroBlock};
-use nimiq_blockchain::{AbstractBlockchain, BlockchainEvent, ChainInfo};
+use nimiq_blockchain::{AbstractBlockchain, BlockchainEvent, ChainInfo, Direction};
 use nimiq_database::Transaction;
 use nimiq_genesis::NetworkId;
 use nimiq_hash::Blake2bHash;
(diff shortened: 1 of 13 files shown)


viquezclaudio (approved):

light-blockchain/src/abstract_blockchain.rs
```diff
@@ -91,12 +99,36 @@ impl AbstractBlockchain for LightBlockchain {
...
+        &self,
+        start_block_hash: &Blake2bHash,
+        count: u32,
+        include_body: bool,
+        direction: Direction,
+        election_blocks_only: bool,
+        _txn_option: Option<&Transaction>,
+    ) -> Option<Vec<Block>> {
+        if include_body {
+            return Some(vec![]);
```
sisou, line 129: This early return breaks the assumption stated in the comment that it returns `None` if start block is not a macro block, as this early return does not check for that.

jsdanielh: This is being addressed in #1167.

PR #1169, opened 2022-12-02
light-blockchain: Check block bodies for macro blocks

jsdanielh:
- Check that macro blocks (not only election blocks) have a valid body for `LightBlockchain`.
- Change the block verification in `LightBlockchain`'s `sync` methods to use the new block verification functions.
- Change the `validator` code to publish block bodies also for checkpoint macro blocks.

## Pull request checklist

### DIFF
--- a/validator/src/validator.rs
+++ b/validator/src/validator.rs
@@ -568,15 +568,11 @@ impl<TNetwork: Network, TValidatorNetwork: ValidatorNetwork>
                 "Failed to publish block"
             );
         }
-        // Empty body before publishing to the block header topic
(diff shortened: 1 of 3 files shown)


validator/src/validator.rs
```diff
@@ -571,12 +571,8 @@ impl<TNetwork: Network, TValidatorNetwork: ValidatorNetwork>
         // Empty body before publishing to the block header topic
```
viquezclaudio: I think we should update this comment

viquezclaudio (approved):

PR #1176, opened 2022-12-08
Fix problem in light sync

viquezclaudio:
The only way to know if a peer is outdated is to request the epoch ids from that peer, because it could be the case that there is no prover node and a peer will always give us an outdated proof.
Fixed some misleading logs

### DIFF
--- a/consensus/src/sync/syncer.rs
+++ b/consensus/src/sync/syncer.rs
@@ -159,13 +159,11 @@ impl<N: Network, M: MacroSync<N::PeerId>, L: LiveSync<N>> Stream for Syncer<N, M
         while let Poll::Ready(result) = self.macro_sync.poll_next_unpin(cx) {
             match result {
                 Some(MacroSyncReturn::Good(peer_id)) => {
+                    debug!(%peer_id, "Macro sync returned good peer");
                     self.move_peer_into_live_sync(peer_id);
                 }
                 Some(MacroSyncReturn::Outdated(peer_id)) => {
-                    debug!(
-                        "History sync returned outdated peer {:?}. Waiting.",
-                        peer_id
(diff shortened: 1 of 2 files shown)


consensus/src/sync/syncer.rs
```diff
@@ -159,13 +159,11 @@ impl<N: Network, M: MacroSync<N::PeerId>, L: LiveSync<N>> Stream for Syncer<N, M
         while let Poll::Ready(result) = self.macro_sync.poll_next_unpin(cx) {
             match result {
                 Some(MacroSyncReturn::Good(peer_id)) => {
+                    debug!("Macro sync returned good peer {:?}", peer_id);
```
jsdanielh:
```suggestion
                    debug!(%peer_id, "Macro sync returned good peer");
```

consensus/src/sync/syncer.rs
```diff
@@ -159,13 +159,11 @@ impl<N: Network, M: MacroSync<N::PeerId>, L: LiveSync<N>> Stream for Syncer<N, M
...
                     self.move_peer_into_live_sync(peer_id);
                 }
                 Some(MacroSyncReturn::Outdated(peer_id)) => {
-                    debug!(
-                        "History sync returned outdated peer {:?}. Waiting.",
-                        peer_id
-                    );
+                    debug!("Macro sync returned outdated peer {:?}. Waiting.", peer_id);
```
jsdanielh:
```suggestion
                    debug!(peer_id, "Macro sync returned outdated peer. Waiting.");
```

jsdanielh (commented):
I'd suggest to use `tracing` style

jsdanielh (approved):

PR #1187, opened 2022-12-14
Add missing modules to logging

paberr:
We were missing a couple of new crates in the modules to be logged.
I used this command to compile the list and **removed tooling/testing related entries**:
```
grep 'name =' $(find . -name Cargo.toml)
```

## Pull request checklist


log/src/lib.rs
```diff
@@ -31,29 +29,27 @@ pub static NIMIQ_MODULES: &[&str] = &[
...
     "nimiq_network_mock",
-    "nimiq_peer_address",
+    "nimiq_pedersen_generators",
     "nimiq_primitives",
     "nimiq_rpc",
     "nimiq_rpc_client",
     "nimiq_rpc_interface",
     "nimiq_rpc_server",
-    "nimiq_signtx",
-    "nimiq_spammer",
```
jsdanielh, line 52: I think this one should stay, shouldn't it?

PR #1205, opened 2022-12-30
Fixing OOM issues in prover

paberr:

- Improve logging for zkp proving process
- Flush prover output before closing process
- Exit main process after running main
- Output full error message for NanoZKPErrors
- Pass command line arguments to child process
- Initialize logger for nimiq test prove binary
- Make timestamps in logs the default

## Pull request checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3179,6 +3179,7 @@ dependencies = [
  "nimiq-test-log",
  "rand 0.8.5",
  "thiserror",
+ "tracing",
 ]
 
 [[package]]
@@ -3715,6 +3716,7 @@ dependencies = [
  "nimiq-genesis-builder",
  "nimiq-hash",
  "nimiq-keys",
+ "nimiq-log",
  "nimiq-macros",
  "nimiq-mempool",
  "nimiq-nano-primitives",
@@ -3735,6 +3737,7 @@ dependencies = [
  "tokio",
  "tokio-stream",
  "tracing",
+ "tracing-subscriber 0.3.16",
 ]
 
 [[package]]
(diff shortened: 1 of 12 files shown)


jsdanielh (approved):

jsdanielh (commented):
The only thing is that I didn't get the purpose of the `prover-log` entry in the config file since it is doing the same as the `log` except that we only use `prover-log` when `prover` is set.

paberr:
The reason for the separate log section is that you can specify different tags to be logged or save them to a different file.
Since it's a separate process, logging into the same file might cause problems.

Alternatively, we could have made additional settings available in the `log` section.
We decided against this, because it would have meant passing the `prover` flag into many places.

jsdanielh:
> The reason for the separate log section is that you can specify different tags to be logged or save them to a different file. Since it's a separate process, logging into the same file might cause problems.
> 
> Alternatively, we could have made additional settings available in the `log` section. We decided against this, because it would have meant passing the `prover` flag into many places.

But from the code I see that we're basically setting the same argument of `initialize_logging`:
```
    initialize_logging(
        Some(&command_line),
        if command_line.prove {
            Some(&config_file.prover_log)
        } else {
            Some(&config_file.log)
        },
    )?;
```
Unless I'm missing something, the same could be achieved by setting the properties of `config_file.log`. So, I don't see the difference of setting something in `config_file.prover_log` or `config_file.log`.

paberr:
Both the prover and the client run using the same config file.
All settings should be shared except for the logging.
Using this approach, we can specify different logging settings for the prover and the client.

PR #1210, opened 2023-01-03
Fix bug in light blockchain

viquezclaudio:
- Fixed bug caused by not storing blocks when the chain ordering is inferior or unknown, as the regular blockchain does

### DIFF
--- a/light-blockchain/src/push.rs
+++ b/light-blockchain/src/push.rs
@@ -51,11 +51,6 @@ impl LightBlockchain {
         // Calculate chain ordering.
         let chain_order = ChainOrdering::order_chains(this.deref(), &block, &prev_info, None);
 
-        // If it is an inferior chain, we ignore it as it cannot become better at any point in time.
-        if chain_order == ChainOrdering::Inferior {
-            return Ok(PushResult::Ignored);
-        }
-
         // Get the intended slot owner.
         let offset = if let Block::Macro(macro_block) = &block {
             if let Some(proof) = &macro_block.justification {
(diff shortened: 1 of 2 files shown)


light-blockchain/src/push.rs
```diff
@@ -105,21 +100,27 @@ impl LightBlockchain {
...
                 return LightBlockchain::extend(this, chain_info, prev_info);
             }
             ChainOrdering::Superior => {
                 return LightBlockchain::rebranch(this, chain_info, prev_info);
             }
-            ChainOrdering::Inferior => unreachable!(),
-            ChainOrdering::Unknown => {}
-        }
+            ChainOrdering::Inferior => {
+                log::debug!(block = %chain_info.head, "Storing block - on inferior chain");
```
jsdanielh, line 111: Seems strange that we're logging this here and not in the other cases. Also I'd change the message to something like:
`log::debug!(block = %chain_info.head, "Ignore block - on inferior chain")`

light-blockchain/src/push.rs
```diff
@@ -105,21 +100,27 @@ impl LightBlockchain {
...
+                PushResult::Ignored
+            }
+            ChainOrdering::Unknown => {
+                log::debug!(block = %chain_info.head, "Storing block - on fork");
```
jsdanielh, line 115: Seems strange that we're logging this here and not in the other cases. Also I'd change the message to something like:
`log::debug!(block = %chain_info.head, "Forked block - on fork")`

light-blockchain/src/push.rs
viquezclaudio: Im following the same pattern that is used in the regular blockchain.
https://github.com/nimiq/core-rs-albatross/blob/albatross/blockchain/src/blockchain/push.rs#L102

Also we are not ignoring the block, we are storing the block as part of the chain info in line 121. 
That was actually the bug that was fixed.

light-blockchain/src/push.rs
viquezclaudio: same as above

light-blockchain/src/push.rs
viquezclaudio: We don't log in the other cases, because those cases are logged by the extend/rebranch functions and not from the caller side

jsdanielh (approved):

PR #1211, opened 2023-01-03
zkp-component: Add prover feature and prune tokio features

jsdanielh:
Add a new `prover` feature for the `zkp-component` crate such that it only requires `process` and `io-utils` features from tokio when enabled.
Also pruned the `tokio` features to stopped using `full` since it doesn't compile for wasm.

## Pull request checklist


ii-cruz (approved):
Looks good to me 👍

PR #1213, opened 2023-01-04
Add some features to `lib` and `network-libp2p` crates

jsdanielh:
Add the following features to the `network-libp2p` crate:
- `dns-tcp`: Feature that enables `tokio-dns` and `tokio-tcp` features of `libp2p`. These two features depend on `tokio` and don't compile for wasm. Also change the code to not use these transports when these features are not enabled.
- Include the `js` feature of `getrandom` when the build target is wasm.

And the following features to the `lib` crate:
- `tokio-console`: To enable the Tokio console and pull the `console-subscribe` crate that doesn't compile for wasm.
- `dns-tcp`: Feature to set the `network-libp2p`'s `dns-tcp` feature.

## Pull request checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1469,8 +1469,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9be70c98951c83b8d2f8f60d7065fa6d5146873094452a1008da8c2f1e4205ad"
 dependencies = [
  "cfg-if",
+ "js-sys",
  "libc",
  "wasi 0.10.2+wasi-snapshot-preview1",
+ "wasm-bindgen",
 ]
 
 [[package]]
@@ -3218,6 +3220,7 @@ dependencies = [
  "bytes",
  "derive_more",
  "futures-util",
+ "getrandom 0.2.6",
  "hex",
  "ip_network",
  "libp2p",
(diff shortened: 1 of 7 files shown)


sisou:
I found that instead of adding a `wasm` feature to relevant crates, which then needs to be added to all crates that use this crate up the dependency tree, it is possible to define specific features and dependencies in a crate based on the compilation target.

So in the `network-libp2p` crate, add this instead of the feature:
```
[target.'cfg(target_family = "wasm")'.dependencies]
getrandom = { version = "0.2.3", features = ["js"] }
```

That avoids adding the `wasm` feature to the `lib` crate just to forward it.

jsdanielh:
> I found that instead of adding a `wasm` feature to relevant crates, which then needs to be added to all crates that use this crate up the dependency tree, it is possible to define specific features and dependencies in a crate based on the compilation target.
> 
> So in the `network-libp2p` crate, add this instead of the feature:
> 
> ```
> [target.'cfg(target_family = "wasm")'.dependencies]
> getrandom = { version = "0.2.3", features = ["js"] }
> ```
> 
> That avoids adding the `wasm` feature to the `lib` crate just to forward it.

Indeed, that is better. Will make the change

viquezclaudio (approved):

PR #1215, opened 2023-01-04
utils: Replace the clear_on_drop dependency for zeroize

jsdanielh:
Replace the `clear_on_drop` dependency in favor of `zeroize` since `clear_on_drop` doesn't compile for wasm.
This change affects related crates such as:
- wallet
- keys

Also pruned the `util`'s `otp` feature from the `transaction-builder` dependencies.

## Pull request checklist


viquezclaudio (approved):

jsdanielh:
The wallet doesn't need to be supported for the wasm target. So this is no longer needed.

PR #1216, opened 2023-01-04
Removing the path of the main process from the prover process command.

ii-cruz:
Fixes the broken pipe error when launching the zk proof generation process. We must skip the first argument of the args array. This way the binary path of the main process is not passed on to the prover process.

## Pull request checklist

### DIFF
--- a/zkp-component/src/proof_gen_utils.rs
+++ b/zkp-component/src/proof_gen_utils.rs
@@ -80,7 +80,7 @@ pub async fn launch_generate_new_proof(
 
     let mut child = Command::new(path)
         .arg("--prove")
-        .args(env::args())
+        .args(env::args().skip(1))
         .stdin(Stdio::piped())
         .stdout(Stdio::piped())
         .spawn()?;


zkp-component/src/proof_gen_utils.rs
```diff
@@ -78,9 +78,12 @@ pub async fn launch_generate_new_proof(
     };
     log::debug!("Launching the prover process at path {:?}", path);
 
+    let mut args = env::args().collect::<Vec<String>>();
+    args.remove(0);
+
```
paberr:
```suggestion
```

zkp-component/src/proof_gen_utils.rs
```diff
@@ -78,9 +78,12 @@ pub async fn launch_generate_new_proof(
...
     let mut child = Command::new(path)
         .arg("--prove")
-        .args(env::args())
+        .args(args)
```
paberr: This should work equally well and avoids `collect`ing the vec :)
```suggestion
        .args(env::args().skip(1))
```

viquezclaudio (approved):

PR #1227, opened 2023-01-06
Do not care if there is no blockchain notifier listeners.

viquezclaudio:
In some clients configuration, there is no mempool or no validator. In those cases, there are no listeners for the blockchain events which is fine, so we should generate those events without logging errors if there is nobody listening to those broadcast channels

### DIFF
--- a/blockchain/src/blockchain/push.rs
+++ b/blockchain/src/blockchain/push.rs
@@ -259,25 +259,15 @@ impl Blockchain {
             "Accepted block",
         );
 
+        // We shouldn't log errors if there are no listeners
         if is_election_block {
-            if let Err(e) = this
+            _ = this
                 .notifier
-                .send(BlockchainEvent::EpochFinalized(block_hash))
-            {
-                log::error!(error = ?e,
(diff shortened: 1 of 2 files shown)


blockchain/src/blockchain/push.rs
```diff
@@ -503,14 +494,10 @@ impl Blockchain {
         this.metrics
             .note_rebranch(&reverted_blocks, &adopted_blocks);
```
jsdanielh, line 495: I'd add a comment on why we don't care about the result

blockchain/src/blockchain/push.rs
```diff
@@ -259,25 +259,16 @@ impl Blockchain {
             "Accepted block",
         );
 
+        // The blockchain notifier is used by the mempool/validator, however in some clients configuration
+        // there is no mempool/validator, so we shouldnt log errors if there are no listeners
```
jsdanielh:
```suggestion
        // there is no mempool/validator, so we shouldn't log errors if there are no listeners
```

jsdanielh (approved):

PR #1237, opened 2023-01-10
Prune epoch doesn't remove *all* micro blocks 

### LINKED ISSUE #1197: Prune epoch doesn't remove *all* micro blocks
`prune_epoch` is supposed to remove all micro blocks of an epoch.
To this end, it iterates over the block numbers and retrieves the blocks from the height index.

However, when there are forks, there could be duplicates at this height (see `get_blocks_at` for how to handle these).
Currently, we disregard such blocks.

Instead, we should iterate over all duplicates for a given height just as in the `get_blocks_at` function.

cud4m:
- Added test for MicroBlock prune
- Added get_block_hashes function. 
- Refactor prune_epoch function to use new get_block_hashes

## Pull request checklist

### DIFF
--- a/blockchain/src/chain_store.rs
+++ b/blockchain/src/chain_store.rs
@@ -277,6 +277,36 @@ impl ChainStore {
         }
     }
 
+    pub fn get_block_hashes_at(
+        &self,
+        block_height: u32,
+        txn_option: Option<&Transaction>,
+    ) -> Vec<Blake2bHash> {
+        let read_txn: ReadTransaction;
+        let txn = match txn_option {
+            Some(txn) => txn,
+            None => {
+                read_txn = ReadTransaction::new(&self.env);
+                &read_txn
+            }
+        };
+
+        let mut hashes: Vec<Blake2bHash> = Vec::new();
+        // Seek the hash of the first block at the given height and add it to our hashes vector
+        let mut cursor = txn.cursor(&self.height_idx);
+        let mut hash = cursor
+            .seek_key::<u32, Blake2bHash>(&block_height)
+            .map(|hash| (0u32, hash));
+
+        // Get any other block hash at the given height and add them to our hashes vector
+        while let Some((_, h)) = hash {
+            hashes.push(h);
+            hash = cursor.next_duplicate::<u32, Blake2bHash>();
+        }
+
+        hashes
+    }
+
     pub fn get_blocks_at(
         &self,
         block_height: u32,
@@ -304,7 +334,7 @@ impl ChainStore {
         while let Ok(block) = self.get_block(&block_hash, include_body, Some(txn)) {
             blocks.push(block);
 
-            // Get next block hash
+            // Get next block hash.
             block_hash = match cursor.next_duplicate::<u32, Blake2bHash>() {
                 Some((_, hash)) => hash,
                 None => break,
@@ -517,12 +547,13 @@ impl ChainStore {
 
         for height in Policy::first_block_of(epoch_number)..Policy::election_block_of(epoch_number)
         {
-            if let Some(hash) = txn.get::<u32, Blake2bHash>(&self.height_idx, &height) {
-                // If we detect a block whose prunable flag is set to false, we don't prune it
-                // Then we need to keep the previous macro block
+            let hashes = self.get_block_hashes_at(height, Some(txn));
+            for hash in hashes {
                 let chain_info: ChainInfo = txn
                     .get(&self.chain_db, &hash)
                     .expect("Corrupted store: ChainInfo referenced from index not found");
+                // If we detect a block whose prunable flag is set to false, we don't prune it
+                // Then we need to keep the previous macro block
                 if chain_info.prunable {
                     txn.remove(&self.chain_db, &hash);
                     txn.remove(&self.block_db, &hash);
--- a/blockchain/tests/mod.rs
+++ b/blockchain/tests/mod.rs
@@ -140,6 +140,106 @@ fn it_can_push_consecutive_view_changes() {
 }
 
 #[test]
+fn prune_epoch_micro_blocks() {
+    // Goal: test that every MicroBlock at a given height is removed when prune_epoch is executed.
+
+    let temp_producer = TemporaryBlockProducer::new();
+    let blockchain = Arc::clone(&temp_producer.blockchain);
+    let producer = temp_producer.producer;
+
+    // Create different MicroBlocks, push them, and then check they do exist.
+    // We ensure more than one MicroBlock at same height exists.
+    let micro_block1 = {
+        let bc_read = blockchain.read();
+        producer.next_micro_block(
+            &bc_read,
+            &bc_read.time.now() + 1_u64 * 1000,
+            vec![],
+            vec![],
+            vec![0x42],
+            None,
+        )
+    };
+    let micro_block2 = {
+        let bc_read = blockchain.read();
+        producer.next_micro_block(
+            &bc_read,
+            bc_read.time.now() + 1_u64 * 100,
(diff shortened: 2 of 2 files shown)


blockchain/src/chain_store.rs
```diff
@@ -1,9 +1,9 @@
+use std::hash::Hash;
```
jsdanielh: Looks like this import is not needed

blockchain/src/chain_store.rs
```diff
@@ -1,9 +1,9 @@
...
 use nimiq_account::Receipts;
 use nimiq_block::Block;
 use nimiq_database::cursor::{ReadCursor, WriteCursor};
-use nimiq_database::{
-    Database, DatabaseFlags, Environment, ReadTransaction, Transaction, WriteTransaction,
-};
+use nimiq_database::{cursor, Database, DatabaseFlags, Environment, ReadTransaction, Transaction,
```
jsdanielh: Looks like the `cursor` import is not needed

blockchain/tests/mod.rs
```diff
@@ -2,13 +2,13 @@ use parking_lot::RwLock;
 use std::sync::Arc;
 use tokio_stream::wrappers::BroadcastStream;
 use tokio_stream::StreamExt;
-
 use nimiq_block::Block;
 use nimiq_block_production::BlockProducer;
-use nimiq_blockchain::PushResult;
+use nimiq_blockchain::{ChainInfo, PushResult};
```
jsdanielh: Looks like the `ChainInfo` import is not needed

blockchain/tests/mod.rs
```diff
@@ -2,13 +2,13 @@ use parking_lot::RwLock;
...
 use nimiq_blockchain::{AbstractBlockchain, Blockchain, BlockchainConfig};
 use nimiq_database::volatile::VolatileEnvironment;
 use nimiq_genesis::NetworkId;
+use nimiq_hash::Blake2bHash;
```
jsdanielh: Looks like this import is not needed

jsdanielh (commented):
Also, could you run `cargo fmt`?. Looks like it isn't passing this CI job.

blockchain/src/chain_store.rs
```diff
@@ -272,6 +272,35 @@ impl ChainStore {
...
+            }
+        };
+
+        let mut hashes: Vec<Blake2bHash> = Vec::new();
+        // Seek the hash of the first block at the given height and add it to our hashes vector
+        let mut cursor = txn.cursor(&self.height_idx);
+        let mut hash = cursor.seek_key::<u32, Blake2bHash>(&block_height)
+            .map(|hash| (0u32, hash));
+
+        //Get any other block hash at the given height and add them to our hashes vector
```
jsdanielh:
```suggestion
        // Get any other block hash at the given height and add them to our hashes vector
```

blockchain/src/chain_store.rs
```diff
@@ -512,16 +540,19 @@ impl ChainStore {
...
-                    txn.remove(&self.chain_db, &hash);
-                    txn.remove(&self.block_db, &hash);
-                    txn.remove_item(&self.height_idx, &height, &hash);
+            if let Ok(hashes) = self.get_block_hashes_at(height,
+                                                         Some(txn)) {
+                for hash in hashes {
+                    let chain_info: ChainInfo = txn
+                        .get(&self.chain_db, &hash)
+                        .expect("Corrupted store: ChainInfo referenced from index not found");
+                     // If we detect a block whose prunable flag is set to false, we don't prune it
```
jsdanielh:
```suggestion
                    // If we detect a block whose prunable flag is set to false, we don't prune it
```

blockchain/src/chain_store.rs
cud4m: Done! I'll be more careful from now on

cud4m:
The conflict is because of the import. There's really no need to have it split in two lines

blockchain/src/chain_store.rs
```diff
@@ -272,6 +272,36 @@ impl ChainStore {
         }
     }
 
+    pub fn get_block_hashes_at(
+        &self,
+        block_height: u32,
+        txn_option: Option<&Transaction>,
+    ) -> Result<Vec<Blake2bHash>, BlockchainError> {
```
paberr: As far as I can see, the function actually cannot return an error. So, a return type of `Vec<Blake2bHash>` would be sufficient.

blockchain/tests/mod.rs
```diff
@@ -1,19 +1,17 @@
...
 use nimiq_genesis::NetworkId;
 use nimiq_primitives::policy::Policy;
 use nimiq_test_log::test;
 use nimiq_test_utils::block_production::TemporaryBlockProducer;
 use nimiq_test_utils::blockchain::{signing_key, voting_key};
 use nimiq_utils::time::OffsetTime;
+use parking_lot::RwLock;
+use std::sync::Arc;
+use tokio_stream::wrappers::BroadcastStream;
+use tokio_stream::StreamExt;
```
paberr: The previous ordering of imports was correct.
We start with a block of external dependencies, then a block of nimiq dependencies, then a block of crate dependencies.

blockchain/tests/mod.rs
```diff
@@ -139,6 +137,106 @@ fn it_can_push_consecutive_view_changes() {
...
+        .is_ok());
+    assert!(bc_read
+        .chain_store
+        .get_chain_info(&micro_block2.hash(), false, None,)
+        .is_ok());
+    assert!(bc_read
+        .chain_store
+        .get_chain_info(&micro_block3.hash(), false, None,)
+        .is_ok());
+    assert_eq!(blockchain.read().block_number(), 1);
```
paberr:
```suggestion
    assert_eq!(bc_read.block_number(), 1);
```

blockchain/src/chain_store.rs
cud4m: Yes, you are completely right. Modified in accordance

blockchain/tests/mod.rs
cud4m: Modified

blockchain/tests/mod.rs
```diff
@@ -139,6 +138,106 @@ fn it_can_push_consecutive_view_changes() {
...
+        .get_chain_info(&micro_block2.hash(), false, None,)
+        .is_ok());
+    assert!(bc_read
+        .chain_store
+        .get_chain_info(&micro_block3.hash(), false, None,)
+        .is_ok());
+    assert_eq!(bc_read.block_number(), 1);
+
+    let mut txs = bc_read.write_transaction();
+    // prune the 3 created MicroBlocks
```
paberr:
```suggestion
    // Prune the 3 created MicroBlocks
```

blockchain/tests/mod.rs
```diff
@@ -139,6 +138,106 @@ fn it_can_push_consecutive_view_changes() {
...
+    bc_read.chain_store.prune_epoch(1, &mut txs);
+
+    // check that they no longer exist
```
paberr:
```suggestion
    // Check that they no longer exist
```

blockchain/src/chain_store.rs
```diff
@@ -298,7 +328,6 @@ impl ChainStore {
         // Iterate until we find all blocks at the same height.
         while let Ok(block) = self.get_block(&block_hash, include_body, Some(txn)) {
             blocks.push(block);
-
```
paberr, line 306: Readd new line

paberr (commented):
Needs rebase + small changes, then it's ready to be merged

PR #1250, opened 2023-01-13
Improve feature support for wasm

jsdanielh:
Improve feature support for `wasm` and also change some of the `wasm` targets in favor of features in crates like `blockchain-proxy` and `consensus`.

## Pull request checklist

### DIFF
--- a/lib/src/client.rs
+++ b/lib/src/client.rs
@@ -7,6 +7,7 @@ use nimiq_primitives::policy::Policy;
 use parking_lot::{Mutex, RwLock};
 
 use nimiq_block::Block;
+#[cfg(feature = "full-consensus")]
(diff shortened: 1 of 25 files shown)


lib/src/client.rs
```diff
@@ -236,7 +243,9 @@ impl ClientInner {
                 let zkp_component = ZKPComponent::new(
                     blockchain_proxy.clone(),
                     Arc::clone(&network),
+                    #[cfg(feature = "zkp-prover")]
```
paberr, line 246: Maybe we should make the feature name more consistent. In the ZKP component it's called 'prover' only. I guess going with `zkp-prover` all along is more descriptive.

viquezclaudio (approved):

PR #1277, opened 2023-01-20
Rename `light-client` to `web-client` and improve logging for wasm

jsdanielh:
- Rename the `light-client` crate to `web-client` and re-write the
  client code to be `wasm` friendly.    
- Add a web logging feature to the `nimiq-lib` crate that makes use of
  the `tracing-web` crate.

## Pull request checklist


web-client/Cargo.toml
```diff
@@ -1,9 +1,9 @@
 [package]
-name = "nimiq-light-client"
+name = "nimiq-web-client"
```
viquezclaudio, line 2: I think we should specify this crate only builds for wasm32

web-client/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+    config::config_file::{ConfigFile, LogSettings, SyncMode},
+    error::Error,
+    extras::{
+        deadlock::initialize_deadlock_detection, metrics_server::NimiqTaskMonitor,
+        panic::initialize_panic_reporting, signal_handling::initialize_signal_handler,
+        web_logging::initialize_web_logging,
+    },
+};
+
+async fn light_client() -> () {
```
viquezclaudio: some Clippy warnings in this file

web-client/Cargo.toml
viquezclaudio: [build]
target = "wasm32-unknown-unknown"

sisou:
Is the reason to rename, as @viquezclaudio indicates, that with the recent changes, this crate _only_ builds for WASM, not for regular targets? Or any other reasoning?

web-client/Cargo.toml
jsdanielh: Unfortunately, AFAIK this is not supported

jsdanielh:
> Is the reason to rename, as @viquezclaudio indicates, that with the recent changes, this crate _only_ builds for WASM, not for regular targets? Or any other reasoning?

It will build for other targets but won't run

viquezclaudio:
Yup... it will panic during runtime if somebody tries to run the binary in not wasm targets, that's why I wanted (if possible) some way or forcing to compile it only for wasm

web-client/Cargo.toml
sisou: Tracking issue: https://github.com/rust-lang/cargo/issues/6179

sisou:
As an alternative to restricting valid targets for cargo, we should at least add a README file to the crate that specifies this incompatibility with non-wasm targets as a major warning.

jsdanielh:
> As an alternative to restricting valid targets for cargo, we should at least add a README file to the crate that specifies this incompatibility with non-wasm targets as a major warning.

Done.

web-client/README.md
```diff
@@ -0,0 +1,24 @@
+# Nimiq web client
+
+This client is a very light client that only includes the necessary dependencies and constructs
```
viquezclaudio, line 3: I think I will re phrase this:
This is a web-client intended to be used in web browsers only (no WASI support).
Although it can be built for other targets, it will panic if it is executed outside a web browser.

viquezclaudio (approved):

PR #1281, opened 2023-01-23
Consensus: use a wasm-friendly sleep

viquezclaudio:


consensus/src/consensus/mod.rs
```diff
@@ -1,3 +1,5 @@
+#[cfg(target_arch = "wasm32")]
+use gloo_timers::future::{sleep, TimeoutFuture};
```
jsdanielh: I'd move this include by line 10 with all non `std` includes

jsdanielh (commented):
Also, could you rebase the branch?

jsdanielh (approved):

PR #1283, opened 2023-01-24
Consensus: Receive a task executor

viquezclaudio:
This is used to make consensus generic over the executor that is used to spawn its tasks. For instance, for all non-web clients we use tokio::spawn. For web-clients we use wasm-bindgen::spawn_local

### DIFF
--- a/primitives/src/lib.rs
+++ b/primitives/src/lib.rs
@@ -14,3 +14,5 @@ pub mod slots;
 pub mod transaction;
 #[cfg(feature = "trie")]
 pub mod trie;
+
+pub mod task_executor;
--- a/primitives/src/task_executor.rs
+++ b/primitives/src/task_executor.rs
@@ -1,0 +1,15 @@
+use std::{future::Future, pin::Pin};
+
+/// This a trait used to abstract the task executor that is used in several modules.
+/// For instance, it can be used to execute tasks with tokio or wasm-bindgen if needed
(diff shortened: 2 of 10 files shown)


primitives/src/lib.rs
```diff
@@ -10,3 +10,5 @@ pub mod networks;
 pub mod policy;
 #[cfg(feature = "slots")]
 pub mod slots;
+
```
jsdanielh, line 17: I'd remove this extra line

primitives/src/task_executor.rs
```diff
@@ -0,0 +1,15 @@
+use std::{future::Future, pin::Pin};
+
+/// This a trait used to abstract the task executor that is used in several modules
```
jsdanielh:
```suggestion
/// This a trait used to abstract the task executor that is used in several modules.
```

primitives/src/task_executor.rs
```diff
@@ -0,0 +1,15 @@
...
+/// For instance, can be used to execute tasks with Tokio or Wasm-bindgen if needed
```
jsdanielh:
```suggestion
/// For instance, it can be used to execute tasks with `tokio` or `wasm-bindgen` if needed
```

jsdanielh (commented):
Also, we need to rebase this branch

jsdanielh (approved):

jsdanielh (approved):

PR #1288, opened 2023-01-25
Create database-storage feature

viquezclaudio:
- This is used to remove the database for wasm compilation
- Create missing light client defaults

### DIFF
--- a/lib/src/config/config.rs
+++ b/lib/src/config/config.rs
@@ -13,12 +13,7 @@ use strum_macros::Display;
 use beserial::Deserialize;
 #[cfg(feature = "validator")]
 use nimiq_bls::{KeyPair as BlsKeyPair, SecretKey as BlsSecretKey};
-#[cfg(any(
-    feature = "full-consensus",
-    feature = "validator",
-    feature = "wallet",
-    feature = "zkp-storage"
-))]
+#[cfg(feature = "database-storage")]
 use nimiq_database::{mdbx::MdbxEnvironment, volatile::VolatileEnvironment, Environment};
 #[cfg(feature = "validator")]
 use nimiq_keys::{Address, KeyPair, PrivateKey};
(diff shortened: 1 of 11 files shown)


lib/src/config/config.rs
```diff
@@ -17,7 +17,7 @@ use nimiq_bls::{KeyPair as BlsKeyPair, SecretKey as BlsSecretKey};
     feature = "full-consensus",
     feature = "validator",
     feature = "wallet",
-    feature = "zkp-storage"
+    feature = "database-storage"
```
jsdanielh: I think here we could remove every feature except `database-storage`

lib/src/config/config.rs
```diff
@@ -651,6 +649,13 @@ impl ClientConfigBuilder {
         self.network_id(NetworkId::TestAlbatross)
     }
 
+    /// Configuration for Light Client
+    pub fn light(&mut self) -> &mut Self {
+        let mut consensus_config = ConsensusConfig::default();
+        consensus_config.sync_mode = SyncMode::Light;
```
jsdanielh: We need to fix this clippy warning

lib/src/config/config.rs
```diff
@@ -588,6 +585,7 @@ pub struct ClientConfig {
 
     /// Database-specific configuration
     ///
+    #[cfg(feature = "database-storage")]
```
jsdanielh, line 583: Shouldn't we gate `storage` as well?

lib/src/error.rs
```diff
@@ -11,7 +11,7 @@ pub enum Error {
         feature = "full-consensus",
         feature = "validator",
         feature = "wallet",
-        feature = "zkp-storage"
+        feature = "database-storage"
```
jsdanielh: I think here we could remove all features except `database-storage`

lib/src/config/config.rs
viquezclaudio: No because storage is needed to specify Volatiles environments, and I think this is fine: you could have a database storage and a volatile storage

jsdanielh (approved):

PR #1295, opened 2023-01-26
Add missing Volatile configuration for WASM

viquezclaudio:
- The Storage configuration should be set to Volatile for web environments
- Removed unused enum type, since volatile is sufficient for the browser

### DIFF
--- a/lib/src/config/config.rs
+++ b/lib/src/config/config.rs
@@ -249,6 +249,7 @@ impl From<Option<config_file::DatabaseSettings>> for DatabaseConfig {
 pub enum StorageConfig {
     /// This will store the database in a volatile storage. After the client shuts
     /// down all data will be lost.
+    /// This can be used for browser based environments
     ///
     Volatile,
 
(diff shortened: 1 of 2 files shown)


jsdanielh (approved):

lib/src/config/config.rs
```diff
@@ -249,22 +249,14 @@ impl From<Option<config_file::DatabaseSettings>> for DatabaseConfig {
...
     Filesystem(FileStorageConfig),
-
-    /// This will store the database in the browser using *IndexedDB*, and the key files in
-    /// *LocalStorage*. This is only available when run in the browser.
-    ///
-    /// # Notes
-    ///
-    /// There no support for WebAssembly yet.
-    ///
-    Browser,
```
jsdanielh, line 267: Looks like this generated some warnings of `unreachable` code where this `enum` was used for `match`es

jsdanielh (approved):

PR #1301, opened 2023-01-29
Add CI for GHCR image

brantje:
This added an automatic build of a docker image for Nimiq.
After a tag is created, the CI will run and build a image, when building is completed, it will be pushed to Nimiq's package registery.

Please note:
- `GHCR_TOKEN` must be set (This can be done at: https://github.com/nimiq/core-rs-albatross/settings/secrets/actions)
- Only nimiq-client is copied to the docker
- Due libssl1.1 being required ubuntu 20.04 is used.

Steps to create a token:
Go to: https://github.com/settings/tokens/new
Check the following checkboxes:

Copy the token and add it as variable to the repo.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -560,13 +560,13 @@ dependencies = [
 
 [[package]]
 name = "async-trait"
-version = "0.1.66"
+version = "0.1.67"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b84f9ebcc6c1f5b8cb160f6990096a5c127f423fcb6e1ccc46c370cbdfb75dfc"
+checksum = "86ea188f25f0255d8f92797797c97ebf5631fa88178beb1a46fdf5622c9a00e4"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 1.0.109",
+ "syn 2.0.0",
 ]
 
 [[package]]


Dockerfile.prod
```diff
@@ -0,0 +1,46 @@
+FROM ubuntu:20.04
+
+# Install dependencies and tini.
```
maestroi: tini is not installed here

Dockerfile.prod
```diff
@@ -0,0 +1,46 @@
...
+EXPOSE 9100
+
+# Expose RPC port
+EXPOSE 8648
+
+# Expose metrics port
+EXPOSE 9200
+
+# Run CMD so we can use other bin
+CMD [ "nimiq-client" ]
```
maestroi: Use entrypoint except if you want other binaries to be accessable by overwriting the command

Dockerfile.prod
```diff
@@ -0,0 +1,46 @@
...
+USER nimiq
+
+# Change homedir to nimiq
+WORKDIR /home/nimiq
+
+# Create nimiq directory for configuration
+RUN mkdir -p /home/nimiq/.nimiq
+
+# Set default config can be overwritten by mounting
+COPY ./client.toml /home/nimiq/.nimiq
```
maestroi: not sure where this client.toml is

Dockerfile.prod
```diff
@@ -0,0 +1,46 @@
...
+
+COPY ./target/release/nimiq-client /usr/local/bin/nimiq-client
+
+# Copy .ZKP to node.
+COPY ./.zkp /home/nimiq/.zkp/
+
+# Expose sync port
+EXPOSE 9100
```
maestroi: might wanna change this but doesn't matter that much, 9100 mostly node_exporter

brantje: 9100 is by default the p2p port:
```
listen_addresses = [
        "/ip4/0.0.0.0/tcp/9100/ws",
]
```

Dockerfile.prod
brantje: It's in the root of the repo: https://github.com/nimiq/core-rs-albatross/blob/albatross/client.toml

jsdanielh:
@brantje Thanks for the PR. It looks very cool to me. Will ask @jeffesquivels to take another look and if everything is ok he would set up the token for us.

jeffesquivels:
Hi guys!

First of all, thanks for the PR, @brantje, this is very useful functionality (also thanks to @maestroi for the first review and comments).

I did a first quick review and in general it looks very good, I still need to take a more in-depth look of some of the Actions that are used in `github_release_docker.yml`, but in the meantime I have a couple of comments/questions:

1. I don't think `libssl1.1` is a build dependency, only `libssl-dev` is, so the project can be built on Ubuntu 22.04 (and then the dependency on the runtime container image should be `libssl3`, but I'm not 100% percent sure of this).
2. Is there a reason why `tini` was dropped from the runtime container image?

Regards!

brantje:
afaik libssl1.1 is a requirement, i tried running without it, but gives an error upon start up.
However, i didn't try with 22.04 + libssl3. 

No reason for removing tiny, thought it's not needed for running the client directly. 


Update: 
Compiled on 22.04 the binary, then changed the ubuntu version to 22.04 in Dockerfile.prod
Changed `libssl1.1` to `libss3` in the apt command and then build the container, no faults.

However, upon running the container the following error shows:
`nimiq-client: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory` 

It might be possible to work around this by using some `ln` magic but not a big fan of that.

maestroi:
@jeffesquivels anything left here?

jeffesquivels:
> @jeffesquivels anything left here?

Thanks for the patience, somehow other stuff kept popping up that needed to be done before I could finish reviewing this PR :sweat_smile: .

I just rebased the original commits and on top of them added several changes of my own (one per commit, to make it easy to review), please look into my changes, if everything seems OK, then you can squash everything into just one commit and I will merge it then. If you have any questions/issues, let me know.

Thanks again, all of your effort is deeply appreciated :-).

jeffesquivels (approved):

jeffesquivels:
For some reason Github closed this PR automatically and now doesn't allow me to add the final commit to the repo/branch this PR was created from, but I just pushed the commit manually: https://github.com/nimiq/core-rs-albatross/commit/e92d5944087e7af9c017cdff2b658c38612fcb31, so now this is effectively closed.

Thanks @brantje @maestroi !

PR #1305, opened 2023-01-30
Use standard parking lot crate

viquezclaudio:
- The parking lot that we were using is not compatible with WASM targets because of the Instant version that was being used
- The fork was being used because it provided additional details in case of deadlocks, however currently that extra information is not being used

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2998,16 +2998,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "lock_api"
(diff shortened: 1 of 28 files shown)


jsdanielh (commented):
Mmm, I think it's better to use:
`parking_lot = "0.12"` and then make sure the `Cargo.lock` has 0.12.1.

jsdanielh (approved):

PR #1312, opened 2023-01-31
network-libp2p: Add a transport for `wasm`

### LINKED ISSUE #1214: Enable a proper libp2p transport for wasm
Enable a proper libp2p transport for wasm.
PR #1213 introduces a `wasm` feature and also gates the `tokio-dns` and `tokio-tcp` libp2p feature behind a `network-libp2p` feature but doesn't handle properly the case when these `tokio` feature are disabled and `wasm` is enabled.
A proper transport needs to be implemented for this case.

jsdanielh:
- Add a transport for `wasm` gated under `wasm-websocket` feature. This transport uses a forked version of `libp2p-websys-transport` only to use the same forked version of `libp2p` that this crate uses for now.
- Rename the feature `websocket` to `wasm-websocket` since this adds support for a different implementation of a websocket transport.

This closes #1214.

## Pull request checklist


sisou:
With this, are server nodes still able to create websocket servers and accept websockets from browsers? Or does it change _all_ websockets to be wasm?

jsdanielh:
> With this, are server nodes still able to create websocket servers and accept websockets from browsers? Or does it change _all_ websockets to be wasm?

Only those that set the feature `wasm-websocket` and that don't set the `tokio-websocket` feature, which as of now only the `web-client` does.

viquezclaudio (approved):

PR #1315, opened 2023-01-31
Use task executor for light macro sync

viquezclaudio:


jsdanielh:
Shouldn't we pass along the `executor`?

jsdanielh (approved):

Issue #1289, opened 2023-01-26
Update arkworks to 0.4


paberr:
There have been plenty of changes in arkworks and we need to update to the recently released version.

PR #1323, opened 2023-02-03
Making zkp component and proof verification wasm friendly by removing the file system requirement  

### LINKED ISSUE #1296: Make the ZKP component WASM-friendly
Currently the ZKP component expects the prover/verifier keys to be located in some files. 
However, this is not browser friendly since there is no filesystem in the browser.

For browser based environments, the verifier keys should be sufficient. 

So one possible solution for development, for WASM compilation, would be to either generate new key files or import ones (if they already exist) during compilation time and save them in some sort of array or data structure in a source file such that during runtime, this can be imported without having to interact with a filesystem..
However this would require some changes (for instance in the verify function of the NanoZKP component) such that we abstract in some way the location of those keys, either in a file system or imported from a source file. In any case, this is just a suggestion, and there are other possible approaches that might be more suitable and the final user environment should be taken into consideration.

ii-cruz:
Puts the verifying key (we only use one to verify the proofs) into the binary at build time. The verification of proof now uses this key (in memory) rather than using the file system.

To achieve this a couple of things were changed:
- The configuration option for zkp keys path is now only for the proving keys and only used for the zkp prover when it is active. The verifying keys are always in the .zkp (used to be the default folder).
- When building we panick if no verifying keys are present on the default directory.
- At runtime, if the prover node is missing the proving keys, it will have to generate them.
- The setup, circuits, and gadgets of the nano zkp had to be moved to a new crate to avoid cyclic dependencies. This should be revised together with issue #1289]
- Also renamed most "nano" relaated crates to "zkp" crates.

## Pull request checklist


blockchain/src/blockchain/blockchain.rs
```diff
@@ -62,16 +62,13 @@ pub struct BlockchainConfig {
     /// Maximum number of epochs (other than the current one) that the ChainStore will store fully.
     /// Epochs older than this number will be pruned.
     pub max_epochs_stored: u32,
-    /// The path for the ZKP verifying keys.
-    pub keys_path: PathBuf,
```
paberr, line 66: I would probably still keep the path (but only at runtime and only for the proving keys).
That makes it possible to have a more flexible setup with pre-compiled binaries.

consensus/tests/history_sync.rs
```diff
@@ -8,9 +8,10 @@ use futures::StreamExt;
 use nimiq_blockchain_proxy::BlockchainProxy;
 use nimiq_bls::cache::PublicKeyCache;
 use nimiq_consensus::sync::syncer_proxy::SyncerProxy;
+use nimiq_nano_primitives::KEYS_PATH;
 use nimiq_test_log::test;
 use nimiq_test_utils::node::TESTING_BLS_CACHE_MAX_CAPACITY;
-use nimiq_test_utils::zkp_test_data::{zkp_test_exe, KEYS_PATH};
```
paberr, line 13: The path is different in test/run environments. That is because tests are run in a different directory relative to the repository's root.
So, we'll still need two separate constants.

lib/src/client.rs
```diff
@@ -131,30 +137,45 @@ impl ClientInner {
...
-            && !NanoZKP::all_files_created(&config.zkp.setup_keys_path, config.zkp.prover_active)
-        {
-            // If the prover node is disabled, we do not generate keys
-            // but inform the user of having set a wrong location for the verifying keys.
-            if !config.zkp.prover_active {
-                log::error!("Missing ZKP verification keys. Make sure to set the correct `setup_keys_path` inside the config.");
+        // If we are missing the verifying keys in memory there might be an error or we may need to generate them not in run time.
+        // If the Prover is active we need to ensure that the proving keys are present.
+        if ZKP_VERIFYING_KEY.read().is_none() || config.zkp.prover_active {
+            if !config.zkp.prover_active
```
paberr: Why do we check `!config.zkp.prover_active`? Doesn't the prover need the key as well?

lib/src/client.rs
```diff
@@ -131,30 +137,45 @@ impl ClientInner {
...
+                || config.network_id == NetworkId::UnitAlbatross
```
paberr: Are the `||` checks actually what we want?

lib/src/client.rs
```diff
@@ -131,30 +137,45 @@ impl ClientInner {
...
-                1, 0, 52, 0, 0, 0, 0, 0, 1, 0, 10, 0, 22, 32, 0, 0, 2, 0, 55, 49, 0, 11, 0, 0, 3,
-                0, 0, 0, 0, 0, 2, 92,
-            ];
-            NanoZKP::setup(
-                ChaCha20Rng::from_seed(seed),
-                &config.zkp.setup_keys_path,
-                config.zkp.prover_active,
-            )?;
-            log::debug!("Finished Nano ZKP setup.")
+            if config.network_id == NetworkId::DevAlbatross || config.zkp.prover_active {
```
paberr: Check `||`

lib/src/client.rs
```diff
@@ -131,30 +137,45 @@ impl ClientInner {
...
+                        );
+                        setup(
+                            ChaCha20Rng::from_seed(SEED),
+                            &zkp_keys_path,
+                            config.zkp.prover_active,
+                        )?;
+                    }
+                    log::info!("Setting the verification key.");
+                    let vk = load_verifying_key_from_file(&zkp_keys_path)?;
+                    set_verifying_key(vk);
```
paberr: Do we need to set the key?

nano-primitives/Cargo.toml
```diff
@@ -17,14 +17,19 @@ rayon = "^1.6"
...
+rand = { version = "0.8", features = ["small_rng"] }
+thiserror = "1.0"
 
 ark-crypto-primitives = "0.3"
 ark-ec = "0.3"
 ark-ff = "0.3"
 ark-groth16 = "0.3"
 ark-mnt4-753 = "0.3"
 ark-mnt6-753 = "0.3"
+ark-relations = "0.3"
```
paberr, line 28: I'm not sure whether we should put all of this into `nano-primitives`.
Perhaps a new crate would be better?
Also, we should rename all the `nano` related crates. 😆

nano-primitives/src/lib.rs
```diff
@@ -7,13 +7,44 @@ pub use macro_block::*;
...
+
+use std::io;
+
+use ark_relations::r1cs::SynthesisError;
+use ark_serialize::SerializationError;
+use beserial::SerializingError;
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum NanoZKPError {
```
paberr: This is good to stay in the primitives.

nano-primitives/src/setup.rs
```diff
@@ -0,0 +1,484 @@
...
+    circuits::mnt6::{
+        MacroBlockWrapperCircuit, MergerWrapperCircuit, PKTreeNodeCircuit as NodeMNT6,
+    },
+    MacroBlock, NanoZKPError, PK_TREE_BREADTH, PK_TREE_DEPTH,
+};
+use nimiq_bls::utils::bytes_to_bits;
+use nimiq_primitives::policy::Policy;
+
+pub const KEYS_PATH: &str = "../.zkp";
+pub const SEED: [u8; 32] = [
```
paberr: Rename to make clear this seed it only used in devnet albatross!

nano-zkp/build.rs
```diff
@@ -0,0 +1,16 @@
...
+
+use nimiq_compile_zkp_setup::verifying_keys_setup;
+use nimiq_nano_primitives::{NanoZKPError, KEYS_PATH};
+
+fn main() -> Result<(), NanoZKPError> {
+    let keys_dest_path = Path::new(&env::var_os("OUT_DIR").unwrap()).join("verifying_keys.data");
+
+    let keys_source_path = Path::new(KEYS_PATH);
+
+    verifying_keys_setup(keys_source_path, &keys_dest_path)?;
```
paberr: I'm still debating a bit with myself whether we should run the setup on build or whether we should abort with an error. I think I'm slightly in favour of offering a separate binary to setup the keys and failing the client build with an error. That gives a bit more control to the person trying to build the client.
(It could be easily missed that the keys were not there and now building takes ages)

nano-zkp/compile-zkp-setup/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+use nimiq_nano_primitives::{
+    setup::{all_files_created, setup},
+    NanoZKPError, SEED,
+};
+
+pub fn verifying_keys_setup(source_path: &Path, dest_path: &Path) -> Result<(), NanoZKPError> {
+    let is_prover_feature = env::var_os("CARGO_FEATURE_ZKP_PROVER")
+        .map_or(false, |var| var.into_string().unwrap() == 1.to_string());
+
+    // Generates the verifying keys if they don't exist yet.
```
paberr: See comment in build.rs

nano-zkp/compile-zkp-setup/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+    let mut file = File::open(
+        source_path
+            .join("verifying_keys")
+            .join("merger_wrapper.bin"),
+    )?;
+    let vk: VerifyingKey<MNT6<Parameters>> = VerifyingKey::deserialize_unchecked(&mut file)?;
+
+    // Creates the destination files to be included in the binary and writes the key to them.
+    let mut dest_file = File::create(dest_path)?;
+    vk.serialize_unchecked(&mut dest_file)?;
```
paberr: Can't we use the existing file instead of writing it to a different directory?
Probably not easily – just checking.

nano-zkp/src/verifying_key.rs
```diff
@@ -0,0 +1,24 @@
...
+fn init_verifying_key() -> Option<VerifyingKey<MNT6<Parameters>>> {
+    let serialized = include_bytes!(concat!(env!("OUT_DIR"), "/verifying_keys.data"));
+
+    let mut serialized_cursor = Cursor::new(serialized);
+    VerifyingKey::deserialize_unchecked(&mut serialized_cursor).ok()
+}
+
+pub fn set_verifying_key(key: VerifyingKey<MNT6<Parameters>>) {
+    *ZKP_VERIFYING_KEY.write() = Some(key);
+}
```
paberr: I would not allow setting the verifying key and get rid of the `RwLock`.
When the prover generates the keys, I would instead check that the verifying keys match.

Also, I would try getting rid of the `Option`. All our client types require the verifying key.
So, we could panic if the deserialize fails. (I think we do the same for the pedersen generators)

zkp-component/tests/zkp_requests.rs
```diff
@@ -45,7 +45,7 @@ fn blockchain() -> Arc<RwLock<Blockchain>> {
 
 #[test(tokio::test)]
 async fn peers_dont_reply_with_outdated_proof() {
-    NanoZKP::setup(get_base_seed(), Path::new(KEYS_PATH), false).unwrap();
+    setup(get_base_seed(), Path::new(KEYS_PATH), false).unwrap();
```
paberr: Is `setup` the only function that's not part of `NanoZKP` anymore? Probably we should make all of them independent of the unused struct.

paberr (commented):
Generally looks good to me.
Just a few ideas, and possible restructurings.

nano-primitives/Cargo.toml
paberr: Maybe `nano-primitives` -> `zkp_primitives`
The new crate could be `zkp_circuits`.
`nano_zkp` -> `zkp`

nano-primitives/src/setup.rs
ii-cruz: rename Development_seed

nano-zkp/src/nano_zkp/verify.rs
```diff
@@ -36,13 +32,10 @@ impl NanoZKP {
         final_pks: Vec<G2MNT6>,
         // The SNARK proof for this circuit.
         proof: Proof<MNT6_753>,
-        // The path to where the `verifying_keys` folder is stored in.
-        path: &Path,
```
paberr, line 40: Pass key as parameter instead.

nano-zkp/src/nano_zkp/prove.rs
```diff
@@ -1067,9 +1068,10 @@ impl NanoZKP {
         // Optionally verify the proof.
         if debug_mode {
             // Load the proving key from file.
-            let mut file = File::open(verifying_keys.join("merger_wrapper.bin"))?;
```
paberr, line 1070: Pass key as parameter instead.

nano-zkp/compile-zkp-setup/src/lib.rs
ii-cruz: took it away

lib/src/client.rs
```diff
@@ -33,13 +36,17 @@ use nimiq_validator::validator::ValidatorProxy as AbstractValidatorProxy;
 use nimiq_validator_network::network_impl::ValidatorNetworkImpl;
 #[cfg(feature = "wallet")]
 use nimiq_wallet::WalletStore;
+#[cfg(feature = "zkp-prover")]
+use nimiq_zkp::ZKP_VERIFYING_KEY;
+#[cfg(feature = "zkp-prover")]
+use nimiq_zkp_circuits::setup::{all_files_created, load_verifying_key_from_file, setup};
+#[cfg(feature = "zkp-prover")]
+use nimiq_zkp_circuits::{setup::KEYS_PATH, DEVELOPMENT_SEED};
```
paberr: Can be merged in one

lib/src/client.rs
```diff
@@ -131,30 +138,27 @@ impl ClientInner {
...
-        {
-            // If the prover node is disabled, we do not generate keys
-            // but inform the user of having set a wrong location for the verifying keys.
-            if !config.zkp.prover_active {
-                log::error!("Missing ZKP verification keys. Make sure to set the correct `setup_keys_path` inside the config.");
-                return Err(Error::config_error("Missing ZKP verification keys"));
+        #[cfg(feature = "zkp-prover")]
+        // If the Prover is active for devnet we need to ensure that the proving keys are present.
+        if config.network_id == NetworkId::DevAlbatross && config.zkp.prover_active {
+            let zkp_keys_path = PathBuf::from(KEYS_PATH);
```
paberr:
```suggestion
```

lib/src/client.rs
```diff
@@ -131,30 +138,27 @@ impl ClientInner {
...
+            if !all_files_created(&config.zkp.proving_keys_path, config.zkp.prover_active) {
+                log::info!("Setting up zero-knowledge prover keys for devnet.");
+                log::info!("This task only needs to be run once and might take about an hour.");
+                log::info!(
+                    "Alternatively, you can place the keys in this folder: {:?}.",
+                    KEYS_PATH
```
paberr:
```suggestion
                    config.zkp.proving_keys_path
```

lib/src/client.rs
```diff
@@ -131,30 +138,27 @@ impl ClientInner {
...
+                );
+                setup(
+                    ChaCha20Rng::from_seed(DEVELOPMENT_SEED),
+                    &zkp_keys_path,
+                    config.zkp.prover_active,
+                )?;
+                log::info!("Setting the verification key.");
+                let vk = load_verifying_key_from_file(&config.zkp.proving_keys_path)?;
+                assert_eq!(vk, *ZKP_VERIFYING_KEY, "Verifying keys don't match. The build verifying keys don't match the newly generated ones.");
```
paberr:
```suggestion
                assert_eq!(vk, *ZKP_VERIFYING_KEY, "Verifying keys don't match. The built-in verifying keys don't match the newly generated ones.");
```

lib/src/client.rs
paberr:
```suggestion
use nimiq_zkp_circuits::{setup::{all_files_created, load_verifying_key_from_file, setup}, DEVELOPMENT_SEED};
```

lib/src/client.rs
```diff
@@ -260,7 +264,8 @@ impl ClientInner {
                     config.zkp.prover_active,
                     #[cfg(feature = "zkp-prover")]
                     None,
-                    config.zkp.setup_keys_path,
```
paberr, line 263: Maybe we should use the opportunity to get rid of the conditional arguments.
Instead, let's create two functions with different names.

zkp-component/src/zkp_component.rs
```diff
@@ -127,7 +127,7 @@ impl<N: Network> ZKPComponent<N> {
         executor: impl TaskExecutor + Send + 'static,
         #[cfg(feature = "zkp-prover")] is_prover_active: bool,
         #[cfg(feature = "zkp-prover")] prover_path: Option<PathBuf>,
-        keys_path: PathBuf,
+        #[cfg(feature = "zkp-prover")] proving_keys_path: PathBuf,
```
paberr: See suggestion in client

zkp/build.rs
```diff
@@ -0,0 +1,16 @@
+use std::{env, path::Path};
+
+use nimiq_compile_zkp_setup::verifying_keys_setup;
+use nimiq_zkp_primitives::NanoZKPError;
+
```
paberr, line 6: I still think we can just put the function from `compile-zkp-setup` here.
To test this, maybe copy the function into the `build.rs` first and see whether it works.
Only remove the additional crate afterwards.

zkp/src/verifying_key.rs
```diff
@@ -0,0 +1,17 @@
...
+use ark_serialize::CanonicalDeserialize;
+use lazy_static::lazy_static;
+
+lazy_static! {
+    pub static ref ZKP_VERIFYING_KEY: VerifyingKey<MNT6<Parameters>> = {
+        let serialized = include_bytes!(concat!(env!("OUT_DIR"), "/verifying_keys.data"));
+
+        let mut serialized_cursor = Cursor::new(serialized);
+        VerifyingKey::deserialize_unchecked(&mut serialized_cursor)
+            .expect("Missing the verifying key. Please rebuild the client.")
```
paberr:
```suggestion
            .expect("Invalid verifying key. Please rebuild the client.")
```

zkp/src/verifying_key.rs
```diff
@@ -0,0 +1,17 @@
...
+
+use ark_ec::mnt6::MNT6;
+use ark_groth16::VerifyingKey;
+use ark_mnt6_753::Parameters;
+use ark_serialize::CanonicalDeserialize;
+use lazy_static::lazy_static;
+
+lazy_static! {
+    pub static ref ZKP_VERIFYING_KEY: VerifyingKey<MNT6<Parameters>> = {
+        let serialized = include_bytes!(concat!(env!("OUT_DIR"), "/verifying_keys.data"));
```
paberr: Perhaps we should rename the file to `verifying_key.data` (instead of plural variant)

paberr (commented):
A few more details 😅

jsdanielh:
There is one commit that only says `todo`. Could you maybe try to squash with the previous one?

PR #1328, opened 2023-02-07
Rework web client into a struct that can be accessed by JS

sisou:
Change the web client to be a struct, which is represented as a class in Javascript, exposing it's methods to be called from JS. For now it only implements subscription methods, which then use separately exposed-from-JS global listeners to communicate the events back to JS. But it can now be easily extended to provide active methods that return data to JS, or send signed transactions to the network, etc.

Interestingly, since the created `Client` is stored in Rust's heap (or stack, I can never remember which is which) when it's made available to JS with `wasm_bindgen`, it is never dereferenced and lives forever (technically only until `.free()` is called on it from JS-side, which is never). So we do not need the statistics-interval-hack anymore to keep the client in-scope =).

> **Note for reviewers:** It is recommended to hide whitespace changes when reviewing, as most existing code in `lib.rs` has simply been indented by one level.

## Pull request checklist

### DIFF
--- a/web-client/Cargo.toml
+++ b/web-client/Cargo.toml
@@ -27,6 +27,12 @@ log = { package = "tracing", version = "0.1", features = ["log"] }
 wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4"
 
+beserial = { path = "../beserial", features = ["derive"] }
+nimiq-blockchain-interface = { path = "../blockchain-interface" }
+nimiq-blockchain-proxy = { path = "../blockchain-proxy", default-features = false }
+nimiq-consensus = { path = "../consensus", default-features = false }
+nimiq-hash = { path = "../hash" }
+nimiq-network-interface = { path = "../network-interface" }
 nimiq-network-libp2p = { path = "../network-libp2p" }
 
 [dependencies.nimiq]
--- a/web-client/src/lib.rs
+++ b/web-client/src/lib.rs
@@ -16,81 +16,176 @@ pub use nimiq::{
     extras::{panic::initialize_panic_reporting, web_logging::initialize_web_logging},
 };
 
+use beserial::Serialize;
+use nimiq_blockchain_interface::{AbstractBlockchain, BlockchainEvent};
+use nimiq_blockchain_proxy::BlockchainProxy;
+use nimiq_consensus::ConsensusEvent;
+use nimiq_hash::Blake2bHash;
+use nimiq_network_interface::network::{Network, NetworkEvent};
 use nimiq_network_libp2p::Multiaddr;
 
-async fn light_client() {
-    let log_settings = LogSettings {
-        level: Some(LevelFilter::DEBUG),
-        ..Default::default()
-    };
+#[wasm_bindgen]
+pub struct WebClient {
+    #[wasm_bindgen(skip)]
+    pub inner: Client,
+}
 
-    // Initialize logging with config values.
-    initialize_web_logging(Some(&log_settings)).expect("Web logging initialization failed");
+#[wasm_bindgen]
+impl WebClient {
+    pub async fn create() -> WebClient {
+        let log_settings = LogSettings {
+            level: Some(LevelFilter::DEBUG),
+            ..Default::default()
+        };
 
-    // Initialize panic hook.
-    initialize_panic_reporting();
+        // Initialize logging with config values.
+        initialize_web_logging(Some(&log_settings)).expect("Web logging initialization failed");
 
-    // Create config builder.
-    let mut builder = ClientConfig::builder();
+        // Initialize panic hook.
+        initialize_panic_reporting();
 
-    // Finalize config.
-    let mut config = builder
-        .volatile()
-        .light()
-        .build()
-        .expect("Build configuration failed");
+        // Create config builder.
+        let mut builder = ClientConfig::builder();
 
-    let seed = Seed {
-        address: Multiaddr::from_str("/dns4/seed1.v2.nimiq-testnet.com/tcp/8443/ws").unwrap(),
-    };
-    config.network.seeds = vec![seed];
+        // Finalize config.
+        let mut config = builder
+            .volatile()
+            .light()
+            .build()
+            .expect("Build configuration failed");
 
-    log::debug!("Final configuration: {:#?}", config);
+        let seed = Seed {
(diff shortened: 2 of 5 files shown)


web-client/Cargo.toml
```diff
@@ -21,12 +21,18 @@ maintenance = { status = "experimental" }
...
 log = { package = "tracing", version = "0.1", features = ["log"] }
 wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4"
 
+nimiq-hash = { path = "../hash" }
+nimiq-blockchain-interface = { path = "../blockchain-interface" }
+nimiq-blockchain-proxy = { path = "../blockchain-proxy", default-features = false }
+nimiq-consensus = { path = "../consensus", default-features = false }
+nimiq-network-interface = { path = "../network-interface" }
 nimiq-network-libp2p = { path = "../network-libp2p" }
```
jsdanielh: We try to keep them alphabetically ordered:
```suggestion
beserial = { path = "../beserial", features = ["derive"] }
nimiq-blockchain-interface = { path = "../blockchain-interface" }
nimiq-blockchain-proxy = { path = "../blockchain-proxy", default-features = false }
nimiq-consensus = { path = "../consensus", default-features = false }
nimiq-hash = { path = "../hash" }
nimiq-network-interface = { path = "../network-interface" }
nimiq-network-libp2p = { path = "../network-libp2p" }
```

web-client/Cargo.toml
```diff
@@ -21,12 +21,18 @@ maintenance = { status = "experimental" }
 crate-type = ["cdylib"]
 
 [dependencies]
+beserial = { path = "../beserial", features = ["derive"] }
```
jsdanielh: This is actually an internal crate, we regularly try to list them apart (see suggestion below)

web-client/src/lib.rs
```diff
@@ -15,101 +16,211 @@ pub use nimiq::{
     error::Error,
     extras::{panic::initialize_panic_reporting, web_logging::initialize_web_logging},
 };
+use nimiq_hash::Blake2bHash;
 
+use nimiq_blockchain_interface::{AbstractBlockchain, BlockchainEvent};
+use nimiq_blockchain_proxy::BlockchainProxy;
+use nimiq_consensus::ConsensusEvent;
+use nimiq_network_interface::network::{Network, NetworkEvent};
 use nimiq_network_libp2p::Multiaddr;
```
jsdanielh:
```suggestion

use beserial::Serialize;
use nimiq_blockchain_interface::{AbstractBlockchain, BlockchainEvent};
use nimiq_blockchain_proxy::BlockchainProxy;
use nimiq_consensus::ConsensusEvent;
use nimiq_hash::Blake2bHash;
use nimiq_network_interface::network::{Network, NetworkEvent};
use nimiq_network_libp2p::Multiaddr;
```

web-client/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 use std::str::FromStr;
 
+use beserial::Serialize;
```
jsdanielh: `beserial` is an internal crate, we try to list it apart, see suggestion below

web-client/src/lib.rs
```diff
@@ -15,101 +16,211 @@ pub use nimiq::{
...
+        // Create client from config.
+        log::info!("Initializing light client");
+        let mut client: Client = Client::from_config(
+            config,
+            Box::new(|fut| {
+                spawn_local(fut);
+            }),
+        )
+        .await
+        .expect("Client initialization failed ");
```
jsdanielh:
```suggestion
        .expect("Client initialization failed");
```

web-client/src/lib.rs
```diff
@@ -15,101 +16,211 @@ pub use nimiq::{
...
+            .light()
+            .build()
+            .expect("Build configuration failed");
+
+        let seed = Seed {
+            address: Multiaddr::from_str("/dns4/seed1.v2.nimiq-testnet.com/tcp/8443/ws").unwrap(),
+        };
+        config.network.seeds = vec![seed];
+
+        log::debug!("Final configuration: {:#?}", config);
```
jsdanielh: We should change this to `tracing style`:
```suggestion
        log::debug!(?config, "Final configuration");
```

web-client/src/lib.rs
```diff
@@ -15,101 +16,211 @@ pub use nimiq::{
...
+                    );
+                }
+                None => {
+                    break;
+                }
+            }
+        }
+    }
+
+    pub async fn subscribe_peers(&self) {
```
jsdanielh, line 171: I'd suggest to add rustdoc documentation

web-client/src/lib.rs
```diff
@@ -15,101 +16,211 @@ pub use nimiq::{
...
+                None => {
+                    break;
+                }
+            }
+        }
     }
 
-    // Run periodically
-    let interval = IntervalStream::new(statistics_interval as u32 * 1000);
+    pub async fn subscribe_blocks(&self) {
```
jsdanielh, line 124: I'd suggest to add rustdoc documentation

web-client/src/lib.rs
```diff
@@ -15,101 +16,211 @@ pub use nimiq::{
...
-    // Create the "monitor" future which never completes to keep the client alive.
-    // This closure is executed after the client has been initialized.
-    // TODO Get rid of this. Make the Client a future/stream instead.
-    let mut statistics_interval = LogSettings::default_statistics_interval();
-    let mut show_statistics = true;
-    if statistics_interval == 0 {
-        statistics_interval = 10;
-        show_statistics = false;
+#[wasm_bindgen]
+pub struct WebClient {
```
jsdanielh, line 39: I'd suggest to add rustdoc documentation

web-client/src/lib.rs
```diff
@@ -15,101 +16,211 @@ pub use nimiq::{
...
+    #[wasm_bindgen(skip)]
+    pub inner: Client,
+}
+
+#[wasm_bindgen]
+impl WebClient {
+    pub async fn launch() -> WebClient {
```
jsdanielh: I'd suggest to add rustdoc documentation

sisou:
Thank you, @jsdanielh! I overlooked the dependency ordering. Fixed that, plus the other comments. I also renamed `WebClient.launch()` to `WebClient.create()`, as I think it better represents what's happening and discourages calling it twice (which would create two competing WebClient instances).

jsdanielh (approved):

PR #1342, opened 2023-02-15
Use the validator services flag

viquezclaudio:
If a node is configured as a validator, then it is interested in other validator nodes (regardless of the services that are provided by those other validators) However, for the syncing process, we only care about those peers that provide the required services for us

### DIFF
--- a/lib/src/client.rs
+++ b/lib/src/client.rs
@@ -178,8 +178,14 @@ impl ClientInner {
             identity_keypair.public().to_peer_id().to_base58()
         );
 
-        let (provided_services, required_services) =
+        let (mut provided_services, required_services) =
             generate_service_flags(config.consensus.sync_mode);
+
+        // We update the services flags depending on our validator configuration
+        #[cfg(feature = "validator")]
+        if config.validator.is_some() {
+            provided_services |= Services::VALIDATOR;
+        }
 
         // Generate my peer contact from identity keypair and my provided services
         // Filter out unspecified IP addresses since those are not addresses suitable
--- a/network-libp2p/src/discovery/peer_contacts.rs
+++ b/network-libp2p/src/discovery/peer_contacts.rs
@@ -60,7 +60,7 @@ bitflags! {
         ///
         const TRANSACTION_INDEX = 1 << 6;
 
-        /// This node accepts validator related messages.
+        /// This node is configured as a validator, so it is interested for other validator nodes.
         ///
         const VALIDATOR = 1 << 7;
     }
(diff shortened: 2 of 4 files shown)


lib/src/client.rs
```diff
@@ -178,9 +178,15 @@ impl ClientInner {
             identity_keypair.public().to_peer_id().to_base58()
         );
 
-        let (provided_services, required_services) =
+        let (mut provided_services, required_services) =
             generate_service_flags(config.consensus.sync_mode);
 
+        //We update the services flags depending upong our validator configuration
```
jsdanielh:
```suggestion
        // We update the services flags depending on our validator configuration
```

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -60,7 +60,7 @@ bitflags! {
         ///
         const TRANSACTION_INDEX = 1 << 6;
 
-        /// This node accepts validator related messages.
+        /// This node is configured as a validator, so it is interested in other validator nodes.
```
```suggestion
        /// This node is configured as a validator, so it is interested for other validator nodes.
```

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -336,9 +336,25 @@ impl PeerContactBook {
     /// Inserts a peer contact or update an existing using the service filtering.
     /// If the filter matches the services provided by the contact, it is added.
     /// Otherwise it is ignored.
+    /// The services_filter argument to this function contain the services that are required.
```
```suggestion
    /// The services_filter argument to this function contains the services that are required.
```

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -336,9 +336,25 @@ impl PeerContactBook {
...
     pub fn insert_filtered(&mut self, contact: SignedPeerContact, services_filter: Services) {
         let info = PeerContactInfo::from(contact);
-        if info.matches(services_filter) {
+
+        if services_filter.contains(Services::VALIDATOR)
+            && info.services().contains(Services::VALIDATOR)
+        {
+            // If Im configured as a validator, and the peer is also validator, then that peer is interesting to me,
```
```suggestion
            // If I'm configured as a validator, and the peer is also a validator, then that peer is interesting to me,
```

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -336,9 +336,25 @@ impl PeerContactBook {
...
+            // regardles of the services that are provided by that peer.
```
```suggestion
            // regardless of the services that are provided by that peer.
```

jsdanielh (approved):

PR #1344, opened 2023-02-16
Add a way to access peer information

### LINKED ISSUE #357: Access to network peer information (IP, type, etc.)
To properly display the network map in the Wallet, and for general informative purposes about node connectivity, it is required to be able to access information about network peers, such as their **IP**, their **node type**, their **connected protocol**.

Currently, the `Peer` struct has only an ID and a reference to the underlying stream from `libp2p`.
https://github.com/nimiq/core-rs-albatross/blob/albatross/network-libp2p/src/message/peer.rs#L24-L31

jsdanielh:
- Add a way to query peer information from the network. For this the following changes were introduced:
  - Move the `Services` bitflags to the `network-interface` crate.
  - Add a `PeerInfo` struct to the `network-interface` crate.
  - Change the `PeerJoined` network event to also dispatch the `PeerInfo` of the peer joined.
  - Change the `peer_listener` function to also receive more arguments.
  - Implement the changes in the `network-interface` in both of the network implementations we support (`libp2p` and `mock`).
  This closes #357.
- Improve `rustdoc` documentation for the `network-interface` crate.

## Pull request checklist


web-client/src/lib.rs
```diff
@@ -174,11 +177,36 @@ impl WebClient {
...
                 }
                 Some(Ok(NetworkEvent::PeerLeft(peer_id))) => {
-                    peer_listener("left", peer_id.to_string(), network.peer_count());
+                    peer_listener(
+                        "left",
+                        peer_id.to_string(),
+                        String::from(""),
+                        "",
+                        network.peer_count(),
+                    );
```
jsdanielh: @sisou Is there a better way to do this? Now the _join_ event carries more information than the _left_ event. I thought on separating the listener for _join_ and _left_ events but not sure if that's what we want.

sisou: After thinking about it, my suggestion would be to not pass the three peer info parts individually, but to pass a `PeerInfo` object (either as JSON, or as a bridged struct like in my other branch). Then everything but the peer ID can be an optional field.

sisou (approved):
Looks good on the wasm-bindgen side!

viquezclaudio (approved):

PR #1349, opened 2023-02-17
Wasm-JS interactivity, first steps

sisou:

This PR implements proxying structs in the web-client crate that allow JS to create and manipulate native rust structs. It currently implements everything needed to sign and send basic transactions. This includes
- PrivateKey
- PublicKey
- KeyPair
- Signature
- SignatureProof
- Address
- Transaction

It tries to stay as close to the Nimiq 1.0 API for these classes as possible. 

The JS file `module.js` makes a global function available that can be used to test the functionality:
```ts
async function sendTransaction(private_key: string, recipient: string, value: number): Promise<string> {}
```

_Note that light clients don't maintain an accounts tree, so they cannot verify if the sender has enough balance. Also, light clients cannot yet get feedback if the broadcast transaction actually gets accepted by other nodes, nor if it is included in a block._

The PR also streamlines the previously merged `PeerInfo` struct and introduces a build-time environment variable that enables automatic freeing of wasm memory when JS garbage-collects its pointers.

## Pull request checklist


web-client/src/transaction.rs
```diff
@@ -0,0 +1,176 @@
...
+                recipient.native_ref().clone(),
+                Coin::try_from(value)?,
+                Coin::try_from(fee)?,
+                validity_start_height,
+                to_network_id(network_id)?,
+            ),
+        ))
+    }
+
+    /// Computes the transaction's hash, which is is used as its unique identifier on the blockchain.
```
jsdanielh, line 50:
```suggestion
    /// Computes the transaction's hash, which is used as its unique identifier on the blockchain.
```

jsdanielh (approved):

PR #1350, opened 2023-02-19
Wasm-JS interactivity, part two: staking roundtrip

sisou:

Makes it possible to easily create and sign all transactions necessary to start staking, add stake, update delegation, and unstake.

- Add creation methods for different transaction types
  - Introduces a `TransactionBuilder` class that is instantiated from the client and receives the client's `network_id` and a blockchain proxy (to access the current head block number), allowing users to omit these when creating transactions.
  - Adds a `tx.sign(keyPair)` method that automatically signs the transaction with the correct signature(s) format if possible.
  - Adds a constructor function to `Transaction` to allow the creation of arbitrary transactions which the transaction builder does not yet cover.
- Adapt Client and Configuration to feel closer to the core-js client API.
- Fix rustdocs

## Pull request checklist

### DIFF
--- a/web-client/src/transaction.rs
+++ b/web-client/src/transaction.rs
@@ -27,7 +27,8 @@ impl Transaction {
     /// The returned transaction is not yet signed. You can sign it e.g. with `KeyPair.signTransaction`.
     ///
     /// Throws when the numbers given for value and fee do not fit within a u64 or the networkId is unknown.
-    pub fn basic(
+    #[wasm_bindgen(js_name = newBasicTransaction)]
+    pub fn new_basic_transaction(
         sender: &Address,
         recipient: &Address,
         value: u64,
@@ -47,7 +48,7 @@ impl Transaction {
         ))
     }
 
-    /// Computes the transaction's hash, which is is used as its unique identifier on the blockchain.
+    /// Computes the transaction's hash, which is used as its unique identifier on the blockchain.
     pub fn hash(&self) -> String {
         let hash: Blake2bHash = self.inner.hash();
         // TODO: Return an instance of Hash
--- a/web-client/src/transaction.rs
+++ b/web-client/src/transaction.rs
@@ -2,9 +2,12 @@ use wasm_bindgen::prelude::*;
 
 use beserial::Serialize;
 use nimiq_hash::{Blake2bHash, Hash};
-use nimiq_primitives::coin::Coin;
+use nimiq_primitives::{account::AccountType, coin::Coin};
+use nimiq_transaction::TransactionFlags;
+use nimiq_transaction_builder::TransactionProofBuilder;
 
 use crate::address::Address;
+use crate::key_pair::KeyPair;
 use crate::utils::{from_network_id, to_network_id};
 
 /// Transactions describe a transfer of value, usually from the sender to the recipient.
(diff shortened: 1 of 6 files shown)


web-client/src/transaction.rs
```diff
@@ -1,10 +1,13 @@
+use nimiq_transaction::TransactionFlags;
```
jsdanielh: I'd move this `use` to the second block of imports where nimiq crates are listed

web-client/src/transaction.rs
```diff
@@ -14,40 +17,168 @@ use crate::utils::{from_network_id, to_network_id};
...
+    /// transaction is created.
+    ///
+    /// ### Contract creation transactions
+    /// To create a new vesting or HTLC contract, set `flags` to `0b1` and specify the contract
+    /// type as the `recipient_type`: `1` for vesting, `2` for HTLC. The `data` bytes must have
+    /// the correct format of contract creation data for the respective contract type.
     ///
-    /// The returned transaction is not yet signed. You can sign it e.g. with `KeyPair.signTransaction`.
+    /// ### Signalling transactions
+    /// To interact with the staking contract, signalling transaction are often used that do not
```
jsdanielh:
```suggestion
    /// To interact with the staking contract, signalling transaction are often used to not
```

web-client/src/transaction.rs
```diff
@@ -14,40 +17,168 @@ use crate::utils::{from_network_id, to_network_id};
...
+    /// transfer any value, but instead simply _signal_ a state change, such as changing one's stake
```
jsdanielh:
```suggestion
    /// transfer any value, but to simply _signal_ a state change instead, such as changing one's stake
```

web-client/src/transaction_builder.rs
```diff
@@ -0,0 +1,264 @@
+use nimiq_blockchain_interface::AbstractBlockchain;
+use wasm_bindgen::prelude::*;
+
+use nimiq_blockchain_proxy::BlockchainProxy;
+use nimiq_primitives::{account::AccountType, coin::Coin, policy::Policy};
```
jsdanielh:
```suggestion
use wasm_bindgen::prelude::*;

use nimiq_blockchain_interface::AbstractBlockchain;
use nimiq_blockchain_proxy::BlockchainProxy;
use nimiq_primitives::{account::AccountType, coin::Coin, policy::Policy};
```

web-client/src/transaction_builder.rs
```diff
@@ -0,0 +1,264 @@
...
+use nimiq_transaction_builder::Recipient;
+
+use crate::address::Address;
+use crate::transaction::Transaction;
+use crate::utils::to_network_id;
+
+#[wasm_bindgen]
```
jsdanielh, line 14: I think we should add `rustdoc` documentation for this `struct`

jsdanielh (commented):
Looks good. Just some minor comments

viquezclaudio:
LGTM

sisou:
Thanks @jsdanielh, I missed those import orderings! I also applied all your other suggestions :+1:

jsdanielh (approved):

jsdanielh:
@sisou I will be rebasing it to get it merged

PR #1351, opened 2023-02-20
Start state live sync as complete

paberr:
When (re)starting a set of validators with only full nodes, they would not be able to reach consensus.
That is because the state live sync was always initialized as incomplete and consensus thus required passing a macro block.
However, since all nodes are waiting for consensus, no new blocks could be produced.

Now, the state live sync is initialized as complete (if the blockchain state is complete) and only reset into syncing if required (i.e., the blockchain state changed to incomplete).

## Pull request checklist


jsdanielh (approved):
LGTM

Issue #1339, opened 2023-02-13
WASM Client Features


jsdanielh:
- [x] A way to pass configuration to it on launch
  - #1341 
- [x] Proper client library/bundle
- [x] A way to access the actual hostname or IP addresses of connected peers, for our world map.
  - #357 
  - #1344
- [x] Access to PrivateKey, PublicKey, Commitments, ~~SerialBuffer~~, etc. Sign a transaction with my PrivateKey
  - #1349
  - #1350
  - Event listeners/callbacks
    - #1363
    - #1410 
  - HTLC signing support
    - HTLCs are technically supported. They can be created, just have to be signed manually, as the `sign()` method does not support it.
  - Multisig primitives
    - Documented in the last check box.
- [x] Method to tell client to reconnect after all peers lost, or it tries reconnecting itself
  - #1357 
- [x] Query tx history by address
  - #1346
  - #1356 
  - #1385 
  - #1386
  - #1401 
- [x] Query balances for addresses
  - #1408 
  - #1462 
- [x] Subscribe to tx events/logs for addresses
  - #1437 
- [x] Compatibility with 1.0 client
  - #1418 
  - Use 1.0 arguments for `getTransactionsByAddress`: `(address, sinceBlockHeight, knownTransactionDetails, limit)`
    - #1430
  - #1425
    - also adds `getTransactionReceiptsByAddress(address, limit)`
  - Accept `Transaction | PlainTransaction | string` for `sendTransaction`
  - `getTransaction(hash, blockHash, blockHeight)`
  - `getTransactionReceipt(hash)`
  - `getTransactionReceiptsByHashes(hashes)`
- [x] multisig support
  - Will not be addressed in this issue and tracking of this will continue in #1578.

sisou:
So I just tested disconnection/reconnection behavior. Detection of network loss takes around 10-12 seconds (all peers leave and consensus is marked as "lost" when the number of peers drop under the configured minimum). However, when network is reconnected, the client does not start reconnecting by itself. So we need a way to tell it to start connecting again. Or we need to kill it and start a new client, but the first option would be better.

sisou:
As for the export of further primitives and structs, we will run into various problems with Rust things that wasm-bindgen does not (yet) support:
- It cannot generate bindings for trait implementations, so e.g. the `from_hex` method of `PrivateKey` is not supported
[...]

PR #1357, opened 2023-02-23
Implement listener for offline/online events in the web-client, add fixes and documentation

jsdanielh:
- Add `rustdoc` documentation for the connection pool behaviour in the `network-libp2p` crate.
- Implement listener for offline/online events in the web-client
  - Implement listener for `Window` offline and online events in the `web-client`.
  - Add a function to the `network-libp2p` crate to allow a client to instruct the network to try to start new peer connections. This is used in the `web-client` for the `offline` event such that the client doesn't issue new connections when it's offline.
  - Change the `start_connecting` functionality in the connection pool behaviour such that it resets the peers marked as `down` if     the function is called a second time and if no dial attempts or connected peers are found. This is used in the `web-client` to `restart` network connection attempts when the `online` event is detected.
- Fix issue in the connection pool behaviour when closing a connection for a peer that doesn't have the `ip` protocol in its
  `Multiaddr`. The issue was that the code wasn't marking the connection as closed in the behaviour connection state and it wasn't matching the state on the rest of the network.
- Fix the procedure to close a connection in the connection pool behaviour by also marking an `address` as down when the connection is closed.

Relates to #1339.

## Pull request checklist

### DIFF
--- a/web-client/src/lib.rs
+++ b/web-client/src/lib.rs
@@ -2,10 +2,9 @@ use std::{cell::RefCell, collections::HashMap, rc::Rc, str::FromStr};
 
 use futures::StreamExt;
 use js_sys::{Array, Uint8Array};
+use log::level_filters::LevelFilter;
 use wasm_bindgen::prelude::*;
 use wasm_bindgen_futures::spawn_local;
-
-use log::level_filters::LevelFilter;
 
 pub use nimiq::{
     client::Consensus,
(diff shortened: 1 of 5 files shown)


web-client/src/lib.rs
```diff
@@ -187,6 +187,56 @@ impl Client {
         }
     }
 
+    /// Adds event handlers for the window `online` and `offline` events
+    ///
+    /// On `online` event, the network is set to start re-connecting to the seeds
+    /// or peers.
+    /// On `offline` event, the network is set to disconnect from all the previously
+    /// connected peers and to stop trying to connect to any possible peer.
+    pub fn add_offline_online_event_handlers(&self, window: Window) -> Result<(), JsValue> {
```
sisou: It is possible to get a reference to the `window` from `web_sys::window().unwrap()`, not requiring this external trigger method. Then this setup method can be called internally from the static `create()` method.

sisou (changes requested):
Works beautifully! Disconnects all peers @ 'offline' and reconnects directly @ 'online'. :+1:

But I would change the setup of the listeners to not require an external trigger from JS userland, instead attaching them on client creation internally.

sisou (approved):

PR #1363, opened 2023-02-27
WASM Client: Convert subscriptions to listeners

sisou:

- Add various block getter methods that are also available on the 1.0 client
- Convert subscriptions into event listeners with the same API as the 1.0 client
  - Remove the need for global imports from JS to emit subscription events
  - Instead take listener closures from JS and store them in a list
  - Start native rust event streams only once and emit events to the stored listeners
  - When adding a listener, a numeric handle is returned with which this listener can be removed again as well
  - Listener methods are typed correctly with a custom external type with a TS annotation
  - Remove statistic subscription, as the three data points it emitted are emitted by the three other available event sources

## Pull request checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4531,7 +4531,7 @@ version = "0.1.0"
 dependencies = [
  "beserial",
  "futures-util",
- "gloo-timers",
+ "js-sys",
  "nimiq-blockchain-interface",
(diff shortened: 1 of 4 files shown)


jsdanielh (approved):
LGTM

PR #1385, opened 2023-03-03
Array of plain JS transaction objects from getTransactionsByAddress

sisou:

- Adds fully TS-typed plain Transaction and TransactionDetails structs equalling those in the 1.0 client
- Updates `client.getTransactionsByAddress` to return serialized `PlainTransactionDetails[]`, because wasm-bindgen cannot return `Vec<T>` where `T` is a wasm-bindgen annotated struct
- Workaround a serde shortcoming by manually serializing `PlainTransactionDetails` structs to produce JS Objects

## Pull request checklist

### DIFF
--- a/web-client/src/transaction.rs
+++ b/web-client/src/transaction.rs
@@ -440,19 +440,21 @@ impl PlainTransaction {
     }
 }
 
-enum TransactionState {
-    _New,
-    _Pending,
+#[derive(serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub enum TransactionState {
+    New,
+    Pending,
     Included,
     Confirmed,
-    _Invalidated,
-    _Expired,
+    Invalidated,
+    Expired,
 }
 
 pub struct PlainTransactionDetails {
     transaction: PlainTransaction,
 
-    pub state: String,
+    pub state: TransactionState,
     pub execution_result: Option<bool>,
     pub block_height: Option<u32>,
     pub confirmations: Option<u32>,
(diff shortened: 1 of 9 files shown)


viquezclaudio (approved):

web-client/src/transaction.rs
```diff
@@ -320,3 +335,229 @@ impl Transaction {
         &self.inner
     }
 }
+
+#[derive(serde::Serialize, serde::Deserialize)]
```
jsdanielh: I'd suggest to add `rustdoc` documentation

web-client/src/transaction.rs
```diff
@@ -320,3 +335,229 @@ impl Transaction {
...
+pub struct Raw {
+    raw: String,
+}
+
+#[derive(serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "camelCase")]
```
jsdanielh, line 353: I'd suggest to add `rustdoc` documentation (and to the members)

web-client/src/transaction.rs
```diff
@@ -320,3 +335,229 @@ impl Transaction {
...
+            AccountType::Staking => "staking".to_string(),
+            AccountType::StakingValidator => "validator".to_string(),
+            AccountType::StakingValidatorsStaker => "staker".to_string(),
+            AccountType::StakingStaker => "staker".to_string(),
+        }
+    }
+}
+
+#[derive(serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "lowercase")]
```
jsdanielh, line 437: I'd suggest to add `rustdoc` documentation (and to the members)

web-client/src/transaction.rs
```diff
@@ -320,3 +335,229 @@ impl Transaction {
...
+pub enum TransactionState {
+    New,
+    Pending,
+    Included,
+    Confirmed,
+    Invalidated,
+    Expired,
+}
+
```
jsdanielh, line 457: I'd suggest to add `rustdoc` documentation (and to the members)

web-client/src/transaction.rs
```diff
@@ -320,3 +335,229 @@ impl Transaction {
...
+  flags: number,
+  data: { raw: string },
+  proof: { raw: string },
+  size: number,
+  valid: boolean,
+};
+"#;
+
+impl PlainTransaction {
+    pub fn from_transaction(tx: &nimiq_transaction::Transaction) -> Self {
```
jsdanielh, line 409: I'd suggest to add `rustdoc` documentation

web-client/src/transaction.rs
```diff
@@ -320,3 +335,229 @@ impl Transaction {
         &self.inner
     }
 }
+
+#[derive(serde::Serialize, serde::Deserialize)]
+pub struct Raw {
+    raw: String,
+}
```
jsdanielh: Maybe:
```suggestion
pub struct Raw(String);
```
?

jsdanielh (commented):
LGTM. Only some small comments

web-client/src/transaction.rs
sisou: The output format is supposed to be `{ raw: string }` in JS, so I think what I have is correct. This is the format used by the 1.0 library.

viquezclaudio (approved):

jsdanielh (approved):

PR #1398, opened 2023-03-08
Enable the network to ban peers

jsdanielh:
- Use a proper reason for closing a peer connection in several places of the codebase.
- Use a reason as well when disconnecting from every peer.
- Make the web client use the `disconnect` function from the network instead of iterating over all peers.
- Add basic infrastructure for banning a peer.
- Refined network interface close reasons and added a special close reason for banning a peer.
- Change the procedure to close a peer connection when instructed by upper layers. Instead of directly instructing the `swarm` to close the connection, now the network signals the connection pool to close it emitting an error and making use of the `CloseReason` that is sent by the callers.
- Add an enumeration for close connection reasons or errors that can happen in the connection pool handler.
- Properly close connections in the connection pool behaviour by notifying the corresponding connection handler with a reason such that it can then close the connection with a reason.
- Rename discovery and connection pool handler errors to better identify them outside the handler implementation.

## Pull request checklist

### DIFF
--- a/network-libp2p/src/connection_pool/behaviour.rs
+++ b/network-libp2p/src/connection_pool/behaviour.rs
@@ -23,7 +23,7 @@ use rand::thread_rng;
 use wasm_timer::Interval;
 
 use nimiq_macros::store_waker;
-use nimiq_network_interface::peer_info::Services;
+use nimiq_network_interface::{network::CloseReason, peer_info::Services};
 
 use crate::discovery::peer_contacts::PeerContactBook;
 
@@ -337,6 +337,17 @@ impl ConnectionPoolBehaviour {
     /// connection such as in a network outage.
     pub fn stop_connecting(&mut self) {
         self.active = false;
+    }
+
+    /// Closes a peer connection with a reason
+    pub fn close_connection(&mut self, peer_id: PeerId, reason: CloseReason) {
+        self.actions
+            .push_back(NetworkBehaviourAction::NotifyHandler {
+                peer_id,
+                handler: NotifyHandler::Any,
+                event: ConnectionPoolHandlerError::Other(reason),
+            });
+        self.wake();
     }
 
     fn choose_peers_to_dial(&self) -> Vec<PeerId> {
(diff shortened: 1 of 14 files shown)


network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -107,6 +118,24 @@ impl<T: Ord> ConnectionState<T> {
...
+        self.down.remove(&id);
+        self.banned.insert(id);
+    }
+
+    /// Removes a connection ID from the banned set
+    fn unmark_banned(&mut self, id: T) {
+        self.banned.remove(&id);
+    }
+
+    /// Returns wether a connection ID is banned
```
viquezclaudio: whether

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -139,7 +168,10 @@ impl<T: Ord> ConnectionState<T> {
 
     /// Returns wether a specific connection ID can dial another connection ID
     fn can_dial(&self, id: &T) -> bool {
-        !self.dialing.contains(id) && !self.connected.contains(id) && !self.down.contains_key(id)
+        !self.dialing.contains(id)
```
viquezclaudio, line 171: In the comment above: whether instead of "wether"

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -335,10 +364,31 @@ impl ConnectionPoolBehaviour {
...
     /// connection such as in a network outage.
-    pub fn stop_connecting(&mut self) {
+    fn stop_connecting(&mut self) {
         self.active = false;
     }
 
+    /// Closes a peer connection with a reason
+    ///
+    /// This will take actions depending on the close reason. For instance:
+    /// - Malicious peer will ban the connection.
```
viquezclaudio: A malicious peer..

hrxi: > A malicious peer..

That seems misleading or wrong. How about:

The close reason `MaliciousPeer` will cause the peer to be banned.

viquezclaudio (approved):

PR #1280, opened 2023-01-23
Implementing abstract blockchain for light client / creating new tests

paberr:
## What's in this pull request?
- Fix light client by implementing required members on abstract blockchain
- Create new tests for different sync modes
- Fix existing tests where necessary
- Make reward transactions explicit in the macro block (required for state sync since inherents were not really inherent to the *block* only; they required the state)
- Only verify history root if possible
- Move staking contract to 0 address, so that we download it first in state live sync
- Move burn address away from 0 address
- Improve requesting of chunks

## Pull request checklist

- [x] All tests pass. The project builds and runs.
- [x] I have resolved any merge conflicts.
- [x] I have resolved all `clippy` and `rustfmt` warnings.

PR #1401, opened 2023-03-08
Transaction inclusion proofs enhancements 

viquezclaudio:
When generating transaction inclusion proofs there are multiple cases that need to be considered in order to generate a proof that can be verified. From the requester side, transactions can be grouped into: 

1. Transactions that belong to finalized epochs (currently hashes_by_epoch )
2. Transactions that belong to an unfinished epoch, but have a checkpoint block that finalizes them
3. Transactions that belong to the current batch

To prove those transactions:
For 1 we essentially do what we are currently doing: use the election block of that epoch to prove those transactions
For 2 we should use the last_macro_block of the current epoch, and use the current MMR length to generate the proof
For 3 we use the last micro block number of the current batch, and the current MMR length to generate the proof

### DIFF
--- a/consensus/src/consensus/consensus_proxy.rs
+++ b/consensus/src/consensus/consensus_proxy.rs
@@ -95,9 +95,11 @@ impl<N: Network> ConsensusProxy<N> {
 
             match response {
                 Ok(response) => {
+                    log::info!("Obtained response, length {} ", response.receipts.len());
+
                     let blockchain = self.blockchain.read();
 
-                    // Group transaction hashes by the block number that proofs those transactions to reduce the number of requests
+                    // Group transaction hashes by the block number that proves those transactions to reduce the number of requests
                     // There are three categories of block numbers:
                     //  - Finalized epochs: we use the election block number that finalized the respective epoch
                     //  - Finalized batch in the current epoch: We use the latest checkpoint block number
@@ -105,11 +107,8 @@ impl<N: Network> ConsensusProxy<N> {
 
                     // This is the structure where we group transactions by their proving block number
                     let mut hashes_by_block = HashMap::new();
-
                     let election_head_number = blockchain.election_head().block_number();
-
                     let checkpoint_head_number = blockchain.macro_head().block_number();
-
                     let current_head_number = blockchain.head().block_number();
 
                     for (hash, block_number) in response.receipts {
--- a/consensus/src/consensus/consensus_proxy.rs
+++ b/consensus/src/consensus/consensus_proxy.rs
@@ -2,6 +2,7 @@ use std::collections::HashMap;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
+use nimiq_blockchain_interface::AbstractBlockchain;
 use tokio::sync::broadcast::Sender as BroadcastSender;
 use tokio_stream::wrappers::BroadcastStream;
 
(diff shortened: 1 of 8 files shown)


consensus/src/consensus/consensus_proxy.rs
```diff
@@ -94,26 +95,59 @@ impl<N: Network> ConsensusProxy<N> {
...
+
+                    // This is the structure where we group transactions by their proving block number
+                    let mut hashes_by_block = HashMap::new();
+
+                    let election_head_number =
+                        self.blockchain.read().election_head().block_number();
+
+                    let checkpoint_head_number = self.blockchain.read().macro_head().block_number();
+
+                    let current_head_number = self.blockchain.read().head().block_number();
```
sisou: I suggest to use the same `read()` instance for all three blocks to have a consistent state.

consensus/src/consensus/consensus_proxy.rs
```diff
@@ -94,26 +95,59 @@ impl<N: Network> ConsensusProxy<N> {
...
                         if verified_transactions.contains_key(&hash) {
                             continue;
                         }
 
-                        let epoch = Policy::epoch_at(block_number);
-                        hashes_by_epoch.entry(epoch).or_insert(vec![]).push(hash);
+                        if block_number <= election_head_number {
+                            // First Case: Transactions from finalized epochs
+                            hashes_by_block
+                                .entry(Policy::election_block_of(Policy::epoch_at(block_number)))
```
sisou: Isn't there a `election_after_block(height)` method or similar?

consensus/src/consensus/consensus_proxy.rs
```diff
@@ -94,26 +95,59 @@ impl<N: Network> ConsensusProxy<N> {
...
+                                    .entry(current_head_number)
+                                    .or_insert(vec![])
+                                    .push(hash);
+                            }
+                        }
                     }
 
-                    // Now we request proofs for each epoch and its hashes
-                    for (epoch, hashes) in hashes_by_epoch {
+                    // Now we request proofs for each epoch and its hashes, according to its classification
```
sisou:
```suggestion
                    // Now we request proofs for each block and its hashes, according to their classification
```

sisou (approved):
Nice. You still have a clippy warning though to use `else if`.

viquezclaudio:
@sisou I just pushed a new commit that might be relevant for you. Because I discussed with JD and Pascal, that for generating txns proofs of the current epoch all we need is the accumulated txn count up to the point where we want to generate the proof, and there is no need to keep track of that count every checkpoint block.

sisou:
I was not aware that it's possible to get a ChainInfo for a past block. In that case your change is correct, as for macro blocks both tx counters are/were the same.

consensus/src/consensus/consensus_proxy.rs
```diff
@@ -94,26 +95,58 @@ impl<N: Network> ConsensusProxy<N> {
 
             match response {
                 Ok(response) => {
-                    // Group transaction hashes by epoch to reduce number of requested proofs
-                    let mut hashes_by_epoch = HashMap::new();
+                    let blockchain = self.blockchain.read();
+
+                    // Group transaction hashes by the block number that proofs those transactions to reduce the number of requests
```
jsdanielh:
```suggestion
                    // Group transaction hashes by the block number that proves those transactions to reduce the number of requests
```

consensus/src/consensus/consensus_proxy.rs
```diff
@@ -94,26 +95,58 @@ impl<N: Network> ConsensusProxy<N> {
...
+                    //  - Current batch: We use the current head to prove those transactions
+
+                    // This is the structure where we group transactions by their proving block number
+                    let mut hashes_by_block = HashMap::new();
+
+                    let election_head_number = blockchain.election_head().block_number();
+
+                    let checkpoint_head_number = blockchain.macro_head().block_number();
+
+                    let current_head_number = blockchain.head().block_number();
```
jsdanielh: I'd remove the extra lines
```suggestion
                    // This is the structure where we group transactions by their proving block number
                    let mut hashes_by_block = HashMap::new();
                    let election_head_number = blockchain.election_head().block_number();
                    let checkpoint_head_number = blockchain.macro_head().block_number();
                    let current_head_number = blockchain.head().block_number();
```

primitives/block/src/block.rs
```diff
@@ -232,8 +232,7 @@ impl Block {
         }
     }
 
-    /// Return the number of transactions in the block. If the block is a Macro
-    /// block it just returns zero, since Macro blocks don't contain any transactions.
```
jsdanielh: Why did we change it?

jsdanielh (approved):

primitives/block/src/block.rs
viquezclaudio: I thought the function returns only the number of transactions of the block and there was no need to clarify the distinction between micro and macro

sisou: I changed that comment actually in my commit, because as of this commit macro blocks now _do_ include transactions:
https://github.com/nimiq/core-rs-albatross/pull/1280/commits/26c8a848f500757d91f7452082d4636500d95e69#diff-aaa71b99350a00b080a331c7886311a206cfc2475c64d82748c711d9c4f10a04R242-R244

jsdanielh: Ah, but we didn't then update the primitives properly: https://github.com/nimiq/core-rs-albatross/blob/albatross/primitives/block/src/block.rs#L221

viquezclaudio: Ok sorry, Initially I was confused because I didn't remember that I made that change, but then I thought that maybe I just forgot about it :) 
Now it makes sense, I can update the primitives as well

PR #1403, opened 2023-03-09
Upgrade arkworks to 0.4

### LINKED ISSUE #1289: Update arkworks to 0.4

paberr:
- Upgrade arkworks to 0.4
- Generalize some of the circuits
- Remove `Vec<bool>` in favour of actual types and `Vec<u8>`
- Prepare pedersen hash to be replaced by arkworks equivalent
- Fix optimization level for `zkp-circuits`

## Pull request checklist


hrxi:
This implicitly upgrades all dependencies, including e.g. libmdbx to 0.1.12, which we apparently intentionally didn't do: https://github.com/nimiq/core-rs-albatross/pull/1238#pullrequestreview-1246000563.

four_validators.toml
```diff
@@ -0,0 +1,19 @@
+[[validator]]
```
jsdanielh: Was this file included by mistake? If we want to include this test scenario we could add it into `.github/devnet_topologies/`

rpc-interface/Cargo.toml
```diff
@@ -17,8 +17,8 @@ is-it-maintained-open-issues = { repository = "nimiq/core-rs" }
 maintenance = { status = "experimental" }
 
 [dependencies]
-ark-groth16 = { version = "0.3", default-features = false }
-ark-mnt6-753 = "0.3"
+ark-groth16 = "0.4"
```
jsdanielh: Do we actually need `parallel` feature (that would be the default feature we would be pulling now)?

zkp-circuits/Cargo.toml
```diff
@@ -19,22 +19,28 @@ log = { package = "tracing", version = "0.1", features = ["log"] }
...
-ark-mnt4-753 = "0.3"
-ark-mnt6-753 = "0.3"
-ark-relations = "0.3"
-ark-r1cs-std = "0.3.1"
-ark-serialize = "0.3"
-ark-std = "0.3"
+ark-crypto-primitives = { version = "0.4", features = ["crh", "prf", "r1cs"] }
+ark-ec = "0.4"
+ark-ff = "0.4"
+ark-groth16 = "0.4"
```
jsdanielh: Do we actually need `parallel` feature (that would be the default feature we would be pulling now)?

rpc-interface/Cargo.toml
paberr: It is definitely extremely useful for proving. We might not want to enable it for the web client, though.

four_validators.toml
ii-cruz: it was a mistake, we have removed it

jsdanielh (approved):

PR #1406, opened 2023-03-10
Fix Pedersen generator generation

hrxi:
Previously, the generation would output the same generator over and over again. This would have made the hash function weak.


jsdanielh (approved):

jsdanielh:
Rebasing it to get it merged

PR #1408, opened 2023-03-11
Create the accounts proof request

viquezclaudio:
Create the requests and handlers that are necessary to perform account proving.

Depends on #1462.

### DIFF
--- a/consensus/src/messages/handlers.rs
+++ b/consensus/src/messages/handlers.rs
@@ -398,7 +398,7 @@ impl Handle<ResponseTransactionsProof, Arc<RwLock<Blockchain>>> for RequestTrans
                 "Requested txn proof from current batch",
             );
             // If we were provided a block number corresponding to a micro block, it needs to correspond to the current batch
-            // Otherwise, the requester should use the latest checkpoing number
+            // Otherwise, the requester should use the latest checkpoint block number
             if self.block_number < macro_head {
                 log::info!(
                     block_number = self.block_number,
@@ -474,5 +474,30 @@ impl Handle<ResponseTransactionReceiptsByAddress, Arc<RwLock<Blockchain>>>
         }
 
         ResponseTransactionReceiptsByAddress { receipts }
+    }
+}
+
+#[cfg(feature = "full")]
+impl Handle<ResponseAccountsProof, Arc<RwLock<Blockchain>>> for RequestAccountsProof {
+    fn handle(&self, blockchain: &Arc<RwLock<Blockchain>>) -> ResponseAccountsProof {
+        let blockchain = blockchain.read();
+
+        // We only prove accounts that exist in our current state
+        let proof = blockchain.get_accounts_proof(&self.addresses);
+
+        if proof.is_none() {
+            // If we could not generate a proof we respond with an empty result
+            return ResponseAccountsProof {
+                proof: None,
+                block_hash: None,
+            };
+        }
+
+        let block_hash = blockchain.head_hash();
+
+        ResponseAccountsProof {
+            proof,
+            block_hash: Some(block_hash),
+        }
     }
 }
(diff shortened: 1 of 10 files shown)


viquezclaudio:
@sisou Just to let you know. This is failing to compile to WASM because there is a dependency of the accounts crate into the database.
So we need to remove that dependency as part of the accounts refactor in order to be able to use this functionality in the web client.

sisou:
Would it be possible to structure the request message in a way that it allows reusing for querying stakers and validators? Right now it doesn't know the difference and I guess only returns the regular accounts. How can we query staker accounts and validator accounts?

viquezclaudio:
LGTM

consensus/src/messages/handlers.rs
```diff
@@ -476,3 +476,28 @@ impl Handle<ResponseTransactionReceiptsByAddress, Arc<RwLock<Blockchain>>>
...
+#[cfg(feature = "full")]
+impl Handle<ResponseAccountsProof, Arc<RwLock<Blockchain>>> for RequestAccountsProof {
+    fn handle(&self, blockchain: &Arc<RwLock<Blockchain>>) -> ResponseAccountsProof {
+        let blockchain = blockchain.read();
+
+        // We only prove accounts that exist in our current state
+        let proof = blockchain.get_accounts_proof(&self.addresses);
+
+        if proof.is_none() {
+            // If we could not generate a proof we response an empty result
```
sisou:
```suggestion
            // If we could not generate a proof we respond with an empty result
```

consensus/src/messages/handlers.rs
```diff
@@ -476,3 +476,28 @@ impl Handle<ResponseTransactionReceiptsByAddress, Arc<RwLock<Blockchain>>>
...
+            return ResponseAccountsProof {
+                proof: None,
+                block_hash: None,
+            };
+        }
+
+        let block_hash = blockchain.head_hash();
```
sisou, line 496: Is this guaranteed to be the same hash that is used to generate the proof? No chance that the head is updated in this blockchain-read in the meantime?

consensus/src/messages/mod.rs
```diff
@@ -248,3 +249,25 @@ pub struct ResponseTransactionReceiptsByAddress {
...
+    #[beserial(len_type(u16, limit = 128))]
+    /// Addresses for which the accounts trie proof is requested for
+    pub addresses: Vec<Address>,
+}
+
+impl RequestCommon for RequestAccountsProof {
+    type Kind = RequestMarker;
+    const TYPE_ID: u16 = 215;
+    type Response = ResponseAccountsProof;
+    const MAX_REQUESTS: u32 = MAX_REQUEST_TRANSACTIONS_PROOF;
```
sisou: Are we sure we should reuse the TRANSACTIONS_PROOF limit for this request?

sisou (approved):

consensus/src/messages/handlers.rs
jsdanielh: Yup, the read lock is taken while this happens

consensus/src/messages/mod.rs
jsdanielh: Value is correct but name is not

consensus/src/messages/mod.rs
```diff
@@ -248,3 +251,25 @@ pub struct ResponseTransactionReceiptsByAddress {
     #[beserial(len_type(u16, limit = 128))]
     pub receipts: Vec<(Blake2bHash, u32)>,
 }
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct RequestTrieProof {
```
sisou, line 256: The handler for this message still needs to be registered in `consensus/src/consensus/mod.rs`.

jsdanielh: Thanks. I already had forgotten about it

jsdanielh (approved):

PR #1409, opened 2023-03-11
Don't precompute Pedersen hashes

hrxi:
This reduces the WASM blob from 12 MiB to 4.8 MiB, a 60% reduction!

We introduced this precomputation because it took too long, many seconds, I believe. However, it seems that this has resolved itself and it's now fast.


sisou:
mind blown

hrxi:
CI failure is #1412, unrelated to this PR.

jsdanielh (approved):

jsdanielh:
Rebasing it to get it merged

PR #1410, opened 2023-03-11
Serialize blocks to JS as objects

sisou:

Instead of sending byte-serialized blocks to JS, add a proper `PlainBlock` struct and type that can be serialized to a JS object (and is JSON-compatible).

## Pull request checklist

### DIFF
--- a/web-client/src/block.rs
+++ b/web-client/src/block.rs
@@ -1,0 +1,160 @@
+use serde::ser::SerializeStruct;
+use tsify::Tsify;
+use wasm_bindgen::prelude::*;
+
+use beserial::Serialize;
+use nimiq_primitives::policy::Policy;
+
+/// JSON-compatible and human-readable format of blocks.
+#[derive(Tsify)]
+#[serde(rename_all = "camelCase")]
+pub struct PlainBlock {
+    /// The block's unique hash, used as its identifier, in HEX format.
+    pub hash: String,
+    /// The block's on-chain size, in bytes.
+    pub size: u32,
+    /// The block's block height, also called block number.
+    pub height: u32,
+    /// The batch number that the block is in.
+    pub batch: u32,
+    /// The epoch number that the block is in.
+    pub epoch: u32,
+    /// The timestamp of the block. It follows the Unix time and has millisecond precision.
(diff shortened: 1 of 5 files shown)


web-client/src/block.rs
```diff
@@ -0,0 +1,187 @@
...
+pub struct PlainBlock {
+    /// The block's unique hash, used as its identifier, in HEX format.
+    pub hash: String,
+    /// The block's on-chain size, in bytes.
+    pub size: u32,
+    /// The block's block height, also called block number.
+    pub height: u32,
+    /// The batch number that the block is in.
+    pub batch: u32,
+    /// The epoch numbe that the block is in.
```
jsdanielh:
```suggestion
    /// The epoch number that the block is in.
```

web-client/src/block.rs
```diff
@@ -0,0 +1,187 @@
...
+        }
+
+        plain.end()
+    }
+}
+
+impl PlainBlock {
+    /// Creates a PlainBlock struct that can be serialized to JS from a native [nimiq_block::Block].
+    pub fn from_block(block: &nimiq_block::Block) -> Self {
+        // log::info!("{block:#?}");
```
jsdanielh: I'd vote to remove it completely

web-client/src/block.rs
```diff
@@ -0,0 +1,187 @@
+use serde::ser::SerializeStruct;
+use tsify::Tsify;
+use wasm_bindgen::prelude::*;
+
+use beserial::Serialize;
+use nimiq_primitives::policy::Policy;
+
+// use crate::transaction::PlainTransaction;
```
jsdanielh: I'd vote to remove it completely

jsdanielh (commented):
Some small comments.

@sisou I noted there are nested fields in `PlainBlockAdditionalFields` commented. Most of them `Option`s. What's the deal with them? I guess we want them?

sisou:
I removed all commented fields, as I am not sure yet how to make them into plain JSON fields. They are not critical at the moment I believe and can be added later, if necessary.

jsdanielh (approved):

PR #1415, opened 2023-03-11
Accounts DB dependency

viquezclaudio:
Commits necessary to make the DB dependency in the accounts crate optional. 
This is required for WASM compilation


jsdanielh:
Current implementation doesn't work because of the staking contract dependency on the `DataStore`.

PR #1425, opened 2023-03-13
Return PlainTransactionDetails from tx sending methods

sisou:
Just like the 1.0 client, the Albatross WASM client should return `PlainTransactionDetails` from its transaction-sending methods. Since we do not have any way to listen to transaction-inclusion events currently, this PR implements a receipt-polling method. It works, but is obviously inefficient. It will be reworked once transaction-inclusion events can be subscribed to.

To facilitate the receipt-polling and single-transaction proving most efficiently, I restructured the big `request_transactions_by_address` method in consensus_proxy into smaller, reusable chunks.

## Pull request checklist


sisou:
Rebased and fixed conflicts with #1430.

jsdanielh (approved):

PR #1426, opened 2023-03-13
Hash values of hybrid nodes before letting them contribute to the hash

hrxi:
This indirection allows trie proofs to only carry the hash of a hybrid node's data instead of its whole data. This should reduce the size of trie proofs.

Leaf nodes still directly hash their data, so as to not decrease performance by hashing twice unnecessarily.


primitives/src/trie/trie_node.rs
```diff
@@ -260,7 +260,22 @@ impl TrieNode {
...
+                    hasher.write_u8(1).unwrap();
+                    val.serialize::<u16, _>(&mut hasher).unwrap();
+                }
+                (true, Some(val)) => {
+                    hasher.write_u8(2).unwrap();
+                    let val_hash: Blake2bHash = val.hash();
+                    val_hash.serialize(&mut hasher).unwrap();
+                }
+            }
+            self.children.serialize(&mut hasher).unwrap();
```
paberr, line 278: Why didn't you implement this behaviour as part of `serialize_content()` instead?

primitives/src/trie/trie_proof_node.rs
```diff
@@ -0,0 +1,157 @@
...
+                hasher.write_u8(1).unwrap();
+                val.serialize::<u16, _>(&mut hasher).unwrap();
+            }
+            ProofValue::Hash(val_hash) => {
+                hasher.write_u8(2).unwrap();
+                val_hash.serialize(&mut hasher).unwrap();
+            }
+        }
+        self.children.serialize(&mut hasher).unwrap();
+        hasher.finish()
```
paberr: Can we avoid code duplication?

paberr (commented):
Generally LGTM

ii-cruz (approved):

primitives/src/trie/trie_node.rs
hrxi: Did that for the `TrieProofNode`, but here I didn't do it because not all `TrieNode`s are actually hashable.

primitives/src/trie/trie_proof_node.rs
hrxi: Unfortunately, it seem like there's no easy way. Converting the `TrieNode` to a `TrieProofNode` would mean copying its data.

paberr (approved):

jsdanielh:
Rebasing it to get it merged

PR #1428, opened 2023-03-13
Add address filter to get_transactions_by_address

onmax:

This pull request adds the ability to filter transactions by sender and recipient addresses in the `get_transactions_by_address` function of the Nimiq RPC interface. The new functionality allows users to retrieve only transactions that involve specific addresses, which is useful for filtering out irrelevant transactions when performing analysis or debugging.

More parameters could be added in the future such as `block_ranges` (`start_block` and `to_block`).

The current implementation is not the most efficient as it fetches the transaction from `blockchain.history_store.get_ext_tx_by_hash` and then filter them out. Improving the current solution requires a database update.

## Pull request checklist

### DIFF
--- a/rpc-client/src/subcommands/blockchain_subcommands.rs
+++ b/rpc-client/src/subcommands/blockchain_subcommands.rs
@@ -91,6 +91,14 @@ pub enum BlockchainCommand {
         /// The address to query by.
         address: Address,
 
+        /// Filter transactions by recipient addresses. Empty list means no filter.
+        /// If both `from` and `to` are empty, all transactions are returned.
+        from: Vec<Address>,
+
+        /// Filter transactions by sender addresses. Empty list means no filter.
+        /// If both `from` and `to` are empty, all transactions are returned.
+        to: Vec<Address>,
+
         /// Max number of transactions to fetch. If absent it defaults to 500.
         #[clap(long)]
         max: Option<u16>,
@@ -267,6 +275,8 @@ impl HandleSubcommand for BlockchainCommand {
 
             BlockchainCommand::TransactionsByAddress {
                 address,
+                from,
+                to,
                 max,
                 just_hash,
             } => {
@@ -283,7 +293,7 @@ impl HandleSubcommand for BlockchainCommand {
                         "{:#?}",
                         client
                             .blockchain
-                            .get_transactions_by_address(address, max)
+                            .get_transactions_by_address(address, from, to, max)
                             .await?
                     )
                 }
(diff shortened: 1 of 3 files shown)


rpc-server/src/dispatchers/blockchain.rs
```diff
@@ -429,6 +429,8 @@ impl BlockchainInterface for BlockchainDispatcher {
     async fn get_transactions_by_address(
         &mut self,
         address: Address,
+        from: Vec<Address>,
+        to: Vec<Address>,
         max: Option<u16>,
     ) -> RPCResult<Vec<ExecutedTransaction>, (), Self::Error> {
```
sisou, line 435: Instead of adding it in the middle of the existing arguments in the existing method, which is a breaking and backward-incompatible change, I'd instead create a new method that explicitly searches for transactions with filters. Maybe it can be possible to pass a `filters` object as an argument, that is freely extensible, instead of fixed arguments?

sisou (changes requested):

rpc-server/src/dispatchers/blockchain.rs
onmax: We tried that, but we ran into problems serialising / deserialising with clap. I would really like to make this work, but I could not find a solution. I asked @jsdanielh and @ii-cruz, but they came to the same conclusion. 

Another solution was to try to make our own parser, but I don't know how to do that.

And yes, I agree with using an object, so in the future we could add other filters

rpc-client/src/subcommands/blockchain_subcommands.rs
```diff
@@ -91,6 +91,14 @@ pub enum BlockchainCommand {
         /// The address to query by.
         address: Address,
```
onmax, line 93: @sisou Here is where we were facing the issues. 

Maybe for now, we could add the object as you mentioned in the RPC request, but ignore those filters in the cli in the #L296 using ` .get_transactions_by_address(address, None, max)`

rpc-server/src/dispatchers/blockchain.rs
sisou: Then at least I'd make this a new method `get_transactions_by_address_with_filters` and add the new arguments there, as I think they will be an edge-case usage.

onmax: Ok. I thought it would be more common use case.

ii-cruz:
The solution would be to either subscribe to the blockchain logs or implement a function that is very computationally heavy since we are open to changing the history store atm.

Also because Prof Doctor Berrang said no.

PR #1430, opened 2023-03-14
Adapt arguments of getTransactionsByAddress to be 1.0-compatible

sisou:

These new arguments follow the same order, type, and name as the 1.0 client method arguments.

Additionally, they are also a nice feature to have to be able to filter by block height and known transactions, reducing the number of proofs requested from the network.

This PR has conflicts with #1425, so when either is merged, the other will need to be rebased and merge-conflicts solved.

## Pull request checklist

### DIFF
--- a/consensus/src/consensus/consensus_proxy.rs
+++ b/consensus/src/consensus/consensus_proxy.rs
@@ -3,6 +3,7 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Arc;
 
 use nimiq_blockchain_interface::AbstractBlockchain;
+use nimiq_hash::Blake2bHash;
 use tokio::sync::broadcast::Sender as BroadcastSender;
 use tokio_stream::wrappers::BroadcastStream;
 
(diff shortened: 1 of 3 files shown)


consensus/src/consensus/consensus_proxy.rs
```diff
@@ -119,6 +122,14 @@ impl<N: Network> ConsensusProxy<N> {
                     let current_head_number = blockchain.head().block_number();
 
                     for (hash, block_number) in response.receipts {
+                        if block_number < since_block_height {
```
viquezclaudio, line 125: This will work if we have less than max transactions in the chain, but it will be nice to have an efficient way to prove txns from a range of blocks. 
@jsdanielh

sisou: It works in any case. Receipts are sorted and are only the most recent, even if more than max are on chain. This filter just filters it down further.

viquezclaudio: Ok

viquezclaudio (approved):

jsdanielh (approved):

PR #1438, opened 2023-03-14
Limit Tendermint proof verification to available CPU parallelism

hrxi:
The `tokio::spawn_blocking` docs suggest to use `rayon` to do that, we also get aborts on the first failed verification for free.


jsdanielh (approved):

jsdanielh:
Rebased to get it merged

PR #1444, opened 2023-03-15
Add support for block inclusion proofs

fiaxh:


blockchain/src/blockchain/verify.rs
```diff
@@ -254,4 +276,15 @@ impl Blockchain {
 
         Ok(None)
     }
+
+    pub fn next_interlink(&self) -> Vec<Blake2bHash> {
```
sisou: Is this method unused?

fiaxh: Indeed it was, thanks.

consensus/src/messages/mod.rs
```diff
@@ -248,3 +250,22 @@ pub struct ResponseTransactionReceiptsByAddress {
...
+#[derive(Serialize, Deserialize)]
+pub struct ResponseBlocksProof {
+    pub proof: Option<BlockInclusionProof>,
+}
+
+impl RequestCommon for RequestBlocksProof {
+    type Kind = RequestMarker;
+    const TYPE_ID: u16 = 215;
+    type Response = ResponseBlocksProof;
+    const MAX_REQUESTS: u32 = MAX_REQUEST_TRANSACTIONS_PROOF;
```
sisou: Are we sure we are going to reuse the `MAX_REQUEST_TRANSACTIONS_PROOF` number for this message type?

sisou (approved):

consensus/src/messages/mod.rs
fiaxh: No, we should use the MAX_REQUEST_BLOCKS_PROOF number that I already created for that purpose, but failed to actually use :D

consensus/src/messages/handlers.rs
```diff
@@ -6,12 +6,12 @@ use std::{cmp, sync::Arc};
 use parking_lot::RwLock;
 
 use nimiq_block::Block;
+use nimiq_block::BlockInclusionProof;
 #[cfg(feature = "full")]
 use nimiq_blockchain::{Blockchain, CHUNK_SIZE};
 use nimiq_blockchain_interface::{AbstractBlockchain, Direction};
 use nimiq_blockchain_proxy::BlockchainProxy;
 use nimiq_network_interface::request::Handle;
-#[cfg(feature = "full")]
```
jsdanielh: I think we need to leave this as it is and then add:
`#[cfg(feature = "full")]` in the handle below

jsdanielh: Fixed

primitives/block/src/macro_block.rs
```diff
@@ -78,6 +78,38 @@ impl MacroBlock {
         Ok(pk_tree_construct(public_keys).to_vec())
     }
 
+    // Computes the next interlink from self.header.interlink
+    pub fn get_next_interlink(&self) -> Result<Vec<Blake2bHash>, BlockError> {
```
sisou:
```suggestion
    /// Computes the next interlink from self.header.interlink
    pub fn get_next_interlink(&self) -> Result<Vec<Blake2bHash>, BlockError> {
```

jsdanielh: Fixed as well

primitives/block/src/block_proof.rs
```diff
@@ -0,0 +1,121 @@
...
+        for block in self.proof.clone() {
+            number_to_block.insert(block.block_number(), block);
+        }
+
+        let hops = BlockInclusionProof::get_interlink_hops(
+            target.block_number(),
+            election_head.block_number(),
+        );
+
+        // If we're at the 1th or 2nd macro block, there are no block proofs
```
jsdanielh:
```suggestion
        // If we're at the 1st or 2nd macro block, there are no block proofs
```

jsdanielh: Fixed

PR #1448, opened 2023-03-16
Rename `include_transactions` to `include_body` in RPC server, interface and client

onmax:

Update `BlockchainDispatcher` to allow for excluding transactions from block responses when requested but keeping the rest of keys untouched.

## Pull request checklist

### More details

```js
blockByNumber(/*block_number*/ 0, /*include_transactions */ false)
``` 

Given the following RCP request to `https://seed1.v2.nimiq-testnet.com:8648/?secret={SECRET}` and body as:
```{"jsonrpc":"2.0","method":"getBlockByNumber","params":[0,false],"id":1}```

<details><summary> Previously the response was</summary>

```json
{
    "jsonrpc": "2.0",
    "result": {
        "data": {
            "hash": "1ea0fe2c0b2004c046e8d5f386359de51ded1d2a8c4c8eb49d253305c12e51d5",
            "size": 278,
            "batch": 0,
            "epoch": 0,
            "version": 1,
            "number": 0,
            "timestamp": 1678838400,
            "parentHash": "0000000000000000000000000000000000000000000000000000000000000000",
            "seed": {
                "signature": [ "blah blah blah" ]
            },
            "extraData": "",
            "stateHash": "35106825436098c59fa582f973a5f1497b78f35fd0e0852ff381b2e21957f379",
            "bodyHash": "c9279c5fde51032adc643e825d9efeb5f23aee83ea0d396093300937517d035f",
            "historyHash": "0000000000000000000000000000000000000000000000000000000000000000",
            "type": "macro",
            "isElectionBlock": true,
            "parentElectionHash": "0000000000000000000000000000000000000000000000000000000000000000"
        },
        "metadata": null
    },
    "id": 1
}
```
</details> 

<details><summary> With this PR, the response is </summary>

```json
{
    "jsonrpc": "2.0",
    "result": {
        "data": {
            "hash": "b542d3744e6adef801ad230004366a80279a3814cc8a53598187c3902af498e9",
            "size": 1650,
            "batch": 0,
            "epoch": 0,
            "version": 1,
            "number": 0,
            "timestamp": 1626307200,
            "parentHash": "0000000000000000000000000000000000000000000000000000000000000000",
            "seed": {
                "signature": [ "blah blah blah" ]
            },
            "extraData": "",
            "stateHash": "b52c86b31d61def7e7b5be4109dafcb30a4d924363cbed2b6953eb2427889d1b",
            "bodyHash": "c066df2adf1847645851d9475022619ce69c937d37a21c0148b888d6a3adda30",
            "historyHash": "0000000000000000000000000000000000000000000000000000000000000000",
            "type": "macro",
            "isElectionBlock": true,
            "parentElectionHash": "0000000000000000000000000000000000000000000000000000000000000000",
            "slots": [ "slots distribution" ],
            "lostRewardSet": [],
            "disabledSet": []
        },
        "metadata": null
    },
    "id": 1
}
```
</details> 

## What did it change?

The new response include: `slots`, `lostRewardSet` and `disabledSet`

### DIFF
--- a/rpc-client/src/subcommands/blockchain_subcommands.rs
+++ b/rpc-client/src/subcommands/blockchain_subcommands.rs
@@ -39,9 +39,9 @@ pub enum BlockchainCommand {
         #[clap(long)]
         block_number: Option<u32>,
 
-        /// Include transactions
-        #[clap(short = 't', long)]
-        include_transactions: bool,
+        /// Whether to include the block body
+        #[clap(short = 'b', long)]
+        include_body: bool,
     },
 
     /// Query a transaction from the blockchain.
@@ -176,23 +176,20 @@ impl HandleSubcommand for BlockchainCommand {
             BlockchainCommand::Block {
                 block_hash,
                 block_number,
-                include_transactions,
+                include_body,
             } => {
                 let block = if let Some(block_hash) = block_hash {
                     client
                         .blockchain
-                        .get_block_by_hash(block_hash, Some(include_transactions))
+                        .get_block_by_hash(block_hash, Some(include_body))
                         .await
                 } else if let Some(block_number) = block_number {
                     client
                         .blockchain
-                        .get_block_by_number(block_number, Some(include_transactions))
+                        .get_block_by_number(block_number, Some(include_body))
                         .await
                 } else {
-                    client
-                        .blockchain
-                        .get_latest_block(Some(include_transactions))
-                        .await
+                    client.blockchain.get_latest_block(Some(include_body)).await
                 }?;
                 println!("{block:#?}")
             }
--- a/rpc-server/src/dispatchers/blockchain.rs
+++ b/rpc-server/src/dispatchers/blockchain.rs
@@ -35,16 +35,13 @@ impl BlockchainDispatcher {
 fn get_block_by_hash(
     blockchain: &BlockchainReadProxy,
     hash: &Blake2bHash,
-    include_transactions: Option<bool>,
+    include_body: Option<bool>,
 ) -> RPCResult<Block, (), Error> {
-    let include_body =
-        include_transactions.unwrap_or(matches!(blockchain, BlockchainReadProxy::Full(_)));
+    let include_body = include_body.unwrap_or(matches!(blockchain, BlockchainReadProxy::Full(_)));
 
     blockchain
         .get_block(hash, include_body)
-        .map(|block| {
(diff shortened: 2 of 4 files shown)


rpc-server/src/dispatchers/blockchain.rs
```diff
@@ -37,13 +37,23 @@ fn get_block_by_hash(
...
 ) -> RPCResult<Block, (), Error> {
-    let include_body =
+    let include_transactions =
         include_transactions.unwrap_or(matches!(blockchain, BlockchainReadProxy::Full(_)));
 
     blockchain
-        .get_block(hash, include_body)
-        .map(|block| {
-            Block::from_block(blockchain, block, include_transactions.unwrap_or(false)).into()
+        .get_block(hash, true)
```
jsdanielh: I just realized this can't be always `true`. I think we would need to request the block first and then do a logic like the one we did at `get_block_at`

jsdanielh (changes requested):

rpc-server/src/dispatchers/blockchain.rs
sisou: When using RPC with a light consensus node, I believe setting `true` will always error there, too.

sisou:
Maybe a simpler solution is to rename the parameter from `include_transactions` to `include_body`, which it maps to anyway from the first layer, and call it a day? Then it's not misleading.

rpc-server/src/dispatchers/blockchain.rs
jsdanielh: Exactly, that is the main reason why we can't set it to `true`

jsdanielh:
> Maybe a simpler solution is to rename the parameter from `include_transactions` to `include_body`, which it maps to anyway from the first layer, and call it a day? Then it's not misleading.

Yeah, if that actually works for you guys, I'd vote for this

jsdanielh (approved):

rpc-client/src/subcommands/blockchain_subcommands.rs
```diff
@@ -40,8 +40,8 @@ pub enum BlockchainCommand {
         block_number: Option<u32>,
 
         /// Include transactions
```
sisou: Should also update this rustdoc comment to clarify the whole body is included/excluded, which includes other things in macro blocks, as @onmax noticed.

sisou (approved):

ii-cruz (approved):
The only thing I see missing would be the extra comment soren suggested

PR #1449, opened 2023-03-16
Accept various argument types for sending a transaction

sisou:

Make the `client.sendTransaction` method compatible with the 1.0 client by accepting instances of the `Transaction` class, a plain object following the `PlainTransaction` type, or a hex string of the serialized transaction. 

## Pull request checklist

### DIFF
--- a/web-client/Cargo.toml
+++ b/web-client/Cargo.toml
@@ -30,6 +30,7 @@ serde-wasm-bindgen = "0.5"
 tsify = { version = "0.4", features= ["js"] }
 wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4"
+wasm-bindgen-derive = "0.1.1"
 web-sys = { version = "0.3.22", features = ["EventTarget", "Window"]}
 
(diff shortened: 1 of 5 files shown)


viquezclaudio (approved):

web-client/Cargo.toml
```diff
@@ -30,6 +30,7 @@ serde-wasm-bindgen = "0.5"
 tsify = { version = "0.4", features= ["js"] }
 wasm-bindgen = "0.2"
 wasm-bindgen-futures = "0.4"
+wasm-bindgen-derive = { git = "https://github.com/sisou/wasm-bindgen-derive.git" }
```
jsdanielh: We should be able to use 0.1.1 now

jsdanielh (changes requested):

web-client/Cargo.toml
sisou: Correct. Updated.

jsdanielh (approved):

jsdanielh:
Rebasing it to get it merged

PR #1455, opened 2023-03-17
Replace a `todo!()` and normalize some trie code

hrxi:


sisou:
Merged as part of #1484.

PR #1458, opened 2023-03-18
ZK Proof Simulation

ii-cruz:
We can now simulate ZKPs inside the unit tests and don't need to generate them using the binary.
Moreover, we added metadata files to verify loaded verifying keys against the policy.
This way we can detect invalid keys easily.

## Pull request checklist


paberr:
Has been merged.

PR #1460, opened 2023-03-19
Emit "syncing" consensus events

sisou:

In addition to emitting `connecting` and `established` events, this adds the ability for the client to also emit `syncing` events, as the 1.0 client did.

This implementation follows a short talk with @styppo. He is hesitant to add a syncing event to the native consensus, instead suggesting that the client should emit' syncing' events depending on the combination of consensus and peer count.

For this purpose, both when consensus state changes and when the peer count changes, we compare both and emit the respective event accordingly. The threshold for emitting the `syncing` event is set in this PR to `1` connected peer. I can also imagine making it dependant on the consensus' `min_peers`, but since the client already starts syncing the ZKP when connected to the first peer (seed node), I thought it should emit that it's syncing then already.

## Pull request checklist

### DIFF
--- a/web-client/src/lib.rs
+++ b/web-client/src/lib.rs
@@ -5,6 +5,7 @@ use std::{cell::RefCell, collections::HashMap, rc::Rc, str::FromStr};
 use futures::StreamExt;
 use js_sys::{Array, Date, Promise};
 use log::level_filters::LevelFilter;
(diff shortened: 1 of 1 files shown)


jsdanielh (approved):

jsdanielh:
Rebasing it to get it merged

PR #1474, opened 2023-03-21
Include hybrid values that are proven in trie proofs

hrxi:
This allows to use the proof to directly read values from the trie.

### DIFF
--- a/primitives/src/trie/trie_proof_node.rs
+++ b/primitives/src/trie/trie_proof_node.rs
@@ -37,17 +37,18 @@ pub struct TrieProofNode {
 #[derive(Clone, Debug, Deserialize, Serialize)]
 #[repr(u8)]
 enum ProofValue {
-    Value(#[beserial(len_type(u16))] Vec<u8>),
-    Hash(Blake2bHash),
     None,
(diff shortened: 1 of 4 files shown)


primitives/src/trie/trie_proof_node.rs
```diff
@@ -58,6 +59,29 @@ impl From<TrieNode> for TrieProofNode {
...
+
+impl TrieProofNode {
+    pub fn into_value(self) -> Option<Vec<u8>> {
+        match self.value {
+            ProofValue::None => None,
+            ProofValue::LeafValue(value) => Some(value),
+            ProofValue::HybridHash(_) => None,
+            ProofValue::HybridValue(value) => Some(value),
+        }
+    }
```
sisou, line 84: Might it make more sense to not take self but use and return references?

```suggestion
    pub fn value(&self) -> Option<&Vec<u8>> {
        match &self.value {
            ProofValue::None => None,
            ProofValue::LeafValue(value) => Some(value),
            ProofValue::HybridHash(_) => None,
            ProofValue::HybridValue(value) => Some(value),
        }
    }
```

jsdanielh (approved):

PR #1475, opened 2023-03-21
Improve mempool behaviour during and after consensus loss

fiaxh:
Currently, the mempool assumes to always work on a complete accounts tree and expect()s it to be there. However, there can be a short timespan after we had consensus and lost it, where the mempool is not yet deactivated (since the validator and the consensus run on different threads). This PR fixes the potential crash by not expect()ing the account anymore.

Also, the state of the mempool might be outdated after the consensus is regained, since it doesn't get any updates in that time. Also, the state might be inconsistent if we ran into a partial accountstree and had to abort the current action. This PR rechecks all transactions in the mempool and rebuilds the reserved balance.

### DIFF
--- a/validator/src/validator.rs
+++ b/validator/src/validator.rs
@@ -426,6 +426,9 @@ where
             #[cfg(not(feature = "metrics"))]
             tokio::spawn({
                 async move {
+                    // The mempool is not updated while consensus is lost.
+                    // Thus, we need to check all transactions if they are still valid.
+                    mempool.mempool_update_full();
                     mempool.start_executors(network, None, None).await;
                 }
             });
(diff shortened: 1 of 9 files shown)


validator/src/validator.rs
```diff
@@ -787,6 +787,10 @@ where
                             let mempool_monitor = self.mempool_monitor.clone();
                             let ctrl_mempool_monitor = self.control_mempool_monitor.clone();
                             async move {
+                                // The mempool is not updated while consensus is lost.
+                                // Thus, we need to check all transactions if they are still valid.
+                                mempool.mempool_update_full();
```
jsdanielh: We need to add this in the `#[cfg(not(feature = "metrics"))]` case

validator/tests/mock.rs
```diff
@@ -17,14 +17,17 @@ use nimiq_network_interface::{
     request::{MessageMarker, RequestCommon},
 };
 use nimiq_network_libp2p::Network;
-use nimiq_network_mock::MockHub;
+use nimiq_network_mock::{MockHub, MockNetwork};
+use nimiq_primitives::coin::Coin;
+use nimiq_primitives::networks::NetworkId;
```
jsdanielh:
```suggestion
use nimiq_primitives::{coin::Coin, networks::NetworkId};
```

jsdanielh (changes requested):

jsdanielh (approved):

PR #1484, opened 2023-03-22
Allow to prove the non-existence of keys

hrxi:
This is done by including the last existing node on the path to the missing key. In order to make verifying cheap, the mapping from the missing key to the last existing node's key is included in the proof.

Includes the commits from #1455, because it would conflict otherwise.

### DIFF
--- a/primitives/src/trie/trie_proof.rs
+++ b/primitives/src/trie/trie_proof.rs
@@ -1,5 +1,6 @@
 use log::error;
-use std::collections::{BTreeMap, BTreeSet};
+use std::cmp::Ordering;
+use std::collections::BTreeMap;
 
 use beserial::{Deserialize, Serialize};
 use nimiq_hash::{Blake2bHash, Hash};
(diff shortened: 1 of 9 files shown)


primitives/src/trie/trie_proof.rs
```diff
@@ -123,32 +139,105 @@ impl TrieProof {
...
+            .collect();
+        keys.sort_by(|&(k1p, k1), &(k2p, k2)| {
+            k1p.post_order_cmp(k2p).then_with(|| k1.post_order_cmp(k2))
+        });
+
+        eprintln!("{:?}", keys);
+        eprintln!(
+            "{:?}",
+            self.nodes.iter().map(|n| &n.key).collect::<Vec<_>>()
+        );
```
sisou: Debug output?

hrxi: Correct, thank you. Fixed.

primitives/src/trie/trie_proof.rs
```diff
@@ -123,32 +139,72 @@ impl TrieProof {
...
+                    if n.key == *k {
+                        let key = n.key.clone();
+                        let value = n
+                            .into_value()
+                            .map_err(|_| Error("requested hybrid node value missing"))?;
+                        assert!(result.insert(key, value).is_none());
+                        cur_key = keys.next();
+                        cur_node = nodes.next();
+                    } else {
+                        eprintln!("checking {} against {} for negativity", k, n.key);
```
sisou: Still a remaining `eprintln`. Should be a `log::debug`?

hrxi: Hmmm hmmm, not sure how I forgot that one… Removed and added a comment instead.

sisou (approved):

jsdanielh (approved):

jsdanielh:
Rebasing it to get it merged

PR #1488, opened 2023-03-23
Validity window fix

### LINKED ISSUE #1478: Start validator only when necessary data is available
Currently, the validator starts after consensus is established.
However, for full nodes, there are two further conditions that have to be met:
- History tree for the current epoch is extendable (currently equivalent to fully known)
- Transaction validity can be checked (currently equivalent to: we have the history tree over the last 7200 blocks)

While we could speed this up in the future by syncing the necessary data, we should at least wait until the conditions are met.

### LINKED ISSUE #1479: Macro sync only if necessary (full nodes)
The macro sync of full nodes always triggers a state sync.
In some cases, it might be wiser to run a history sync instead.
We should evaluate the threshold in the macro sync based on the number of leaves of the history (which can be proven by a hash).

- [ ] Decide when to live sync instead of macro sync? (To avoid syncing state) -> base on number of txs?
    - [ ] Alternative: fixed amount of macro blocks as threshold
    - [x] Include number of children in hash? Somehow include number of txs
    - [ ] history_root = Hash(num_leaves || peak_hash || root_hash)
    - [ ] We need a function prove_num_leaves() that returns num_leaves and peak_hash and root_hash

ii-cruz:
- Prepares SizeProof to be integrated into history sync
- Start validator only when conditions for validity window are fulfilled
- Optimize transition from macro to live sync for full nodes

## Pull request checklist

### DIFF
--- a/consensus/src/sync/light/sync_stream.rs
+++ b/consensus/src/sync/light/sync_stream.rs
@@ -20,6 +20,8 @@ use crate::sync::{
     syncer::{MacroSync, MacroSyncReturn},
 };
 
+use super::full_sync_threshold;
+
 impl<TNetwork: Network> LightMacroSync<TNetwork> {
     // This function is the one that starts the LightMacroSync process,
     // by adding peers into the MacroSync component.
@@ -66,11 +68,36 @@ impl<TNetwork: Network> LightMacroSync<TNetwork> {
                         // Apply a newer proof to the blockchain
                         let result = match self.blockchain {
                             #[cfg(feature = "full")]
-                            BlockchainProxy::Full(ref full_blockchain) => Blockchain::push_zkp(
-                                full_blockchain.upgradable_read(),
-                                Block::Macro(block),
-                                proof.proof.expect("Expected a zkp proof"),
-                            ),
+                            BlockchainProxy::Full(ref full_blockchain) => {
+                                let blockchain_urg = full_blockchain.upgradable_read();
+                                if block
+                                    .block_number()
+                                    .saturating_sub(blockchain_urg.block_number())
+                                    <= full_sync_threshold()
+                                {
+                                    // We deem this too close to do a macro sync, thus we are not pushing the zkp. This would
+                                    // clear the state and history store. Instead, we request the epoch ids from this peer.
+                                    log::debug!(
+                                        peer_id = ?peer_id,
+                                        "Peer is sufficiently close not to apply the zkp."
+                                    );
+
+                                    let future = Self::request_epoch_ids(
+                                        self.blockchain.clone(),
+                                        Arc::clone(&self.network),
+                                        peer_id,
+                                    )
+                                    .boxed();
+                                    self.epoch_ids_stream.push(future);
+
+                                    continue;
+                                }
+                                Blockchain::push_zkp(
+                                    blockchain_urg,
+                                    Block::Macro(block),
+                                    proof.proof.expect("Expected a zkp proof"),
+                                )
+                            }
                             BlockchainProxy::Light(ref light_blockchain) => {
                                 LightBlockchain::push_zkp(
                                     light_blockchain.upgradable_read(),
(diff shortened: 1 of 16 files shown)


consensus/src/sync/light/sync_stream.rs
```diff
@@ -66,11 +68,36 @@ impl<TNetwork: Network> LightMacroSync<TNetwork> {
...
-                            ),
+                            BlockchainProxy::Full(ref full_blockchain) => {
+                                let blockchain_urg = full_blockchain.upgradable_read();
+                                if block
+                                    .block_number()
+                                    .saturating_sub(blockchain_urg.block_number())
+                                    <= full_sync_threshold()
+                                {
+                                    // We deem this too close, thus we are not pushing the zkp.
+                                    // Instead we request the epoch ids from this peer.
```
nibhar: Please add a comment here explaining why not to push the zkp here, as even when pushing the zkp the epoch ids would still be requested  in lines 101+ of this file. The reason why this change set makes a difference is not immediately clear.

primitives/mmr/src/mmr/utils.rs
```diff
@@ -31,6 +33,48 @@ pub(crate) fn bagging<H: Merge, I: Iterator<Item = Result<(H, usize), Error>>>(
     Ok(root)
 }
 
+/// Computes a size proof for an iterator over peak positions.
+/// This function fails if there is less than 2 peaks.
```
nibhar: Seems like it can prove a single Peek.

primitives/mmr/src/mmr/mod.rs
```diff
@@ -143,6 +146,28 @@ impl<H: Merge + Clone, S: Store<H>> MerkleMountainRange<H, S> {
...
+        // Return the history root.
+        if self.is_empty() {
+            return Ok(SizeProof::EmptyTree);
+        }
+
+        let it = self.rev_peaks().map(|peak_pos| {
+            Ok((
+                self.store
+                    .get(peak_pos.index)
+                    .ok_or(Error::InconsistentStore)?,
```
nibhar, line 163: If I understand this correctly, adding a Single `Err` into the iterator will make the subsequent call to `prove_num_leaves` fail anyways. Maybe we can consider failing here already and returning `Err(Error::InconsistentStore)` without calling `prove_num_leaves`.

primitives/mmr/src/mmr/proof.rs
```diff
@@ -6,6 +6,31 @@ use crate::mmr::peaks::PeakIterator;
 use crate::mmr::position::{leaf_number_to_index, Position};
 use crate::mmr::utils::bagging;
 
+#[derive(Clone, Debug)]
+pub enum SizeProof<H: Merge, T: Hash<H>> {
+    EmptyTree,
+    SingleItem(u64, T),
+    MultipleItem(u64, H, H),
```
nibhar:
```suggestion
    MultipleItems(u64, H, H),
```

validator/src/validator.rs
```diff
@@ -432,8 +477,11 @@ where
     }
 
     fn on_blockchain_history_adopted(&mut self, _: &Blake2bHash) {
-        self.mempool.mempool_clean_up();
-        debug!("Performed a mempool clean up because new history was adopted");
+        // Mempool updates are only done once we can be active.
+        if self.can_be_active() {
```
nibhar, line 483: Note on these:
As I understand it, the inconsistency of the mempool will subsequently fixed by #1475

validator/src/validator.rs
```diff
@@ -816,11 +854,12 @@ where
...
-            let consensus_established = self.consensus.is_established();
-            trace!(?event, consensus_established, "blockchain event");
-            if consensus_established {
-                let latest_hash = event.get_newest_hash();
-                self.on_blockchain_event(event);
+            self.update_can_enforce_validity_window();
+            let can_be_active = self.can_be_active();
+            trace!(?event, can_be_active, "blockchain event");
+            let latest_hash = event.get_newest_hash();
+            self.on_blockchain_event(event);
```
nibhar, line 863: It seems fine to me, but just to make sure:
This change executes `on_blockchain_event(event)` for every event, even when `!can_be_active`. Previously it would only execute with `consensus_established`. Is that by design?

ii-cruz: It is yes

primitives/mmr/src/mmr/mod.rs
paberr: This would only work if we first allocate it as a Vec (via collect).
However, that's not really needed. We thus only lazily evaluate that error.

jsdanielh (approved):

PR #1437, opened 2023-03-14
Tx events subscription

viquezclaudio:
Ability to subscribe to some specific transaction addresses.
This functionality is mainly used by the web client, and should perform those subscription requests to full node and history nodes

PR #1489, opened 2023-03-23
Various WASM client updates

sisou:

- Fix `PlainBlock` generated Typescript types, which were invalid before
- Make all client methods `async`, like they have been in the 1.0 client and to be consistent (because the tx listener method must be async as it does network requests to subscribe at peers)
- Implement `getAccount` and `getAccounts` methods on the client (enabled by #1408 & #1484)
- Implement transaction listeners by subscribing to addresses (enabled by #1437)
- Add `serialize` method to Address for compatibility
- Avoid concurrent access locks by avoiding any `&mut self` methods
- Prevent `addTransactionListener` from failing when called without peers

## Pull request checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4574,6 +4574,7 @@ dependencies = [
  "futures-util",
  "hex",
  "js-sys",
+ "nimiq-account",
  "nimiq-block",
  "nimiq-blockchain-interface",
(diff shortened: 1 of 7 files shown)


jsdanielh:
Rebasing it to get it merged

PR #1492, opened 2023-03-23
consensus: Add a way of getting staking contract elements

jsdanielh:
Add a way of getting staking contract elements to a remote endpoint. For this `RemoteDataStore` was implemented in consensus such that it builds the appropriate `TrieProofRequest` to then verify the received proofs and then deserialize the data appropriately. The implementation differs for now on `wasm` and non `wasm` targets since in `wasm` right now everything runs in a single thread and we can't block it. Therefore, in `wasm` the implementation is for now leaking the `StakingContractStore` instead of just using sync interface of `RemoteDataStore` for the staking contract.

## Pull request checklist

### DIFF
--- a/consensus/src/consensus/remote_data_store.rs
+++ b/consensus/src/consensus/remote_data_store.rs
@@ -1,0 +1,332 @@
+use std::collections::{BTreeMap, HashMap};
+use std::sync::Arc;
+
+use beserial::Deserialize;
+use nimiq_account::{
+    Account, DataStoreReadOps, Staker, StakingContract, StakingContractStore, Tombstone, Validator,
+};
+use nimiq_blockchain_interface::AbstractBlockchain;
+use nimiq_blockchain_proxy::BlockchainProxy;
+use nimiq_keys::Address;
+use nimiq_network_interface::{
+    network::{CloseReason, Network},
+    peer_info::Services,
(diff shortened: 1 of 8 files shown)


consensus/src/consensus/remote_data_store.rs
```diff
@@ -0,0 +1,332 @@
...
+use nimiq_primitives::{key_nibbles::KeyNibbles, policy::Policy};
+
+use crate::messages::RequestTrieProof;
+
+/// The Remote Data Store is a component to remotely request data from the staking
+/// contract such as:
+/// - Validators
+/// - Stakers
+/// - Tombstones
+/// It also serves an utility function to extract any type from the Accounts Trie
```
viquezclaudio: It also serves *as* a utility function...

consensus/src/consensus/remote_data_store.rs
```diff
@@ -0,0 +1,332 @@
...
+            self.blockchain.clone(),
+            &[key.clone()],
+            self.min_peers,
+        )
+        .await?;
+
+        if accounts.len() != 1 {
+            log::error!(
+                len = accounts.len(),
+                "Expected only an account for the staking contract"
```
viquezclaudio: Expected only one account for the staking contract

viquezclaudio (approved):

PR #1495, opened 2023-03-24
Include add_stake transaction notifications

viquezclaudio:


consensus/src/consensus/remote_event_dispatcher.rs
```diff
@@ -253,25 +274,38 @@ impl<N: Network> Future for RemoteEventDispatcher<N> {
...
+                        self.add_notification_receipts(
+                            &txn.recipient,
+                            txn.hash(),
+                            block.block_number(),
+                            &mut peer_receipts,
+                        );
+
+                        // Process staking transaction (which are a special case)
+                        if txn.recipient_type == AccountType::Staking {
+                            // Parse transaction data.
```
jsdanielh:
```suggestion
                            // Parse transaction data
```
For consistency with the rest of the comments?

consensus/src/consensus/remote_event_dispatcher.rs
```diff
@@ -213,6 +215,25 @@ impl<N: Network> RemoteEventDispatcher<N> {
             network_event_rx: network_events,
         }
     }
+
+    // This is a helper function to determine if we need to create notifications
```
jsdanielh:
```suggestion
    /// This is a helper function to determine if we need to create notifications
```
Not strictly necessary but since it is already a comment on the function, I'd suggest to add it

jsdanielh (changes requested):
Also there appears to be a clippy warning

jsdanielh (approved):

consensus/src/consensus/remote_event_dispatcher.rs
```diff
@@ -253,27 +274,38 @@ impl<N: Network> Future for RemoteEventDispatcher<N> {
...
+                            &txn.recipient,
+                            txn.hash(),
+                            block.block_number(),
+                            &mut peer_receipts,
+                        );
+
+                        // Process staking transaction (which are a special case)
+                        if txn.recipient_type == AccountType::Staking {
+                            // Parse transaction data
+                            let proof = IncomingStakingTransactionData::parse(txn).unwrap();
```
sisou, line 296: This is actually the `data` of the transaction, not the proof.

PR #1496, opened 2023-03-24
consensus: Use the block number to generate a transaction proof

jsdanielh:
- Use the block number in the handler of `RequestTransactionsProof` to generate the proof. This avoids race conditions where the client is behind and can't verify the proof for a block that is still unknown to it.
- Allow transaction proof requests of micro blocks that are within a batch of difference of the current block number in the handler of `RequestTransactionsProof`.
- Make `prove_transactions_from_receipts` fail if the block number that is passed to generate the proof is in the future.

## Pull request checklist

### DIFF
--- a/consensus/src/messages/handlers.rs
+++ b/consensus/src/messages/handlers.rs
@@ -369,6 +369,7 @@ impl<N: Network> Handle<N, ResponseTransactionsProof, Arc<RwLock<Blockchain>>>
                 len = hashes.len(),
                 "Requested txn proof from finalized checkpoint block",
             );
+
             // If we were provided a block number corresponding to a checkpoint block, it needs to correspond to the current epoch
             // Otherwise, the requester should use the latest epoch number.
             if self.block_number < election_head {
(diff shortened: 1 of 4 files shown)


consensus/src/messages/handlers.rs
```diff
@@ -405,9 +406,9 @@ impl<N: Network> Handle<N, ResponseTransactionsProof, Arc<RwLock<Blockchain>>>
                 len = hashes.len(),
                 "Requested txn proof from current batch",
             );
-            // If we were provided a block number corresponding to a micro block, it needs to correspond to the current batch
+            // If we were provided a block number corresponding to a micro block, it needs to correspond at least one previous batch
             // Otherwise, the requester should use the latest checkpoint block number
```
sisou:
```suggestion
            // If we were provided a block number corresponding to a micro block, it needs to correspond to at least the previous batch (we allow the previous batch also, to not fail when the client is a few blocks behind and we are already at or over a macro block).
            // If the requested block is older than the previous batch, the requester should use the latest checkpoint block number instead
```

sisou (approved):

PR #1497, opened 2023-03-24
Prove transaction by hash

viquezclaudio:
Currently, we have the possibility to request a transaction proof by hash and block number. In this PR a new variant is introduced, which adds the possibility to request a proof by providing only the transaction hash. In other words, the proving block is determined from the server side and not the client side.

### DIFF
--- a/consensus/src/consensus/consensus_proxy.rs
+++ b/consensus/src/consensus/consensus_proxy.rs
@@ -146,6 +146,7 @@ impl<N: Network> ConsensusProxy<N> {
             .filter(|(hash, block_number)| {
                 block_number > &since_block_height && !ignored_hashes.contains(hash)
             })
+            .map(|(hash, block_number)| (hash, Some(block_number)))
             .collect();
 
         self.prove_transactions_from_receipts(receipts, min_peers)
@@ -158,7 +159,24 @@ impl<N: Network> ConsensusProxy<N> {
         block_number: u32,
         min_peers: usize,
     ) -> Result<ExtendedTransaction, RequestError> {
-        let receipts = vec![(tx_hash, block_number)];
+        let receipts = vec![(tx_hash, Some(block_number))];
+        let mut txs = self
+            .prove_transactions_from_receipts(receipts, min_peers)
+            .await?;
+        match txs.pop() {
+            Some(tx) => Ok(tx),
+            None => Err(RequestError::OutboundRequest(
+                OutboundRequestError::NoReceiver,
+            )),
+        }
+    }
+
+    pub async fn request_transaction_by_hash(
+        &self,
+        tx_hash: Blake2bHash,
+        min_peers: usize,
+    ) -> Result<ExtendedTransaction, RequestError> {
+        let receipts = vec![(tx_hash, None)];
         let mut txs = self
             .prove_transactions_from_receipts(receipts, min_peers)
             .await?;
@@ -192,15 +210,9 @@ impl<N: Network> ConsensusProxy<N> {
 
     pub async fn prove_transactions_from_receipts(
         &self,
-        receipts: Vec<(Blake2bHash, u32)>,
+        receipts: Vec<(Blake2bHash, Option<u32>)>,
         min_peers: usize,
     ) -> Result<Vec<ExtendedTransaction>, RequestError> {
-        // Group transaction hashes by the block number that proves those transactions to reduce the number of requests
-        // There are three categories of block numbers:
-        //  - Finalized epochs: we use the election block number that finalized the respective epoch
-        //  - Finalized batch in the current epoch: We use the latest checkpoint block number
-        //  - Current batch: We use the current head to prove those transactions
-
         let blockchain = self.blockchain.read();
         let election_head = blockchain.election_head();
         let checkpoint_head = blockchain.macro_head();
@@ -212,7 +224,7 @@ impl<N: Network> ConsensusProxy<N> {
 
         if receipts
             .iter()
-            .any(|(_, block_number)| block_number > &current_block_number)
+            .any(|(_, block_number)| block_number.unwrap_or(0) > current_block_number)
         {
             log::error!(
                 head = current_block_number,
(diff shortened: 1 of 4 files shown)


consensus/src/consensus/consensus_proxy.rs
```diff
@@ -210,10 +222,13 @@ impl<N: Network> ConsensusProxy<N> {
...
-        if receipts
-            .iter()
-            .any(|(_, block_number)| block_number > &current_block_number)
-        {
+        if receipts.iter().any(|(_, block_number)| {
+            if let Some(block_number) = block_number {
+                block_number > &current_block_number
+            } else {
+                false
+            }
```
sisou: Can maybe be shortened to `block_number.unwrap_or(0)`:

```suggestion
            block_number.unwrap_or(0) > &current_block_number
```

consensus/src/consensus/consensus_proxy.rs
```diff
@@ -239,33 +254,51 @@ impl<N: Network> ConsensusProxy<N> {
...
+                    } else {
+                        // Third Case: Transanctions from the current batch
+                        hashes_by_block
+                            .entry(Some(current_head.block_number()))
+                            .or_insert(vec![])
+                            .push(hash.clone());
+                    }
                 } else {
-                    // Third Case: Transanctions from the current batch
+                    //Case B: We are not provided a block_number
```
sisou:
```suggestion
                    // Case B: We are not provided a block_number
```

consensus/src/consensus/consensus_proxy.rs
```diff
@@ -239,33 +254,51 @@ impl<N: Network> ConsensusProxy<N> {
...
                     hashes_by_block
-                        .entry(current_head.block_number())
+                        .entry(None)
```
sisou, line 283: Haha, that is possible? An entry for `None`? lol :D

consensus/src/consensus/consensus_proxy.rs
```diff
@@ -239,33 +254,51 @@ impl<N: Network> ConsensusProxy<N> {
...
-                log::debug!(
-                block_number=%block_number,
-                "Performing txn proof request for block number",
-                    );
+                if let Some(block_number) = block_number {
+                    log::debug!(
+                    block_number=%block_number,
+                    "Performing txn proof request for block number");
+                } else {
+                    log::debug!("Performing txn proof request");
```
sisou:
```suggestion
                    log::debug!("Performing txn proof request without block number");
```

consensus/src/messages/handlers.rs
```diff
@@ -304,144 +304,190 @@ impl<N: Network> Handle<N, ResponseChunk, Arc<RwLock<Blockchain>>> for RequestCh
...
+    };
+
+    let block = blockchain
+        .chain_store
+        .get_block_at(proving_block_number, false, None)
+        .ok();
+
+    if let Some(ref block) = block {
+        // We have some extra work for a macro block
+        if Policy::is_macro_block_at(proving_block_number) {
```
sisou: Does this not also return `true` for election blocks, where we don't need it?

```suggestion
        if Policy::is_macro_block_at(proving_block_number) && !Policy::is_election_block(proving_block_number) {
```

consensus/src/messages/handlers.rs
```diff
@@ -304,144 +304,190 @@ impl<N: Network> Handle<N, ResponseChunk, Arc<RwLock<Blockchain>>> for RequestCh
...
+    ResponseTransactionsProof { proof, block }
+}
+
+#[cfg(feature = "full")]
+fn prove_transaction(
+    blockchain: &Arc<RwLock<Blockchain>>,
+    transaction: &Blake2bHash,
+) -> ResponseTransactionsProof {
+    let blockchain = blockchain.read();
+    //let hashes: Vec<&Blake2bHash> = transactions.iter().collect();
```
sisou: Commented line

consensus/src/messages/handlers.rs
```diff
@@ -304,144 +304,190 @@ impl<N: Network> Handle<N, ResponseChunk, Arc<RwLock<Blockchain>>> for RequestCh
...
+        let proving_block_number = if block_number <= election_head {
+            // If the txn is in a finalized epoch, we use the last election block
+            election_head
+        } else if block_number <= macro_head {
+            // If the txn is in a finalized batch in the current epoch, we use the last checkpoint block
+            macro_head
+        } else {
+            // If the txn is in the current batch, we use the head.
+            // This might be problematic if the requester is behind and does not have the current head (yet)
+            current_head
```
sisou: Instead of the current head, could use the `ext_tx.block_number`. It's at most the current_head, otherwise older, and older is better. If it's in the current batch, the requester should have it.

consensus/src/messages/handlers.rs
```diff
@@ -304,144 +304,190 @@ impl<N: Network> Handle<N, ResponseChunk, Arc<RwLock<Blockchain>>> for RequestCh
...
+
+        let block = blockchain
+            .chain_store
+            .get_block_at(election_head, false, None)
+            .ok();
 
-            if let Some(block) = block.clone() {
+        if let Some(ref block) = block {
+            // If it is a checkpoint block, we have some extra work to do
+            if Policy::is_macro_block_at(proving_block_number) {
```
sisou: Same question as before. Does this return `true` for election block numbers?

consensus/src/messages/handlers.rs
```diff
@@ -456,6 +502,38 @@ impl<N: Network> Handle<N, ResponseTransactionsProof, Arc<RwLock<Blockchain>>>
...
+                proof: None,
+                block: None,
+            };
+        }
+
+        // Handle the different cases: if we are provided a block number (to generate the proof) or not
+        if let Some(block_number) = self.block_number {
+            prove_txns_with_block_number(blockchain, &self.hashes, block_number)
+        } else {
+            prove_transaction(blockchain, &self.hashes[0])
```
sisou, line 536: So this only works for requesting one transaction, not for multiple without block. Is that on purpose? Could it also prove multiple hashes, or is that too much of a can of worms to open?

web-client/src/lib.rs
```diff
@@ -877,8 +877,15 @@ impl Client {
                     }
                 }
 
+                let receipts = notification
+                    .receipts
+                    .iter()
+                    .cloned()
```
sisou: Do they have to be cloned? Could you use `into_iter()` instead?

consensus/src/consensus/consensus_proxy.rs
viquezclaudio: Yes :) , they key type is an Option

consensus/src/messages/handlers.rs
viquezclaudio: True! good catch

consensus/src/messages/handlers.rs
viquezclaudio: Because if we support multiple txns, then we would need to group them by finalized epoch/current epoch/current batch and potentially it would require returning multiple blocks, etc..

PR #1507, opened 2023-03-25
lib: Add an optional `advertised_addresses` for `NetworkSettings`

jsdanielh:
Add an optional `advertised_addresses` for `NetworkSettings` in the configuration file  such that the client can use them as the addresses in the `PeerContact` which end up being advertised to other peers.

## Pull request checklist

### DIFF
--- a/lib/src/config/config_file/client.example.toml
+++ b/lib/src/config/config_file/client.example.toml
@@ -28,6 +28,15 @@ seed_nodes = [
         { address = "/dns4/seed1.v2.nimiq-testnet.com/tcp/8443/ws" }
 ]
 
+# Optionally specify address(es) that will be advertised to peers instead of the ones in `listen_addresses`
+#
+# This can be used to advertise the public URL and port that peers should connect to, while
+# `listen_addresses` contains the loopback IP and port that this nodes listens on, which may
+# not be publicly reachable.
+#advertised_addresses = [
+#        "/dns4/my.public.domain.com/tcp/8443/wss",
+#]
+
 # User Agent
 #
 # String that describes what kind of node is running.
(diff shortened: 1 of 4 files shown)


lib/src/config/config_file/client.example.toml
```diff
@@ -28,6 +28,12 @@ seed_nodes = [
         { address = "/dns4/seed1.v2.nimiq-testnet.com/tcp/8443/ws" }
 ]
 
+# Optionally use advertised addresses to advertised other peers addresses different
```
viquezclaudio: Optionally use advertised addresses to *advertise* other peers ...

viquezclaudio (changes requested):

lib/src/config/config_file/client.example.toml
```diff
@@ -28,6 +28,12 @@ seed_nodes = [
         { address = "/dns4/seed1.v2.nimiq-testnet.com/tcp/8443/ws" }
 ]
 
+# Optionally use advertised addresses to advertise different addresses different
+# from the ones in `listen_addresses`
```
sisou:
```suggestion
# Optionally specify address(es) that will be advertised to peers instead of the ones in `listen_addresses`
#
# This can be used to advertise the public URL and port that peers should connect to, while
# `listen_addresses` contains the loopback IP and port that this nodes listens on, which may
# not be publicly reachable.
```

lib/src/config/config_file/client.example.toml
```diff
@@ -28,6 +28,12 @@ seed_nodes = [
...
+#advertised_addresses = [
+#        "/ip4/0.0.0.0/tcp/9100/ws",
```
sisou:
```suggestion
#        "/dns4/my.public.domain.com/tcp/8443/wss",
```

viquezclaudio (approved):

PR #1515, opened 2023-03-27
Do not disconnect if txn proof does not verify

viquezclaudio:
For now, if we request a transaction proof from a peer and it does not verify we log a warning We can then continue requesting other proofs to other peers.


jsdanielh:
Do we need it after the fix in #1517?

PR #1517, opened 2023-03-28
Various fixes for recent transaction proving

sisou:

Together with @jsdanielh we finally figured out why fetching recent transactions doesn't work reliably. It's again because of a mismatch of current head block between the browser client and the history nodes it requests from.

When we changed which block the full node is using to answer a tx-proof request from its current head to the block requested by the client in 058a9ec09e815d720fb201fa892ffa5374d0a538, we failed to see that this also affects micro blocks in the current batch that are older than the node's head block.

Closes #1515.

## Pull request checklist

### DIFF
--- a/consensus/src/messages/handlers.rs
+++ b/consensus/src/messages/handlers.rs
@@ -396,10 +396,8 @@ fn prove_txns_with_block_number(
         .ok();
 
     if let Some(ref block) = block {
-        // We have some extra work for a macro block
-        if Policy::is_macro_block_at(proving_block_number)
-            && !Policy::is_election_block_at(proving_block_number)
-        {
+        // We have some extra work in the current epoch, if the block we are proving is not our current head
+        if block.block_number() > election_head && block.block_number() < current_head {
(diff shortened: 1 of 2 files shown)


consensus/src/messages/handlers.rs
```diff
@@ -465,7 +463,7 @@ fn prove_transaction(
 
         let block = blockchain
             .chain_store
-            .get_block_at(election_head, false, None)
+            .get_block_at(proving_block_number, false, None)
```
sisou, line 466: This one is also important, but I just found it as a drive-by. It's was probably a copy-paste oversight.

viquezclaudio: Yes indeed, we need to use `proving_block_number`

viquezclaudio (approved):
Nice

viquezclaudio:
Do we still want to close #1515 ? 
I don't know if for now it is better to just log the warning without closing the connection, and leave the handling of malicious behaviour as part of hardening?

PR #1240, opened 2023-01-11
Update the `libp2p` dependency to 0.50.0

jsdanielh:
Bumps [libp2p](https://github.com/libp2p/rust-libp2p) from 0.46.1 to 0.50.0.
- [Release notes](https://github.com/libp2p/rust-libp2p/releases)
- [Changelog](https://github.com/libp2p/rust-libp2p/blob/master/CHANGELOG.md)
- [Commits](libp2p/rust-libp2p@v0.46.1...v0.50.0)

This requires to:
- Update some other inner dependencies such as:
  - asynchronous-codec
  - futures-rustls
  - ipnet
- Move out to the network the members of the `NimiqBehaviour` that were previously skipped by the `NetworkBehaviour` derive macro.
- Move away from some deprecated `libp2p` functions.
- Use the new `tcp` and `dns` libp2p features along with the `tokio` feature instead of the old `tcp-tokio` and `dns-tokio` features.     

Also rename the `dns-tcp` feature to `websocket`.
Add the `protobuf-compiler` dependency in the project's `README.md` file and also change the GH actions workflow description files to include it.

## Pull request checklist

- [X] All tests pass. Demo project builds and runs.
- [X] I have resolved any merge conflicts.

PR #1530, opened 2023-03-29
network-libp2p: Finish DHT queries as soon as one result is available

### LINKED ISSUE #1499: GetRecord query result for unknown query ID
In recent runs the WARN `GetRecord query result for unknown query ID` could be observed multiple times. 

Adding to the log the step information as in commit 55a28c64aea331d4f644dd6f5b54766cb84dbb0c shows that those results are multi step results, likely removed when the first step of them was received.

jsdanielh:
Finish `DHT` queries as soon as one result is reported. This replicates the previous behavior where for getting a record from the `DHT` a quorum of one was used. This behavior got changed in #1240 and was provoking several queries received for the same record and thus unexpected error messages such as the ones reported in #1499. A TODO was added such that this policy of using a quorum of 1 is revisited later and in case of changing this a proper implementation of verifying and selecting the best value must be added.

This fixes #1499.

## Pull request checklist

### DIFF
--- a/network-libp2p/src/network.rs
+++ b/network-libp2p/src/network.rs
@@ -93,7 +93,7 @@ pub(crate) enum NetworkAction {
     },
     DhtGet {
         key: Vec<u8>,
-        output: oneshot::Sender<Result<Option<Vec<u8>>, NetworkError>>,
+        output: oneshot::Sender<Result<Vec<u8>, NetworkError>>,
     },
     DhtPut {
         key: Vec<u8>,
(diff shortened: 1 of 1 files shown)


network-libp2p/src/network.rs
```diff
@@ -686,21 +686,46 @@ impl Network {
                 match event {
                     NimiqEvent::Dht(event) => {
                         match event {
-                            KademliaEvent::OutboundQueryProgressed { id, result, .. } => {
+                            KademliaEvent::OutboundQueryProgressed {
```
nibhar, line 689: Unrelated, but we could include the KadmliaEvent type already in NimiqEvent match to reduce the indentations a bit, as they grow quite large.

network-libp2p/src/network.rs
```diff
@@ -686,21 +686,46 @@ impl Network {
...
+                                                .dht
+                                                .query_mut(&id)
+                                                .unwrap()
+                                                .finish();
+                                            if output.send(Ok(record.record.value)).is_err() {
                                                 error!(query_id = ?id, error = "receiver hung up", "could not send get record query result to channel");
                                             }
                                         } else {
-                                            warn!(query_id = ?id, "GetRecord query result for unknown query ID");
+                                            warn!(query_id = ?id, ?step, "GetRecord query result for unknown query ID");
```
nibhar, line 714: Is this in fact still a warn, or expected behaviour?

nibhar (approved):

network-libp2p/src/network.rs
jsdanielh: It would be unexpected, thus the warning

PR #1538, opened 2023-04-05
Add additional conditions to omit block proof

sisou:

https://github.com/nimiq/core-rs-albatross/pull/1526 added a condition to omit network request roundtrips if the block we are trying to prove is already proven by the election head's interlink. But as @fiaxh pointed out (privately), the check was missing the condition when the block to prove is the parent election block, which is referenced by hash in our election head already. I then also added a check for when the block to be proven is the same as our election head (which can happen, as the block number comparison above is `<=`).

## Pull request checklist


viquezclaudio (approved):

fiaxh (approved):

jsdanielh (approved):

jsdanielh:
Rebasing it to get it merged

PR #1547, opened 2023-04-12
Add retire RPC method and log

### LINKED ISSUE #1542: Add `retire_validator` method to the RPC interface
# Description

With the addition of a new state called `retire` in the validators, it has become necessary to implement the corresponding method in the RPC server, interface, and client. These implementations were not considered during the initial implementation of the RPC interface. Therefore, they need to be added now.

onmax:

- Added RPC method to `createRetireValidator` and `sendRetireValidator`
- Logs `RetireValidator` event

### Example of execution using TS

```ts
  const client = new Client(new URL(rpcClient.url));
  await testRpc(client);

  const created = await monkeyChaos.actions.createValidator()
  const validatorKeys: ValidatorKeys = (created as any).validator_keys

  await unlockKey(client, validatorKeys.signing_key)

  await monkeyChaos.actions.sendTx('deactivate', validatorKeys)

  const { next } = await client.logs.subscribe({addresses: [validatorKeys.validator_address]})
  next(console.log) // print logs
  const { next: nextBlock } = await client.block.subscribe({ filter: 'FULL' })
  await new Promise(resolve => {
    nextBlock(block => {
      if((block.data as MacroBlock).isElectionBlock) resolve(block)
    })
  })
  await sleep(2000)
  await monkeyChaos.actions.deleteValidator(validatorKeys)
```

This outputs the log:
```json
{
   "data":{
      "type":"applied-block",
      "inherents":[
         
      ],
      "timestamp":1681332905656,
      "transactions":[
         {
            "hash":"78671e42f19c27a48b1c76a2785f6cc49b5533d761576cd7d437a638ab86a121",
            "logs":[
               {
                  "type":"pay-fee",
                  "from":"NQ81 VAYU Q7NL P8CK M3MG B46L KF8Y A608 F9U8",
                  "fee":0
               },
               {
                  "type":"transfer",
                  "from":"NQ81 VAYU Q7NL P8CK M3MG B46L KF8Y A608 F9U8",
                  "to":"NQ77 0000 0000 0000 0000 0000 0000 0000 0001",
                  "amount":0
               },
               {
                  "type":"retire-validator",
                  "validatorAddress":"NQ81 VAYU Q7NL P8CK M3MG B46L KF8Y A608 F9U8"
               }
            ],
            "failed":false
         }
      ]
   }
}
```

## Pull request checklist


primitives/account/src/logs.rs
```diff
@@ -121,6 +121,9 @@ pub enum Log {
         new_validator_address: Option<Address>,
     },
 
+    #[cfg_attr(feature = "serde-derive", serde(rename_all = "camelCase"))]
+    RetireValidator { validator_address: Address },
```
paberr, line 125:
```suggestion
    RetireValidator { validator_address: Address, newly_inactivated: bool },
```
@ii-cruz and I would suggest to add a bool to the log that is set to true if `retire` also inactivates the validator (when it is called from an active validator).
That way, when a block with a retire transaction is reverted, the listener can know which state the validator goes back into.

sisou: When an active validator is retired, a `DeactivateValidator` log is emitted, too. Both are then emitted.

sisou: https://github.com/nimiq/core-rs-albatross/pull/1547/files#diff-92583b503b91c6afdf466239768975d1071b25a871fdc50ab664652a6b686f6dR538

onmax: What we don't log is `UnparkValidator` when we retire the validator. Even though we unpark it [here](https://github.com/nimiq/core-rs-albatross/pull/1547/files#diff-92583b503b91c6afdf466239768975d1071b25a871fdc50ab664652a6b686f6dR544).

@paberr I already have the fix to your comment in my local. Let me know what you decided given the comments from @sisou 😃

paberr: True, I missed the Deactivate log.

But then indeed we should do three things:

- We should also emit `RetireValidator` in the revert function.
- We should also emit `UnparkValidator` in the commit and revert for the retire (if it was parked).
- We don't need the boolean in `RetireValidator`, so remove it. Sorry about this 😅

onmax: Regarding your bullet points: 
1. Emitting the log `RetireValidator` implies that the validator has changed to `retired` state. In the `revert` function, what we do is removing the state. Therefore, does it make sense to log `RetireValidator` when we change to `validator.retired = false`? I suggest to create a new log called `RevertRetireValidator`.
2. Not sure what you mean by the `commit` function. Are you referring to `retire_validator`?
3. Ok

sisou: 1. No, they are using the same logs. But the fact that the log will be in an array called something with "reverted_logs", will make it clear that it was in fact reverted. See all the `revert_*` methods in `staking_contract/validator.rs`. So just add the same lines pushing the log into the logger in the `revert_retire_validator` method.
2. The "commit" method is the one applying the change (forward) so yes the `retire_validator` method.

sisou (approved):

ii-cruz (approved):
LGTM

onmax:
Ready to merge then. I don't have permission

PR #1552, opened 2023-04-19
validator-network: Use `RwLock` instead of `Mutex`

jsdanielh:
- Use `parking_lot`'s `RwLock` instead of `future`'s `Mutex` to lock the state in the validator network. This is necessary since the       validator network is this unique instance for example in `Tendermint` where all messages goes to. In the case of `Tendermint` this means multiple handel aggregations will use this network for sending multiple concurrent messages and a `Mutex` will block other send opetations to complete only for the sake of reading (most of the time) the validator network internal cache. So this change replaces the `Mutex` for a `RwLock` such that we only block when we need to change the state which will be mostly once at       the beggining of the epoch (calling `set_validators`) or when resolving for the first time a `validator_id` to a `peer_id`.
- Make the `send_to` function in the validator network to only receive one recipient since callers always pass a single element array.
- Add some `rustdoc` to the `ValidatorNetwork` trait.

## Pull request checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4525,6 +4525,7 @@ dependencies = [
  "nimiq-bls",
  "nimiq-network-interface",
  "nimiq-utils",
+ "parking_lot 0.12.1",
  "thiserror",
  "tokio",
  "tracing",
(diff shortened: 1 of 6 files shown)


viquezclaudio (approved):

sisou:
You could add to the description _why_ this change is necessary: Mutex was blocking concurrent read accesses to the network's state, which prevented messages from being sent, which in turn prevented the network from progressing. RwLock however allows concurrent read access, so this unblocks the network.

jsdanielh:
> You could add to the description _why_ this change is necessary: Mutex was blocking concurrent read accesses to the network's state, which prevented messages from being sent, which in turn prevented the network from progressing. RwLock however allows concurrent read access, so this unblocks the network.

Done

PR #1563, opened 2023-04-25
Optimize ZKP verification and rely on result of ZKP component

paberr:
This is a tiny optimization preventing us from verifying ZKPs twice.
The ZKP component always verifies the proofs before forwarding them to the syncer.

## Pull request checklist


sisou (commented):
The method `push_zkp` is not called anywere else, and now both invocations set the `trusted_proof` flag to `true`. Is that because the method _might_ be called from somewhere else in the future?

paberr:
Yes exactly. Alternatively we were considering to remove the proof altogether from the functions and call them `push_block_from_zkp` or something like that.

PR #1567, opened 2023-04-27
Add handel store unittest

viquezclaudio:
Create a new handel store unittest

## Pull request checklist

### DIFF
--- a/handel/src/store.rs
+++ b/handel/src/store.rs
@@ -488,4 +488,139 @@ mod tests {
             }
         }
     }
+
+    #[test]
+    fn it_doesnt_get_better_contribution() {
+        let mut rng = rand::thread_rng();
+        let num_ids = rng.gen_range(8..512);
+
+        let node_id = rng.gen_range(0..num_ids);
+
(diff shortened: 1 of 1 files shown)


handel/src/store.rs
```diff
@@ -488,4 +488,139 @@ mod tests {
...
+            level,
+            Arc::new(TestRegistry {}),
+            (),
+        );
+
+        {
+            let store = store.read();
+            let current_best_contribution = store.best(level);
+
+            if let Some(best) = current_best_contribution {
```
jsdanielh: We should `assert` if this is `None`

handel/src/store.rs
```diff
@@ -488,4 +488,139 @@ mod tests {
...
```
jsdanielh: We should `assert` if this is `None`

handel/src/store.rs
```diff
@@ -488,4 +488,139 @@ mod tests {
...
```
jsdanielh: We should `assert` if this is `None`

jsdanielh (approved):

PR #1571, opened 2023-04-28
Encode genesis supply in genesis block

### LINKED ISSUE #1506: Genesis builder does not encode genesis supply in genesis block
The genesis block's `extra_data` field is supposed to encode the genesis supply in luna. It does not currently.

sisou:

Calculates and encodes the genesis supply from the config into the genesis block.

## Pull request checklist


sisou:
This should have a test. I am also not sure if the sum of all accounts does not count the staking contract balance and _also_ stakers directly, which would sum all staked balance double.

jsdanielh (approved):

PR #1573, opened 2023-05-02
Increase max databases in tests to 20

hrxi:
This fixes the problem that we're currently at the maximum of 10 DBs in the volatile environment, adding any more will cause `DbsFull` errors.


jsdanielh:
Merged in f567549.

PR #1585, opened 2023-05-03
Conflicting default ports in `client.example.toml`

Eligioo:
The default setup of the `client.example.toml` have conflicting port values. It will have the least impact to change the host listing port rather then the `metrics server` port.

**edit** When giving it a second thought, it can break Docker configurations when the listening port is exposed in the container. Thus have more impact.

...

## Pull request checklist

### DIFF
--- a/lib/src/config/config_file/client.example.toml
+++ b/lib/src/config/config_file/client.example.toml
@@ -21,7 +21,7 @@
 [network]
 
 listen_addresses = [
-        "/ip4/0.0.0.0/tcp/9100/ws",
+        "/ip4/0.0.0.0/tcp/8443/ws",
 ]
 
 seed_nodes = [


Eligioo:
@jsdanielh @viquezclaudio thoughts?

sisou:
We usually expose nodes on `8443`, which i think should be the default listen address.

jsdanielh:
> We usually expose nodes on `8443`, which i think should be the default listen address.

Yeah, I'd be more in favor of `8443`

Eligioo:
@jsdanielh updated to 8443

sisou:
Instead of merging in Albatross, we rebase on Albatross. This PR should be one commit, not 4 ;)

jsdanielh (commented):
LGTM but could you rebase `albatross` to make this PR a single commit?

Eligioo:
@jsdanielh squashed it

sisou (approved):

jsdanielh (approved):

PR #1616, opened 2023-05-18
Fix #1249

### LINKED ISSUE #1249: 'tokio-runtime-worker' panicked at 'attempt to multiply with overflow'
Multiplication may fail with an overflow as the result may exceed u32 capacity.   

https://github.com/nimiq/core-rs-albatross/blob/7967eab0a95c8e9a6a734a17ead0387c48a224ba/primitives/src/policy.rs#L249-L274

One way to reproduce it, is using nimiq-rpc-client  and asking for an epoch big enough: 

`nimiq-rpc   -U super -P secret election-block-of 500000
`
This will cause the nimiq server to fail with the following output:
`ERROR panic                | thread 'tokio-runtime-worker' panicked at 'attempt to multiply with overflow': primitives/src/policy.rs:267`

## Proposed solution:

Use checked_mul() instead of just multiplying, whenever values that we can't ensure will  result on a number in the range of u32 are used. 

## Implications: 
checked_mul() returns an Option. Which is None if an overflow happens. This means that any code which relies on this functions to calculate its result, must be refactored

viquezclaudio:
Used checked_mul() in some policy functions to avoid overflows.

## Pull request checklist

- [] All tests pass. The project builds and runs.
- [] I have resolved any merge conflicts.
- [] I have resolved all `clippy` and `rustfmt` warnings.

### DIFF
--- a/primitives/src/policy.rs
+++ b/primitives/src/policy.rs
@@ -264,30 +264,40 @@ impl Policy {
     }
 
     /// Returns the block number of the first block of the given epoch (which is always a micro block).
-    pub fn first_block_of(epoch: u32) -> u32 {
+    /// If the index is out of bounds, None is returned
+    pub fn first_block_of(epoch: u32) -> Option<u32> {
         if epoch == 0 {
             panic!("Called first_block_of for epoch 0");
         }
-        (epoch - 1) * Self::blocks_per_epoch() + 1
+
+        (epoch - 1)
+            .checked_mul(Self::blocks_per_epoch())?
+            .checked_add(1)
     }
 
     /// Returns the block number of the first block of the given batch (which is always a micro block).
-    pub fn first_block_of_batch(batch: u32) -> u32 {
+    /// If the index is out of bounds, None is returned
+    pub fn first_block_of_batch(batch: u32) -> Option<u32> {
         if batch == 0 {
             panic!("Called first_block_of_batch for batch 0");
         }
-        (batch - 1) * Self::blocks_per_batch() + 1
+
+        (batch - 1)
+            .checked_mul(Self::blocks_per_batch())?
+            .checked_add(1)
     }
 
     /// Returns the block number of the election macro block of the given epoch (which is always the last block).
-    pub fn election_block_of(epoch: u32) -> u32 {
-        epoch * Self::blocks_per_epoch()
+    /// If the index is out of bounds, None is returned
+    pub fn election_block_of(epoch: u32) -> Option<u32> {
+        epoch.checked_mul(Self::blocks_per_epoch())
     }
 
     /// Returns the block number of the macro block (checkpoint or election) of the given batch (which
     /// is always the last block).
-    pub fn macro_block_of(batch: u32) -> u32 {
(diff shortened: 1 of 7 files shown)


primitives/src/policy.rs
```diff
@@ -264,30 +264,37 @@ impl Policy {
...
+    pub fn first_block_of(epoch: u32) -> Option<u32> {
         if epoch == 0 {
             panic!("Called first_block_of for epoch 0");
         }
-        (epoch - 1) * Self::blocks_per_epoch() + 1
+
+        match (epoch - 1).checked_mul(Self::blocks_per_epoch()) {
+            Some(election) => election.checked_add(1),
+            None => None,
+        }
```
jsdanielh:
```suggestion
        (epoch - 1).checked_mul(Self::blocks_per_epoch())?.checked_add(1)
```

primitives/src/policy.rs
```diff
@@ -264,30 +264,37 @@ impl Policy {
...
     }
 
     /// Returns the block number of the first block of the given batch (which is always a micro block).
```
jsdanielh, line 278: Update documentation

primitives/src/policy.rs
```diff
@@ -264,30 +264,37 @@ impl Policy {
     }
 
     /// Returns the block number of the first block of the given epoch (which is always a micro block).
```
jsdanielh, line 266: Update documentation

primitives/src/policy.rs
```diff
@@ -264,30 +264,37 @@ impl Policy {
...
-    pub fn first_block_of_batch(batch: u32) -> u32 {
+    pub fn first_block_of_batch(batch: u32) -> Option<u32> {
         if batch == 0 {
             panic!("Called first_block_of_batch for batch 0");
         }
-        (batch - 1) * Self::blocks_per_batch() + 1
+        match (batch - 1).checked_mul(Self::blocks_per_batch()) {
+            Some(checkpoint) => checkpoint.checked_add(1),
+            None => None,
+        }
```
jsdanielh: (batch - 1).checked_mul(Self::blocks_per_batch())?.checked_add(1)

primitives/src/policy.rs
```diff
@@ -264,30 +264,37 @@ impl Policy {
...
     }
 
     /// Returns the block number of the election macro block of the given epoch (which is always the last block).
```
jsdanielh, line 290: Update documentation

primitives/src/policy.rs
```diff
@@ -264,30 +264,37 @@ impl Policy {
...
-    pub fn election_block_of(epoch: u32) -> u32 {
-        epoch * Self::blocks_per_epoch()
+    pub fn election_block_of(epoch: u32) -> Option<u32> {
+        epoch.checked_mul(Self::blocks_per_epoch())
     }
 
     /// Returns the block number of the macro block (checkpoint or election) of the given batch (which
     /// is always the last block).
```
jsdanielh, line 297: Update documentation

jsdanielh (approved):

PR #1635, opened 2023-06-07
Add a default way to format imports

hrxi:
This will allow us to forget entirely about it and not require manual changes anymore. Unfortunately, these configs require nightly, so they need to be formatted using `cargo +nightly fmt` if you are on stable by default.


hrxi:
I don't care too much about the exact settings.

I see multiple options to group the imports: https://rust-lang.github.io/rustfmt/?version=v1.5.1#imports_granularity

It's either by crate (currently in this PR), module, or some other possibilities.

jsdanielh (approved):

PR #1637, opened 2023-06-07
Fix usage of lost_rewards set in macro blocks

### LINKED ISSUE #1274: Block production builds on wrong assumptions
Our block production (and the reward distribution) might build on wrong assumptions.

https://github.com/nimiq/core-rs-albatross/blob/576c9f4dd9e8e353d5aaed5d3ad27a6b54744ca3/block-production/src/lib.rs#L295-L305

The code assumes that the state has already been updated, which is *not* the case.
We think that using the `previous_lost_rewards` is actually what we want to do despite the comment, though.

fiaxh:
- Include current lost_rewards instead of previous one when producing a block
- When verifying macro blocks, verify the staking-contract-related fields in blocks before clearing the stalking contract (and use the current sets)

### DIFF
--- a/block-production/src/lib.rs
+++ b/block-production/src/lib.rs
@@ -274,18 +274,10 @@ impl BlockProducer {
         let staking_contract = blockchain.get_staking_contract();
 
         // Calculate the disabled set for the current validator set.
-        // TODO: Update comment, see issue #1274.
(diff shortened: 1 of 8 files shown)


paberr (approved):
LGTM

PR #1643, opened 2023-06-13
Reward calculation fix with disabled slots

sisou:

> **Note**
> This PR should be reviewed by individual commit. Explainer of every commit below:

While investigating why my validator consistently receives exactly 0 luna reward transactions in the current epoch, I stumbled upon a combination of two issues. The first is already fixed by #1637, which was that the wrong lost-rewards and disabled-slots were included in and shown for blocks, namely the lists of the previous batch **and epoch**. That brings us to the next issue I found which this PR addresses:
> The `previous_disabled_slots` are **not** the ones of the previous batch as their explainer comment and association with `previous_lost_rewards` suggests. Instead `current_disabled_slots` and therefore `previous_disabled_slots` are only reset/overwritten on **epoch finalization**. 

### 1st commit

To make this differentiation clear, in the first commit of this PR, I have renamed these two properties and their getters in the staking contract to `previous_batch_lost_rewards` and `previous_epoch_disabled_slots` instead.

### 2nd commit

The second commit then fixes the reward calculation (which is the second issue I found) to only use the `previous_epoch_disabled_slots` for the first batch of an epoch (which still pays out rewards for the last batch of the previous epoch). The following batches will then correctly use the `current_disabled_slots` to omit rewards.

The reason why my validator is not receiving rewards in this epoch, although it was not slashed, is that it was assigned a slot that was slashed and therefore disabled in the previous epoch, but since the reward-calculation always uses the previous epoch's disabled slots, no reward is paid the whole current epoch.

### 3rd commit

The third commit then removes the removal of a validator from this `previous_epoch_disabled_slots` set during unparking, as I think it's not necessary (and unparking should also not change the history of the previous epoch at all). I am open to remove this last commit though, if you think the behavior was correct. 

## Pull request checklist

### DIFF
--- a/blockchain/src/blockchain/inherents.rs
+++ b/blockchain/src/blockchain/inherents.rs
@@ -162,7 +162,7 @@ impl Blockchain {
             state
                 .current_slots
                 .as_ref()
-                .expect("Slots for last batch are missing")
+                .expect("Slots for current batch are missing")
         };
 
         // Calculate the slots that will receive rewards.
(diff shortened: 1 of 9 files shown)


paberr (approved):
LGTM
If I recall correctly, the lost rewards set is needed to prevent a validator from getting the reward in the batch it was slashed even if they unpark, right?

ii-cruz (approved):

PR #1647, opened 2023-06-15
Remove parked state

ii-cruz:
With automatic reactivation, the deactivated state became equivalent to the park state. This removes the park state for the validator.

## Pull request checklist


rpc-interface/src/blockchain.rs
```diff
@@ -97,9 +97,6 @@ pub trait BlockchainInterface {
         &mut self,
     ) -> RPCResult<SlashedSlots, BlockchainState, Self::Error>;
 
-    async fn get_parked_validators(&mut self)
```
jsdanielh, line 100: Shouldn't we change it for `get_inactive_validators`?

metrics-server/src/chain.rs
```diff
@@ -24,16 +24,11 @@ impl BlockMetrics {
...
 
-        let bc = blockchain.clone();
+        let bc = blockchain;
         let closure = NumericClosureMetric::new_gauge(Box::new(move || {
             bc.read().get_staking_contract().active_validators.len() as i64
         }));
         sub_registry.register("active_validators", "Number of active validators", closure);
-
-        let closure = NumericClosureMetric::new_gauge(Box::new(move || {
-            blockchain.read().get_staking_contract().parked_set.len() as i64
```
jsdanielh: Shouldn't we change it in favor of `inactive_validators`?

rpc-interface/src/blockchain.rs
ii-cruz: We no longer have an easy way to list the validators that are inactive, it would be a very expensive task to iterate over all validators to them list them.

metrics-server/src/chain.rs
ii-cruz: Same reason as above

jsdanielh (approved):

PR #1658, opened 2023-06-23
Jailing, 2-step unstake, Slash and Penalty

ii-cruz:
### Jailing
After being "slashed", a validator goes to "jail" and is forced to be inactive for a certain time frame.
During this period, the validator cannot be reactivated.

### 2-step unstake
Stakers need to inactivate stake before being able to withdraw it.
Inactive stake is locked up for a certain time frame before becoming available.

Re-delegation requires the whole stake to be inactive first.

The jailing also applies to the staker.
Thus a staker can only re-delegate or unstake after the jail period and the lock-up period for inactive stake have passed.

### Slashing and Penalizing
The old slashing is renamed to penalizing. Penalties are applied to slots and deactivate the validator. The penalties are a consequence of skip blocks.
Slashing is applied to all validator slots and jails the validator. Slashes are a consequence of forks (other malicious behavior still needs to be integrated).

### Clean up of bit sets in the staking contract and macro block
Inside the staking contract, the sets are organized in a separate struct. Additionally, the macro block now only contains the necessary information for syncing.

## Pull request checklist


primitives/account/src/account/staking_contract/receipts.rs
```diff
@@ -15,6 +15,7 @@ pub struct SlashReceipt {
     pub newly_deactivated: bool,
     pub newly_disabled: bool,
     pub newly_lost_rewards: bool,
+    pub old_jail_release: Option<u32>,
```
jsdanielh, line 38: Could you add `rustdoc` to the members?

primitives/account/src/account/staking_contract/receipts.rs
```diff
@@ -27,6 +28,22 @@ pub struct UpdateValidatorReceipt {
 }
 convert_receipt!(UpdateValidatorReceipt);
 
+#[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
+pub struct JailValidatorReceipt {
```
jsdanielh, line 60: Could you add `rustdoc` to the `struct` and its members?

primitives/account/src/account/staking_contract/receipts.rs
```diff
@@ -48,6 +65,7 @@ pub struct DeleteValidatorReceipt {
     pub reward_address: Address,
     pub signal_data: Option<Blake2bHash>,
     pub inactive_since: u32,
+    pub jail_release: Option<u32>,
```
jsdanielh, line 110: Could you add `rustdoc` to the `struct` and its members?

primitives/account/src/account/staking_contract/receipts.rs
```diff
@@ -56,3 +74,17 @@ pub struct StakerReceipt {
     pub delegation: Option<Address>,
 }
 convert_receipt!(StakerReceipt);
+
+#[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
+pub struct SetInactiveStakeReceipt {
```
jsdanielh, line 126: Could you add `rustdoc` to the `struct` and its members?

primitives/account/src/account/staking_contract/receipts.rs
```diff
@@ -56,3 +74,17 @@ pub struct StakerReceipt {
...
+    pub old_inactive_release: Option<u32>,
+    pub old_active_balance: Coin,
+}
+convert_receipt!(SetInactiveStakeReceipt);
+
+#[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
+pub struct RemoveStakeReceipt {
```
jsdanielh, line 137: Could you add `rustdoc` to the `struct` and its members?

web-client/src/transaction.rs
```diff
@@ -663,6 +672,13 @@ pub struct PlainUpdateStakerData {
     pub new_delegation: Option<String>,
 }
 
+#[derive(Clone, serde::Serialize, serde::Deserialize, Tsify)]
+#[serde(rename_all = "camelCase")]
+pub struct PlainSetInactiveStakeData {
```
jsdanielh, line 690: Could you add `rustdoc` to the `struct` and its members?

paberr:
We will add rustdocs as part of a separate PR that will update how the sets are being used.

web-client/src/client/account.rs
```diff
@@ -139,11 +133,20 @@ impl PlainAccount {
...
+    /// The staker's active balance.
     balance: u64,
     /// The address of the validator for which the staker is delegating its stake for. If it is not
     /// delegating to any validator, this will be set to None.
     delegation: Option<String>,
+    /// The staker's inactive balance. Only released inactive balance can be withdrawn from the staking contract.
+    /// Stake can only be re-delegated if the whole balance of the staker is inactive and released
+    /// (or if there was no prior delegation). For inactive balance to be released, the maximum of
+    /// the `inactive_release` and the validator's `jail_release` must have passed.
+    inactive_balance: Coin,
```
sisou: This is a `PlainStaker` object, and `Coin` does not translate cleanly to a number in JSON. Please set this as `u64` like the `balance` field.

sisou (changes requested):

web-client/src/client/account.rs
hrxi: What does `Coin` translate to in JSON? Perhaps it should translate to a simple number.

sisou (changes requested):
Please add a method for setting inactive stake (e.g. `new_set_inactive_stake`) in the web-client: `web-client/src/primitives/transaction_builder.rs`.

web-client/src/client/account.rs
paberr: Didn't #1681 change the JSON serialization of coin to just a simple number?

paberr:
@sisou We addressed both of your comments.

web-client/src/client/account.rs
hrxi: Yea, looks like it: https://github.com/nimiq/core-rs-albatross/pull/1681/commits/a666be6d6deb6900660be084f44960fc6d563fc8#diff-ac8e9d0ded515f48f601114227b21f3b7e850e8a850a7fdc8aad2103f8c8609e.

sisou: Well, then not JSON, but to WASM. Because what I get now when compiling to WASM is these types:
```
export interface PlainStaker {
    balance: number;
    delegation: string | undefined;
    inactiveBalance: Coin;
    inactiveRelease: number | undefined;
}
```
where `Coin` is not defined and thus a type error.

sisou (approved):

jsdanielh:
Rebasing it to get it merged

PR #1668, opened 2023-06-28
Add support for SHA512 in HTLCs

jsdanielh:
Add support for the SHA512 algorithm in HTLCs.

## Pull request checklist

### DIFF
--- a/primitives/transaction/src/account/htlc_contract.rs
+++ b/primitives/transaction/src/account/htlc_contract.rs
@@ -1,4 +1,4 @@
-use std::borrow::Cow;
+use std::{borrow::Cow, str::FromStr};
 
 use log::error;
 use nimiq_hash::{
@@ -71,13 +71,12 @@ impl AccountTransactionVerification for HashedTimeLockedContractVerifier {
     }
 }
 
-#[derive(Clone, Debug, Deserialize, Serialize, Ord, PartialOrd, Eq, PartialEq, Hash)]
-#[serde(tag = "algorithm", content = "hash", rename_all = "lowercase")]
+#[derive(Clone, Debug, Ord, PartialOrd, Eq, PartialEq, Hash)]
 #[repr(u8)]
 pub enum AnyHash {
-    Blake2b(AnyHash32) = 1,
-    Sha256(AnyHash32) = 3,
-    Sha512(AnyHash64) = 4,
+    Blake2b(AnyHash32),
+    Sha256(AnyHash32),
+    Sha512(AnyHash64),
 }
 
 impl Default for AnyHash {
@@ -110,6 +109,60 @@ impl AnyHash {
             AnyHash::Blake2b(hash) => &hash.0,
             AnyHash::Sha256(hash) => &hash.0,
             AnyHash::Sha512(hash) => &hash.0,
+        }
+    }
+}
+
+#[derive(Clone, Debug, Ord, PartialOrd, Eq, PartialEq, Hash)]
+#[repr(u8)]
+pub enum PreImage {
+    PreImage32(AnyHash32),
+    PreImage64(AnyHash64),
+}
+
+impl Default for PreImage {
+    fn default() -> Self {
+        PreImage::PreImage32(AnyHash32::default())
+    }
+}
+
+impl From<Blake2bHash> for PreImage {
+    fn from(value: Blake2bHash) -> Self {
+        PreImage::PreImage32(AnyHash32(value.into()))
+    }
+}
+
+impl From<Sha256Hash> for PreImage {
+    fn from(value: Sha256Hash) -> Self {
+        PreImage::PreImage32(AnyHash32(value.into()))
+    }
+}
+
+impl From<Sha512Hash> for PreImage {
+    fn from(value: Sha512Hash) -> Self {
+        PreImage::PreImage64(AnyHash64(value.into()))
+    }
+}
+
+impl FromStr for PreImage {
+    type Err = nimiq_macros::hex::FromHexError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s.len() == AnyHash32::SIZE * 2 {
+            Ok(PreImage::PreImage32(AnyHash32::from_str(s)?))
+        } else if s.len() == AnyHash64::SIZE * 2 {
+            Ok(PreImage::PreImage64(AnyHash64::from_str(s)?))
+        } else {
+            Err(nimiq_macros::hex::FromHexError::InvalidStringLength)
+        }
+    }
+}
+
+impl PreImage {
+    pub fn as_bytes(&self) -> &[u8] {
+        match self {
+            PreImage::PreImage32(hash) => &hash.0,
+            PreImage::PreImage64(hash) => &hash.0,
         }
     }
 }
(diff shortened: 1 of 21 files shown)


sisou (commented):
I see you are adding an additional byte to the HTLC that just encodes `00` for a 32-byte hash and `01` for a 64-byte hash. Is that really necessary? The length of the hash can be determined from the hash algorithm field in the previous byte, like in Nimiq 1.0. Or is this necessary for Rust serialization?

jsdanielh:
> I see you are adding an additional byte to the HTLC that just encodes `00` for a 32-byte hash and `01` for a 64-byte hash. Is that really necessary? The length of the hash can be determined from the hash algorithm field in the previous byte, like in Nimiq 1.0. Or is this necessary for Rust serialization?

Yeah, unfortunately it's necessary for serialization. But there is another solution which would imply removing the `hashAlgorithm` `enum` and encode it into `AnyHash`. This would be more efficient in terms of the space indeed. Let me come up with a possible implementation of this.

jsdanielh:
@sisou Ok, done. With this implementation the `data` field length will remain invariant in terms of the size for 32 byte hashes. The proof will increase in a byte, but it's better than the previous change where it was increasing in 2.

primitives/transaction/tests/htlc_contract_verify.rs
```diff
@@ -128,70 +127,63 @@ fn it_can_verify_regular_transfer() {
...
     // regular: valid Blake-2b
     let proof = OutgoingHTLCTransactionProof::RegularTransfer {
-        hash_algorithm: HashAlgorithm::Blake2b,
         hash_depth: 1,
-        hash_root: AnyHash::from(<[u8; 32]>::from(
-            Blake2bHasher::default().digest(&[0u8; 32]),
-        )),
-        pre_image: AnyHash::from([0u8; 32]),
+        hash_root: AnyHash::from(Blake2bHasher::default().digest(&[0u8; 32])),
+        pre_image: AnyHash::Blake2b(AnyHash32::from([0u8; 32])),
```
sisou: Technically, the preimage is not necessarily a hash of any kind. It's just random data, not even technically a _hash_ at all. A secret basically.

primitives/transaction/tests/htlc_contract_verify.rs
```diff
@@ -128,70 +127,63 @@ fn it_can_verify_regular_transfer() {
...
     assert_eq!(AccountType::verify_outgoing_transaction(&tx), Ok(()));
 
     // regular: valid SHA-256
     let proof = OutgoingHTLCTransactionProof::RegularTransfer {
-        hash_algorithm: HashAlgorithm::Sha256,
         hash_depth: 1,
-        hash_root: AnyHash::from(<[u8; 32]>::from(Sha256Hasher::default().digest(&[0u8; 32]))),
-        pre_image: AnyHash::from([0u8; 32]),
+        hash_root: AnyHash::from(Sha256Hasher::default().digest(&[0u8; 32])),
+        pre_image: AnyHash::Sha256(AnyHash32::from([0u8; 32])),
```
sisou: Should also add a test for sha512 I think.

rpc-interface/src/consensus.rs
```diff
@@ -114,7 +114,6 @@ pub trait ConsensusInterface {
         htlc_recipient: Address,
         hash_root: AnyHash,
         hash_count: u8,
-        hash_algorithm: HashAlgorithm,
```
sisou, line 117: How are these RPC methods going to work? Does the `hash_root` field (hex string over JSON) now have to be prefixed by a byte representing the algo, so needs to be 33 bytes long?

rpc-interface/src/types.rs
```diff
@@ -95,16 +93,7 @@ impl FromStr for ValidityStartHeight {
...
-
-#[allow(clippy::from_over_into)]
-impl Into<HTLCContractHashAlgorithm> for HashAlgorithm {
-    fn into(self) -> HTLCContractHashAlgorithm {
-        match self {
-            HashAlgorithm::Blake2b => HTLCContractHashAlgorithm::Blake2b,
-            HashAlgorithm::Sha256 => HTLCContractHashAlgorithm::Sha256,
-        }
-    }
+    Sha512 = 4,
```
sisou, line 96: `HashAlgorithm` is unused now?

rpc-interface/src/consensus.rs
jsdanielh: No, this should work the same way we send `enums`. A tag field must be inserted to specify the `hashAlgorithm`. This is why I added [this](https://github.com/nimiq/core-rs-albatross/pull/1668/files#diff-d1b20c99015ddf82b249f66f2236dd9541bfdbba5b728f8eae881b34bdff1436R73).

rpc-interface/src/types.rs
jsdanielh: Its actually used by the `rpc-client` [here](https://github.com/nimiq/core-rs-albatross/pull/1668/files#diff-c37ba4db67fea5930db56228ef619010a93f6144fb2d4427c61e8992d5a4346bR10).

primitives/transaction/tests/htlc_contract_verify.rs
jsdanielh: Yeah, this is not elegant. One solution could be to use another type for the `pre_image` instead of `AnyHash`.

rpc-interface/src/consensus.rs
sisou: OK, I tested this. And while I think that it's not very ergonomic, it does work.

Still, for the `preimage` setting the hash algo doesn't make much sense usually.

jsdanielh: @sisou Moved the `pre_image` to a separate type

primitives/transaction/src/account/htlc_contract.rs
```diff
@@ -269,52 +323,219 @@ impl OutgoingHTLCTransactionProof {
...
+            if serializer.is_human_readable() {
+                match self {
+                    PreImage::PreImage32(hash) => Serialize::serialize(hash, serializer),
+                    PreImage::PreImage64(hash) => Serialize::serialize(hash, serializer),
+                }
+            } else {
+                let mut state = serializer.serialize_struct("PreImage", PREIMAGE_FIELDS.len())?;
+                match self {
+                    PreImage::PreImage32(pre_image) => {
+                        state.serialize_field(PREIMAGE_FIELDS[0], &0u8)?;
```
sisou: I believe it's usual for the byte to express the length of the following field. So this number could be `32u8`, and for `PreImage64` it could be `64u8`. I think that makes more sense than `0` and `1`.

rpc-interface/src/consensus.rs
```diff
@@ -138,7 +138,7 @@ pub trait ConsensusInterface {
         wallet: Address,
         contract_address: Address,
         recipient: Address,
-        pre_image: AnyHash,
+        pre_image: PreImage,
```
sisou, line 141: So do I understand the deserializer for `PreImage` correctly, that in the JSON-RPC call just a regular HEX string is now expected for the `pre_image` field?

jsdanielh: Exactly!

primitives/transaction/src/account/htlc_contract.rs
jsdanielh: Ok, yeah, for vectors it is true. I will change it

jsdanielh: Done @sisou

sisou (approved):

PR #1688, opened 2023-07-06
Add RPC methods to dump the blockchain state

jsdanielh:
- Add method to get an `Account`s chunk
  - Add method in the full blockchain to get a chunk of `Account`s.
  - Add `AccountsChunk` to return a chunk of accounts including the `end_key` for iteration purposes and the decoded `Account`s and their addresses.
  - Add a RPC method to query all of the accounts in the accounts tree along with annotations that it is currently very expensive computationally.
- Use `Accounts` to get a proof for accounts
  - Change the blockchain `get_accounts_proof` method to use an `Accounts` method instead of accesing directly the inner `AccounsTrie` for better encapsulation.
  - Add a method on `Accounts` to get a proof from a set of keys.
  - Add `rustdoc` documentation to the blockchain `get_accounts_proof` function and the chain of calls up to `AccountsTrie`.
- Add interface for getting all validators
  - Add an interface to the staking contract to be able to query all validators. This was added with the warning that the query requires the blockchain read lock and its computationally expensive.
  - Add RPC method to get all validators in the staking contract along with the warning on the expensive operation it is.

## Pull request checklist

### DIFF
--- a/blockchain/src/blockchain/accounts.rs
+++ b/blockchain/src/blockchain/accounts.rs
@@ -1,11 +1,22 @@
-use nimiq_account::{Accounts, BlockLogger, BlockState, TransactionOperationReceipt};
+use nimiq_account::{Account, Accounts, BlockLogger, BlockState, TransactionOperationReceipt};
 use nimiq_block::{Block, BlockError, SkipBlockInfo};
 use nimiq_blockchain_interface::PushError;
-use nimiq_database::{traits::Database, WriteTransactionProxy};
+use nimiq_database::{traits::Database, TransactionProxy, WriteTransactionProxy};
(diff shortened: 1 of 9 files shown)


primitives/src/trie/trie_proof_node.rs
```diff
@@ -65,6 +65,7 @@ impl TrieProofNode {
     }
 }
 
+#[derive(Debug)]
```
Eligioo: Still necessary?

jsdanielh: You're right. It is no longer needed. Just removed it

viquezclaudio (approved):
LGTM

PR #1695, opened 2023-07-11
genesis-builder: Fix initial supply calculation

jsdanielh:
- Fix initial supply calculation in the `genesis-builder` crate.
- Use tracing style for logging in the `genesis-builder` crate.

## Pull request checklist


sisou:
What was wrong with the initial supply number?

jsdanielh:
> What was wrong with the initial supply number?

So far I haven't seen that the number is incorrect but with the current way of calculating it, it may lead to incorrect results if iterating `accounts` we hit a non-account node such as anything within a contract.

viquezclaudio (approved):

PR #1708, opened 2023-07-15
Parse HistoryStore's last_leaf_table key from big endian for usage

sisou:

`u32`s converted to big-endian with `.to_be()` are not equal to their not-converted counterparts, as evident by [this Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=57ebe4dfdb90b6b7a220545b87f03e62).

So to be able to use the key of the `last_leaf_table` in comparisons and also to calculate epoch numbers, it must be parsed with `u32::from_be()`. Alternatively, we can replace the previous usage of `.to_be()` with `.to_le()` to convert to low-endian, which has the same effect I believe. But I think, using `u32::from_be()` is more stable.

## Pull request checklist

### DIFF
--- a/blockchain/src/history/history_store.rs
+++ b/blockchain/src/history/history_store.rs
@@ -112,12 +112,12 @@ impl HistoryStore {
         match cursor.seek_range_key::<u32, u32>(&block_number.to_be()) {
             // If it exists, we simply get the last leaf index for the block. We increment by 1
             // because the leaf index is 0-based and we want the number of leaves.
-            Some((n, i)) if n == block_number => i + 1,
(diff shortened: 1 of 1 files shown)


paberr:
Instead of converting to/from BE all the time, the store should simply use `UINT_KEYS`.
See here: https://github.com/nimiq/core-rs-albatross/blob/0478de393d94fb4dbea4d4b54d92dacd5340dbeb/blockchain/src/chain_store.rs#L41

paberr:
Ok, we've decided to go with this non-breaking change for now and then implement the proper fix for the next reset.

paberr (approved):

PR #1707, opened 2023-07-14
Sync fixes

paberr:
## What's in this pull request?

- Fixing an off-by-one error in the verification of subsequent missing blocks requests.
- Re-requesting missing blocks after failures.
- The validator cached a flag for whether it can verify the transaction validity window and never updated it after being true. It now updates the flag when necessary.
- The calculation of the transaction validity window flag in the blockchain used a history store function, whose comment promised functionality that was not correctly implemented. @nibhar fixed this.
- The history store is using `u32`s in BE format specifically. There was a bug in the conversion that @sisou fixed.

This fixes #1692

## Pull request checklist

- [x] All tests pass. The project builds and runs.
- [x] I have resolved any merge conflicts.
- [x] I have resolved all `clippy` and `rustfmt` warnings.

PR #1710, opened 2023-07-17
History store integer keys

paberr:
This PR removes manual big-endian conversions for the history store in favor of using the integer keys flag of mdbx.
It is based on #1707 .

**This PR is breaking the database structure and thus should only be merged with the next reset.**

## Pull request checklist


sisou:
Does it actually affect the protocol and requires all nodes to update, or can this change be made by nodes individually, just by clearing their database?

jsdanielh (approved):

PR #1719, opened 2023-07-25
Fix SizeProof in MMR and add tests

ii-cruz:
Fixes the proving of number of leaves on the history store.

## Pull request checklist


sisou:
Does this fix have any effect on syncing or networking? Or is it a purely internal thing in a node?

ii-cruz:
This is not used right now. The idea would be to use it for switching between light and macro syncing.

PR #1730, opened 2023-07-28
Add missing genesis block number init

viquezclaudio:


blockchain/src/blockchain/blockchain.rs
```diff
@@ -118,6 +118,12 @@ impl Blockchain {
             return Err(BlockchainError::InvalidGenesisBlock);
         }
 
+        let mut policy_config = Policy::default();
```
jsdanielh: We need to move this to the beginning

blockchain/src/blockchain/blockchain.rs
```diff
@@ -118,6 +118,12 @@ impl Blockchain {
...
+        // The genesis block number must be set accordingly
+        policy_config.genesis_block_number = genesis_block.block_number();
+
+        let _ = Policy::get_or_init(policy_config);
```
jsdanielh: Could we error out if what we set is not what we got?

jsdanielh (changes requested):
Also, there is a clippy warning

jsdanielh (changes requested):
We need to do this also in `spammer`

viquezclaudio:
Its not needed for the spammer because the spammer is calling `Client = Client::from_config`

jsdanielh:
> Its not needed for the spammer because the spammer is calling `Client = Client::from_config`

Yeah, you're right. But I think I found a place where it is needed: `zkp/examples/prover/prove.rs`

jsdanielh (approved):

jsdanielh (approved):

PR #1739, opened 2023-08-03
policy: Do not panic for epoch 0

jsdanielh:
Do not panic for epoch 0 in the following two functions in policy:
- `epoch_index_at`
- `batch_index_at`

These functions are required to properly return values for the history store to be able to retrieve transactions that are prior the genesis block (in epoch 0).

## Pull request checklist

### DIFF
--- a/primitives/src/policy.rs
+++ b/primitives/src/policy.rs
@@ -199,7 +199,7 @@ impl Policy {
     pub fn epoch_index_at(block_number: u32) -> u32 {
         // Any block before the genesis is considered to be part of epoch 0
         if block_number < Self::genesis_block_number() {
-            panic!("There are no epochs prior to the genesis block");
+            block_number
         } else {
             let blocks_per_epoch = Self::blocks_per_epoch();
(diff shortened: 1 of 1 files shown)


viquezclaudio (approved):

primitives/src/policy.rs
```diff
@@ -199,7 +199,7 @@ impl Policy {
     pub fn epoch_index_at(block_number: u32) -> u32 {
         // Any block before the genesis is considered to be part of epoch 0
         if block_number < Self::genesis_block_number() {
-            panic!("There are no epochs prior to the genesis block");
+            block_number
```
sisou, line 202: This should probably return `0` instead of the input block number? The comment says anything before the genesis block is epoch 0.

sisou (commented):
@jsdanielh

primitives/src/policy.rs
jsdanielh: Not really, the epoch index is a bad name for what it is: _the epoch index is the number of a block relative to the epoch it is in_

PR #1743, opened 2023-08-07
Web-client sender_data adjustments

sisou:

- Fix `sender_data` handling in web-client transaction
- Rename `recipient_data` back to `data` in plain web-client transactions, for backward-compatibility and ease-of-use

## Pull request checklist

### DIFF
--- a/web-client/src/transaction.rs
+++ b/web-client/src/transaction.rs
@@ -446,7 +446,7 @@ impl Transaction {
                     PlainTransactionSenderData::Raw(PlainRawData { raw: raw_data.raw })
                 }
             },
-            recipient_data: {
+            data: {
                 if self.inner.recipient_type == AccountType::Staking {
                     // Parse transaction data
                     let data = IncomingStakingTransactionData::parse(&self.inner).unwrap();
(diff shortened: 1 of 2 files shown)


web-client/src/transaction.rs
```diff
@@ -782,15 +782,15 @@ pub struct PlainTransaction {
...
+    /// The `data` field of a transaction serves different purposes based on the transaction's recipient type.
     /// For transactions to "basic" address types, this field can contain up to 64 bytes of unstructured data.
     /// For transactions that create contracts or interact with the staking contract, the format of this field
     /// must follow a fixed structure and defines the new contracts' properties or how the staking contract is
     /// changed.
-    pub recipient_data: PlainTransactionRecipientData,
-    /// The `sender_data` field serves a purpose based on the transaction's recipient type.
-    /// It is currently only used for extra information in transactions from the staking contract.
-    pub sender_data: PlainTransactionSenderData,
+    pub data: PlainTransactionRecipientData,
```
jsdanielh, line 793: What is the reason to rename it back to `data`?

sisou: IMO the recipient data field is the more used & read of the two, almost exclusively. For that reason, and to make it backward compatible with the JS API, I think it's better when it's named `data`. The other field is still sufficiently named differently IMO.

jsdanielh:
Rebasing it to get it merged

jsdanielh (approved):

PR #1752, opened 2023-08-09
Fix mempool check for unstake

paberr:
The mempool used the wrong balance to check unstake transactions.

## Pull request checklist


sisou (approved):
I think it would make sense to add a test for this.

jsdanielh (approved):

PR #1758, opened 2023-08-10
Fix increasing and decreasing numStakers for a validator

sisou:
This field was not correctly increased/decreased when the setting of inactive balance means that a previous zero active balance now becomes non-zero (and in reverse when reverting). Because when the active balance is zero, the staker is not counted towards the validator's stakers. But when it changes to non-zero, the staker needs to be added to the counter (with `add_staker_to_validator`), not just it's balance added (with `update_stake_for_validator`).

This has the `testnet-restart` label, because the `numStakers` counter affects the accounts tree hash, which will lead to a network split if only some nodes update.

## Pull request checklist


ii-cruz (approved):
LGTM
Perhaps adding a test for this.

ii-cruz:
This was fixed in #1764 instead.

PR #1759, opened 2023-08-10
Allow to set inactive balance with the update-staker functionality

sisou:
When changing delegation with `update_staker`, the pre-condition is that either the active balance delegated to the current validator is zero (and the inactive stake is released), or the current delegation is `None`. The first case is more common, as users mostly want to change delegation without unstaking first.

Currently, this process requires two consecutive transactions: `update_staker()` for changing delegation, then `set_inactive_balance()` to zero to activate the staking balance again.

For convenience, it will be good if the `update_staker()` function can just take care of both of these steps in one.

This PR includes an optional `reactivate_all_stake` parameter for the `update_staker()` method. When true, all previous inactive stake is activated.

This PR needs to be merged after #1765.

## Pull request checklist

### DIFF
--- a/primitives/account/src/account/staking_contract/receipts.rs
+++ b/primitives/account/src/account/staking_contract/receipts.rs
@@ -117,6 +117,10 @@ convert_receipt!(DeleteValidatorReceipt);
 pub struct StakerReceipt {
     // the delegation before this transaction is applied
     pub delegation: Option<Address>,
+    // the active balance before this transaction is applied
+    pub active_balance: Coin,
+    // the inactive release before this transaction is applied
+    pub inactive_release: Option<u32>,
 }
 convert_receipt!(StakerReceipt);
 
--- a/primitives/account/src/account/staking_contract/receipts.rs
+++ b/primitives/account/src/account/staking_contract/receipts.rs
@@ -34,7 +34,7 @@ pub struct JailReceipt {
     pub old_previous_batch_punished_slots: BitSet,
     /// the current batch punished slots of the affected validator before this jail is applied
     pub old_current_batch_punished_slots: Option<BTreeSet<u16>>,
-    // the jail release before this jail is applied
+    /// the jail release before this jail is applied
     pub old_jail_release: Option<u32>,
 }
 convert_receipt!(JailReceipt);
@@ -43,11 +43,11 @@ convert_receipt!(JailReceipt);
 /// these transactions.
 #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
 pub struct UpdateValidatorReceipt {
-    // the signing key before this transaction is applied
+    /// the signing key before this transaction is applied
     pub old_signing_key: SchnorrPublicKey,
-    // the voting key before this transaction is applied
+    /// the voting key before this transaction is applied
     pub old_voting_key: BlsPublicKey,
-    // the reward address before this transaction is applied
+    /// the reward address before this transaction is applied
     pub old_reward_address: Address,
     // the signal data before this transaction is applied
     pub old_signal_data: Option<Blake2bHash>,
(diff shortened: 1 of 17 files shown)


primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -276,30 +276,52 @@ impl StakingContract {
...
+            staker.inactive_release = None;
         }
 
+        // If we are now delegating to a validator we add ourselves to it.
+        if let Some(validator_address) = &staker.delegation {
+            self.register_staker_on_validator(store, validator_address, false);
+            if reactivate_all_stake {
+                self.increase_stake_to_validator(store, validator_address, staker.balance);
+            }
+        }
```
sisou: I believe `increase_stake_to_validator` must also be called when `reactivate_all_stake` is false, but the active stake is not zero. That can happen when delegation was `None`.

```diff
-             if reactivate_all_stake {
+             if !staker.balance.is_zero() {
```

Technically, the condition is not even needed and `increase_stake_to_validator` can just always be called, even when `staker.balance` is `0`.

Add a test where delegation is None but active balance is something, then update delegation to Some, then check the validator balance.

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -316,25 +338,38 @@ impl StakingContract {
...
+        let total_balance = staker.balance + staker.inactive_balance;
 
         // Remove ourselves from the current delegation, if it exists.
-        if staker.delegation.is_some() && !staker.balance.is_zero() {
-            self.remove_staker_from_validator(store, &staker)
-                .expect("inconsistent contract state");
+        if let Some(validator_address) = &staker.delegation {
+            if receipt.reactivate_all_stake {
+                self.decrease_stake_from_validator(store, validator_address, staker.balance)
+            }
```
sisou: Same here, but in reverse. When unregistering a staker from a validator, its balance must always be decreased, too. I'd remove the condition here and always call it. Cannot hurt.

sisou:
Somehow Github is stuck basing this PR on the branch of `albatross` _before_ PRs #1764 and #1765. Can someone fix that, so it only shows 2 commits for this PR?

jsdanielh:
> Somehow Github is stuck basing this PR on the branch of `albatross` _before_ PRs #1764 and #1765. Can someone fix that, so it only shows 2 commits for this PR?

Yeah, this is strange. I'll try closing & opening the PR since the branch looks good to me.

jsdanielh:
> > Somehow Github is stuck basing this PR on the branch of `albatross` _before_ PRs #1764 and #1765. Can someone fix that, so it only shows 2 commits for this PR?
> 
> Yeah, this is strange. I'll try closing & opening the PR since the branch looks good to me.

Looks like it worked.

primitives/account/src/account/staking_contract/receipts.rs
```diff
@@ -117,6 +117,12 @@ convert_receipt!(DeleteValidatorReceipt);
 pub struct StakerReceipt {
     // the delegation before this transaction is applied
     pub delegation: Option<Address>,
+    // the active balance before this transaction is applied
+    pub active_balance: Coin,
+    // the inactive release before this transaction is applied
+    pub inactive_release: Option<u32>,
+    // if the staker should reactivate all its stake
+    pub reactivate_all_stake: bool,
```
jsdanielh:
```suggestion
    /// the active balance before this transaction is applied
    pub active_balance: Coin,
    /// the inactive release before this transaction is applied
    pub inactive_release: Option<u32>,
    /// if the staker should reactivate all its stake
    pub reactivate_all_stake: bool,
```
The one before this one was wrong as well.

primitives/account/src/account/staking_contract/staker.rs
ii-cruz: Great catch!

sisou (commented):
LGTM :+1:

PR #1764, opened 2023-08-15
Fixing the validator's counters for number of stakers

ii-cruz:
This fixes the stall of yesterday.

Changes:
- Made the validator counter increase and decrease only when removing a staker or changing delegation.
- When we inactivate stake we only change the validators balance, not the counter. 
- Added a test of updating the staker with the same delegation.
- Added tests for the validator's staker counter.

The validator's counter for the number of stakers represents the number of stakers that have delegation set to that address.
The validator's stake balance represents the sum of the current active stake its stakers have delegated to the validator.

## Pull request checklist

### DIFF
--- a/primitives/account/src/account/staking_contract/staker.rs
+++ b/primitives/account/src/account/staking_contract/staker.rs
@@ -129,7 +129,7 @@ impl StakingContract {
 
         // If we are delegating to a validator, we need to update it.
         if let Some(validator_address) = &staker.delegation {
-            self.decrease_stake_from_validator(store, &validator_address, staker.balance);
+            self.decrease_stake_from_validator(store, validator_address, staker.balance);
             self.unregister_staker_from_validator(store, validator_address);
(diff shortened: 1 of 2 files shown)


primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -127,13 +128,13 @@ impl StakingContract {
         self.balance -= value;
 
         // If we are delegating to a validator, we need to update it.
-        if staker.delegation.is_some() {
-            self.remove_staker_from_validator(store, &staker)
-                .expect("inconsistent contract state");
+        if let Some(validator_address) = &staker.delegation {
+            self.decrease_stake_from_validator(store, &validator_address, staker.balance);
```
jsdanielh: Clippy warning:
```suggestion
            self.decrease_stake_from_validator(store, validator_address, staker.balance);
```

PR #1828, opened 2023-09-06
pow-migration: Add some integration tests

jsdanielh:
- Add some integration tests
  - Add some integration tests gated under the `pow-migration-tests` feature because they require an external setup: Having a PoW client running and passing the RPC URL using the `POW_CLIENT_URL` environment variable.
    The feature warrants that the tests won't get run by default and only when needed these tests will be run.
- Move some functionality to the library
  - Move some functionality from the `pow-migration` binary into its library such that some of these components can be easily tested from integration tests.
- Do not run test with all features
  - Change GH Actions workflows to only selectively enable required features instead of enabling them all.
    This is necessary to avoid enabling the `pow-migration-tests` when running tests in the CI since they require an external setup.

## Pull request checklist

### DIFF
--- a/pow-migration/src/lib.rs
+++ b/pow-migration/src/lib.rs
@@ -3,8 +3,27 @@ pub mod history;
 pub mod monitor;
 pub mod state;
 
+use std::{path::PathBuf, process::Command, time::Duration};
+
+use nimiq_database::DatabaseProxy;
+use nimiq_genesis_builder::config::GenesisConfig;
+use nimiq_keys::Address;
 use nimiq_primitives::networks::NetworkId;
+use nimiq_rpc::Client;
 use thiserror::Error;
+use tokio::time::sleep;
+
+use crate::{
+    genesis::{
+        get_pos_genesis,
+        types::{PoSRegisteredAgents, PoWRegistrationWindow},
+    },
+    monitor::{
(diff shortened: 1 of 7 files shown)


pow-migration/src/lib.rs
```diff
@@ -82,12 +101,323 @@ pub enum Error {
...
+/// Get the block windows according to the specified network ID.
 pub fn get_block_windows(network_id: NetworkId) -> Result<&'static BlockWindows, Error> {
     match network_id {
         NetworkId::TestAlbatross => Ok(TESTNET_BLOCK_WINDOWS),
         NetworkId::MainAlbatross => Ok(MAINET_BLOCK_WINDOWS),
         _ => Err(Error::InvalidNetworkID(network_id)),
     }
 }
+
+/// Performs the PoS migration from PoS by parsing transactions and state of the PoW
```
viquezclaudio: Performs the PoS migration from PoW...

pow-migration/src/main.rs
```diff
@@ -186,6 +172,7 @@ async fn main() {
         exit(1);
     };
 
+    // Check to see if the client already have consensus
```
viquezclaudio: if the client already has consensus

viquezclaudio (approved):

Issue #1481, opened 2023-03-22
Check that fork proofs are not included in more than one block


ii-cruz:
Currently, we check that a single block does not contain duplicate fork proofs.
However, it seems we do not check whether the same fork proof was part of another block in the reporting window.

Thus, fork proofs could be included more than once to re-park a validator despite an unparking transaction.

If we check this, we need to make sure a validator is able to enforce it (i.e., knows the already included fork proofs in that window).

hrxi:
Fixed by #1882.

PR #1832, opened 2023-09-08
Deduplicate equivocation proofs by block height etc.

hrxi:
As an example, this means that if a validator produces 5 fork blocks at a given height, only one fork proof can be included in a subsequent block, not 5*4/2 proofs as before.

CC #1481.


paberr (approved):

blockchain/src/blockchain/verify.rs
```diff
@@ -218,26 +219,13 @@ impl Blockchain {
...
-                    proof.verify(&validators)?;
-                }
-            }
+            let validators = &self
+                .get_validators_for_epoch(
+                    Policy::epoch_at(equivocation_proof.block_number()),
+                    Some(txn),
+                )
+                .expect("Couldn't calculate validators");
+            equivocation_proof.verify(&validators)?;
```
sisou, line 228: Ups, missed a clippy warning.

PR #1835, opened 2023-09-08
Expose policy info and utility methods to be available in the web-client

Eligioo:
Annotate Policy primitives including the static utility functions with `wasm-bindgen` so they can be used by the `web-client`.

## Pull request checklist

### DIFF
--- a/primitives/src/policy.rs
+++ b/primitives/src/policy.rs
@@ -8,32 +8,34 @@ use wasm_bindgen::prelude::*;
 /// Global policy
 static GLOBAL_POLICY: OnceCell<Policy> = OnceCell::new();
 
-#[cfg_attr(
-    feature = "ts-types",
-    derive(nimiq_serde::Serialize),
-    serde(rename_all = "camelCase"),
-    wasm_bindgen::prelude::wasm_bindgen
-)]
+#[cfg_attr(feature = "ts-types", cfg_eval::cfg_eval, wasm_bindgen)]
 #[derive(Clone, Copy)]
 pub struct Policy {
     /// Length of a batch including the macro block
+    #[cfg_attr(feature = "ts-types", wasm_bindgen(skip))]
     pub blocks_per_batch: u32,
     /// How many batches constitute an epoch
+    #[cfg_attr(feature = "ts-types", wasm_bindgen(skip))]
     pub batches_per_epoch: u16,
     /// Tendermint's initial timeout, in milliseconds.
     ///
     /// See <https://arxiv.org/abs/1807.04938v3> for more information.
+    #[cfg_attr(feature = "ts-types", wasm_bindgen(skip))]
     pub tendermint_timeout_init: u64,
     /// Tendermint's timeout delta, in milliseconds.
     ///
     /// See <https://arxiv.org/abs/1807.04938v3> for more information.
+    #[cfg_attr(feature = "ts-types", wasm_bindgen(skip))]
     pub tendermint_timeout_delta: u64,
     /// Maximum size of accounts trie chunks.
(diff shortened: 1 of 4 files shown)


primitives/src/policy.rs
```diff
@@ -6,6 +6,12 @@ use once_cell::sync::OnceCell;
 /// Global policy
 static GLOBAL_POLICY: OnceCell<Policy> = OnceCell::new();
 
+#[cfg_attr(
+    feature = "ts-types",
+    derive(nimiq_serde::Serialize),
+    serde(rename_all = "camelCase"),
```
sisou: It seems this rename rule has no effect, in the types the properties are still snake_cased.

primitives/src/policy.rs
```diff
@@ -180,8 +186,17 @@ impl Policy {
             .state_chunks_max_size
     }
 
+    #[inline]
+    pub fn get_or_init(policy: Policy) -> Policy {
+        *GLOBAL_POLICY.get_or_init(|| policy)
+    }
+}
+
+#[cfg_attr(feature = "ts-types", wasm_bindgen::prelude::wasm_bindgen)]
```
sisou: Instead of having the long fully qualified import path everywhere, you can conditionally import a package only when a feature is active. That way you can use it with it's short direct name.

sisou (changes requested):

primitives/src/policy.rs
sisou: Sleeping over it, I noticed where our confusion may come from.

The serde attributes are only respected by `tsify`, not by `wasm-bindgen`. Unfortunately, wasm-bindgen doesn't have any rule like that ([issue still open](https://github.com/rustwasm/wasm-bindgen/issues/1818)), so you must add a `js_name` rule to all properties instead :/

sisou:
I fixed a few things, now the only thing remaining is to add getters for the `pub const`s when the "ts-types" feature is enabled.

sisou:
Added the getters for the `const`s. However, now we have two types of getters. `CAPITAL_SNAKE_CASED` and `pascalCased` ones:

image

I am torn between:
1. Changing the `pascalCased` ones to `CAPITAL_SNAKE_CASE`, because that would be consistent with best-practices in JS/TS for static, never-changing (during runtime) variables. This renaming would only affect the JS-glue code, no renaming in Rust is required.
2. Keeping them as-is, to keep them consistent with how they are defined in Rust.

Any opinions?

primitives/src/policy.rs
```diff
@@ -445,16 +501,161 @@ impl Policy {
...
+// wasm_bindgen does not support exposing `pub const` struct fields, so we reimplement those consts
+// as getters when compiling for WASM.
+#[cfg(feature = "ts-types")]
+#[wasm_bindgen]
+#[allow(non_snake_case)]
+impl Policy {
+    /// This is the address for the staking contract. Corresponds to
+    /// 'NQ77 0000 0000 0000 0000 0000 0000 0000 0001'
+    #[cfg_attr(feature = "ts-types", wasm_bindgen(getter = STAKING_CONTRACT_ADDRESS))]
+    pub fn WASM_STAKING_CONTRACT_ADDRESS() -> String {
```
jsdanielh: Can we actually have the function name in lower case (and keeping the `wasm_bindgen` `getter` property in upper case)?

sisou: Yes

sisou: Done in 70eacbe9323e7c8223dde6ff57b78a86cc8fe055

jsdanielh (approved):

sisou (approved):

PR #1863, opened 2023-09-19
Incorporate the block-production crate in the blockchain crate

nibhar:
This PR removes the Block production crate and moves all of the contents unchanged (except for imports and dependencies) into the blockchain crate and the test utils.

While not strictly necessary, this opens up the blockchain to use the block production which previously would have led to a cyclic dependency.


jsdanielh (commented):
We also have to remove it from [here](https://github.com/nimiq/core-rs-albatross/blob/nibhar/remove-block-production-crate/log/src/lib.rs#L20).

nibhar:
> We also have to remove it from [here](https://github.com/nimiq/core-rs-albatross/blob/nibhar/remove-block-production-crate/log/src/lib.rs#L20).

Done

jsdanielh (approved):

jsdanielh:
Rebasing and signing commits

PR #1866, opened 2023-09-20
Various small fixes and web-client features

sisou:

Various fixes for small issues I found during work on #<next PR>, plus two small features for the web-client.
See commits for a list.

## Pull request checklist

### DIFF
--- a/web-client/src/primitives/hash.rs
+++ b/web-client/src/primitives/hash.rs
@@ -1,0 +1,33 @@
+use nimiq_serde::Serialize;
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+pub struct Hash;
+
+#[wasm_bindgen]
+impl Hash {
+    /// Computes a 32-byte Blake2b hash from the input data. Blake2b is used for example to compute a public key's address.
+    #[wasm_bindgen(js_name = computeBlake2b)]
+    pub fn compute_blake2b(data: &[u8]) -> Vec<u8> {
+        nimiq_hash::Hasher::digest(nimiq_hash::Blake2bHasher::default(), data).serialize_to_vec()
+    }
+
+    /// Computes a 32-byte Argon2 hash from the input data.
+    #[wasm_bindgen(js_name = computeArgon2d)]
+    pub fn compute_argon2d(data: &[u8]) -> Vec<u8> {
+        nimiq_hash::Hasher::digest(nimiq_hash::Argon2dHasher::default(), data).serialize_to_vec()
+    }
+
+    /// Computes a 32-byte SHA256 hash from the input data.
+    #[wasm_bindgen(js_name = computeSha256)]
+    pub fn compute_sha256(data: &[u8]) -> Vec<u8> {
+        nimiq_hash::Hasher::digest(nimiq_hash::Sha256Hasher::default(), data).serialize_to_vec()
+    }
+
+    /// Computes a 64-byte SHA512 hash from the input data.
+    #[wasm_bindgen(js_name = computeSha512)]
+    pub fn compute_sha512(data: &[u8]) -> Vec<u8> {
+        nimiq_hash::Hasher::digest(nimiq_hash::sha512::Sha512Hasher::default(), data)
+            .serialize_to_vec()
+    }
+}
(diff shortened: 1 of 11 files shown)


web-client/src/primitives/hash.rs
```diff
@@ -0,0 +1,29 @@
+use nimiq_serde::Serialize;
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+pub struct Hash;
+
+#[wasm_bindgen]
+impl Hash {
+    #[wasm_bindgen(js_name = computeBlake2b)]
+    pub fn compute_blake2b(data: &[u8]) -> Vec<u8> {
```
jsdanielh, line 11: Could you add some small `rustdoc`?

web-client/src/primitives/hash.rs
```diff
@@ -0,0 +1,29 @@
...
+        nimiq_hash::Hasher::digest(nimiq_hash::Blake2bHasher::default(), data).serialize_to_vec()
+    }
+
+    #[wasm_bindgen(js_name = computeArgon2d)]
+    pub fn compute_argon2d(data: &[u8]) -> Vec<u8> {
```
jsdanielh, line 17: Could you add some small `rustdoc`?

web-client/src/primitives/hash.rs
```diff
@@ -0,0 +1,29 @@
...
+        nimiq_hash::Hasher::digest(nimiq_hash::Argon2dHasher::default(), data).serialize_to_vec()
+    }
+
+    #[wasm_bindgen(js_name = computeSha256)]
+    pub fn compute_sha256(data: &[u8]) -> Vec<u8> {
```
jsdanielh, line 23: Could you add some small `rustdoc`?

web-client/src/primitives/hash.rs
```diff
@@ -0,0 +1,29 @@
...
+        nimiq_hash::Hasher::digest(nimiq_hash::Sha256Hasher::default(), data).serialize_to_vec()
+    }
+
+    #[wasm_bindgen(js_name = computeSha512)]
+    pub fn compute_sha512(data: &[u8]) -> Vec<u8> {
```
jsdanielh, line 29: Could you add some small `rustdoc`?

sisou: Haha, I'm tempted to just put `/// small` now everywhere 😅

I don't have my laptop/SSH key now, but I'll formulate the rustdoc here and then you can add it, ok?

web-client/src/primitives/hash.rs
sisou: Do we really need a rustdoc? The method name is pretty self-explanatory?

sisou: > Computes a 32-byte Blake2b hash from the input data. Blake2b is used for example to compute a public key's address.

web-client/src/primitives/hash.rs
sisou: > Computes a 32-byte Argon2 hash from the input data. Argon2d hashes are used for example for [purpose].

Please find out for what, @jsdanielh 😅

web-client/src/primitives/hash.rs
sisou: > Computes a 32-byte SHA256 hash from the input data. SHA256 is a common hash algorithm.

web-client/src/primitives/hash.rs
sisou: > Computes a 64-byte SHA512 hash from the input data.

jsdanielh: > I don't have my laptop/SSH key now, but I'll formulate the rustdoc here and then you can add it, ok?

Yup, no problem

web-client/src/primitives/hash.rs
jsdanielh: We actually don't have a specific usage for it :smile:

jsdanielh (approved):

web-client/src/primitives/hash.rs
sisou: I think it was the PoW hash.

jsdanielh: You are right

PR #1867, opened 2023-09-20
Webauthn signature support

sisou:
**Demo deployment**: https://webauthn.pos.nimiqwatch.com

Support for signing and validating transactions with Webauthn ES256 signatures (the most common **Passkey** type) and also with existing Ed25519 signatures (currently only supported by Yubikey Series 5). 

ES256 is a COSE standard and is defined as ECDSA over P-256 with SHA-256. It is supported by all modern Passkey implementations (Android, iOS) and modern hardware authenticators. It _excludes_ Windows Hello on Windows 10 though, because that only supports RS256, which uses RSA. Windows Hello on Windows 11 supports ES256. 

Passkey platform support: https://passkeys.dev/device-support/

### New structs

- `ES256PublicKey`: ES256 public keys cannot disregard the y-coordinate parity byte in compressed form as I understand it, so these public keys are 33 bytes long (compressed). This alone made a change in proof serialization necessary.
- `ES256Signature`: a signature over the P256 curve, to be verified by `ES256PublicKey`

### Changed structs

- `SignatureProof`: to cover both the existing EdDSA (Schnorr) signature proofs and the new ECDSA (Webauthn) signature proofs, this struct as been extended to take enums of the public key and signature. Additionally, it receives an optional `webauthn_fields` struct field, because webauthn signatures are made over more than the `data` and thus the proof needs to carry this additional information, more information below.

### Webauthn Signature Proofs

Webauthn signatures are made over more than the `data` they ultimately sign. They are instead made over joined information like this: `authenticatorData || sha256(clientDataJSON)`. The actual `data` that is signed is included in `clientDataJSON` as the `challenge` field and base64Url encoded. According to @mar-v-in, authenticators can be limited in the length of this challenge, so instead of putting the whole serialized content of a transaction in it, we instead sign a Blake2b hash of that content, which is actually the transaction hash.

To verify that the signature was made over the right challenge (transaction hash), we need to include the cleartext client data in the proof. Since that is a lot of bytes, we instead include only the information necessary to recreate [`authenticatorData`](https://www.w3.org/TR/webauthn-2/#sctn-authenticator-data) and [`clientDataJSON`](https://www.w3.org/TR/webauthn-2/#dictionary-client-data), which boils down to:

- The `host` part of the URL - the host part includes the full domain and the port (if different from 443 (or 80 for localhost)). From that we can recreate both the `hostname` for the `authenticatorData` (without port) and `origin` for the `clientDataJSON`.
- `authenticatorData` suffix, i.e. the trailing data after the first 32-byte `hostname` hash.
- `clientDataJSON` extra fields, as only 4 fields are defined as mandatory by the standard, but additional fields can be added by authenticators. Google Chrome actually adds an extra field only sometimes with the explicit purpose of triggering signature errors when implementors are not handling additional fields.
- `clientDataJSON` flags, because some authenticators (looking at you, [Android Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=1233616)), encode the `clientDataJSON` non-standard.

From these pieces of information, we can recreate both `authenticatorData` and `clientDataJSON`, allowing us to verify the signature.

## Open issues, decisions and TODOs

- [x] Naming: Instead of calling it `WebauthnPublicKey`, we can name it `ES256PublicKey`, as it more accurately describes the encoded public key and would leave the door open to support other Webauthn public key schemas in the future (e.g. RS256 for Windows Hello) without having to rename it only then.
- [x] Naming: Instead of calling the existing public key now `EdDSAPublicKey`, we can call it `SchnorrPublicKey` (as is already done in some imports), or `Ed25519PublicKey`, or something else.
- [ ] ~Refactor: Instead of changing the name everywhere `PublicKey` changed to `EdDSAPublicKey`, we can rename it back to `PublicKey` in each file's import statement.~
- [x] Reusability: Should the `WebauthnSignatureProof` have an algorithm identifier, to reuse it for other public key schemas in the future? The information besides the public key and signature is identical I believe, so it would be reusable for other public key types as well?
- [x] Consistency: I have not added a `WebauthnSignature`, instead I am reusing the existing ed25519 `Signature` that already exists, since both are 64 bytes and thus (de)serialize the same. Only at the verification am I re-serializing it as an P256 signature. To be consistent, this could be changed to use a P256 signature struct wherever appropriate, but would introduce overhead in casting and enum matching.
- [ ] Security: For P256 cryptography I am using the [p256](https://crates.io/crates/p256) crate, which explicitly says it has never been independently audited. It is however [used widely across crypto rust](https://crates.io/crates/p256/reverse_dependencies).
- [x] Should we support Webauthn signatures for basic transactions, too?
- [x] Does the merkle-path math work with P256 public keys, e.g. for multisig, @paberr? I have omitted adding Webauthn public keys for multisig for now.
- [x] TODO: Improve DX of WASM `SignatureProof.webauthnSingleSig()` by taking the raw `clientDataJSON` bytes and figuring out the flags and extra fields internally, instead of requiring the library user do that. That would also obviate the need to pass the `host` in seperately.
- [ ] TODO: Add getters for `authenticatorData` and `clientDataJSON` on `WebauthnSignatureProof`?

## Pull request checklist


paberr:
> WebauthnPublicKey: ES256 public keys cannot disregard the y-coordinate parity byte in compressed form as I understand it, so these public keys are 33 bytes long (compressed). This alone made a change in proof serialization necessary.

Yes, the public keys take 256 bits, so 32 bytes. Thus, there is no space to put the parity byte into this encoding and we need another byte.

keys/src/address.rs
```diff
@@ -6,7 +6,7 @@ use nimiq_hash::{hash_typed_array, Blake2bHash, Blake2bHasher, Hasher};
 use nimiq_macros::create_typed_array;
 use thiserror::Error;
 
-use crate::{key_pair::KeyPair, PublicKey};
+use crate::{key_pair::KeyPair, EdDSAPublicKey, PublicKey, WebauthnPublicKey};
```
paberr: I agree; we should rename `WebauthnPublicKey` to something specific about the curve/scheme.

keys/src/address.rs
```diff
@@ -159,16 +159,33 @@ impl From<Blake2bHash> for Address {
...
+        Address::from(hash)
+    }
+}
+
+impl<'a> From<&'a WebauthnPublicKey> for Address {
+    fn from(public_key: &'a WebauthnPublicKey) -> Self {
         let hash = Blake2bHasher::default().digest(public_key.as_bytes());
         Address::from(hash)
     }
 }
```
paberr: Is this a problem? We now effectively assign an address to double the amount of possible keys.
We have 2^160 addresses. Previously we had 2^256 keys, thus 2^96 keys per address.
Now it would be 2^257 keys, thus 2^97 keys per address.

keys/src/multisig.rs
```diff
@@ -10,7 +10,7 @@ use nimiq_utils::key_rng::{CryptoRng, RngCore, SecureGenerate};
 use rand::Rng;
 use sha2::{self, Digest, Sha512};
 
-use crate::{KeyPair, PublicKey, Signature};
+use crate::{EdDSAPublicKey, KeyPair, Signature};
```
paberr: Should we rename `KeyPair` and `Signature` to `EdDSA...`?

keys/src/webauthn_public_key.rs
```diff
@@ -0,0 +1,160 @@
...
+
+#[derive(Clone, Copy)]
+#[cfg_attr(feature = "serde-derive", derive(nimiq_hash_derive::SerializeContent))]
+pub struct WebauthnPublicKey(pub p256::EncodedPoint);
+
+impl WebauthnPublicKey {
+    pub const SIZE: usize = 33;
+
+    pub fn verify(&self, signature: &Signature, data: &[u8]) -> bool {
+        let signature = p256::ecdsa::Signature::from_slice(&signature.to_bytes()).unwrap();
```
paberr: I think I prefer a more clear-cut separation between them, i.e., having a separate struct for each sig type.

keys/src/webauthn_public_key.rs
```diff
@@ -0,0 +1,160 @@
...
+    fn cmp(&self, other: &WebauthnPublicKey) -> Ordering {
+        self.0.cmp(&other.0)
+    }
+}
+
+impl PartialOrd for WebauthnPublicKey {
+    fn partial_cmp(&self, other: &WebauthnPublicKey) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
```
paberr: You can probably derive these instead of implementing them manually.

primitives/account/src/account/staking_contract/validator.rs
```diff
@@ -488,7 +488,7 @@ impl StakingContract {
         }
 
         // Check that the signer is correct.
-        if *signer != Address::from(&validator.signing_key) {
+        if *signer != Address::from(&PublicKey::EdDSA(validator.signing_key)) {
```
paberr: Maybe we should implement `Address::from` not only for `PublicKey` but also the different pk types. Then we don't need to wrap the key just for getting the address.

primitives/account/tests/accounts.rs
```diff
@@ -833,10 +834,10 @@ fn it_commits_valid_and_failing_txns() {
         1,
         NetworkId::Dummy,
     );
-    tx.sender_type = AccountType::Basic;
+    // tx.sender_type = AccountType::Basic;
```
paberr, line 853: Why did you comment this?

primitives/transaction/src/account/basic_account.rs
```diff
@@ -41,8 +41,14 @@ impl AccountTransactionVerification for BasicAccountVerifier {
         assert_eq!(transaction.sender_type, AccountType::Basic);
 
         // Verify signer & signature.
-        let signature_proof: SignatureProof =
-            Deserialize::deserialize_from_vec(&transaction.proof[..])?;
+        let signature_proof: SignatureProof = match transaction.format() {
+            TransactionFormat::Basic => SignatureProof::EdDSA(
+                EdDSASignatureProof::deserialize_from_vec(&transaction.proof[..])?,
```
paberr: If we go through with these changes, I think we should go all the way and implement it for basic txns.

@jsdanielh @viquezclaudio What are the implications for changing the format of transactions w.r.t. the old history? It's not a new problem (i.e., we have a solution), right?

primitives/transaction/src/lib.rs
```diff
@@ -135,16 +138,199 @@ impl SignatureProof {
...
+        const ESCAPED_ORIGIN_SLASHES = 1 << 1;
+
+        // const HAS_EXTRA_FIELDS = 1 << 7; // TODO Replace client_data_extra_fields length null byte when no extra fields are present
+    }
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct WebauthnSignatureProof {
+    pub public_key: WebauthnPublicKey,
+    pub merkle_path: Blake2bMerklePath,
```
paberr: Having the Merkle path should already allow to have an address controlled by either an EdDSA or Webauthn key. 🙂

primitives/transaction/src/lib.rs
```diff
@@ -135,16 +138,199 @@ impl SignatureProof {
...
+    pub signature: Signature,
+    pub host: String,
```
paberr: In practical terms, what would the host be? Is it `nimiq.com` for the wallet? Isn't that a privacy concern as one can now identify which wallet has been used?

paberr: Actually, you can probably also infer things about the browser etc.

If we think that this is an issue, I have some thoughts about how to work around it.

web-client/src/primitives/signature.rs
```diff
@@ -25,7 +25,34 @@ impl Signature {
     /// Serializes the signature to a byte array.
     #[wasm_bindgen(js_name = toBytes)]
     pub fn serialize(&self) -> Vec<u8> {
-        self.inner.as_bytes().to_vec()
+        self.inner.to_bytes().to_vec()
+    }
+
+    #[wasm_bindgen(js_name = fromAsn1)]
+    pub fn from_asn1(bytes: &[u8]) -> Result<Signature, JsError> {
```
paberr, line 36: What is `asn1`?

paberr (commented):
Here are a couple of comments

keys/src/address.rs
mar-v-in: Due to the multi sig scheme / merkle path, every key is essentially valid for every address, you "just" have to find the bytes to combine with your public key to make the the merkle root match the address.

What matters is not how many keys there are per address, but the probability of a collision, which is not affected by this.

primitives/transaction/src/lib.rs
mar-v-in: You already leak information about your wallet and browser when connecting to the network through user-agent and origin headers in websocket connections.

web-client/src/primitives/signature.rs
mar-v-in: https://en.wikipedia.org/wiki/ASN.1

keys/src/webauthn_public_key.rs
sisou: Copied it from the existing public key. Will try.

primitives/account/tests/accounts.rs
sisou: I think because it's technically not needed, as `Basic` is the default.

primitives/transaction/src/lib.rs
sisou: Yes. The data over which webauthn makes the signature includes the origin. Host is the origin without protocol, so `wallet.nimiq.com` in our case.

If we would omit any identifying information from the proof, we'd have to include the authenticatorData and clientDataJSON directly, which 1. takes more bytes on average and does not let us verify that the challenge is actually the tx hash.

So if you have ideas how to avoid, please share.

@mar-v-in I'd say there's a difference between the two, as websocket info is only shared with nodes and not put on-chain to be visible by anybody.

web-client/src/primitives/signature.rs
sisou: The serialization format in which signatures are returned by the Webauthn API.

primitives/account/src/account/staking_contract/validator.rs
sisou: It's already implemented, but these occurances seem to be leftovers from an early draft where they were not implemented.

keys/src/address.rs
sisou: Would you prefer `ECDSAPublicKey` for the new name, or `ES256PublicKey`? I tend to the later one, as it describes the whole thing including the key length, not just the scheme. 

In a similar vain, `EdDSAPublicKey` should probably be named `Ed25519PublicKey` instead. Or `SchnorrPublicKey`, as it's already renamed in some imports?

keys/src/multisig.rs
sisou: Or `Ed25519...` like mentioned above?

primitives/transaction/src/lib.rs
sisou: Yes, 1-of-x multisig works with both pubkey types :+1:

keys/src/webauthn_public_key.rs
sisou: I implemented a separate `ES256Signature` in this PR's branch and did the dublication/separation wherever necessary. But I think it's not clean. It also leads to a problem later, when we want to use the `WebauthnSignatureProof` also for Ed25519 keys and signatures (because apparently that is in the spec and is supported by YubiKey Series 5 [I ordered one and will try it]). Then we need to add an additional ident byte in the signature proof that distinguishes the two key types, and also the two signature types.

I implemented another approach in this commit: https://github.com/nimiq/core-rs-albatross/commit/50abce07287f682e8f8720ce9c1ff7da530083d7
It generalizes the `Signature` to be a simple typed array, just like an `Address` is internally represented by. It then has getter methods to get it as a Ed25519 signature or a P256 signature.

What do you think about that solution? I think it works very well, and we can get rid of the duplicate implementation and matching between the two variants, leading to cleaner and easier understood code.

paberr: I checked the commit, but I don't particularly like it either.
The reason is that semantically speaking, the two signatures are constructed entirely differently (on a mathematical level). Also, it is rather a coincidence that both have the same length.

My original idea was to have structs for each variant (`Ed25519PublicKey`, `Ed25519Signature`,...) and then enums unifying those (`PublicKey`, `Signature`,...) so that we can have type safety over which can interact with which.
Also, so far, we've usually been hiding the concrete implementation as much as possible, so that we can swap out libraries if necessary. Thus, methods like `pub fn to_ed25519(&self) -> ed25519_zebra::Signature` should be avoided.

primitives/transaction/src/lib.rs
paberr: As mentioned on our call, we could avoid putting the data on-chain using a zk-proof (in the future).

keys/src/address.rs
sisou: Renamed them now to `Ed25519PublicKey` and `ES256PublicKey` respectively. I find that represents their properties best.

keys/src/webauthn_public_key.rs
sisou: This is now implemented in a single `SignatureProof` that has support for any curve public key and signature, as well as optional webauthn fields. Through the use of a single "type and flags" byte at the start of the serialization, this is efficiently serialized with only one additional byte, compared to before.

primitives/transaction/src/account/basic_account.rs
sisou: I have now added webauthn support also to basic transactions. To not change the serialization format completely, I have changed the manual implementation of the de/serialization for it to read and write the proof "type and flags" byte at the beginning (in front of the sender pubkey) and handle the rest of the de/serialization accordingly. It's a bit messy (at least messier than before), but that's what you get when you want a custom space-saving serialization for basic txs.

wallet/src/multisig_account.rs
```diff
@@ -157,23 +162,21 @@ impl MultiSigAccount {
             Vec::with_capacity(raw_aggregated_signature.len() + raw_aggregated_commitment.len());
         combined.extend_from_slice(&aggregated_commitment.to_bytes());
         combined.extend_from_slice(aggregated_signature.as_bytes());
```
Eligioo: @sisou  A small oversight from my initial implementation that you can pick up in this PR. `raw_aggregated_signature` and `raw_aggregated_commitment` can be used to extend the `combined` slice. The byte representation is now computed twice.

sisou:
I think this PR is complete, ready for real review.

I have deployed the current state to the demo page, it allows signing with ES256 keys (which most Passkey implementations support) and also Ed25519 keys (which AFAIK only Yubikeys Series 5 support):
https://webauthn.pos.nimiqwatch.com

According to @paberr one more optimization can be made: ECDSA signatures (the new ES256 ones) allow extracting the signing public key from the signature when the signed data is known (which it is in our case). So for ES256 the public key could be omitted in the signature proof, which would save 32 bytes per transaction (64 bytes for double-signed staking transactions!).

primitives/transaction/src/lib.rs
```diff
@@ -655,27 +647,67 @@ mod serde_derive {
...
             A: SeqAccess<'de>,
         {
-            let public_key: PublicKey = seq
+            // Read type field to determine public_key and signature algorithm and if webauthn_fields exists
+            let proof_type_and_flags: u8 = seq
                 .next_element()?
                 .ok_or_else(|| serde::de::Error::invalid_length(0, &self))?;
+            let (algorithm, flags) =
+                SignatureProof::parse_type_and_flags_byte(proof_type_and_flags);
+            let public_key: PublicKey = if algorithm == 0 {
```
styppo: It would be better to parse the algorithm into an Enum instead of comparing with magic constants here.

web-client/src/primitives/signature_proof.rs
```diff
@@ -15,15 +18,55 @@ pub struct SignatureProof {
...
+        let js_value: &JsValue = public_key.unchecked_ref();
+        let public_key = PublicKey::try_from(js_value).map_or_else(
+            |_| {
+                ES256PublicKey::try_from(js_value).map_or_else(
+                    |_| Err(JsError::new("Invalid public key")),
+                    |key| Ok(nimiq_keys::PublicKey::ES256(*key.native_ref())),
+                )
+            },
+            |key| Ok(nimiq_keys::PublicKey::Ed25519(*key.native_ref())),
+        )?;
```
styppo:
```suggestion
        let public_key = if let Ok(key) = PublicKey::try_from(js_value) {
            nimiq_keys::PublicKey::Ed25519(*key.native_ref())
        } else if let Ok(key) = ES256PublicKey::try_from(js_value) {
            nimiq_keys::PublicKey::ES256(*key.native_ref())
        } else {
            return Err(JsError::new("Invalid public key"));
        };
```

web-client/src/primitives/signature_proof.rs
```diff
@@ -15,15 +18,55 @@ pub struct SignatureProof {
...
+        let js_value: &JsValue = signature.unchecked_ref();
+        let signature = Signature::try_from(js_value).map_or_else(
+            |_| {
+                ES256Signature::try_from(js_value).map_or_else(
+                    |_| Err(JsError::new("Invalid signature")),
+                    |sig| Ok(nimiq_keys::Signature::ES256(sig.native_ref().clone())),
+                )
+            },
+            |sig| Ok(nimiq_keys::Signature::Ed25519(sig.native_ref().clone())),
+        )?;
```
styppo:
```suggestion
        let signature = if let Ok(sig) = Signature::try_from(js_value) {
            nimiq_keys::Signature::Ed25519(sig.native_ref().clone())
        } else if let Ok(sig) = ES256Signature::try_from(js_value) {
            nimiq_keys::Signature::ES256(sig.native_ref().clone())
        } else {
            return Err(JsError::new("Invalid signature"));
        };
```

web-client/src/primitives/signature.rs
```diff
@@ -47,11 +56,35 @@ impl Signature {
 }
 
 impl Signature {
-    pub fn from_native(signature: nimiq_keys::Signature) -> Signature {
+    pub fn from_native(signature: nimiq_keys::Ed25519Signature) -> Signature {
```
styppo: Consider implementing the `From<Ed25519Signature>` trait for `Signature` instead of having this function.

web-client/src/primitives/signature.rs
```diff
@@ -47,11 +56,35 @@ impl Signature {
...
         Signature { inner: signature }
     }
 
-    pub fn native_ref(&self) -> &nimiq_keys::Signature {
+    pub fn native_ref(&self) -> &nimiq_keys::Ed25519Signature {
```
styppo, line 65: 2 of 3 call sites to this function clone the return value. Consider implementing `Into<Ed25519>` or adding a function like `into_inner(self)` to avoid cloning.

primitives/transaction/src/signature_proof.rs
```diff
@@ -0,0 +1,512 @@
...
+        }
+
+        // Append extra clientData fields before the final closing bracket
+        if !webauthn_fields.client_data_extra_fields.is_empty() {
+            client_data_json = format!(
+                "{},{}}}",
+                &client_data_json[..client_data_json.len() - 1],
+                webauthn_fields.client_data_extra_fields
+            );
+        }
```
styppo: I'm not a big fan of first using the `json!` macro and then fixing up the resulting string manually. I think it would be simpler to construct the JSON string directly using string concatenation and without the `json!` macro.

primitives/transaction/src/signature_proof.rs
```diff
@@ -0,0 +1,512 @@
...
+        // Hash the clientDataJSON
+        let client_data_hash = Sha256Hasher::default().digest(client_data_json.as_bytes());
+
+        // 5. Concat authenticatorData and clientDataHash to build the data signed by Webauthn
+        let signed_data = [authenticator_data.as_slice(), client_data_hash.as_slice()].concat();
+
+        self.do_verify(signed_data.as_slice())
+    }
+
+    fn do_verify(&self, message: &[u8]) -> bool {
```
styppo:
```suggestion
    fn verify_signature(&self, message: &[u8]) -> bool {
```

primitives/transaction/src/signature_proof.rs
```diff
@@ -0,0 +1,512 @@
...
+
+    pub fn parse_type_and_flags_byte(byte: u8) -> (u8, SignatureProofFlags) {
+        // The algorithm is encoded in the lower 4 bits
+        let algorithm = byte & 0b0000_1111;
+        // The flags are encoded in the upper 4 bits
+        let flags = SignatureProofFlags::from_bits_truncate(byte >> 4);
+
+        (algorithm, flags)
+    }
+}
```
styppo: As mentioned, parsing the algorithm into an Enum here is less error-prone and improves locality.

primitives/transaction/src/signature_proof.rs
```diff
@@ -0,0 +1,512 @@
...
+
+        // Check if the suffix contains extra fields
+        if suffix.len() > 1 {
+            // Cut off first comma and last curly brace
+            client_data_extra_fields = suffix[1..suffix.len() - 1].to_string();
+        }
+
+        if origin.contains(r":\/\/") {
+            client_data_flags.insert(WebauthnClientDataFlags::ESCAPED_ORIGIN_SLASHES);
+        }
```
styppo, line 115: Consider using a JSON parser here instead of parsing `client_data_json` by hand. Maybe put the function behind a feature flag to not carry the dependency along everywhere. At least we should take care of the `FIXME`s to not crash the client if the caller sends junk.

sisou:
I have only rebased the branch on top of `albatross` so far, fixed all the merge conflicts. Now waiting for tests to be green before starting with working on the review comments.

web-client/src/primitives/signature.rs
sisou: The reason we clone objects sent to or from JS is that when we move them instead, they become inaccessible null-pointers on the JS side. So the web API is taking care to never move a shared struct, always clone.

primitives/transaction/src/signature_proof.rs
sisou: The problem with JSON parsers is that they delete the backward-slashes that escape the forward slashes in Android Chrome's `origin` field (`https:\/\/`), which are nonetheless relevant to the signature! So when they are removed, it becomes impossible to know if they should be there or not, making signatures from Android Chrome invalid.

I can of course keep searching for a parser that handles it "correctly", or fork one, but since the JSON in client_data_JSON is otherwise standardized, I think we can also parse it manually.

sisou: I have pushed a version with a lightweight JSON parser here: https://github.com/nimiq/core-rs-albatross/commit/fb8a9d87fbeeb1fde42b159a0c71afa090d5f5d8

Please let me know if this is good, then I'll add it to this PR.

jsdanielh:
Rebasing branch to get it merged

PR #1869, opened 2023-09-21
Inferior chain macro block verification

nibhar:
This PR adds the ability to verify a macro block proposal whose predecessor is no longer on the main chain (i.e it was rebranched by a skip block in its place). Prior to this the state verification would fail, as the state was not altered from main chain head to the actual predecessor of the macro block.

Macro blocks with a non main chain predecessor are not invalid and thus their verification must succeed, given the micro blocks leading up to it.

### DIFF
--- a/blockchain/src/blockchain/rebranch_utils.rs
+++ b/blockchain/src/blockchain/rebranch_utils.rs
@@ -1,0 +1,261 @@
+use std::error::Error;
+
+use nimiq_account::{BlockLog, BlockLogger};
+use nimiq_blockchain_interface::{ChainInfo, PushError};
+use nimiq_database::{TransactionProxy, WriteTransactionProxy};
+use nimiq_hash::Blake2bHash;
+use nimiq_primitives::trie::trie_diff::TrieDiff;
+
+use crate::Blockchain;
+
+impl Blockchain {
+    /// Finds the common ancestor between the current main chain in the context of `txn` and the fork chain given by
+    /// its chain info and the block hash.
+    ///
+    /// Returns the common ancestor as .0 and the chain leading to the common ancestor backwards from `block_hash` as .1
+    /// The block with `block_hash` is included.
+    pub(super) fn find_common_ancestor(
+        &self,
+        block_hash: Blake2bHash,
+        chain_info: ChainInfo,
+        diff: Option<TrieDiff>,
+        txn: &TransactionProxy,
+    ) -> Result<
+        (
+            (Blake2bHash, ChainInfo, Option<TrieDiff>),
+            Vec<(Blake2bHash, ChainInfo, Option<TrieDiff>)>,
+        ),
+        PushError,
+    > {
+        // Walk up the fork chain until we find a block that is part of the main chain.
+        // Store the chain along the way
+        let target = chain_info.head.header();
+
+        // Collects the chain on the way back to the common ancestor
+        let mut fork_chain = vec![];
+
+        // Keeps track of the currently investigated block
+        let mut current: (Blake2bHash, ChainInfo, Option<TrieDiff>) =
+            (block_hash, chain_info, diff);
+
+        // Check if the currently checked block is on main chain. If so it is the common ancestor.
+        while !current.1.on_main_chain {
+            // If not keep on moving backwards so get the prev hash
+            let prev_hash = current.1.head.parent_hash().clone();
+
+            // Using the prev hash get the chain info of the previous block
+            let prev_info = self
+                .chain_store
+                .get_chain_info(&prev_hash, true, Some(txn))
+                .expect("Corrupted store: Failed to find fork predecessor while rebranching");
+
+            // Get the prev diff for that block as well.
+            let prev_diff = self
+                .chain_store
+                .get_accounts_diff(&prev_hash, Some(txn))
+                .ok();
+
+            // Store the current block as part of the fork
+            fork_chain.push(current);
+
+            // Update with the information regarding the previous block
+            current = (prev_hash, prev_info, prev_diff);
+        }
+
+        // Check if ancestor is in current batch.
+        if current.1.head.block_number() < self.state.macro_info.head.block_number() {
+            warn!(
+                block = %target,
+                reason = "ancestor block already finalized",
+                ancestor_block = %current.1.head,
+                "Rejecting block",
+            );
+            return Err(PushError::InvalidFork);
+        }
+
+        // Return the ancestor and the part of the chain used to get there.
+        Ok((current, fork_chain))
+    }
+
+    /// Reverts all blocks until the common ancestor given as an argument is reached.
+    /// After that applies all blocks given as target_chain in reverse order or until a block fails
+    /// to be applied.
+    ///
+    /// Returns the reverted chain as `.1` and the block logs as `.2` or the blocks which are on a faulty fork.
+    /// It does _not_ deal with the faulty blocks.
+    pub(super) fn rebranch_to(
+        &self,
+        target_chain: &mut Vec<(Blake2bHash, ChainInfo, Option<TrieDiff>)>,
+        ancestor: &mut (Blake2bHash, ChainInfo, Option<TrieDiff>),
+        write_txn: &mut WriteTransactionProxy,
+    ) -> Result<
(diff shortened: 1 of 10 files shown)


validator/src/lib.rs
```diff
@@ -5,5 +5,5 @@ pub mod aggregation;
 mod jail;
 mod r#macro;
```
paberr, line 6: Is this a typo?

nibhar: Not a typo. 

It is a raw identifier, as `macro` is a keyword. We could rename the file to not having to use it here, but it was not part of this PR.

blockchain/src/blockchain/rebranch_utils.rs
```diff
@@ -0,0 +1,261 @@
...
+                    let remove_chain = vec![block.clone()]
+                        .into_iter()
+                        .chain(target_chain_iter.cloned())
+                        .collect();
+                    return Err(remove_chain);
+                }
+            }
+        }
+
+        // Unset onMainChain flag / mainChainSuccessor on the current main chain up to (excluding) the common ancestor.
```
hrxi:
```suggestion
        // Unset on_main_chain flag / main_chain_successor on the current main chain up to (excluding) the common ancestor.
```

blockchain/src/blockchain/rebranch_utils.rs
```diff
@@ -0,0 +1,261 @@
...
+        for reverted_block in revert_chain.iter_mut() {
+            reverted_block.1.on_main_chain = false;
+            reverted_block.1.main_chain_successor = None;
+
+            self.chain_store
+                .put_chain_info(write_txn, &reverted_block.0, &reverted_block.1, false);
+        }
+
+        // Update the mainChainSuccessor of the common ancestor block.
```
hrxi:
```suggestion
        // Update the main_chain_successor of the common ancestor block.
```

blockchain/src/blockchain/rebranch_utils.rs
```diff
@@ -0,0 +1,261 @@
...
+        ancestor.1.main_chain_successor = Some(target_chain.last().unwrap().0.clone());
+        self.chain_store
+            .put_chain_info(write_txn, &ancestor.0, &ancestor.1, false);
+
+        // Set onMainChain flag / mainChainSuccessor on the fork.
```
hrxi:
```suggestion
        // Set on_main_chain flag / main_chain_successor on the fork.
```

blockchain/src/blockchain/verify.rs
```diff
@@ -296,4 +301,237 @@ impl Blockchain {
...
+        round: u32,
+        valid_round: Option<u32>,
+        signed_data: Vec<u8>,
+        signature: &Signature,
+        signature_only: bool,
+    ) -> Result<MacroBody, PushError> {
+        // If the accounts tree is not complete nothing can be done
+        if !self.accounts_complete() {
+            // Maybe introduce a proper error here. Is this even needed?
+            return Err(PushError::MissingAccountsTrieDiff);
```
hrxi:
```suggestion
            return Err(PushError::IncompleteAccountsTrie);
```

blockchain/src/blockchain/verify.rs
```diff
@@ -296,4 +301,237 @@ impl Blockchain {
...
+        proposed_block: MacroBlock,
+        round: u32,
+        valid_round: Option<u32>,
+        signed_data: Vec<u8>,
+        signature: &Signature,
+        signature_only: bool,
+    ) -> Result<MacroBody, PushError> {
+        // If the accounts tree is not complete nothing can be done
+        if !self.accounts_complete() {
+            // Maybe introduce a proper error here. Is this even needed?
```
hrxi, line 329: Theoretically, the validator isn't active while the trie isn't complete, so this should not happen?

A validator will just vote none if this function returns an error, but that's not a problem in terms of the protocol, right?

validator/src/tendermint.rs
```diff
@@ -330,134 +329,40 @@ where
...
-        }
-
-        // Abort if the blockchain is not in the correct state anymore.
-        let blockchain = self.blockchain.read();
-        if !blockchain.accounts_complete() {
-            return Err(ProposalError::Other);
-        }
+        // verify_macro_block_proposal could create and commit a write transaction, thus take an
+        // upgradable read lock on blockchain.
+        let blockchain = self.blockchain.upgradable_read();
```
hrxi: This can go below so that the lock is only held for the minimum time.

blockchain/src/blockchain/rebranch_utils.rs
```diff
@@ -0,0 +1,261 @@
...
+        // Revert to common ancestor is done.
+
+        // Next, push each block of the target chain.
+
+        // Pushing must happen in reverse.
+        let mut target_chain_iter = target_chain.iter().rev();
+
+        while let Some(block) = target_chain_iter.next() {
+            // Collect logs for the upcoming push.
+            let mut block_logger = BlockLogger::new_applied(
```
hrxi, line 184: Sneaky change, could maybe go into another commit or in the commit message.

blockchain/src/blockchain/verify.rs
```diff
@@ -296,4 +301,237 @@ impl Blockchain {
...
+            .map_err(PushError::BlockchainError)?
+            .header();
+
+        // Get the validator for this round.
+        let mut proposer = self
+            .get_proposer_at(
+                proposed_block.block_number(),
+                round,
+                prev_header.seed().entropy(),
+                None,
```
hrxi: Nit, but not required: We pass `None` for the transaction here and below, if we created a read transaction here, we wouldn't need to open two to get two proposers (also existed before this PR).

validator/src/tendermint.rs
```diff
@@ -330,134 +329,40 @@ where
...
+                log::debug!(?error, ?proposal, "Proposal verification failed",);
+                // Special case for invalid signatures, as that means the proposer is not the signer.
+                // It is needed as the proposer if he produces a faulty block might be subjected to
+                // some sort of punishment or mitigation whereas with an invalid signature it would be
+                // the one relaying the message, who would potentially be subjected to protective measures.
+                if error == PushError::InvalidBlock(BlockError::InvalidJustification) {
+                    ProposalError::InvalidSignature
+                } else {
+                    ProposalError::InvalidProposal
+                }
```
hrxi, line 364: Just an idea, not mandatory: Maybe `PushError` is the wrong error type for `verify_macro_block_proposal`, maybe it could return a `ProposalError` directly?

hrxi (commented):
Looks very good, I only had a couple of nits that aren't required to be fixed and two questions.

blockchain/src/blockchain/verify.rs
nibhar: Correct. Deeming a proposal invalid in the case where you cannot verify its validity is the only safe option. I do not quite know about wether or not a validator with an incomplete tree can consider itself active, as i have not worked on it. I would assume, it is not fit for validating with an incomplete tree, but i would have to check with who worked on that.
This might thus be `unreachable!()`.

validator/src/tendermint.rs
nibhar: `ProposalError` is a fairly unspecific error type. I chose to only convert it here such that it can be logged before the conversion to hopefully help with debugging and for the, albeit highly unlikely chance, it would potentially see use in a different location. 

`ProposalError` is also an export of `nimiq-tendermint` which `nimiq-blockchain` currently only uses as a dev dependency, which I wanted to maintain.

Personally I would rather keep the `PushError` as it also is in line with the remaining functions present in `blockchain/src/blockchain/verify.rs` but changing it is not a big deal if it is considered better form.

blockchain/src/blockchain/verify.rs
nibhar: I will also include the same `txn` in the query for the preceding block.

blockchain/src/blockchain/rebranch_utils.rs
nibhar: Done

hrxi:
Rebased to get it signed and pushed to master: https://github.com/nimiq/core-rs-albatross/commit/08e0d65a344585ef0af1b27dd84a71f306e3f8d0.

PR #1881, opened 2023-10-04
Fix signing of InStaking transaction with web-client's WASM API

### LINKED ISSUE #1879: Invalid Proof in WebClient Transaction::Sign
Signing InStaking transactions is broken in the webclient.

## Problem
Instead of returning the transaction the proof is set in place.
But not on both transactions

## Reproduce
Use the webclient to create a transaction using  
`Nimiq.TransactionBuilder.newCreateValidator`
then use `sign` on this transaction and try to   
send it using the `sendTransaction` of the client

## Change that introduced it
https://github.com/nimiq/core-rs-albatross/pull/1755

## The Code
https://github.com/nimiq/core-rs-albatross/blob/albatross/web-client/src/transaction.rs#L185-L199

sisou:

Fix signing of `InStaking` transactions with the web-client's WASM API.

## Pull request checklist

### DIFF
--- a/web-client/src/lib.rs
+++ b/web-client/src/lib.rs
@@ -10,58 +10,56 @@ mod transaction;
 mod utils;
 
 #[cfg(test)]
-use wasm_bindgen::JsValue;
-#[cfg(test)]
-use wasm_bindgen_test::wasm_bindgen_test;
+mod tests {
+    use wasm_bindgen::JsValue;
+    use wasm_bindgen_test::wasm_bindgen_test;
 
-#[cfg(test)]
-use crate::{
-    address::Address,
-    primitives::{
-        bls_key_pair::BLSKeyPair, key_pair::KeyPair, transaction_builder::TransactionBuilder,
-    },
-};
+    use crate::{
+        address::Address,
+        primitives::{
+            bls_key_pair::BLSKeyPair, key_pair::KeyPair, transaction_builder::TransactionBuilder,
+        },
+    };
 
-#[cfg(test)]
-#[wasm_bindgen_test]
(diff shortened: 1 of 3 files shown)


web-client/src/transaction.rs
```diff
@@ -192,11 +192,14 @@ impl Transaction {
                 builder.sign_with_key_pair(key_pair.native_ref());
                 let mut builder = builder.generate().unwrap().unwrap_basic();
                 builder.sign_with_key_pair(key_pair.native_ref());
-                builder.generate().unwrap()
+                let tx = builder.generate().unwrap();
+                // Set the recipient data to the data with the added signature
+                self.set_recipient_data(tx.recipient_data);
```
sisou, line 197: This is the relevant line that fixes the issue.

web-client/src/lib.rs
```diff
@@ -8,3 +8,60 @@ mod client_configuration;
 mod primitives;
 mod transaction;
 mod utils;
+
+#[cfg(test)]
```
jsdanielh, line 12: I'd add a `mod test` and only gate the module behind the `test` feature:
```
#[cfg(test)]
mod tests {
  use wasm_bindgen::JsValue;
  use wasm_bindgen_test::wasm_bindgen_test;
...
  #[wasm_bindgen_test]
  pub fn it_can_create_and_sign_basic_transactions() {
...
}

sisou: Ah, that's how it works! I wanted to do that, but only tried opening braces.

Will do!

jsdanielh (approved):

PR #1882, opened 2023-10-06
Make extended/historic transactions more precise and include equivocation violations

hrxi:


hrxi:
Still needs some work.

- [x] Remove transaction signatures from the historic transaction store.
- [x] Use the history store's equivocation locators to reject newer proofs for the same violation.
- [x] Figure out whether changing the RPC methods' names is okay.

blockchain/src/history/history_store.rs
```diff
@@ -960,22 +970,24 @@ impl HistoryStore {
...
+                txn.put(
+                    &self.address_table,
+                    &ev.reward_address,
+                    &OrderedHash {
+                        index: index_tx_recipient,
+                        hash: tx_hash,
+                    },
+                );
             }
+            // Do not index equivocation events.
```
nibhar: Why not?

primitives/transaction/src/historic_transaction.rs
```diff
@@ -129,10 +161,8 @@ impl HistoricTransaction {
     pub fn tx_hash(&self) -> Blake2bHash {
         match &self.data {
             HistoricTransactionData::Basic(tx) => tx.hash(),
-            HistoricTransactionData::Inherent(v) => match v {
-                Inherent::Reward { .. } => self.clone().into_transaction().unwrap().hash(),
-                _ => v.hash(),
-            },
+            // TODO: check for freedom of hash prefixes
```
nibhar, line 165: An issue should be created for this, following the merger of this PR otherwise this might get lost.

nibhar (changes requested):
While I think this correctly does what it is supposed to do, I would like the documentation to be improved.

blockchain/src/history/history_store.rs
hrxi: I haven't seen a use case for it yet.

primitives/transaction/src/historic_transaction.rs
hrxi: I think so, too.

blockchain/src/history/history_store.rs
hrxi: Added it to the comment.

nibhar (approved):

hrxi:
Rebased.

primitives/transaction/src/historic_transaction.rs
```diff
@@ -0,0 +1,274 @@
...
+            .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))
+    }
+}
+
+/// An enum specifying the type of transaction and containing the necessary data to represent that
+/// transaction.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize, SerializeContent)]
+pub enum HistoricTransactionData {
+    /// A basic transaction. It simply contains the transaction as contained in the block.
+    Basic(ExecutedTransaction),
```
sisou, line 229: I think calling all transactions (as opposed to inherents) `Basic` is misleading, as transactions themselves are also split into 'basic' and 'extended' formats, and this `Basic` here represents both of those formats.

hrxi: Maybe "normal transaction" would work?

Note that this pull request didn't introduce this terminology, it just kept using it.

sisou: Yes, I already noticed it a while ago.

Maybe just 'transaction' vs 'inherent'?

PR #1884, opened 2023-10-09
web-client: Add basic Merkle Tree and multisig APIs

sisou:

Adds `MerkleTree` and `MerklePath` APIs to the web-client WASM library.

This is useful for basic 1-of-x "multi"sigs.

## Pull request checklist

### DIFF
--- a/web-client/Cargo.toml
+++ b/web-client/Cargo.toml
@@ -49,6 +49,7 @@ nimiq-primitives = { workspace = true, features = ["coin", "networks", "ts-types
 nimiq-serde = { workspace = true }
 nimiq-transaction = { workspace = true, features = ["ts-types"] }
 nimiq-transaction-builder = { workspace = true }
+nimiq-utils = { workspace = true, features = ["merkle"] }
 
 [dependencies.nimiq]
 package = "nimiq-lib"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4611,6 +4611,7 @@ dependencies = [
  "nimiq-serde",
  "nimiq-transaction",
  "nimiq-transaction-builder",
+ "nimiq-utils",
  "nimiq-zkp-component",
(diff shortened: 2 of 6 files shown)


web-client/Cargo.toml
```diff
@@ -46,6 +46,7 @@ nimiq-primitives = { workspace = true, features = ["coin", "networks", "ts-types
 nimiq-serde = { workspace = true }
 nimiq-transaction = { workspace = true, features = ["ts-types"] }
 nimiq-transaction-builder = { workspace = true }
+nimiq-utils = { workspace = true, features = ["merkle"]}
```
jsdanielh:
```suggestion
nimiq-utils = { workspace = true, features = ["merkle"] }
```

jsdanielh (commented):
Just a very small suggestion.
Also, I wish we shouldn't have needed to expose Merkle tree/path to the web-client but again this could be myself not seeing the whole picture here.

web-client/Cargo.toml
sisou: Interesting that `Cargo.toml` files are not covered by `cargo fmt`.

sisou:
The `MerklePath` could be created only internally in the `SignatureProof::single_sig` method (or an extended version of it) and would not need to be exposed.

The `MerkleTree`, or rather it's currently only static method, is needed to compute the tree root which is the multisig's address.

sisou:
@jsdanielh I have updated the PR, rebased it, and removed `MerklePath` from the public API. Instead, I have added `*_multi_sig` method variants to the `SignatureProof` API, which handle the creation of the MerklePath internally.

The `MerkleTree` should stay though, as that is needed to compute the root of a merkle tree, which in turn enables calculating the address of a multisig-wallet, as there is no other way to get it.

jsdanielh (approved):
LGTM

PR #1886, opened 2023-10-09
Fix issue #1876

### LINKED ISSUE #1876: History Sync: Panic 'Election block must have validators'
In the private testnet (8xhistory-validators), the following panic was observed in validator-2:

`
thread 'tokio-runtime-worker' panicked at 'Election block must have validators': /home/nimiq/core-rs-albatross/consensus/src/sync/history/cluster.rs:223
`
The test conditions were as follows:

- Validator7  & Validator8 are tainted validators: They will respond with erroneous information to the different consensus requests (history chunk, head, missing blocks, etc. )
- All other validators are normal.
- Validator 2 was stopped, then its state was erased and then it was restarted. During the history sync process, this validator panicked. 

It is important to note that many of the validator-2 chunks that this validator requested failed to verify due to the tainted validators:
`
HistoryChunk failed to verify
`
So this validator was trying to re-request many times the history chunks and then it panicked.

viquezclaudio:
## Pull request checklist

### DIFF
--- a/consensus/src/sync/history/cluster.rs
+++ b/consensus/src/sync/history/cluster.rs
@@ -218,9 +218,13 @@ impl<TNetwork: Network + 'static> SyncCluster<TNetwork> {
                 verify_state.predecessor = Block::Macro(macro_block.clone());
 
                 if macro_block.is_election_block() {
-                    verify_state.validators = macro_block
-                        .get_validators()
-                        .expect("Election block must have validators");
+                    match macro_block.get_validators() {
+                        Some(validators) => verify_state.validators = validators,
+                        None => {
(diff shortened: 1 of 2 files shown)


sisou:
Can you explain why it's safe to ignore missing validators?

viquezclaudio:
This was found during a tainted validator testing, where one of the tainted validators, on purpose, was replying with incorrect information to other validators. 
So from the requestor side, if we obtain an invalid batch set, we just need to return `false` within the verify function (which is what the code is now doing) instead of panicking, since we cannot trust what other peers respond to us.

jsdanielh (approved):

PR #1889, opened 2023-10-10
Implement multisig account

Eligioo:
Implementation of a multisig account.

## Pull request checklist

### DIFF
--- a/wallet/src/multisig_account.rs
+++ b/wallet/src/multisig_account.rs
@@ -1,0 +1,206 @@
+use std::num::NonZeroU8;
+
+use curve25519_dalek::EdwardsPoint;
+use itertools::Itertools;
+use nimiq_hash::Blake2bHasher;
+use nimiq_keys::{
+    multisig::{hash_public_keys, Commitment, CommitmentPair, PartialSignature, RandomSecret},
+    Address, KeyPair, PublicKey, SecureGenerate, SecureRng, Signature,
+};
+use nimiq_primitives::{coin::Coin, networks::NetworkId};
+use nimiq_serde::Serialize;
+use nimiq_transaction::{SignatureProof, Transaction};
+use nimiq_utils::merkle::{compute_root_from_content_slice, MerklePath};
+use thiserror::Error;
+
+/// A multi-signature account is an account that requires multiple signatures to authorize outgoing transactions.
+#[derive(Debug)]
+pub struct MultiSigAccount {
+    /// The public address that is used to interact with other accounts.
+    pub address: Address,
+    /// The keypair owning this account.
+    pub key_pair: KeyPair,
+    /// Minimum number of required signatures.
+    pub min_signatures: NonZeroU8,
+    /// A list of all aggregated public keys.
+    pub public_keys: Vec<PublicKey>,
+}
+
+impl MultiSigAccount {
+    /// Returns a new MultiSignature Account.
+    ///
+    /// # Arguments
+    ///
+    /// * `key_pair` - Keypair owning this account.
+    /// * `min_signatures` - Number of signatures required.
+    /// * `public_keys` - A list of all owners' public keys. The public key of the `key_pair` must at least be one of the elements.
+    pub fn from_public_keys(
+        key_pair: &KeyPair,
+        min_signatures: NonZeroU8,
+        public_keys: &[PublicKey],
+    ) -> Result<Self, MultiSigAccountError> {
+        if public_keys.is_empty() {
+            return Err(MultiSigAccountError::PublicKeysNotEmpty);
+        } else if !public_keys.contains(&key_pair.public) {
+            return Err(MultiSigAccountError::KeyPairNotPartOfList);
+        }
+
+        let mut sorted_public_keys = public_keys.to_vec();
+        sorted_public_keys.sort();
+
+        let multi_sig_keys: Vec<PublicKey> = sorted_public_keys
+            .into_iter()
+            .combinations(min_signatures.get() as usize)
+            .map(|pk| MultiSigAccount::aggregate_public_keys(&pk))
+            .collect();
+
+        Ok(Self::new(key_pair, min_signatures, &multi_sig_keys))
+    }
+
+    /// Returns a new MultiSignature Account. This method expects that the provided public keys are already delinearized and aggregated.
+    /// Use `MultiSigAccount.from_public_keys` otherwise.
+    ///
+    /// # Arguments
+    ///
+    /// * `key_pair` - Keypair owning this account.
+    /// * `min_signatures` - Number of signatures required.
+    /// * `public_keys` - A list of all aggregated public keys.
+    pub fn new(key_pair: &KeyPair, min_signatures: NonZeroU8, public_keys: &[PublicKey]) -> Self {
+        Self {
+            address: Address::from(compute_root_from_content_slice::<Blake2bHasher, PublicKey>(
+                public_keys,
+            )),
+            key_pair: key_pair.clone(),
+            min_signatures,
+            public_keys: public_keys.to_vec(),
+        }
+    }
+
+    /// Generates a new commitment pair containing a commitment and a random secret.
+    #[inline]
+    pub fn create_commitment(&self) -> CommitmentPair {
+        CommitmentPair::generate(&mut SecureRng::default())
+    }
+
+    /// Creates an unsigned transaction.
+    pub fn create_transaction(
+        &self,
+        recipient: Address,
+        value: Coin,
+        fee: Coin,
+        validity_start_height: u32,
+        network_id: NetworkId,
+    ) -> Transaction {
+        Transaction::new_basic(
+            self.address.clone(),
+            recipient,
+            value,
+            fee,
+            validity_start_height,
+            network_id,
+        )
+    }
+
+    /// Utility method that delinearizes and aggregates the provided slice of public keys.
+    pub fn aggregate_public_keys(public_keys: &[PublicKey]) -> PublicKey {
+        let mut sorted_public_keys = public_keys.to_vec();
+        sorted_public_keys.sort();
+
+        let public_keys_hash = hash_public_keys(&sorted_public_keys);
+        let delinearized_pk_sum: EdwardsPoint = public_keys
+            .iter()
+            .map(|public_key| public_key.delinearize(&public_keys_hash))
+            .sum();
+
+        let mut public_key_bytes: [u8; PublicKey::SIZE] = [0u8; PublicKey::SIZE];
+        public_key_bytes.copy_from_slice(delinearized_pk_sum.compress().as_bytes());
+        PublicKey::from(public_key_bytes)
+    }
+
+    /// Creates a partial signature of the provided transaction.
+    pub fn partially_sign_transaction(
+        &self,
+        transaction: &Transaction,
+        public_keys: &[PublicKey],
+        commitments: &[Commitment],
+        secret: &RandomSecret,
+    ) -> PartialSignature {
+        let mut sorted_public_keys = public_keys.to_vec();
+        sorted_public_keys.sort();
+
+        self.key_pair
+            .partial_sign(
+                &sorted_public_keys,
+                secret,
+                commitments,
+                transaction.serialize_content().as_slice(),
+            )
+            .0
+    }
+
+    /// Creates a signature proof.
+    pub fn create_proof(
+        &self,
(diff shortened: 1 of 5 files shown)


wallet/src/multisig_account.rs
```diff
@@ -0,0 +1,232 @@
...
+use nimiq_keys::{
+    multisig::{hash_public_keys, Commitment, CommitmentPair, PartialSignature, RandomSecret},
+    Address, KeyPair, PublicKey, SecureGenerate, SecureRng, Signature,
+};
+use nimiq_primitives::{coin::Coin, networks::NetworkId};
+use nimiq_serde::Serialize;
+use nimiq_transaction::{SignatureProof, Transaction};
+use nimiq_utils::merkle::{compute_root_from_content_slice, MerklePath};
+
+pub struct MultiSigAccount {
```
jsdanielh, line 18: Could you add `rustdoc` to the struct and its members?

wallet/src/multisig_account.rs
```diff
@@ -0,0 +1,232 @@
...
+    pub address: Address,
+    pub key_pair: KeyPair,
+    pub min_signatures: u8,
```
jsdanielh: How about using [NonZeroU8](https://doc.rust-lang.org/std/num/struct.NonZeroU8.html#)?

wallet/src/multisig_account.rs
```diff
@@ -0,0 +1,232 @@
...
+        public_keys: &[PublicKey],
+    ) -> Self {
+        assert!(
+            !public_keys.is_empty(),
+            "Provided public keys may not be empty."
+        );
+        assert!(
+            min_signatures > 0,
+            "Minimal signatures must be greater than 0."
+        );
```
jsdanielh: If we use [NonZeroU8](https://doc.rust-lang.org/std/num/struct.NonZeroU8.html#), then this check wouldn't be needed

wallet/src/multisig_account.rs
```diff
@@ -0,0 +1,232 @@
...
+    /// * `min_signatures` - Number of signatures required. Panics if value is less than 1.
+    /// * `public_keys` - A list of all owners' public keys. Panics if list is missing the public key of `key_pair`.
+    pub fn from_public_keys(
+        key_pair: &KeyPair,
+        min_signatures: u8,
+        public_keys: &[PublicKey],
+    ) -> Self {
+        assert!(
+            !public_keys.is_empty(),
+            "Provided public keys may not be empty."
```
jsdanielh:
```suggestion
            "Provided public keys must not be empty."
```

wallet/src/multisig_account.rs
```diff
@@ -0,0 +1,232 @@
...
+        );
+        assert!(
+            min_signatures > 0,
+            "Minimal signatures must be greater than 0."
+        );
+        assert!(
+            public_keys.contains(&key_pair.public),
+            "Public key of keypair must be part of provided public keys."
```
jsdanielh: From an interface perspective, I think we could maybe get rid of this check? i.e we're passing the same data twice?

wallet/src/multisig_account.rs
```diff
@@ -0,0 +1,232 @@
...
+        );
+
+        let mut sorted_public_keys = Vec::with_capacity(public_keys.len());
+        public_keys.clone_into(&mut sorted_public_keys);
+        sorted_public_keys.sort();
```
jsdanielh: I think we should be able to do:
```suggestion
        let mut sorted_public_keys = public_keys.clone();
        sorted_public_keys.sort();
```

wallet/src/multisig_account.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    /// Returns a new MultiSignature Account. This method expects that the provided public keys are already delinearized and aggregated.
+    /// Use `MultiSigAccount.from_public_keys` otherwise.
+    ///
+    /// # Arguments
+    ///
+    /// * `key_pair` - KeyPair owning this account.
+    /// * `min_signatures` - Number of signatures required.
+    /// * `public_keys` - A list of all aggregated public keys.
+    pub fn new(key_pair: &KeyPair, min_signatures: u8, public_keys: &[PublicKey]) -> Self {
```
jsdanielh:
```suggestion
    pub fn new(key_pair: &KeyPair, min_signatures: NonZeroU8, public_keys: &[PublicKey]) -> Self {
```

wallet/src/multisig_account.rs
```diff
@@ -0,0 +1,232 @@
...
+            validity_start_height,
+            network_id,
+        )
+    }
+
+    /// Utility method that delinearizes and aggregates the provided slice of public keys.
+    pub fn aggregate_public_keys(public_keys: &[PublicKey]) -> PublicKey {
+        let mut sorted_public_keys = Vec::with_capacity(public_keys.len());
+        public_keys.clone_into(&mut sorted_public_keys);
+        sorted_public_keys.sort();
```
jsdanielh:
```suggestion
        let mut sorted_public_keys = public_keys.clone();
        sorted_public_keys.sort();
```

wallet/src/multisig_account.rs
```diff
@@ -0,0 +1,232 @@
...
+    pub fn partially_sign_transaction(
+        &self,
+        transaction: &Transaction,
+        public_keys: &[PublicKey],
+        commitments: &[Commitment],
+        secret: &RandomSecret,
+    ) -> PartialSignature {
+        let mut sorted_public_keys = Vec::with_capacity(public_keys.len());
+        public_keys.clone_into(&mut sorted_public_keys);
+        sorted_public_keys.sort();
```
jsdanielh:
```suggestion
        let mut sorted_public_keys = public_keys.clone();
        sorted_public_keys.sort();
```

wallet/src/multisig_account.rs
```diff
@@ -0,0 +1,232 @@
...
+    pub fn create_proof(
+        &self,
+        aggregated_public_key: &PublicKey,
+        aggregated_commitment: &Commitment,
+        signatures: &[PartialSignature],
+    ) -> SignatureProof {
+        assert!(
+            signatures.len() == self.min_signatures as usize,
+            "Number of signatures must be the same as the minimal signatures."
+        );
```
jsdanielh: How about returning `Option<SignatureProof>` instead of panicking? I think it's better to return an error or `None` in this case.

wallet/src/multisig_account.rs
```diff
@@ -0,0 +1,232 @@
...
+
+        let aggregated_signature: PartialSignature = signatures.iter().sum();
+        let raw_aggregated_signature = aggregated_signature.as_bytes();
+        let raw_aggregated_commitment = aggregated_commitment.to_bytes();
+
+        let mut combined =
+            Vec::with_capacity(raw_aggregated_signature.len() + raw_aggregated_commitment.len());
+        combined.extend_from_slice(&aggregated_commitment.to_bytes());
+        combined.extend_from_slice(aggregated_signature.as_bytes());
+        let signature = Signature::from_bytes(&combined).expect("Invalid signature constructued.");
```
jsdanielh: If we return a `None` or error, we should remove this `expect` and convert it to the appropriate return type

wallet/src/multisig_account.rs
```diff
@@ -0,0 +1,232 @@
...
+            partial_signatures,
+        );
+
+        let mut signed_transaction = transaction.clone();
+        signed_transaction.proof = proof.serialize_to_vec();
+
+        signed_transaction
+    }
+
+    fn combinations<T: Clone>(list: &[T], k: usize) -> Vec<Vec<T>> {
```
jsdanielh: Can't we use [itertool's combinations](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.combinations)?

wallet/src/multisig_account.rs
```diff
@@ -0,0 +1,232 @@
...
+            public_keys.contains(&key_pair.public),
+            "Public key of keypair must be part of provided public keys."
+        );
+
+        let mut sorted_public_keys = Vec::with_capacity(public_keys.len());
+        public_keys.clone_into(&mut sorted_public_keys);
+        sorted_public_keys.sort();
+
+        let combinations =
+            MultiSigAccount::combinations(&sorted_public_keys, min_signatures as usize);
```
jsdanielh: Can't we use [itertool's combinations](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.combinations)?

wallet/tests/multisig.rs
```diff
@@ -0,0 +1,65 @@
+use hex::FromHex;
+use nimiq_keys::{multisig::Commitment, Address, KeyPair, PrivateKey, PublicKey};
+use nimiq_primitives::{coin::Coin, networks::NetworkId};
+use nimiq_wallet::MultiSigAccount;
+
+#[test]
```
jsdanielh, line 14: I'd add a couple of test for the cases where code should panic

wallet/src/multisig_account.rs
Eligioo: Currently it makes it more explicit which keypair actually is going to sign the transactions.

Also if we would allow the public key of the keypair not to be part of the `public_key` slice, this method has to add it as an extra step and creates confusion what the value of `min_signatures` must be.

wallet/src/multisig_account.rs
Eligioo: By calling `public_keys.clone()` you clone the reference instead of the actual values. So each item has to be cloned.
You could also do the following, but that would clone each item individually
```
let mut sorted_public_keys: Vec<PublicKey> = public_keys.iter().cloned().collect();
sorted_public_keys.sort();
```

jsdanielh: Mmm... I think the problem is that you want a mutable copy of the array elements which can't be done with `clone`. But then I suggest to do:
```
let mut sorted_public_keys = public_keys.to_vec();
sorted_public_keys.sort()
```

wallet/src/multisig_account.rs
jsdanielh: Ok, I see the point

wallet/src/multisig_account.rs
```diff
@@ -0,0 +1,209 @@
...
+        signed_transaction.proof = proof.serialize_to_vec();
+
+        Ok(signed_transaction)
+    }
+}
+
+/// Possible multi-sig account errors.
+#[derive(Debug, Error)]
+pub enum MultiSigAccountError {
+    #[error("Invalid signature constructed.")]
```
jsdanielh:
```suggestion
    #[error("Invalid signature constructed")]
```

wallet/src/multisig_account.rs
```diff
@@ -0,0 +1,209 @@
...
+    InvalidSignatureConstructed,
+    #[error("Failed to construct a valid signature based on the provided signatures and aggregated commitment.")]
```
jsdanielh:
```suggestion
    #[error("Failed to construct a valid signature based on the provided signatures and aggregated commitment")]
```

wallet/src/multisig_account.rs
```diff
@@ -0,0 +1,209 @@
...
+    InvalidSignatureFromBytes(#[from] nimiq_keys::SignatureError),
+    #[error("Number of signatures must be the same as the minimal signatures.")]
```
jsdanielh:
```suggestion
    #[error("Number of signatures must be the same as the minimal signatures")]
```

wallet/src/multisig_account.rs
```diff
@@ -0,0 +1,209 @@
...
+    InvalidSignaturesLength,
+    #[error("The public key of keypair must be part of provided public keys.")]
```
jsdanielh:
```suggestion
    #[error("The public key of keypair must be part of provided public keys")]
```

wallet/src/multisig_account.rs
```diff
@@ -0,0 +1,209 @@
...
+    KeyPairNotPartOfList,
+    #[error("The provided public keys must not be empty.")]
```
jsdanielh:
```suggestion
    #[error("The provided public keys must not be empty")]
```

jsdanielh (commented):
I'd also suggest to not use a period on the error description

Eligioo:
@jsdanielh suggestions regarding the error description and sorting the public keys has been addressed.

jsdanielh (approved):

jsdanielh:
Rebasing and squashing commits to get it merged

PR #1952, opened 2023-11-13
Validity Window in terms of batches

viquezclaudio:
- Change the Policy::TransactionValidityWindow constant to batches instead of blocks
- Introduce a new transaction_validity_window_blocks() function to obtain the transaction validity window in terms of blocks

## Pull request checklist

### DIFF
--- a/primitives/src/policy.rs
+++ b/primitives/src/policy.rs
@@ -30,8 +30,7 @@ pub struct Policy {
     /// Maximum size of accounts trie chunks.
     #[cfg_attr(feature = "ts-types", wasm_bindgen(skip))]
     pub state_chunks_max_size: u32,
-    /// Number of blocks a transaction is valid with Albatross consensus.
-    /// This should be a multiple of `blocks_per_batch`.
+    /// Number of batches a transaction is valid with Albatross consensus.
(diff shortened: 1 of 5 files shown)


jsdanielh (approved):

primitives/src/policy.rs
```diff
@@ -664,7 +672,7 @@ impl Default for Policy {
             tendermint_timeout_init: 1000,
             tendermint_timeout_delta: 1000,
             state_chunks_max_size: 200, // #Nodes/accounts 200, TODO: Simulate with different sizes
-            transaction_validity_window: 7200,
+            transaction_validity_window: 120,
```
sisou, line 675: Should add a comment here, specifying that this is in batches.

sisou:
First question: but why?

Second question: Does this change how long transactions are valid, when their validity-start-height is in the middle of a batch? Is the tx then valid until the _end_ of the 120th following batch? Or still valid only until the middle of that batch?

viquezclaudio:
This is to simplify some calculations and cases for an upcoming change related to syncing the transactions that have occurred during the current validity window: currently when full nodes validators are syncing, they do not have the history of the validity window, so they cannot properly enforce the "tx in validity window" check when producing blocks. 
And no, this does not change how long transactions are valid.

PR #1953, opened 2023-11-13
Fix comment in history chunk proof

paberr:
The `prove_chunk` method was poorly documented, especially the `verifier_block_number` argument.
When discussing with @viquezclaudio, we discovered that it is used to calculate the boundaries of the chunk proof.

If the block is set to a point before the chunk starts, the proof will always be empty.
The block is supposed to mark the end of the sync and the root the chunks will be verified against.

This PR mainly updates the comments to reflect this.

## Pull request checklist

### DIFF
--- a/blockchain/src/history/history_store.rs
+++ b/blockchain/src/history/history_store.rs
@@ -782,6 +782,9 @@ impl HistoryStore {
     /// Returns the `chunk_index`th chunk of size `chunk_size` for a given epoch.
     /// The return value consists of a vector of all the historic transactions in that chunk
     /// and a proof for these in the MMR.
+    /// The `verifier_block_number` is the block the chunk proof should be verified against.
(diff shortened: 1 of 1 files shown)


viquezclaudio (approved):

blockchain/src/history/history_store.rs
```diff
@@ -782,6 +782,9 @@ impl HistoryStore {
     /// Returns the `chunk_index`th chunk of size `chunk_size` for a given epoch.
     /// The return value consists of a vector of all the historic transactions in that chunk
     /// and a proof for these in the MMR.
+    /// The `verifier_block_number` is the block the chunk proof should be verified against.
+    /// That means that no leaf beyond this block is returned and that the prove should be
```
sisou:
```suggestion
    /// That means that no leaf beyond this block is returned and that the proof should be
```

jsdanielh (approved):

PR #1957, opened 2023-11-14
Correctly check the signer of the VRF seed for macro block proposals.

### LINKED ISSUE #1956: Proposal Verification must use the correct seed signer
Currently the `valid_round` proposer is used to verify the seed signing if present. That is in most cases okay, but the valid round could already be a re-proposed version of that proposal, making the proposer for that round not the original signer of the seed. The header does have the round it was proposed in set and that value should be used instead.

Additionally any proposal whose round and `header.round` differ must have a valid round set for it to be valid, as they are  re-proposed.

nibhar:
Previously the vrf seed would be checked against the proposer for the round that is given in valid_round. That is however not necessarily correct as that round could already be a re-proposed proposal, making that proposer not the signer of the seed.

Now the round given in the header is taken for seed verification and proposals which do not correctly state a VR if they should are being rejected.

Fixes #1956

### DIFF
--- a/blockchain/src/blockchain/verify.rs
+++ b/blockchain/src/blockchain/verify.rs
@@ -368,13 +368,28 @@ impl Blockchain {
             return Ok(proposed_block.body.unwrap());
         }
 
-        // If a valid round is set, the VRF seed will be signed by the proposer of that round.
-        // So get that proposer.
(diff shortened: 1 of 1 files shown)


hrxi:
Rebased to sign it.

PR #1960, opened 2023-11-14
Verify contributions of future round aggregations

### LINKED ISSUE #1434: Tendermint: Contributions to future rounds are not verified.
Currently the contributions to future rounds in Tendermint are only tracked using their signatories BitSet. That is not sufficient as it is easy to fake. The signature of the message must be verified for a valid skip ahead to the round of the contribution.

hrxi:
Fixes #1434.

### DIFF
--- a/tendermint/src/tendermint.rs
+++ b/tendermint/src/tendermint.rs
@@ -23,11 +23,8 @@ use crate::{
     protocol::{Aggregation, Protocol, SignedProposalMessage, TaggedAggregationMessage},
     state::State,
     utils::{Return, Step},
-    Proposal,
+    AggregationMessage, Proposal,
 };
-
-// TODO:
-//  * Signatures for future rounds.
 
 /// Main Tendermint structure.
 ///
@@ -68,11 +65,13 @@ pub struct Tendermint<TProtocol: Protocol> {
     pub(crate) aggregations: SelectAll<BoxStream<'static, ((u32, Step), TProtocol::Aggregation)>>,
 
     /// Keeps track of contributions to future aggregations.
-    ///
-    /// TODO This is not correct behavior yet. The signatures are unauthenticated.
-    /// Instead look into already aggregated contributions of f+1 vote weight or more. Only consider single Signatures,
-    /// as their verification is quite costly. Potentially introduce a threshold of signatures that may compose f+1.
     pub(crate) future_contributions: BTreeMap<u32, BitSet>,
+
+    /// Keeps at most one future aggregation per validator, while it is waiting for verification.
+    future_round_messages: BTreeMap<u16, TaggedAggregationMessage<TProtocol::AggregationMessage>>,
+
+    /// The future round aggregation that is currently being verified.
+    future_round_verification: Option<BoxFuture<'static, Result<(u32, BitSet), ()>>>,
 
     /// In case a timeout is required it will be stored here until elapsed or no longer necessary.
     /// Must be cleared in both cases.
@@ -117,6 +116,8 @@ impl<TProtocol: Protocol> Tendermint<TProtocol> {
             aggregations: SelectAll::default(),
             aggregation_senders: BTreeMap::default(),
             future_contributions: BTreeMap::default(),
+            future_round_messages: BTreeMap::default(),
+            future_round_verification: None,
             timeout: None,
             decision: false,
             state_return_pending: false,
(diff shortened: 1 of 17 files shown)


tendermint/src/tendermint.rs
```diff
@@ -390,52 +391,104 @@ impl<TProtocol: Protocol> Tendermint<TProtocol> {
...
+                    self.future_round_verification = None;
+
+                    // If the signature was found to be valid, add the contributing validators to
+                    // the set of contributors for that future round.
+                    if let Ok((round, new_contributors)) = verification_result {
+                        if round > self.state.current_round {
+                            // future contributions need to be tracked separately, as once a round has f+1 contributions
+                            // it can be fast tracked
+                            let contributors = self.future_contributions.entry(round).or_default();
+                            *contributors |= new_contributors;
```
nibhar, line 414: Maybe consider sticking with the Entry API here. 
```suggestion
                            self.future_contributions.entry(round)
                                .and_modify(|contributors| *contributors |= new_contributors)
                                .or_insert(new_contributors);
```

tendermint/src/tendermint.rs
```diff
@@ -390,52 +391,104 @@ impl<TProtocol: Protocol> Tendermint<TProtocol> {
...
-            } else if let Some(sender) = self.aggregation_senders.get(&id) {
-                match sender.try_send(message.aggregation) {
-                    Ok(()) => {}
-                    Err(mpsc::error::TrySendError::Closed(_aggregation_message)) => {
-                        panic!("Aggregations must not be dropped.");
+            }
+
+            // If no future round aggregation verification is currently running, start a new one.
+            if self.future_round_verification.is_none() && !self.future_round_messages.is_empty() {
+                // TODO: choose at random
```
nibhar: Or choose with weight. Not every contribution is equally good.|

Picking the first one is likely not vey good as that will (in all nodes) prefer low index validators over others. While weighting the contributions is maybe not necessary I think random selection is.

tendermint/src/tendermint.rs
```diff
@@ -390,52 +391,104 @@ impl<TProtocol: Protocol> Tendermint<TProtocol> {
...
+                        }
+                        Entry::Occupied(mut o) => {
+                            // Generally: Newer is better. If the new message comes
+                            // from the same round as a previous one, assume that
+                            // it is better.
+                            if id.0 >= o.get().tag.0 {
+                                o.insert(message);
+                            }
+                        }
                     }
```
nibhar, line 461: I like the Entry API better here as well.
```suggestion
                    match self
                        .future_round_messages
                        .entry(message.aggregation.sender())
                        .and_modify(|m| {
                            if id.0 > m.tag.0 {
                                *m = message
                            }
                        })
                        .or_insert(message)
```

validator-network/src/network_impl.rs
```diff
@@ -43,37 +45,30 @@ where
...
-                    _ => {}
-                }
-            }
-        };
-
-        tokio::time::timeout(Duration::from_secs(5), future)
-            .await
-            .map_err(|_| NetworkError::Unreachable)?
+    /// Returns the local validator ID.
+    fn local_validator_id(&self) -> u16 {
```
nibhar: This will panic for non elected, but active validators.

validator-network/src/network_impl.rs
```diff
@@ -178,60 +196,18 @@ where
         validator_id: usize,
         msg: M,
```
nibhar: `M` likely no longer needs to be `Clone`. Can't mention it there, but it is 3 lines up.

validator-network/src/network_impl.rs
```diff
@@ -178,60 +196,18 @@ where
...
+            self.network
+                .message(msg, peer_id)
+                .map_err(NetworkError::Request)
+                .await
+        } else {
+            Err(NetworkError::Unreachable)
+        }
     }
 
     async fn request<TRequest: Request>(
```
nibhar, line 209: With `request` and `send_to` now being virtually the same, we may want to look into unifying them into one.

nibhar (changes requested):
Looks good! Mostly minor comments, except for the selection of the next verification task. I think randomness there is important, as the selection otherwise is based on a value for which the entire network has the same bias.

tendermint/src/tendermint.rs
hrxi: I also like that better, but it's not possible due to ownership rules, unfortunately. The closure would take ownership of `message`, but it's needed afterwards.

tendermint/src/tendermint.rs
hrxi: Same ownership problem, but here, I also actually like the current version better.

validator-network/src/network_impl.rs
hrxi: Not so easy because of the different trait requirements.

nibhar: Just looked a bit more into it and seams not worth the effort if even worth it at all.

validator-network/src/network_impl.rs
hrxi: Made it a `Result` instead.

validator-network/src/network_impl.rs
hrxi: Done.

tendermint/src/tendermint.rs
hrxi: Now chosen at random.

nibhar (approved):
LGTM

PR #1961, opened 2023-11-15
Improve DHT publishing

fiaxh:
Previously, the own DHT entry was published on every new epoch. That means
- It might try to publish before the DHT finished setting up
- It republishes the DHT although we configured the library to automatically do that every 60 seconds.

This PR adds an event to the Network that is fired when DHT is done bootstrapping. The validator listens for this event and publishes the own record upon receiving it.

### DIFF
--- a/consensus/src/consensus/remote_event_dispatcher.rs
+++ b/consensus/src/consensus/remote_event_dispatcher.rs
@@ -337,7 +337,7 @@ impl<N: Network> Future for RemoteEventDispatcher<N> {
                     // Remove the peer from internal data structures.
                     self.state.write().remove_peer(&peer_id);
                 }
-                Ok(NetworkEvent::PeerJoined(_peer_id, _)) => {}
+                Ok(_) => {}
                 Err(_) => return Poll::Pending,
             }
         }
--- a/validator/src/validator.rs
+++ b/validator/src/validator.rs
@@ -26,7 +26,7 @@ use nimiq_keys::{Address, KeyPair as SchnorrKeyPair, Signature as SchnorrSignatu
 use nimiq_macros::store_waker;
 use nimiq_mempool::{config::MempoolConfig, mempool::Mempool};
 use nimiq_network_interface::{
-    network::{MsgAcceptance, Network, PubsubId, Topic},
+    network::{MsgAcceptance, Network, NetworkEvent, PubsubId, SubscribeEvents, Topic},
     request::request_handler,
 };
 use nimiq_primitives::{coin::Coin, policy::Policy};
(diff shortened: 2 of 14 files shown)


validator/src/validator.rs
```diff
@@ -892,6 +897,17 @@ where
             }
         }
 
+        // Check if DHT is bootstrapped and we can publish our record
+        while let Poll::Ready(Some(result)) = self.network_event_rx.poll_next_unpin(cx) {
+            match result {
+                Ok(NetworkEvent::DhtBootstrapped) => {
```
hrxi, line 905: AFAICT, the event isn't sent anywhere.

validator/src/validator.rs
```diff
@@ -322,19 +338,9 @@ where
             .iter()
             .map(|validator| validator.voting_key.clone())
             .collect();
-        let key = self.voting_key();
         let network = Arc::clone(&self.network);
 
-        // TODO might better be done without the task.
```
hrxi, line 328: Is the "might better be done without the task." fixed by this commit? If not, it should probably stay.

hrxi (changes requested):

consensus/src/consensus/remote_event_dispatcher.rs
```diff
@@ -337,7 +337,7 @@ impl<N: Network> Future for RemoteEventDispatcher<N> {
                     // Remove the peer from internal data structures.
                     self.state.write().remove_peer(&peer_id);
                 }
-                Ok(NetworkEvent::PeerJoined(_peer_id, _)) => {}
+                Ok(_) => {}
```
nibhar, line 340: I think adding information is better than removing it. Very much personal preference though.

```suggestion
                Ok(NetworkEvent::PeerJoined(_peer_id, _peer_info)) => {}
                Ok(NetworkEvent::DhtBootstrapped => {}
```

validator/src/validator.rs
```diff
@@ -249,8 +251,22 @@ where
         self.control_mempool_monitor.clone()
     }
 
+    pub(crate) fn publish_dht(&self) {
```
nibhar: Looks to me like the `pub(crate)` might not be necessary.

validator/src/validator.rs
```diff
@@ -892,6 +898,17 @@ where
...
         }
 
+        // Check if DHT is bootstrapped and we can publish our record
+        while let Poll::Ready(Some(result)) = self.network_event_rx.poll_next_unpin(cx) {
+            match result {
+                Ok(NetworkEvent::DhtBootstrapped) => {
+                    self.publish_dht();
+                }
+                Ok(_) => {}
+                Err(_) => break,
```
nibhar: What does this case mean? Is this safe to just ignore? Will this recover?

consensus/src/consensus/remote_event_dispatcher.rs
hrxi: I think for a "listen for events" it makes sense to ignore all that you don't want to listen for. Otherwise you'll need to update each location when you add another event.

validator/src/validator.rs
fiaxh: That error is thrown by the broadcast stream in the following case:

> The receiver lagged too far behind. Attempting to receive again will return the oldest message still retained by the channel.

Thus, it would be better to continue the loop instead of breaking (although that would have also continued in the next poll). I adjusted the code accordingly.

validator/src/validator.rs
fiaxh: Indeed

hrxi (approved):
LGTM.

PR #1964, opened 2023-11-16
Fix bundler worker type

faberto:
Fixes:

`Uncaught SyntaxError: Cannot use import statement outside a module (at worker.js?type=classic&worker_file:5:1)`

when the worker trys to import Comlink when using the web client ESM with vite bundler.


sisou (approved):
Good find! This makes the `bundler` target finally work with vite!

I tested with both vite and the previously working webpack bundling, and both are fine with this :+1:

PR #1966, opened 2023-11-16
Add RPC support for devnet script

jeffesquivels:


scripts/devnet/validator.py
```diff
@@ -26,7 +26,7 @@ class Validator(Node):
 
     def __init__(self, name: str, listen_port: int,
                  topology_settings: TopologySettings, sync_mode: str = "full",
-                 metrics: Optional[dict] = None,
+                 rpc: Optional[dict] = None, metrics: Optional[dict] = None,
```
jsdanielh, line 31: Could you add it to the class documentation also?

scripts/devnet/node.py
```diff
@@ -384,11 +397,11 @@ class RegularNode(Node):
 
     def __init__(self, name: str, listen_port: int,
                  topology_settings: TopologySettings, sync_mode: str = "full",
-                 metrics: Optional[dict] = None,
+                 rpc: Optional[dict] = None, metrics: Optional[dict] = None,
```
jsdanielh, line 402: Could you also add it to the class documentation?

scripts/devnet/node.py
```diff
@@ -505,12 +512,12 @@ class Seed(Node):
...
```
jsdanielh, line 519: Could you also add it to the class documentation?

jsdanielh (commented):
I also think you missed to add it to the `spammer` node subclass

jsdanielh (approved):

PR #1967, opened 2023-11-16
Move Proposal verification. Make sure predecessor exists.

nibhar:
This PR moves the proposal verification into the `ProposalBuffer`. From the viewpoint of Tendermint this results in all proposals being properly authenticated. Additionally all proposal which enter Tendermint have an existing Microblock predecessor.

### DIFF
--- a/blockchain/src/blockchain/verify.rs
+++ b/blockchain/src/blockchain/verify.rs
@@ -5,7 +5,6 @@ use nimiq_database::{
     traits::{ReadTransaction, WriteTransaction},
     TransactionProxy as DBTransaction, WriteTransactionProxy,
 };
-use nimiq_keys::Signature;
 use nimiq_primitives::policy::Policy;
 
 use crate::{blockchain_state::BlockchainState, BlockProducer, Blockchain};
@@ -311,29 +310,16 @@ impl Blockchain {
     }
 
     /// Verifies a `proposed_block`, given the round it is proposed in as well as its valid round if applicable.
-    ///
-    /// For signature verification the signature as well as the supposedly signed payload are needed. They are
-    /// different from the eventual justification as the signature here is a Schnorr signature from the proposer,
-    /// whereas the final proof will be an aggregated bls key of all contributing validators.
     pub fn verify_macro_block_proposal(
         &self,
         proposed_block: MacroBlock,
         round: u32,
         valid_round: Option<u32>,
-        signed_data: Vec<u8>,
-        signature: &Signature,
-        signature_only: bool,
     ) -> Result<MacroBody, PushError> {
         // If the accounts tree is not complete nothing can be done
         if !self.accounts_complete() {
             // Maybe introduce a proper error here. Is this even needed?
             return Err(PushError::IncompleteAccountsTrie);
-        }
-
-        // No inherent was given, but signature verification is the only required step, independent from
-        // the result this can never work, as the inherent would have to be returned.
-        if proposed_block.body.is_none() && signature_only {
-            return Err(PushError::InvalidBlock(BlockError::MissingBody));
         }
 
         // Get the seed of the preceding block.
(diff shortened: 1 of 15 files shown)


tendermint/src/protocol.rs
```diff
@@ -10,8 +10,6 @@ use crate::utils::Step;
 pub enum ProposalError {
     /// Invalid proposals may be used to punish the proposer.
     InvalidProposal,
-    /// Proposals with invalid signatures must be ignored, but the peer who broadcasted it may be banned.
-    InvalidSignature,
     /// Collectively used for all problems not accounted for by InvalidProposal and InvalidSignature.
```
hrxi:
```suggestion
    /// Collectively used for all problems not accounted for by InvalidProposal
```

blockchain/src/blockchain/verify.rs
```diff
@@ -311,71 +310,45 @@ impl Blockchain {
...
-        // value provided if `valid_round` is Some(vr). Importantly though, that very vr is not necessarily
+        // The round is specified within the header itself, and it should only ever differ from the round
+        // value provided if valid_round = Some(vr). Importantly though, that very vr is not necessarily
         // the same as the round given in the header, as it might not be the first time this proposal is
         // re-proposed.
-        if proposed_block.header.round != round {
+        if let Some(vr) = valid_round {
+            if vr < proposed_block.header.round || vr >= round {
+                // A VR only makes sense if it is at least the original round of the proposal,
+                // while also not being in the future.
```
hrxi, line 339: Can it actually be the same round as the original proposal?

validator/src/proposal_buffer.rs
```diff
@@ -0,0 +1,340 @@
...
+///     1. The predecessor exists
+///     2. The proposer of the proposal calculated with the predecessor vrf seed was indeed supposed
+///         to be who was stated in the original message.
+///
+/// Any message that fails any of the above should be dropped, however depending on why it failed, additional
+/// action may be advised. If the signature was incorrect for the assumed signer, the peer who relayed the
+/// message relayed invalid data and should be banned. If the assumed proposer does not match the actual proposer,
+/// the assumed proposer may be punished (as he produced faulty data, proven by the signature and the predecessor vrf).
+/// If the predecessor is unavailable (even after requesting it from the peer who originated the proposal or relayed
+/// the proposal) they both could be banned asd they failed to produce proper data upon being asked to do so.
```
hrxi:
```suggestion
/// the proposal) they both could be banned as they failed to produce proper data upon being asked to do so.
```

validator/src/proposal_buffer.rs
```diff
@@ -0,0 +1,340 @@
...
+    ) -> (
+        ProposalSender<TValidatorNetwork>,
+        ProposalReceiver<TValidatorNetwork>,
+    ) {
+        // Create the shared buffer
+        let buffer = Self {
+            buffer: LinkedHashMap::new(),
+            waker: None,
+        };
+        // TODO does this need to be a rwlock? Mutex? should we upgradable lock in ProposalSender::send()
```
hrxi: Should probably be a `Mutex` because we have at most on reader.

validator/src/proposal_buffer.rs
```diff
@@ -0,0 +1,340 @@
...
+                    // Don't propagate the proposal
+                    self.network
+                        .validate_message::<ProposalTopic<TValidatorNetwork>>(
+                            id,
+                            MsgAcceptance::Reject,
+                        );
+                    None
+                }
+                Err(_) => {
+                    // Block is not known. Request it. (todo)
```
hrxi:
```suggestion
                    // Block is not known. Request it. (TODO)
```
More easily found.

validator/src/proposal_buffer.rs
```diff
@@ -0,0 +1,340 @@
...
+                        .expect("Must be able to calculate block producer.")
+                        .validator;
+
+                    // Compare the expected and the actual validator
+                    if *assumed_validator == actual_validator {
+                        // Validators match. The signature was already verified. Proposal is good to go.
+                        // DO NOT validate message the proposal, as there is much more to the validity than the signature.
+                        Some(entry.remove())
+                    } else {
+                        // Assumed validator is not the proposer. Punish the validator (todo).
```
hrxi:
```suggestion
                        // Assumed validator is not the proposer. Punish the validator (TODO).
```

validator/src/proposal_buffer.rs
```diff
@@ -0,0 +1,340 @@
...
+                        // Proposal is removed as it is not a valid proposal.
+                        log::debug!("Found MacroHeader Proposal where validators do not match.");
+                        let (_proposal, id) = entry.remove();
+
+                        // The peer relaying this proposal relayed a faulty proposal, disconnect him.
+                        let source = id.propagation_source();
+                        let nw = Arc::clone(&self.network);
+                        tokio::task::spawn(async move {
+                            nw.disconnect_peer(source, CloseReason::MaliciousPeer).await;
+                        });
```
hrxi: Theoretically, we're already in an `async` function, so no need to `tokio::spawn` another task for this. A list of disconnect-peer-futures could be stored in the `ProposalReceiver` and polled every time `poll_next` is called.

Might be too annoying to implement though. 🤷‍♀️

hrxi (commented):
Looks generally good to me, no blocking concerns.

blockchain/src/blockchain/verify.rs
nibhar: Yes it can and usually will be. 

For any proposal that is re-proposed the first time it will have VR being the round of the original proposal. Only once this iteration gets re-proposed it will have as the VR the round it was re-proposed in.

tendermint/src/protocol.rs
nibhar: I removed `Other` entirely as it is unused and I consider it generally not good to have around. If needed in the future a new enum variant should be added. Thanks for making me aware.

validator/src/proposal_buffer.rs
nibhar: Done

validator/src/proposal_buffer.rs
nibhar: Added it to the `ProposalBuffer` instead, as it made the most sense to have it there. It is polled from the `ProposalReceiver` though.

hrxi (approved):
LGTM, rebasing to get it signed.

PR #1976, opened 2023-11-17
Fix historic transaction hash collisions

### LINKED ISSUE #1935: The data hashed for `HistoricTransaction` isn't necessarily unique
Likely not an issue in practice, but should be fixed regardless. `HistoricTransaction` delegates hashing to its contained `Transaction` if it's a `HistoricTransaction::Basic`. This might collide with other variants of `HistoricTransaction`.

This is done so that the `HistoricTransaction` corresponding to a known `Transaction` can be looked up easily.

To fix this, I propose to instead hash the hash of the `HistoricTransaction` together with a unique prefix so that you can still look up the `Transaction` by hash, but the hash becomes unique.

CC #1882.

hrxi:
Include block number and network ID in the hash. Also make sure that they can't clash with normal transactions.

Fixes #1935.

### DIFF
--- a/primitives/transaction/src/historic_transaction.rs
+++ b/primitives/transaction/src/historic_transaction.rs
@@ -1,7 +1,7 @@
-use std::{error, fmt, io, ops::Range};
+use std::{error, fmt, io, io::Write, ops::Range};
 
 use nimiq_database_value::{FromDatabaseValue, IntoDatabaseValue};
-use nimiq_hash::{Blake2bHash, Hash};
+use nimiq_hash::{Blake2bHash, Blake2bHasher, Hash, Hasher};
 use nimiq_hash_derive::SerializeContent;
 use nimiq_keys::Address;
 use nimiq_mmr::hash::Hash as MMRHash;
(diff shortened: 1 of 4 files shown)


primitives/transaction/src/historic_transaction.rs
```diff
@@ -159,15 +159,33 @@ impl HistoricTransaction {
...
-            // TODO: check for freedom of hash prefixes
-            _ => self.data.hash(),
+            _ => {
+                let block_height = if !matches!(self.data, HistoricTransactionData::Equivocation(_))
+                {
+                    self.block_number
+                } else {
+                    // Equivocations are not distinguished by block height.
+                    0
+                };
```
sisou: I'd turn this condition around. It makes more sense if the if block handles the match with the specifically mentioned data type, and the else block handles all other cases.

```suggestion
                let block_height = if matches!(self.data, HistoricTransactionData::Equivocation(_))
                {
                    // Equivocations are not distinguished by block height.
                    0
                } else {
                    self.block_number
                };
```

primitives/transaction/src/historic_transaction.rs
```diff
@@ -159,15 +159,33 @@ impl HistoricTransaction {
...
+                hasher.write_all(b"\x00\x00").unwrap(); // recipient data len
+                {} // recipient data
+                hasher.write_all(b"\x00\x00\x00\x00\x00\x00\x00\x00").unwrap(); // sender
+                hasher.write_all(b"\x00\x00\x00\x00\x00\x00\x00\x00").unwrap(); // sender
+                hasher.write_all(b"\x00\x00\x00\x00").unwrap(); // sender
+                hasher.write_all(b"\xff").unwrap(); // sender_type (HistoricTransaction = 0xff)
+                self.network_id.serialize_to_writer(&mut hasher).unwrap();
+                block_height.serialize_to_writer(&mut hasher).unwrap();
+                self.data.serialize_to_writer(&mut hasher).unwrap();
+                hasher.finish()
```
sisou, line 187: I don't understand the comment. Do you want to adhere to a transaction format, or don't? Because you write you don't, but then start exactly like a transaction would. But then after the sender type serialize completely different. So it is _not_ a transaction serialization, in which case the static data length and sender stuff make no sense and can be left out.

```suggestion
                let mut hasher = Blake2bHasher::new();
                block_height.serialize_to_writer(&mut hasher).unwrap();
                self.network_id.serialize_to_writer(&mut hasher).unwrap();
                self.data.serialize_to_writer(&mut hasher).unwrap();
                hasher.finish()
```

primitives/transaction/src/historic_transaction.rs
hrxi: Done.

viquezclaudio:
Just verified this solves the collisions I was having in some of my tests 👍

viquezclaudio (approved):
LGTM

primitives/transaction/src/historic_transaction.rs
```diff
@@ -159,15 +159,37 @@ impl HistoricTransaction {
...
+                    // Equivocations are not distinguished by block height.
+                    0
+                } else {
+                    self.block_number
+                };
+                let mut hasher = Blake2bHasher::new();
+                // Ensure that the data here doesn't form a valid transaction
+                // by adhering to the same serialization structure as
+                // `Transaction::serialize_content`.
+                hasher.write_all(b"\x00\x00").unwrap(); // recipient data len
```
sisou, line 177: I didn't understand what this comment meant until Philipp explained it to me. I would reword more like

> This ensures a separate hashing-namespace for inherents, by preventing `self.data` from serializing to a valid transaction (accidentally or on purpose), by following the serialization of a transaction until an impossible byte (sender type).

sisou: However, as Philipp also mentioned, tx serialization might change to include execution result, and there is no compiler link between that and the manual bytes here. So how do we ensure that if the serialization changes, this code here is also updated and not forgotten?

A comment at the tx serialization?

hrxi: Did both.

hrxi:
Thanks for the hints about documentation, @sisou. :)

primitives/transaction/src/historic_transaction.rs
jsdanielh: Couldn't we actually implement something that puts zeroes based on the expected types. I'm thinking on something like this:
```
write_all(0_u8 * size_of(Address))
```

PR #1977, opened 2023-11-17
Upgrade libp2p to version 0.53.1

jsdanielh:
Upgrade libp2p to version 0.53.1 using a forked version for now due to the lack of support of workers in the `websocket-websys` transport.
This also modifies most of the network code to use the adapted libp2p naming standard.

## Pull request checklist


hrxi:
Depending on https://github.com/libp2p/rust-libp2p/pull/4889 to get rid of our fork, I guess.

jsdanielh:
> Depending on [libp2p/rust-libp2p#4889](https://github.com/libp2p/rust-libp2p/pull/4889) to get rid of our fork, I guess.

Yes, thanks for adding the reference.

network-libp2p/src/behaviour.rs
```diff
@@ -1,178 +1,178 @@
...
             config.discovery.required_services,
         );
 
         // Request Response behaviour
-        let codec = MessageCodec::default();
-        let protocol = ReqResProtocol::Version1;
-        let config = RequestResponseConfig::default();
-        let request_response =
-            RequestResponse::new(codec, iter::once((protocol, ProtocolSupport::Full)), config);
+        let protocol = StreamProtocol::new("/nimiq/reqres/0.0.1");
```
hrxi, line 97: We could probably look into giving this a better version number at some point.

network-libp2p/src/dispatch/codecs/typed.rs
```diff
@@ -8,50 +8,33 @@
...
+impl request_response::Codec for MessageCodec {
+    type Protocol = StreamProtocol;
     type Request = IncomingRequest;
     type Response = OutgoingResponse;
 
     async fn read_request<T>(&mut self, _: &Self::Protocol, io: &mut T) -> io::Result<Self::Request>
     where
         T: AsyncRead + Unpin + Send,
     {
-        let bytes = upgrade::read_length_prefixed(io, MAX_REQUEST_SIZE).await?;
```
hrxi, line 53: AFAICT, this is changing the network in an incompatible way, is this desired?

network-libp2p/src/dispatch/codecs/typed.rs
```diff
@@ -8,50 +8,33 @@
...
-        Ok(bytes)
+        let mut vec = Vec::new();
+        let _ = io.take(MAX_REQUEST_SIZE).read_to_end(&mut vec).await?;
```
hrxi: This will not error on overlong messages, even though it probably should.

network-libp2p/Cargo.toml
```diff
@@ -51,8 +51,9 @@ nimiq-utils = { workspace = true, features = [
 ] }
 nimiq-validator-network = { workspace = true }
 
+# For now we need to use a fork
 [target.'cfg(not(target_family = "wasm"))'.dependencies]
-libp2p = { git = "https://github.com/jsdanielh/rust-libp2p.git", default-features = false, features = [
+libp2p = { git = "https://github.com/jsdanielh/rust-libp2p.git", branch = "libp2p_0.53.1", default-features = false, features = [
```
hrxi, line 57: It'd probably be nicer to stay on master. Cargo.lock will ensure that you don't accidentally change commits.

hrxi (commented):
Amazing work! :)

I left a couple of comments and questions.

network-libp2p/Cargo.toml
jsdanielh: You mean this fork master? It is because that fork's master has changes that are no longer necessary and are not part of libp2p in any way so I can't rebase it

hrxi: I see.

network-libp2p/src/dispatch/codecs/typed.rs
jsdanielh: Yeah, I didn't think it through. I changed it back to be compatible. Unfortunately libp2p no longer support these functions so needed to me it to our code

network-libp2p/src/dispatch/codecs/typed.rs
jsdanielh: I changed it back to the previous implementation

network-libp2p/src/behaviour.rs
jsdanielh: Followed discovery protocol naming

network-libp2p/src/behaviour.rs
```diff
@@ -1,178 +1,178 @@
...
-                    GossipsubHandlerError,
-                >,
-                std::io::Error,
-            >,
-            PingFailure,
-        >,
-        ConnectionPoolHandlerError,
-    >,
-    ConnectionHandlerUpgrErr<std::io::Error>,
->;
```
thomaseizinger, line 54: It is satisfying to see stuff like this finally gone. For more information, see https://github.com/libp2p/rust-libp2p/issues/3591.

network-libp2p/src/behaviour.rs
```diff
@@ -1,178 +1,178 @@
...
 
 #[derive(Debug)]
-pub enum NimiqEvent {
-    Dht(KademliaEvent),
-    Discovery(DiscoveryEvent),
-    Gossip(GossipsubEvent),
-    Identify(IdentifyEvent),
-    Ping(PingEvent),
-    Pool(ConnectionPoolEvent),
+pub enum Event {
```
thomaseizinger: FYI: You can have this and the below `From` impls auto-generated if you remove the `#[behaviour(out_event = "Event")]` attribute.

network-libp2p/src/connection_pool/handler.rs
thomaseizinger, line 1: :partying_face:

network-libp2p/src/connection_pool/mod.rs
```diff
@@ -1,2 +1,27 @@
...
+/// Connection Pool errors
+#[derive(Clone, Debug, Error)]
+pub enum Error {
+    /// Ip is banned
+    #[error("IP is banned")]
+    BannedIp,
+
+    /// Peer is banned
+    #[error("Peer is banned")]
+    BannedPeer,
```
thomaseizinger, line 14: FYI, there is also a module that supports peer banning: https://docs.rs/libp2p/latest/libp2p/allow_block_list/index.html

network-libp2p/src/discovery/protocol.rs
```diff
@@ -86,7 +86,7 @@ pub enum DiscoveryMessage {
 pub struct DiscoveryProtocol;
```
thomaseizinger, line 86: You might want to use `ReadyUpgrade` here instead.

network-libp2p/src/dispatch/codecs/typed.rs
```diff
@@ -27,30 +23,106 @@ pub struct MessageCodec {}
...
+    Ok(buf)
+}
+
 #[async_trait::async_trait]
-impl RequestResponseCodec for MessageCodec {
-    type Protocol = ReqResProtocol;
+impl request_response::Codec for MessageCodec {
+    type Protocol = StreamProtocol;
     type Request = IncomingRequest;
     type Response = OutgoingResponse;
```
thomaseizinger, line 119: Whilst certainly possible, this trait was designed with the idea in mind that you'd use _typed_ messages here and not just (length-prefixed) bytes :)

thomaseizinger: Sending bytes back and forth over a "generic" request-response protocol kind of negates the idea of using protocols for multiplexing because you need another, nested layer that allows you to differentiate, which message got sent to you.

YMMV but it might be easier to have multiple instances of `libp2p-request-response`, all with a different codec and more specific protocols that only sent specific messages back and forth (or perhaps only a single message-type per protocol).

network-libp2p/src/network.rs
```diff
@@ -370,60 +361,54 @@ impl Network {
...
-                .authenticate(noise::NoiseConfig::xx(noise_keys).into_authenticated())
+                .authenticate(noise::Config::new(keypair).unwrap())
                 .multiplex(yamux)
                 .timeout(std::time::Duration::from_secs(20))
                 .boxed())
         }
     }
 
-    fn new_swarm(
+    async fn new_swarm(
```
thomaseizinger: This `async` seems to be unused.

network-libp2p/src/network.rs
```diff
@@ -370,60 +361,54 @@ impl Network {
...
-            Box::new(move |fut| {
-                executor.exec(fut);
-            }),
-        )
-        .connection_limits(limits)
-        .build()
+        #[cfg(not(target_family = "wasm"))]
+        let swarm = SwarmBuilder::with_existing_identity(keypair)
+            .with_tokio()
+            .with_other_transport(|_| transport)
```
thomaseizinger, line 398: A big benefit of the new `SwarmBuilder` is that it enforces at compile-time, that you compose transports together correctly. For example, DNS needs to wrap all other transports (except relay).

By using `with_other_transport`, you are not leveraging any of this. I would suggest to refactor this into two functions, one of them for `wasm` and the other one `non-wasm` and use the entire `SwarmBuilder` in there, without any further conditionally-compiled code.

network-libp2p/src/network.rs
```diff
@@ -915,34 +914,27 @@ impl Network {
...
-                                log::trace!(?event.peer, "Responded Ping from peer");
-                            }
-                            Ok(PingSuccess::Ping { rtt }) => {
-                                log::trace!(
-                                    ?event.peer,
-                                    ?rtt,
-                                    "Sent Ping and received response to/from peer",
-                                );
+                            Ok(duration) => {
+                                log::trace!(?event.peer, ?duration, "Successful ping from peer");
```
thomaseizinger, line 923: FYI: `libp2p-ping` already has such a log on `debug` level.

thomaseizinger: Also, version `v0.53` is now using `tracing` instead of `log` so if you are wondering, why you aren't seeing any messages, it is because you need to set up a `tracing-subscriber`.

network-libp2p/tests/discovery.rs
```diff
@@ -77,10 +72,20 @@ impl TestNode {
...
-            DiscoveryBehaviour::new(config, keypair, Arc::clone(&peer_contact_book), clock);
+        let behaviour = discovery::Behaviour::new(
+            config,
+            keypair.clone(),
+            Arc::clone(&peer_contact_book),
+            clock,
+        );
 
-        let mut swarm = Swarm::with_threadpool_executor(transport, behaviour, peer_id);
+        let mut swarm = SwarmBuilder::with_existing_identity(keypair)
```
thomaseizinger, line 82: There is also `libp2p-swarm-test` which allows for quick, in-memory testing of `Swarm`s.

thomaseizinger (commented):
`rust-libp2p` maintainer here. Upgrade looks clean, well done.

I've left a couple of comments :)

hrxi:
Thanks for taking a look at this PR, @thomaseizinger.

network-libp2p/src/network.rs
hrxi: We're also using `tracing`, actually, we just renamed it to `log` so as to not touch every file.

> FYI: libp2p-ping already has such a log on debug level.

Since we cap other crates at the `info!` level usually and sometimes increase the logging level for our crates only, this can still make sense.

thomaseizinger: > We're also using `tracing`, actually, we just renamed it to `log` so as to not touch every file.
> 
> > FYI: libp2p-ping already has such a log on debug level.
> 
> Since we cap other crates at the `info!` level usually and sometimes increase the logging level for our crates only, this can still make sense.

Of course! I just mentioned it so you know you could achieve the same thing with configuration instead of writing the code yourself :)

Each connection also has a span that adds context such as the remote's `PeerId` and address: https://github.com/libp2p/rust-libp2p/blob/0ceb6580941c010edff2a5689ada2e9e33c773cb/swarm/src/connection/pool.rs#L534

Hence, all logs from within the `ConnectionHandler` are automatically scoped to this :)

We've only recently transitioned to this and I'd be very curious to learn how helpful it is to users!

jsdanielh:
> `rust-libp2p` maintainer here. Upgrade looks clean, well done.
> 
> I've left a couple of comments :)

Hey, thank you very much for taking the time to review it and kudos for the recent improvements in `rust-libp2p`.
I already started implementing some of your suggestions.

network-libp2p/src/behaviour.rs
jsdanielh: Yes, thanks. Already removed it.

network-libp2p/src/behaviour.rs
jsdanielh: Yeah, it's already making our lives easier :smile:

PR #1987, opened 2023-11-20
ZKP updates (decoupling circuits)

paberr:
So far, each parent circuit had the children's verifying keys embedded as constants.
This creates a chain of dependencies that prohibits running the phase2 ceremony for all of them in parallel.

This PR eliminates the problem by having the VK as _private inputs_ and passing a Pedersen commitment to the VK as _public input_.
That way, the verifier still binds the prover to a specific set of keys and circuits.

Since we need to commit to all verifying keys and want to minimize the number of constraints and public inputs, our commitment consists of two sub-commitments that are concatenated: `C1 || C2` (resulting in 3 public inputs instead of 4)
We have one sub-commitment per pairing (MNT4/6) and use respective Pedersen parameters.
We define `Ci = pedersen_hash(vk1 || vk2 || ...)` over all verifying keys on the respective curve and prove only the keys required on a specific circuit.

## Pull request checklist


zkp-component/tests/zkp_prover.rs
```diff
@@ -18,10 +18,8 @@ async fn can_launch_process_and_parse_output() {
 
     let result = launch_generate_new_proof(recv, proof_input, Some(zkp_test_exe())).await;
 
-    assert_eq!(
-        result,
-        Err(ZKProofGenerationError::NanoZKP("invalid block".to_string()))
-    );
+    // Should fail because verifying keys cannot be loaded.
+    assert!(matches!(result, Err(ZKProofGenerationError::NanoZKP(_))));
```
hrxi, line 22: This seems like it's not a clippy fix, this weakens the test. I talked with @paberr and this is intended.

hrxi (approved):
I haven't looked at the changes in-depth, that would require a lot more time. From a high-level overview, they look good though.

hrxi:
Rebased for merging.

PR #1989, opened 2023-11-20
Validator network: Invalidate validator key->peer id cache on connect…

fiaxh:


validator-network/src/network_impl.rs
```diff
@@ -127,6 +127,19 @@ where
...
     }
+
+    /// Clears the validator->peer_id cache.
+    /// The cached entry should be cleared when the peer id might have changed.
+    fn clear_validator_peer_id_cache(&self, validator_id: usize) {
+        let state = self.state.write();
+        let validator_key = state.validator_keys.get(validator_id);
+        if let Some(validator_key) = validator_key {
+            self.state
+                .write()
```
hrxi: Releasing and immediately acquiring the write lock again sounds wrong to me.

validator-network/src/network_impl.rs
```diff
@@ -199,7 +212,12 @@ where
         if let Ok(peer_id) = self.get_validator_peer_id(validator_id).await {
             self.network
                 .message(msg, peer_id)
-                .map_err(NetworkError::Request)
+                .map_err(|e| {
```
hrxi, line 214: Would a distinction between different kind of errors be feasible?

hrxi (commented):
Looks mostly good, I left two comments.

validator-network/src/network_impl.rs
nibhar: Looks like a reentrant lock here, this would deadlock I believe as the write taken in L134 would not be dropped. I looked a bit at the logs produced by the reconnect test and they support that at least on the surface.

I think keep using the already taken state should fix it.

fiaxh: Indeed that was supposed to use the existing write lock. Fixed.

validator-network/src/network_impl.rs
fiaxh: It looks to me like a validator that changed its peer id should result in it not being dial-able anymore, thus resulting in a libp2p `OutboundFailure::DialFailure`, which we convert into a `RequestError::OutboundRequest(OutboundRequestError::DialFailure)`. However the network implementation isn't trivial and I'm worried that a changed peer id might also result in other errors in some cases or after code adjustments.

I'd lean towards clearing the cache on any error. That will result in some false-positives but also might be a more robust implementation.

fiaxh: ~It looks to me like a validator that changed its peer id should result in it not being dial-able anymore, thus resulting in a libp2p `OutboundFailure::DialFailure`, which we convert into a `RequestError::OutboundRequest(OutboundRequestError::DialFailure)`.~ The scenario currently results in a `InboundRequestError::NoReceiver`. However the network implementation isn't trivial and I'm worried that a changed peer id might also result in other errors in some cases or after code adjustments.

I'd lean towards clearing the cache on any error. That will result in some false-positives but also might be a more robust implementation.

hrxi: Sounds good.

hrxi:
Rebasing to get it merged.

PR #1990, opened 2023-11-20
Drop outdated proposals earlier

nibhar:
Instead of delivering proposals to Tendermint to drop them later outdated proposals can be dropped immediately in the proposal buffer and this PR adds that functionality.


hrxi (approved):
LGTM.

PR #2001, opened 2023-11-22
Missing blocks request improvements

### LINKED ISSUE #458: Retry failed missing block requests
If a missing block request failed for all synced peers, we currently give up. That leads to any further blocks just being buffered, but they can't be pushed to the blockchain due to the missing block. We should re-request the missing block.

### LINKED ISSUE #1535: Missing block request hardening
With the new ordering of missing block responses (#1527), we introduced a new attack vector.
A peer can announce a block in the future, and we will request missing blocks to get there.

Let's say our current block height is 20.
By always returning a set of successive blocks from block 20 that never lead to the target block, the peer can make us request a lot of data.

Previously, when returning blocks backwards, it was guaranteed that the missing blocks requests/responses would terminate (because there can be only one chain leading towards the target block).
Now, there are infinite chains starting at one of our blocks and there is no way to check they actually lead to the target block.

paberr:
This PR improves the missing block requests in three dimensions:

1. We first try to request from the peer who announced the given block.
2. We improve the checks on verifying the response:
    * We still check the hash chain.
    * We additionally check that the last block in the response is either the target block or a macro block.
    * We verify the signature on the macro block (this makes the response valid as long as the security assumption holds).
3. We retry if we get an invalid response.

## Pull request checklist

### DIFF
--- a/consensus/src/sync/history/cluster.rs
+++ b/consensus/src/sync/history/cluster.rs
@@ -194,10 +194,13 @@ impl<TNetwork: Network + 'static> SyncCluster<TNetwork> {
                 validators: blockchain.current_validators().unwrap(),
             }
         };
-
+        let epoch_ids_queue = epoch_ids
+            .iter()
+            .map(|epoch_id| (epoch_id.clone(), None))
+            .collect();
         let batch_set_queue = SyncQueue::with_verification(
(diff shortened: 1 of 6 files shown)


hrxi (approved):
LGTM, rebasing. :)

PR #2003, opened 2023-11-23
Change policy constants

nibhar:
Change the following policy constants:
- Validator deposit: Increased to 100000 NIM.
- Batches per epoch: Increased to 720 batches.


hrxi:
Needs some test fixes.

PR #2009, opened 2023-11-23
Reduce reliance on blake2 internals, move `RequestCommon::*serialize*`

hrxi:


hrxi:
Requires regenerating the ZKP keys.

Maybe we can get rid of the blake2-rfc dependency entirely once the `blake2` crate supports blake2x: https://github.com/RustCrypto/hashes/issues/83.

jsdanielh (approved):

PR #2026, opened 2023-12-01
Internally tag all web-client plain transaction data variants

sisou:

This PR adds a `type` field to all plain transaction data and proof variants in the web-client, to let users easily distinguish between data types. This will be especially helpful in our Keyguard to distinguish between staking transaction types, without having to hard-code the variant ID in the Keyguard. Experience showed that these IDs sometimes shifted, requiring manual updates. With this change, the update would be included in the client update.

## Pull request checklist


jsdanielh (approved):

jsdanielh:
Rebasing it to merge it

PR #2032, opened 2023-12-06
Use Rust 1.74 feature for lints

hrxi:


hrxi:
Not sure if we still want to wait with this for a bit, since it requires the latest stable Rust version (1.74), released Nov 16.

jsdanielh (approved):

PR #2033, opened 2023-12-06
Equivocation proofs inclusion to history store

### LINKED ISSUE #1978: `HistoricTransactionData::Equivocation` is not inserted into the history store
Thanks to @styppo for noticing. A test would also be nice.

ii-cruz:
Makes us store equivocation proofs in the history store when we push a micro block. It adds a test for this specific scenario as well.

This PR also adds tests pushing and removing different inherents and equivocation proofs to the history store test.

## Pull request checklist

### DIFF
--- a/primitives/transaction/src/historic_transaction.rs
+++ b/primitives/transaction/src/historic_transaction.rs
@@ -56,6 +56,7 @@ impl HistoricTransaction {
         block_time: u64,
         transactions: Vec<ExecutedTransaction>,
         inherents: Vec<Inherent>,
+        equivocation_locator: Vec<EquivocationLocator>,
     ) -> Vec<HistoricTransaction> {
         let mut hist_txs = vec![];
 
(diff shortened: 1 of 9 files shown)


primitives/transaction/src/historic_transaction.rs
```diff
@@ -115,10 +116,23 @@ impl HistoricTransaction {
...
+        }
+
         hist_txs
     }
 
-    pub fn count(num_transactions: usize, inherents: &[Inherent]) -> usize {
+    pub fn count(
+        num_transactions: usize,
+        inherents: &[Inherent],
+        equivocation_locator: Vec<EquivocationLocator>,
```
hrxi, line 134:
```suggestion
        equivocation_locator: &[EquivocationLocator],
```

primitives/transaction/src/historic_transaction.rs
```diff
@@ -56,6 +56,7 @@ impl HistoricTransaction {
         block_time: u64,
         transactions: Vec<ExecutedTransaction>,
         inherents: Vec<Inherent>,
+        equivocation_locator: Vec<EquivocationLocator>,
```
hrxi, line 59:
```suggestion
        equivocation_locator: Vec<EquivocationProof>,
```
All call sites have `EquivocationProof`s, not equivocation locators. Maybe it'd be better to get it as such, makes the call sites easier.

hrxi (commented):
Looks good to me, just two minor style nits. Could be merged without addressing those.

hrxi:
We also talked about putting the equivocation locators before the inherents into the history store.

PR #2038, opened 2023-12-08
network-libp2p: Expose a configuration for the desired number of peers

jsdanielh:
- Expose a configuration for the desired number of peers for the connection pool to realize when to try to connect to more peers. This configuration is exposed in the network and is set in the client to be `consensus.min_peers`.
- Make the `ConnectionState` aware of the desired number of connections such that when doing `housekeeping` it resets the connections marked as down after 1s when the number of connections is below the desired number. This changes how eagerly connections are re-tried when the number of connected peers is less than the desired value.

## Pull request checklist


network-libp2p/src/config.rs
```diff
@@ -27,6 +27,7 @@ pub struct Config {
     pub memory_transport: bool,
     pub required_services: Services,
     pub tls: Option<TlsConfig>,
+    pub desired_peer_count: usize,
```
hrxi, line 30:
```suggestion
    pub desired_peer_count: u32,
```
I don't see a good reason to make the size of this integer platform-specific, the desired peer count doesn't have anything to do with whether the processor is 32 bit or 64 bit.

sisou: `u8` or `u16` should be enough though.

hrxi: Yep, at least `u16` should be enough.

I think Rust and Google coding guidelines say that you should mostly use 32-bit and 64-bit integers though, and don't think too hard about smaller sizes.

hrxi: @jsdanielh and I discussed this offline and figured out that this is too little of an issue in practice to deal with changing the type.

lib/src/client.rs
```diff
@@ -297,6 +297,7 @@ impl ClientInner {
             false,
             required_services,
             tls_config,
+            config.consensus.min_peers,
```
sisou, line 300: Should this not pass in the _desired_ peer count?

PR #2039, opened 2023-12-11
Add fallen behind peers to the timeout collection as well.

nibhar:
The `outdated_timeouts` are used to re add peers to the macro sync once elapsed. Not being in there meant that the specific peer will stay in the `outdated_peers` set forever once it fell behind.


hrxi:
Rebased to sign it.

jsdanielh (approved):

PR #2054, opened 2023-12-14
network-libp2p: Add the autonat behaviour

jsdanielh:
- Add the `autonat` behaviour to detect when a peer is behind a private `NAT`. If in the other hand a peer has a public rachable IP, then the behaviour will fire the `ExternalAddrConfirmed` event and will put the `kad` behaviour (DHT) into server mode.
  This removes the previous setting where the DHT was being set always in server mode regardless of the node reachability except for the memory transport where Autonat doesn't work (only works with IPs).
- Add a configuration file option to allow `Autonat` to use non global IPs. This is useful in our `devnet` testing scenarios since it
  regularly consists of nodes with non global IPs (like loopback addresses, etc).

## Pull request checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -488,6 +488,19 @@ dependencies = [
 
 [[package]]
 name = "asynchronous-codec"
+version = "0.6.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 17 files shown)


hrxi (approved):
Looks good to me. :)

hrxi:
Rebasing.

PR #2063, opened 2023-12-19
Fix interlink hop calculation for block inclusion proofs

### LINKED ISSUE #2059: Web-client cannot fetch transactions from past epochs
## New issue checklist

- [x] I have read all of the [`README`](https://github.com/nimiq-network/core/blob/master/README.md) 
- [x] I have searched [existing issues](https://github.com/nimiq-network/core/issues?q=is%3Aissue+sort%3Acreated-desc) and **this is not a duplicate**.

### General information

- Library version(s): 0.18.0
- Browser version(s): Chrome
- Devices/Simulators/Machine affected:
- Related issues:

#### Expected behavior

Client can fetch and return transactions history of an address across epochs.

#### Actual behavior

Client can only fetch transactions from current epoch, past epoch tx proof do not verify. 

#### Steps to reproduce

Fetch tx history for an address that has a tx in a past epoch. 

I'm still investigating the root cause, but so far I have found that the extra block proof that is requested for past epoch txs fails to verify, specifically when checking that all necessary interlink hops are included in the proof. The proof does not contain any blocks, so contains no hops, but in my example two hops are required to be proven. 

None

sisou:

Block inclusion proofs for past epochs were failing, because the response contained no blocks. The reason is that the calculation for which blocks must be included (hops) was not updated to non-zero genesis block numbers and still used the naive blocknumber-divided-by-blocks-per-epoch approach.

A second commit adds an early "return" (actually a `continue`) for the case when the `HistoryTreeProof` already fails to verify. If it does, the `verification_result` cannot become `true` anymore, so in order to prevent the fetching of block inclusion proofs (both for accidental and purposefully invalid proofs) in this case and to reduce time spent in the loop, we can already skip to the next loop iteration, as it would anyway at the end of the loop body where the `verification_result` is checked.

## Pull request checklist

### DIFF
--- a/consensus/src/consensus/consensus_proxy.rs
+++ b/consensus/src/consensus/consensus_proxy.rs
@@ -337,6 +337,12 @@ impl<N: Network> ConsensusProxy<N> {
                                     .verify(block.history_root().clone())
                                     .map_or(false, |result| result);
 
+                                if !verification_result {
+                                    // If the proof didn't verify, we continue with another peer
+                                    log::warn!(peer=%peer_id, "The transaction history proof from this peer did not verify");
+                                    continue;
+                                }
+
                                 // Verify that the transaction proof fits to the chain
                                 if block.block_number() <= election_head.block_number() {
                                     let block_hash = block.hash();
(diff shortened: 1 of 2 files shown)


jsdanielh (approved):
LGTM

viquezclaudio (approved):

PR #2064, opened 2023-12-20
Simplify `Handle` trait

hrxi:
The response is no longer an extra type parameter but instead derived from the new `Request` bound.


hrxi:
Superseded by #2099.

PR #2070, opened 2023-12-21
Update the web-client README

viquezclaudio:
Update the web-client README to be consistent with the latest scripts

### DIFF
--- a/web-client/README.md
+++ b/web-client/README.md
@@ -10,11 +10,11 @@ as its Javascript environments.
 To run the example, first build the web-client by running the following script from this directory:
 
 ```sh
-./scripts/build.sh
+./scripts/build.sh --only web
 ```
 
-This script builds the `web`, `bundler`, and `node` targets and generates the corresponding JS
-bindings required to run the client in a web browser.
+This script builds the `web` wasm-bindgen target and generates the corresponding JS bindings required
+to run the client in a web browser.
 
 After the client has been built, this directory can be served with a web server (e.g. `python3 -m http.server`)
 and then the `/example/index.html` file can be loaded from e.g. http://localhost:8000/example/index.html.
--- a/web-client/README.md
+++ b/web-client/README.md
@@ -2,39 +2,19 @@
 
 This client is a very light client that only includes the necessary dependencies and constructs
 to compile a client to WebAssembly and to run it in a web browser. This is a web-client intended
-to be used in web browsers only (no WASI support). Although it can be built for other targets,
-it will panic if it is executed outside a web browser.
+to be used in web browsers only (no WASI support). It currently supports modern browsers and NodeJS
+as its Javascript environments.
 
-## Building the example
+## Running the example
 
-To build the client, the recommended way is to use [`wasm-pack`](https://rustwasm.github.io/wasm-pack/).
-This tool can be obtained using:
-
-```
-cargo install wasm-pack
-```
-
-or via their website. Once installed, the client can be built from this directory:
+To run the example, first build the web-client by running the following script from this directory:
 
 ```sh
-./scripts/build-web.sh && ./scripts/build-launcher.sh
+./scripts/build.sh
 ```
 
-The above command will compile the Rust code to WebAssembly and generate the corresponding JS
+This script builds the `web`, `bundler`, and `node` targets and generates the corresponding JS
 bindings required to run the client in a web browser.
 
 After the client has been built, this directory can be served with a web server (e.g. `python3 -m http.server`)
 and then the `/example/index.html` file can be loaded from e.g. http://localhost:8000/example/index.html.
-
-## Publishing to NPM
-
-> **Warning**
-> You must be using wasm-pack version >= 0.11.0.
-
-To publish this package to NPM, run this command:
-
-```sh
-./scripts/npm-publish.sh
-```
-
-This script builds the `web`, `bundler`, and `node` targets, the types and bindings and publishes the package.
--- a/web-client/scripts/build.sh
+++ b/web-client/scripts/build.sh
@@ -1,20 +1,43 @@
+#!/bin/bash
+
 set -e
 
 # TODO: Disable rust compiler warnings with [lints] table in Cargo.toml when it becomes stable in 1.75 (https://github.com/rust-lang/cargo/issues/12115)
 
+# Defaults
 CARGO_PROFILE="release-wasm"
 CARGO_TARGET="wasm32-unknown-unknown"
-CARGO_OUTPUT="../target/$CARGO_TARGET/$CARGO_PROFILE/nimiq_web_client.wasm"
+TARGETS="bundler,web,nodejs"
+RUN_WASM_OPT=true
+BUILD_TYPES=true
+BUILD_LAUNCHER=true
+
+# Parse command line arguments - all arguments are optional
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --cargo-profile) CARGO_PROFILE="$2"; shift ;;
+        --cargo-target) CARGO_TARGET="$2"; shift ;;
+        --cargo-dir) CARGO_OUTPUT="$2/nimiq_web_client.wasm"; shift ;;
+        -o|--only) TARGETS="$2"; shift ;;
+        --skip-wasm-opt) RUN_WASM_OPT=false ;;
+        --skip-types) BUILD_TYPES=false ;;
+        --skip-launcher) BUILD_LAUNCHER=false ;;
+        *) echo "Unknown argument: $1"; exit 1 ;;
+    esac
+    shift
+done
+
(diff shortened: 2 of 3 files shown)


jsdanielh (approved):

web-client/README.md
```diff
@@ -17,24 +17,11 @@ cargo install wasm-pack
 or via their website. Once installed, the client can be built from this directory:
 
 ```sh
-./scripts/build-web.sh && ./scripts/build-launcher.sh
+./scripts/build.sh && ./scripts/build-launcher.sh
```
sisou: build-launcher.sh is included in build.sh now.

web-client/scripts/build.sh
```diff
@@ -51,25 +76,43 @@ then
...
+fi
 
 # Types
-printf "Building types...\n"
-compile "client,primitives"
-wasm-bindgen --weak-refs --target web --out-name web --out-dir dist/types $CARGO_OUTPUT
-wasm-bindgen --weak-refs --target bundler --out-name bundler --out-dir dist/types $CARGO_OUTPUT
-find dist/types ! -name 'web.d.ts' ! -name 'bundler.d.ts' -type f -exec rm {} +
+if [ "$BUILD_TYPES" = "true" ] ; then
+    printf "Building types...\n"
```
jsdanielh:
```suggestion
    echo "Building types..."
```

web-client/scripts/build.sh
```diff
@@ -51,25 +76,43 @@ then
...
+    wasm-bindgen --weak-refs --target web --out-name web --out-dir dist/types $CARGO_OUTPUT
+    wasm-bindgen --weak-refs --target bundler --out-name bundler --out-dir dist/types $CARGO_OUTPUT
+    find dist/types ! -name 'web.d.ts' ! -name 'bundler.d.ts' -type f -exec rm {} +
+fi
 
 # Build launcher
-printf "Building launcher...\n"
-./scripts/build-launcher.sh
+if [ "$BUILD_LAUNCHER" = "true" ] ; then
+    printf "Building launcher...\n"
```
jsdanielh:
```suggestion
    echo "Building launcher..."
```

web-client/scripts/build.sh
```diff
@@ -51,25 +76,43 @@ then
...
+    generate "worker" "nodejs"
+fi
 
 # Types
-printf "Building types...\n"
-compile "client,primitives"
-wasm-bindgen --weak-refs --target web --out-name web --out-dir dist/types $CARGO_OUTPUT
-wasm-bindgen --weak-refs --target bundler --out-name bundler --out-dir dist/types $CARGO_OUTPUT
-find dist/types ! -name 'web.d.ts' ! -name 'bundler.d.ts' -type f -exec rm {} +
+if [ "$BUILD_TYPES" = "true" ] ; then
```
jsdanielh:
```suggestion
if $BUILD_TYPES" ; then
```

web-client/scripts/build.sh
```diff
@@ -33,10 +55,13 @@ function generate() {
 
     # Generate JS bindings (also generates a GC'd WASM file)
     wasm-bindgen --weak-refs --target $TARGET --out-name index --out-dir $OUT_DIR --no-typescript $CARGO_OUTPUT
-    # Optimize bindgen's WASM output for size
-    wasm-opt -Os -o $OPT_OUTPUT $BINDGEN_OUTPUT
-    # Replace bindgen's WASM output with the optimized one
-    mv $OPT_OUTPUT $BINDGEN_OUTPUT
+
+    if [ "$RUN_WASM_OPT" = "true" ] ; then
```
jsdanielh:
```suggestion
    if $RUN_WASM_OPT ; then
```

web-client/scripts/build.sh
```diff
@@ -51,25 +76,43 @@ then
...
+    compile "client,primitives"
+    wasm-bindgen --weak-refs --target web --out-name web --out-dir dist/types $CARGO_OUTPUT
+    wasm-bindgen --weak-refs --target bundler --out-name bundler --out-dir dist/types $CARGO_OUTPUT
+    find dist/types ! -name 'web.d.ts' ! -name 'bundler.d.ts' -type f -exec rm {} +
+fi
 
 # Build launcher
-printf "Building launcher...\n"
-./scripts/build-launcher.sh
+if [ "$BUILD_LAUNCHER" = "true" ] ; then
```
jsdanielh:
```suggestion
if $BUILD_LAUNCHER ; then
```

web-client/scripts/build.sh
```diff
@@ -1,17 +1,39 @@
...
+    shift
+done
 
 # Compile web-client to WASM for a given feature set
 # $1: features to enable
 function compile() {
     FEATURES=$1
 
-    printf "Compiling WASM for $FEATURES...\n"
+    echo "Compiling WASM for $FEATURES...\n"
```
hrxi: Here and below: If you change `printf` to `echo`, you need to remove the final `\n`.

jsdanielh (approved):

web-client/scripts/build.sh
```diff
@@ -29,47 +51,68 @@ function generate() {
...
+    generate "worker" "nodejs"
+fi
 
 # Types
-printf "Building types...\n"
-compile "client,primitives"
-wasm-bindgen --weak-refs --target web --out-name web --out-dir dist/types $CARGO_OUTPUT
-wasm-bindgen --weak-refs --target bundler --out-name bundler --out-dir dist/types $CARGO_OUTPUT
-find dist/types ! -name 'web.d.ts' ! -name 'bundler.d.ts' -type f -exec rm {} +
+if $BUILD_TYPES ; then
```
hrxi: This runs the command in `$BUILD_TYPES`, I'd consider that unexpected. Could you change the variable to contain 0 or 1 instead and compare with `[[` or `[`?

web-client/scripts/build.sh
```diff
@@ -29,47 +51,68 @@ function generate() {
     BINDGEN_OUTPUT="$OUT_DIR/index_bg.wasm"
     OPT_OUTPUT="$BINDGEN_OUTPUT.opt"
 
-    printf "Generating $TYPE WASM for target $TARGET...\n"
+    echo "Generating $TYPE WASM for target $TARGET...\n"
 
     # Generate JS bindings (also generates a GC'd WASM file)
     wasm-bindgen --weak-refs --target $TARGET --out-name index --out-dir $OUT_DIR --no-typescript $CARGO_OUTPUT
```
hrxi: Here and below, needs double quotes around strings involving variables such as `$CARGO_OUTPUT`.

web-client/scripts/build.sh
```diff
@@ -1,17 +1,39 @@
+#!/bin/bash
+
 set -e
 
 # TODO: Disable rust compiler warnings with [lints] table in Cargo.toml when it becomes stable in 1.75 (https://github.com/rust-lang/cargo/issues/12115)
 
+# Defaults
 CARGO_PROFILE="release-wasm"
 CARGO_TARGET="wasm32-unknown-unknown"
 CARGO_OUTPUT="../target/$CARGO_TARGET/$CARGO_PROFILE/nimiq_web_client.wasm"
```
hrxi: This would need to be set again if `$CARGO_TARGET` or `$CARGO_PROFILE` changes.

hrxi (commented):
Couple of shell scripting gotchas.

web-client/scripts/build.sh
sisou: This was actually changed by comment from JD: https://github.com/nimiq/core-rs-albatross/pull/2070#discussion_r1434679800

I'd also prefer my original version with string comparison.

jsdanielh: `true` and `false` are valid commands. I'd say this is expected and the string comparison is not needed

hrxi: The current statement works.

I found it unexpected that the command inside a variable is run, instead of compared as string. If I need to assess the security of this, I need to find out all places where the variable is set, to ensure that nothing unexpected can happen. If it's a string comparison, I don't even need to look.

jsdanielh: Mmm, I see your point but in the script there are only two places where we're setting this variable and they are only set to `true` and `false`. Having said this, I don't have a strong opinion on this so we could go to the previous version.

web-client/scripts/build.sh
sisou: You are right. But they don't change, so I renamed them to `# Configuration`, as that better describes what these are.

sisou: Oh wait, I did not read my own code :D They change with command line arguments. Hold on...

sisou: I moved the definition of `CARGO_OUTPUT` to after the arguments-handling, setting the default if it was not set from an argument. Is that right?

hrxi: Looks correct, thanks!

hrxi (approved):
LGTM.

hrxi:
Rebased for merging.

PR #2106, opened 2024-01-05
network-libp2p: Filter valid addresses and fix discovery protocol

### LINKED ISSUE #2052: Peer Filtering
Especially for the WebClient there needs to be an option do forbid connections to insecure sockets.  
Otherwise the browser will flag the page as insecure.

Suggested filter rules:
- disable Insecure: true/false    ->disables ws://
- disable local IPs: true/false    ->disables all non public IPS

jsdanielh:
- Add several changes to filter valid addresses:
  - Remove the `identify` behaviour since it was returning peer addresses that sometimes are not valid such as secure websocket and IPs multiaddresses.
  - Let know the discovery behaviour about new listen addresses to mimic the `identify` behaviour and be able to discover more of our own addresses.
  - Add some filtering in the `PeerContactBook` such that the `discovery` and the `connection_pool` behaviours only return valid addresses (in `handle_pending_outbound_connection`).
  - Add a configuration knob in the network to allow only secure websocket connections and another one for allowing loopback addresses.
  - Only add an address for a peer in the DHT behaviour if it passes the validity filters such that it doesn't return invalid addresses when the swarm dials a peer.
  This fixes #2052.
- Fix the discovery protocol to always periodically send our own `PeerContact` (along with some of our contacts) which is updated frequently to avoid being aged.

## Pull request checklist

### DIFF
--- a/network-libp2p/src/config.rs
+++ b/network-libp2p/src/config.rs
@@ -31,6 +31,8 @@ pub struct Config {
     pub required_services: Services,
     pub tls: Option<TlsConfig>,
     pub autonat_allow_non_global_ips: bool,
+    pub only_secure_ws_connections: bool,
+    pub allow_loopback_addresses: bool,
 }
 
 impl Config {
@@ -43,6 +45,8 @@ impl Config {
         required_services: Services,
         tls_settings: Option<TlsConfig>,
         autonat_allow_non_global_ips: bool,
+        only_secure_ws_connections: bool,
+        allow_loopback_addresses: bool,
     ) -> Self {
         // Hardcoding the minimum number of peers in mesh network before adding more
         // TODO: Maybe change this to a mesh limits configuration argument of this function
@@ -58,7 +62,7 @@ impl Config {
                 let mut s = DefaultHasher::new();
                 message.topic.hash(&mut s);
                 message.data.hash(&mut s);
-                gossipsub::MessageId::from(s.finish().to_string())
+                gossipsub::MessageId::from(s.finish().to_be_bytes())
             })
             .build()
             .expect("Invalid Gossipsub config");
@@ -76,13 +80,19 @@ impl Config {
             keypair,
             peer_contact,
             seeds,
-            discovery: discovery::Config::new(genesis_hash, required_services),
+            discovery: discovery::Config::new(
+                genesis_hash,
+                required_services,
+                only_secure_ws_connections,
+            ),
             kademlia,
             gossipsub,
             memory_transport,
             required_services,
             tls: tls_settings,
             autonat_allow_non_global_ips,
+            only_secure_ws_connections,
+            allow_loopback_addresses,
         }
     }
 }
--- a/network-libp2p/src/config.rs
+++ b/network-libp2p/src/config.rs
@@ -1,12 +1,9 @@
-use std::{
-    collections::hash_map::DefaultHasher,
-    hash::{Hash, Hasher},
-    time::Duration,
-};
+use std::time::Duration;
 
 use libp2p::{gossipsub, identity::Keypair, kad, Multiaddr};
 use nimiq_hash::Blake2bHash;
 use nimiq_network_interface::peer_info::Services;
+use sha2::{Digest, Sha256};
 
 use crate::discovery::{self, peer_contacts::PeerContact};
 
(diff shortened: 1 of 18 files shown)


network-libp2p/src/discovery/handler.rs
```diff
@@ -210,14 +210,18 @@ impl Handler {
 
     /// Get peer contacts from our contact book to send to this peer. The contacts are filtered according to the peer's
     /// protocols and service filters, they are limited to the number of peers specified by the peer.
-    fn get_peer_contacts(&self, peer_contact_book: &PeerContactBook) -> Vec<SignedPeerContact> {
-        let n = self.peer_list_limit.unwrap() as usize;
-
+    /// This list also includes our own contact which should be already filtered since we already have
+    /// connections to these peers.
+    fn get_peer_contacts(
+        &self,
```
hrxi, line 216:
```suggestion
```
No longer necessary.

network-libp2p/src/config.rs
```diff
@@ -58,7 +62,7 @@ impl Config {
                 let mut s = DefaultHasher::new();
                 message.topic.hash(&mut s);
                 message.data.hash(&mut s);
-                gossipsub::MessageId::from(s.finish().to_string())
+                gossipsub::MessageId::from(s.finish().to_be_bytes())
```
hrxi: Unrelated to this PR: `DefaultHasher` is not guaranteed to stay the same across Rust versions. Is it important that this is the same for all of the network? Couldn't immediately see that from the libp2p-gossipsub documentation: [`message_id_fn`](https://docs.rs/libp2p-gossipsub/0.46.1/libp2p_gossipsub/struct.ConfigBuilder.html#method.message_id_fn).

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -299,35 +316,43 @@ impl PeerContactBook {
...
+    pub fn insert_filtered(
+        &mut self,
+        contact: SignedPeerContact,
+        services_filter: Services,
+        only_secure_connections: bool,
+    ) {
         let info = PeerContactInfo::from(contact);
 
-        if self
+        // A peer is interested to us in two cases:
```
hrxi:
```suggestion
        // A peer is interesting to us in two cases:
```
Looks like grammar but inverts the meaning.

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -299,35 +316,43 @@ impl PeerContactBook {
...
+        // - I'm configured as a validator, and the peer is also a validator, then that peer is interesting to me
+        //   regardless of the services that are provided by that peer.
+        // - There is a match in the services provided by the peer and our services
```
hrxi:
```suggestion
        // - The services provided by the peer are a superset of the requested services.
```
It was unclear to me what a "match" was supposed to mean. Digging into the source code, it seems to mean "superset".

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -452,6 +491,69 @@ impl PeerContactBook {
...
+
+    /// Returns true if an address is a secure websocket connection.
+    /// If address doesn't have the websocket protocol, it will return `false`.
+    fn is_address_secure(address: &Multiaddr) -> bool {
+        address.into_iter().any(|p| matches!(p, Protocol::Wss(_)))
+    }
+
+    /// Returns true if an address is valid for dialing.
+    /// It performs basic checks against unsupported addresses.
+    pub fn is_address_valid(&self, address: &Multiaddr) -> bool {
```
hrxi:
```suggestion
    pub fn is_address_dialable(&self, address: &Multiaddr) -> bool {
```
`valid` doesn't really tell me what this function does. Perhaps `dialable` would fit?

hrxi (commented):
On a more general note, maybe we shouldn't call the wss connections "secure". As far as I know, all of our connections are secure. The difference to wss is that it has two layers that try to secure the connection, the transport and the libp2p protocol itself. I'm not sure what to call it, maybe "secure transport" or simply "secure websocket"? Maybe the latter.

network-libp2p/src/discovery/handler.rs
jsdanielh: It's still required for `self.service_filter`

network-libp2p/src/config.rs
jsdanielh: Mmm, I'd say it is important. `libp2p-gossipsub` documentation is not very clear on this. But then, how would you propose to go since we can't use the inner `SipHashers` as they are deprecated?

hrxi (approved):
This looks good to me now. Does it require waiting for a testnet restart? @jsdanielh

network-libp2p/src/config.rs
hrxi: We could use a standard cryptographic hash function, I guess.

jsdanielh:
> This looks good to me now. Does it require waiting for a testnet restart? @jsdanielh

Mmm, I'd say that the change in gossipsub might need it

network-libp2p/src/discovery/handler.rs
hrxi: Oops.

network-libp2p/src/config.rs
jsdanielh: I just changed it to use `sha256`

hrxi: Nice. :)

PR #2114, opened 2024-01-09
Retire Stake and Minimum stake

### LINKED ISSUE #2000: Support minimum stake
So we don't spam staking contract too much.

Maybe 100 NIM.

ii-cruz:

- Retire stake was added to the staker. It is required so that we are sure there are enough funds to deduct the remove stake fee on a commit failed transaction.
- Introduces the Minimum Stake of 100 NIM. This is to prevent a blown-up state size for the Staking Contract. Thus we add two minimum stake invariants to the staker: 
Total balance of a staker (active + inactive + retire balances) >= Minimum Stake OR == 0
Non retired balances of a staker (active + inactive balances) >= Minimum Stake OR == 0
- Adds all the test case scenarios testing the new invariants and the new retired balance state.
- Restricts the remove stake to only allow the removal of all retired balance.
- Finishes renaming all `unstake` transactions to `remove_stake` to make it standard across the codebase.
- Finishes renaming all `stake` transactions to `add_stake` to make it standard across the codebase.

## Pull request checklist


web-client/src/transaction.rs
```diff
@@ -837,6 +846,14 @@ pub struct PlainSetActiveStakeData {
     pub new_active_balance: u64,
 }
 
+/// JSON-compatible and human-readable format of set inactive stake data.
```
sisou: Copy-pasta
```suggestion
/// JSON-compatible and human-readable format of retire stake data.
```

sisou: Please also fix the comment for `PlainSetActiveStakeData`, to read "set active stake data".

ii-cruz: Fixed! thank you so much for the good eye🕵️‍♂️

sisou:
From only reading the PR description, not any code yet: these to conditions seem like they could be simplified:

> Total balance of a staker (active + inactive + retire balances) >= Minimum Stake OR == 0
> Non retired balances of a staker (active + inactive balances) >= Minimum Stake OR == 0

Since the second condition is a subset of the first, the first condition can be reduced to simply "retired balance >= minimum stake OR == 0" and does not have to include the two other ones.

primitives/src/policy.rs
```diff
@@ -105,6 +105,12 @@ impl Policy {
     /// in proof-of-work.
     pub const VALIDATOR_DEPOSIT: u64 = 10_000_000_000;
 
+    /// The stake necessary to create a staker in Lunas (1 NIM = 100,000 Lunas).
+    /// This minimum is applied to the amount of:
+    ///     - non retired stake
+    ///     - total stake balance
+    pub const MINIMUM_STAKE: u64 = 10_000_000;
```
sisou, line 114: Offtopic comment:
I always find this way of separating by thousands weird, because 10k luna doesn't tell me anything useful. I would prefer this whole file separates by 5 zeros at the end first, and then the NIM part by thousands:
```suggestion
    pub const MINIMUM_STAKE: u64 = 100_00000;
```

docs/becoming_validator.md
```diff
@@ -184,12 +184,12 @@ validator:inactivate <validator_address> <validator_address> <signing_private_ke
 ```
 
 
-### 5.2 Unstake delegated Stake
+### 5.2 Remove Stake
```
sisou, line 187: I don't quite understand this step in the documentation. Is it for validators, or for stakers? In any case, the `retire` step is missing.

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -9,6 +9,7 @@ use nimiq_primitives::coin::Coin;
 use nimiq_primitives::policy::Policy;
 use serde::{Deserialize, Serialize};
 
+use crate::RetireStakeReceipt;
```
sisou: Can this import be moved into the `use crate` group below?

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -21,11 +22,13 @@ use crate::{
...
 /// Struct representing a staker in the staking contract.
-/// The staker's balance is divided into active and inactive stake.
+/// The staker's balance is divided into active, inactive and retired stake.
 ///
 /// Actions concerning a staker are:
 /// 1. Create:           Creates a staker.
-/// 2. Stake:            Adds coins from any outside address to a staker's active balance.
+/// 2. AddStake:         Adds coins from any outside address to a staker's active balance.
+///                      This action is only possible if:
+///                        (a) if the resulting non retired funds respect the invariant 1 - minimum stake for non retired funds.
```
sisou: Double "if" (already present in the intro above).
```suggestion
///                        (a) the resulting non-retired funds respect the invariant 1 - minimum stake for non-retired funds.
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -35,17 +38,34 @@ use crate::{
 ///                      This action is only possible if:
 ///                        (a) the stake was previously not delegated (i.e., delegated to `None`)
 ///                        (b) the active balance is zero and the inactive balance has been released (*).
-/// 4. Unstake:          Removes coins from a staker's balance to outside the staking contract.
-///                      Only inactive stake that has been released can be withdrawn (*).
+/// 4. RetireStake:      Permanently marks the given balance for future withdrawal. Only inactive funds can be retired.
+///                      This action is only possible if:
+///                        (a) if the resulting non retired funds respect the invariant 1 - minimum stake for non retired funds.
```
sisou:
```suggestion
///                        (a) the resulting non-retired funds respect the invariant 1 - minimum stake for non-retired funds.
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -35,17 +38,34 @@ use crate::{
...
+///                        (b) the inactive funds are already released (*).
```
sisou:
```suggestion
///                        (b) the inactive funds are released (*).
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -35,17 +38,34 @@ use crate::{
...
+/// 5. RemoveStake:      Removes coins from a staker's balance to outside the staking contract.
+///                      The retired balance can always be withdrawn as long as the staker still respects minimum
+///                      total stake or there are no more funds left. On the latter case, this action will result in
+///                      the deletion of the staker.
+///                      This action is only possible if:
+///                         (a) if the resulting total stake respect the invariant 2 - minimum stake for total funds.
```
sisou:
```suggestion
/// 5. RemoveStake:      Removes retired balance from a staker to outside the staking contract.
///                      Retired balance can always be withdrawn as long as the staker respects the minimum
///                      total stake or all balance is removed. The latter case results in
///                      the deletion of the staker.
///                      This action is only possible if:
///                         (a) the resulting total stake respects the invariant 2 - minimum stake for total funds.
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -35,17 +38,34 @@ use crate::{
...
+///         potential jail period must have passed.
+/// (**)    The staker has a set of invariants:
+///             (invariant 1) Min Stake for non retired balances:
+///                           active + inactive balances should be == 0 or >= Minimum Stake
+///             (invariant 2) Min Stake for total staker balances:
+///                           active + inactive + retired balances should be == 0 or >= Minimum Stake
+///             (invariant 3) Inactivation block height is always associated with some inactive balance:
+///                          `inactive_from` is Some(_) && `inactive_balance` > 0
+///             (invariant 4) When there are no inactive funds, there is no inactivation block height:
+///                          `inactive_from` is None && `inactive_balance` == 0
```
sisou:
```suggestion
///             (invariant 1) Min stake for non-retired balances:
///                           active + inactive balance must be == 0 or >= minimum stake
///             (invariant 2) Min stake for total staker balance:
///                           active + inactive + retired balance must be == 0 or >= minimum stake
///             (invariant 3) Inactivation block height is always associated with some inactive balance:
///                          `inactive_from` is Some(_) && `inactive_balance` > 0
///             (invariant 4) When there are no inactive funds, there is no inactivation block height:
///                          `inactive_from` is None && `inactive_balance` == 0
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -55,12 +75,17 @@ pub struct Staker {
...
     /// the inactive and the validator's jailed periods must have passed.
     pub inactive_balance: Coin,
-    /// The block number at which the inactive balance was last inactivated for withdrawal or re-delegation.
+    /// The block number at which the inactive balance was last inactivated.
     /// If the stake is currently delegated to a jailed validator, the maximum of its jail release
     /// and the inactive release is taken. Re-delegation requires the whole balance of the staker to be inactive.
     /// The stake can only effectively become inactive on the next election block. Thus, this may contain a
     /// future block height.
     pub inactive_from: Option<u32>,
+    /// The staker's retired balance. Retired balance can only be withdrawn, thus this action is irreversible.
```
sisou:
```suggestion
    /// The staker's retired balance. Retired balance can only be withdrawn, thus retiring is irreversible.
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -55,12 +75,17 @@ pub struct Staker {
...
+    /// Only released inactive balance can be retired, so the maximum of the inactive and the validator's jailed
+    /// periods must have passed.
+    /// Once retired, the funds are immediately available to be withdrawn.
```
sisou:
```suggestion
    /// Once retired, the funds are immediately available to be withdrawn (removed).
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -167,17 +220,31 @@ impl StakingContract {
         // Get the staker.
         let mut staker = store.expect_staker(staker_address)?;
 
+        // Fail if the minimum stake would be violated for the non retired funds (invariant 1).
```
sisou:
```suggestion
        // Fail if the minimum stake would be violated for the non-retired funds (invariant 1).
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -247,10 +312,10 @@ impl StakingContract {
             store.expect_validator(new_validator_address)?;
         }
 
-        // If stake is currently delegated, it must be inactive.
-        if let Some(validator_address) = &staker.delegation {
+        // If stake is currently delegated, than the active balance should be 0.
```
sisou: Alternatively, `then`.
```suggestion
        // If stake is currently delegated, the active balance must be 0.
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -259,27 +324,14 @@ impl StakingContract {
...
-            if let Some(validator) = store.get_validator(validator_address) {
-                if validator.is_jailed(block_number) {
-                    debug!(
-                        ?staker_address,
-                        "Tried to update staker that currently delegates to a jailed validator"
-                    );
-                    return Err(AccountError::InvalidForRecipient);
-                }
-            }
+            // Fail if the the funds are in cooldown or jailed.
```
sisou: First time the word "cooldown" is used. I suggest using the same words as before to describe the situation. "not yet released" and "locked" come to mind.

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -259,27 +324,14 @@ impl StakingContract {
...
+            if !self.is_inactive_stake_released(store, &staker, block_number) {
+                debug!(
+                    ?staker_address,
+                    "Tried to update staker with inactive balance locked (cooldown or jail)"
```
sisou:
```suggestion
                    "Tried to update staker with inactive balance locked or jailed"
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -384,10 +435,10 @@ impl StakingContract {
         Ok(())
     }
 
-    /// Changes the active and consequentially the inactive balances of staker.
+    /// Changes the active and consequentially the inactive balances of the staker.
     /// The active balance will be set immediately.
-    /// The inactive balance will be available to withdraw after a lock up period and, if applicable,
-    /// until the jail period is finished.
+    /// The inactive balance will be available to retire after a lock up period and, if applicable,
```
sisou:
```suggestion
    /// The inactive balance will be available to retire after a lockup period and, if applicable,
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -384,10 +435,10 @@ impl StakingContract {
...
+    /// after the validator's jail period has finished.
     /// If the staker has already some inactive stake, the corresponding balance will be overwritten and
```
sisou:
```suggestion
    /// If the staker already has some inactive stake, the corresponding balance will be overwritten and
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -384,10 +435,10 @@ impl StakingContract {
...
     /// the lock up period will be reset. This corresponds to editing the active stake balance
```
sisou:
```suggestion
    /// the lockup period will be reset. This corresponds to editing the active stake balance
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -418,6 +469,16 @@ impl StakingContract {
         let old_active_balance = staker.balance;
         let new_inactive_balance = total_balance - new_active_balance;
 
+        // Update the staker's balance.
```
sisou:
```suggestion
        // Update the staker's balances.
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -500,96 +551,175 @@ impl StakingContract {
         Ok(())
     }
 
-    pub(crate) fn can_remove_stake<S: StakingContractStoreReadOps>(
-        &self,
-        store: &S,
-        staker: &Staker,
+    /// Adds to the retired balance and consequentially changes the inactive balance of the staker.
+    /// The balance can only be retired if the lock up period and associated validator's jail period have passed.
```
sisou:
```suggestion
    /// The balance can only be retired if the lockup period and associated validator's jail period have passed.
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -500,96 +551,175 @@ impl StakingContract {
...
+    /// The retire fails if the invariant 1 for the non retired stake is violated.
```
sisou:
```suggestion
    /// The retire fails if the invariant 1 for the non-retired stake is violated.
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -500,96 +551,175 @@ impl StakingContract {
...
-            }
+        // Fail if staker does not have sufficient funds to retire.
+        if staker.inactive_balance < retire_stake {
+            return Err(AccountError::InsufficientFunds {
+                needed: retire_stake,
+                balance: staker.inactive_balance,
+            });
         }
+
+        // Fail if the minimum stake would be violated for the non retired funds (invariant 1).
```
sisou:
```suggestion
        // Fail if the minimum stake would be violated for non-retired funds (invariant 1).
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -500,96 +551,175 @@ impl StakingContract {
...
+                active_balance=?staker.balance,
+                inactive_balance=?staker.inactive_balance,
+                retired_balance=?staker.retired_balance,
+                ?retire_stake,
+                "Tried to retire stake that would violate the minimum stake"
+            );
+            return Err(AccountError::InvalidCoinValue);
+        }
+
+        // Fail if the the funds are in cooldown or jailed.
```
sisou: "not yet released" or "locked"

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -500,96 +551,175 @@ impl StakingContract {
...
+        if !self.is_inactive_stake_released(store, &staker, block_number) {
+            debug!(
+                ?staker_address,
+                "Tried to retire stake with inactive balance locked (cooldown or jail)"
```
sisou:
```suggestion
                "Tried to retire stake with inactive balance locked or jailed"
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -500,96 +551,175 @@ impl StakingContract {
...
+            return Err(AccountError::InvalidForRecipient);
+        }
+
+        // All checks passed, not allowed to fail from here on!
+
+        // Store old values for receipt.
+        let old_inactive_from = staker.inactive_from;
+
+        // Update the staker's balances.
+        staker.inactive_balance -= retire_stake;
```
sisou: You created this variable above, might as well use it.
```suggestion
        staker.inactive_balance = new_inactive_balance;
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -500,96 +551,175 @@ impl StakingContract {
...
+
+        // Update the staker entry.
+        store.put_staker(staker_address, staker);
+
         Ok(())
     }
 
-    /// Removes coins from a staker's balance. If the entire staker's balance is removed then the
-    /// staker is deleted.
+    /// Removes coins from the retired balance of the staker.
```
sisou:
```suggestion
    /// Removes balance from the retired stake of the staker.
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -637,6 +770,106 @@ impl StakingContract {
...
+        Ok(())
+    }
+
+    /*  Helpers for checking release block heights and invariants */
+
+    /// Checks if the minimum stake invariants are respected.
+    /// This function expects as input the final balances of a given operation.
+    /// Invariants:
+    ///         (1) active + inactive balances should be == 0 or >= Minimum Stake
+    ///         (2) active + inactive + retired balances should be == 0 or >= Minimum Stake
```
sisou:
```suggestion
    ///         (1) active + inactive balances must be == 0 or >= minimum stake
    ///         (2) active + inactive + retired balances must be == 0 or >= minimum stake
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -637,6 +770,106 @@ impl StakingContract {
...
+    fn is_min_stake_enforced(
+        active_balance: Coin,
+        inactive_balance: Coin,
+        retired_balance: Coin,
+    ) -> bool {
+        // Check invariant 1: that total non retired balance (active+inactive) does not violate the minimum stake.
```
sisou:
```suggestion
        // Check invariant 1: total non-retired balances (active+inactive) do not violate the minimum stake.
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -637,6 +770,106 @@ impl StakingContract {
...
+            return false;
+        }
+
+        // Check invariant 2: total stake does not violate minimum stake.
+        // If all funds are retired, then retired funds must be 0 or at least minimum stake.
+        if non_retired_balances.is_zero()
+            && retired_balance > Coin::ZERO
+            && retired_balance < Coin::from_u64_unchecked(Policy::MINIMUM_STAKE)
+        {
+            log::error!("{:?}", retired_balance);
```
sisou: leftover logging from debugging?

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -637,6 +770,106 @@ impl StakingContract {
...
+            return false;
+        }
+
+        true
+    }
+
+    /// Returns true if the inactive funds are released both from the cooldown period and any potential jail
```
sisou:
```suggestion
    /// Returns true if the inactive funds are released both from the lockup period and any potential jail
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -637,6 +770,106 @@ impl StakingContract {
...
+        staker: &Staker,
+        block_number: u32,
+    ) -> bool {
+        // If stake is currently delegated we check if funds are released.
+        // We only need to check cooldown of inactive funds if there is a delegation.
+        // This is because the purpose of the cooldown is to ensure that we wait for the reporting window
+        // to finish and thus that the validator will not be jailed for misbehavior.
+        if let Some(validator_address) = &staker.delegation {
+            // Funds are not released if release block height has not passed yet.
+            if let Some(inactive_from) = staker.inactive_from {
```
sisou: Didn't you state earlier that an `inactive_from` number is only set when the staker has a delegation? So the check for delegation is unnecessary, as a set `inactive_from` field implies a delegation anyway?

primitives/account/src/account/staking_contract/traits.rs
```diff
@@ -604,13 +640,17 @@ impl AccountTransactionInteraction for StakingContract {
...
                 let staker_address = proof.compute_signer();
 
+                // Fetch the staker.
                 let staker = store.expect_staker(&staker_address)?;
 
-                self.can_remove_stake(&store, &staker, block_state.number)?;
+                // Verify that the stake can actually be removed.
+                self.can_remove_stake(&staker, transaction.total_value())?;
+                // Verify that the fee by itself can be removed without violating the minimum stake.
+                self.can_remove_stake(&staker, transaction.fee)?;
```
sisou: If the second check is always done, even if the first check passes, then that makes it impossible to remove stake which is exactly 100 NIM with a tx with a fee. Only a 0-fee tx would then be allowed, or you'd have to fund the staker with more NIM beforehand.

primitives/account/src/account/staking_contract/validator.rs
```diff
@@ -643,14 +643,35 @@ impl StakingContract {
...
+        if value != validator.deposit {
+            return Err(AccountError::InvalidCoinValue);
+        }
+
+        self.enforce_retire_and_release(validator, block_number)?;
+
+        Ok(())
+    }
+
+    /// Checks if a validator is retired and release in order to have a fee or a deposit deduction.
```
sisou:
```suggestion
    /// Checks if a validator is retired and released in order to have a fee or a deposit deduction.
```

primitives/account/src/account/staking_contract/validator.rs
```diff
@@ -643,14 +643,35 @@ impl StakingContract {
...
+    /// For this, the validator must be retired and have enough funds and be
+    /// released from jail.
```
sisou: This sentence is the same as the previous sentence in meaning.

primitives/account/src/account/staking_contract/validator.rs
```diff
@@ -643,14 +643,35 @@ impl StakingContract {
...
+    pub(crate) fn enforce_retire_and_release(
         &self,
         validator: &Validator,
         block_number: u32,
     ) -> Result<(), AccountError> {
         // Check that the validator is retired.
         if !validator.retired {
-            debug!(?validator.address, "Tried to delete active validator");
+            debug!(?validator.address, "Tried to delete a non retired validator");
```
sisou:
```suggestion
            debug!(?validator.address, "Tried to delete a non-retired validator");
```

primitives/src/policy.rs
```diff
@@ -105,6 +105,12 @@ impl Policy {
     /// in proof-of-work.
     pub const VALIDATOR_DEPOSIT: u64 = 10_000_000_000;
 
+    /// The stake necessary to create a staker in Lunas (1 NIM = 100,000 Lunas).
+    /// This minimum is applied to the amount of:
+    ///     - non retired stake
```
sisou:
```suggestion
    ///     - non-retired stake
```

primitives/transaction/src/account/staking_contract/structs.rs
```diff
@@ -161,17 +165,23 @@ impl IncomingStakingTransactionData {
...
                 }
 
                 // Check that the signature is correct.
                 verify_transaction_signature(transaction, proof, true)?
             }
             IncomingStakingTransactionData::AddStake { .. } => {
-                // No checks needed.
+                // Adding stake should be at least greater than 0.
+                if transaction.value.is_zero() {
+                    warn!("Add stake transactions must actually have higher than 0 value. The offending transaction is the following:\n{:?}", transaction);
```
sisou:
```suggestion
                    warn!("Add stake transactions must have higher than 0 value. The offending transaction is the following:\n{:?}", transaction);
```

rpc-client/src/subcommands/transactions_subcommands.rs
```diff
@@ -72,8 +72,9 @@ pub enum TransactionCommand {
         tx_commons: TxCommonWithValue,
     },
 
-    /// Sends a staking transaction from the address of a given `wallet` to a given `staker_address`.
-    Stake {
+    /// Sends a `add_stake` transaction from the address of a given `wallet` to a given `staker_address`.
+    /// This transaction result must have the sum of active and inactive stake is >= minimum stake, otherwise it fails.
```
sisou:
```suggestion
    /// This transaction must result in the sum of active and inactive stake to be >= minimum stake, otherwise it fails.
```

rpc-client/src/subcommands/transactions_subcommands.rs
```diff
@@ -123,23 +124,46 @@ pub enum TransactionCommand {
...
-    Unstake {
-        /// The stake will be sent from this wallet. The sender wallet must be unlocked prior to this action.
-        sender_wallet: Address,
+    /// Sends a `retire_stake` transaction to the network. You can pay the transaction fee from a basic
+    /// account (by providing the sender wallet) or from the staker account's balance (by not providing a sender wallet).
+    /// Only already inactivated and released funds can be retired.
+    /// This transaction can only result in all funds being retired or the sum of active and new inactive
+    /// balances being >= minimum stake. Otherwise it fails.
+    RetireStake {
+        /// The fee will be payed by this wallet if any is provided. In such case the sender wallet must be unlocked prior to this action.
```
sisou:
```suggestion
        /// The fee will be paid by this wallet if any is provided. In such case the sender wallet must be unlocked prior to this action.
```

rpc-client/src/subcommands/transactions_subcommands.rs
```diff
@@ -123,23 +124,46 @@ pub enum TransactionCommand {
...
+        /// If absent the fee is payed by the stakers account.
```
sisou:
```suggestion
        /// If absent, the fee is paid by the staker's account.
```

test-utils/src/transactions.rs
```diff
@@ -219,10 +222,25 @@ impl<R: Rng + CryptoRng> TransactionsGenerator<R> {
             if fail_sender {
                 value += Coin::from_u64_unchecked(1);
             }
+        } else if matches!(sender, OutgoingType::RemoveStake) {
+            value = Coin::from_u64_unchecked(2 * Policy::MINIMUM_STAKE)
+                .checked_sub(fee)
+                .expect("Fee should not be higher than staker stake");
```
sisou:
```suggestion
                .expect("Fee must not be higher than staker stake");
```

transaction-builder/src/lib.rs
```diff
@@ -883,7 +883,7 @@ impl TransactionBuilder {
     ///                             address will be derived from this key pair.
     ///  - `delegation`:            The (optional) delegation to a validator.
     ///  - `value`:                 The value for the initial stake. This is sent from the account
-    ///                             belonging to `key_pair`.
+    ///                             belonging to `key_pair`. Value must be >= to minimum stake.
```
sisou:
```suggestion
    ///                             belonging to `key_pair`. Value must be >= minimum stake.
```

transaction-builder/src/lib.rs
```diff
@@ -925,7 +925,7 @@ impl TransactionBuilder {
         }
     }
 
-    /// Creates a staking transaction from the address of a given `key_pair` to a specified
+    /// Creates a transaction to add more stake from the address of a given `key_pair` to a specified
```
sisou:
```suggestion
    /// Creates a transaction to add stake from the address of a given `key_pair` to a specified
```

transaction-builder/src/lib.rs
```diff
@@ -934,7 +934,8 @@ impl TransactionBuilder {
     ///                             stake is sent from the basic account belonging to this key pair.
     ///  - `staker_address`:        The address of the staker that we are sending the stake to.
     ///  - `value`:                 The value of the stake. This is sent from the account
-    ///                             belonging to `key_pair`.
+    ///                             belonging to `key_pair`. This transaction can fail if the value violates
+    ///                             minimum non retired stake
```
sisou:
```suggestion
    ///                             belonging to `key_pair`. This transaction can fail if the resulting staker violates
    ///                             the minimum non-retired stake.
```

transaction-builder/src/recipient/staking_contract.rs
```diff
@@ -160,6 +160,18 @@ impl StakingRecipientBuilder {
         self
     }
 
+    /// This method allows to retire the staker's balance. The retired funds are deducted from the inactive
+    /// stake. Retiring funds is only allowed after the inactive funds are released (i.e. cooldown and jail
+    /// periods have passed). Additionally, minimum total stake for non retired funds must still be respected.
+    /// It needs to be signed by the key pair corresponding to the staker address.
```
sisou:
```suggestion
    /// This method allows to retire stake from the staker's inactive balance.
    /// Retiring funds is only allowed after the inactive funds are released (i.e. lock-up and jail
    /// periods have passed). Additionally, the minimum total stake for non-retired funds must still be respected.
    /// It needs to be signed by the key pair corresponding to the staker address.
```

web-client/example/module.js
```diff
@@ -148,14 +148,14 @@ init().then(async () => {
      * @param {number} [fee]
      * @returns {Promise<string>}
      */
-    window.sendUnstakeTransaction = async (privateKey, amount, fee = 0) => {
+    window.sendRemoveStakeTransaction = async (privateKey, amount, fee = 0) => {
```
sisou, line 151: For this to make sense, we also need `window.sendSetActiveStakeTransaction` and  `window.sendRetireStakeTransaction` methods now.

web-client/src/primitives/transaction_builder.rs
```diff
@@ -259,6 +256,40 @@ impl TransactionBuilder {
         Ok(Transaction::from_native(tx))
     }
 
+    /// Remove stake from the staking contract and transfers `value` amount of luna (NIM's smallest unit)
```
sisou:
```suggestion
    /// Removes stake from the staking contract and transfers `value` amount of luna (NIM's smallest unit)
```

web-client/src/primitives/transaction_builder.rs
```diff
@@ -228,22 +225,22 @@ impl TransactionBuilder {
...
     /// signaling transaction and as such does not transfer any value.
     ///
     /// The returned transaction is not yet signed. You can sign it e.g. with `tx.sign(keyPair)`.
     ///
-    /// Throws when the number given for fee does not fit within a u64 or the networkId is unknown.
-    #[wasm_bindgen(js_name = newSetActiveStake)]
-    pub fn new_set_active_stake(
+    /// Throws when the numbers given for fee and `retire_stake` do not fit within a u64 or the networkId is unknown.
+    #[wasm_bindgen(js_name = newRetireStake)]
+    pub fn retire_stake(
```
sisou, line 235:
```suggestion
    pub fn new_retire_stake(
```

sisou (approved):
The logic seems to mostly be correct :+1:
I have however several wording and documentation clarity suggestions.

primitives/account/src/account/staking_contract/traits.rs
ii-cruz: Yes, correct. That was a trade-off we agreed upon to actually enforce the minimum stake despite the possibility of a commit failed tx and its consequential fee deduction.
Since it's an edge case that only happens if a staker is only holding the minimum stake (which is a really small amount) and fees are currently 0 on our network, we decided to enforce the minimum stake at the expense of this edge case.
Otherwise, we would have a minimum stake but it could be broken, similar to the validator deposit. Calling it a minimum stake would be rather pointless, even implementing a "minimum stake" on these conditions would be rather pointless.

sisou: I don't fully understand. Is this code here the ultimate test if the transaction will succeed? Or can it still fail later on?

If it can still fail later on, I agree that this test has to be as it is.

But if this is the ultimate test, and we have already established that the whole tx will go through with the first check (e.g. removing 99 NIM with 1 NIM fee), then why do we still need to test for a failing tx, which is the second check?

primitives/account/src/account/staking_contract/staker.rs
ii-cruz: We can have delegation set to None and have an inactive balance>0 and, thus inactive_from = Some(...).
The invariant is related to inactive_balance and inactive_from, not about the delegation.

ii-cruz: However, if we have inactive funds delegated to nobody, we don't need to bother waiting for lockup or jail periods. 
The jail is self-explanatory, there is no delegation so the delegation cannot be jailed.
For the lockup checking it is unnecessary because the purpose of the lockup is to prevent the staker from retiring stake and potentially removing the stake before the reporting window for the misbehavior of a validator has passed. However, there is no delegation, so nobody relevant can misbehave. Thus waiting for lockup is not needed.

docs/becoming_validator.md
ii-cruz: conclusion: this needs updating but later. Currently, the documentation seems to be confusing the validator and stake inactivation. However, we need to first update the arpl and then update this file.

web-client/example/module.js
ii-cruz: conclusion: this needs updating but later. this is an example file and is missing other transactions besides the ones already mentioned.

primitives/account/src/account/staking_contract/traits.rs
ii-cruz: We cannot be 100% sure that the tx will be successful at the reserve balance level. So no, this is not the ultimate test, it can still fail later on. 
It can fail on the receiver for example, by trying to transfer to another contract. That's just one of the cases, there are also some other scenarios where it can still fail later at the commit failed tx despite being accepted here.

primitives/src/policy.rs
hrxi: I guess it'd need a separate PR, but I generally agree.

primitives/account/tests/accounts.rs
```diff
@@ -988,6 +988,16 @@ fn can_revert_transactions() {
 
                     let receipts = accounts.test(&[tx], &[], &block_state);
                     if fail_sender || fail_recipient {
+                        for aa in receipts.transactions.iter() {
```
hrxi:
```suggestion
                        for receipt in receipts.transactions.iter() {
```

primitives/account/tests/accounts.rs
```diff
@@ -988,6 +988,16 @@ fn can_revert_transactions() {
...
+                            match aa {
+                                OperationReceipt::Ok(tx) => {
+                                    error!("Ok {:?}", tx);
+                                }
+                                OperationReceipt::Err(tx, reason) => {
+                                    error!("Ok {:?} - {:?}", tx, reason);
```
hrxi:
```suggestion
                                    error!("Err {:?} - {:?}", tx, reason);
```

primitives/account/tests/staking_contract/staker.rs
```diff
@@ -1602,6 +1606,44 @@ fn unstake_works() {
     );
 }
 
+#[test]
+fn unstake_does_not_allow_nim_stake_to_be_violated() {
```
hrxi: `nim_stake` ^^ (Gone in a later commit)

primitives/transaction/src/account/staking_contract/structs.rs
```diff
@@ -161,17 +165,23 @@ impl IncomingStakingTransactionData {
...
                 }
 
                 // Check that the signature is correct.
                 verify_transaction_signature(transaction, proof, true)?
             }
             IncomingStakingTransactionData::AddStake { .. } => {
-                // No checks needed.
+                // Adding stake should be at least greater than 0.
+                if transaction.value.is_zero() {
+                    warn!("Add stake transactions must have higher than 0 value. The offending transaction is the following:\n{:?}", transaction);
```
hrxi:
```suggestion
                    warn!("Add stake transactions must have positive value. The offending transaction is the following:\n{:?}", transaction);
```
or non-zero.

primitives/transaction/src/account/staking_contract/structs.rs
```diff
@@ -181,38 +191,38 @@ impl IncomingStakingTransactionData {
                 // Check that the signature is correct.
                 verify_transaction_signature(transaction, proof, true)?
             }
+            IncomingStakingTransactionData::RetireStake {
+                proof,
+                retire_stake,
+            } => {
+                // Check that retire is greater than 0.
+                if retire_stake.is_zero() {
+                    warn!("Retire stake transactions must have higher than 0 amount to retire. The offending transaction is the following:\n{:?}", transaction);
```
hrxi:
```suggestion
                    warn!("Retire stake transactions must retire a non-zero amount of stake. The offending transaction is the following:\n{:?}", transaction);
```

primitives/account/tests/staking_contract/staker.rs
```diff
@@ -346,388 +425,505 @@ fn stake_works() {
...
+    let (mut staker_setup, validator_address2, tx) =
+        prepare_second_validator_for_redelegation(ValidatorState::Active, 0, 150_000_000, 0);
+
     let data_store = staker_setup
         .accounts
         .data_store(&Policy::STAKING_CONTRACT_ADDRESS);
     let mut db_txn = staker_setup.env.write_transaction();
     let mut db_txn = (&mut db_txn).into();
-    let mut data_store_write = data_store.write(&mut db_txn);
+    _ = data_store.write(&mut db_txn);
```
hrxi: What does this do? (Also the further occurrences in this file.)

test-utils/src/transactions.rs
```diff
@@ -215,14 +218,29 @@ impl<R: Rng + CryptoRng> TransactionsGenerator<R> {
...
         // Ensure that the value is the validator deposit if we are creating a validator.
+        // Ensure that the value is the minimum stake is respected in the staker operations.
         if matches!(recipient, IncomingType::CreateValidator) {
             value = Coin::from_u64_unchecked(Policy::VALIDATOR_DEPOSIT);
+        } else if matches!(recipient, IncomingType::CreateStaker) {
+            value = Coin::from_u64_unchecked(Policy::MINIMUM_STAKE);
+        } else if matches!(recipient, IncomingType::SetActiveStake) {
+            value = Coin::ZERO;
+        } else if matches!(recipient, IncomingType::RetireStake) {
+            value = Coin::from_u64_unchecked(Policy::MINIMUM_STAKE);
```
hrxi: This looks like it could benefit from an actual `match` instead of many `if matches!`.

hrxi (approved):
Looks generally good to me. I added a couple of nits.

primitives/account/tests/accounts.rs
ii-cruz: Ups those were debug prints, took them out :pray:

primitives/account/tests/staking_contract/staker.rs
ii-cruz: This is a bit annoying. So without that write, the .into() from the line above infers the wrong type because the first operation performed in these tests using the `db_txn` is a `get_staker`. This is a read operation on the datastore and thus when we use it to write (on the commit incoming transaction) we have a type mismatch. 
This is totally a hardcoded way of dealing with the problem, let me know if there is a cleaner way.

hrxi: I guess a type annotation on the second `db_txn` would work?

```rs
use nimiq_trie::WriteTransactionProxy;

let mut db_txn = staker_setup.env.write_transaction();
let mut db_txn: WriteTransactionProxy = (&mut db_txn).into();
```

ii-cruz: This was so damn stupid xD I had tried it but the rust analyzer was complaining still, I probably had the wrong import. 🙈

ines-toupeira (approved):

primitives/account/src/account/staking_contract/staker.rs
sisou: Ok :+1:

primitives/account/src/account/staking_contract/traits.rs
sisou: Thanks for clarifying :bulb:

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -78,6 +101,30 @@ impl StakingContract {
...
+            delegation,
+            inactive_balance,
+            inactive_from,
+            Coin::ZERO,
+            tx_logger,
+        )
+    }
+
+    /// Creates a new staker with retired funds.
+    fn create_staker_with_retired(
```
styppo:
```suggestion
    fn create_staker_with_retired_balance(
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -637,6 +768,105 @@ impl StakingContract {
...
+
+        Ok(())
+    }
+
+    /// Checks if the minimum stake invariants are respected.
+    /// This function expects as input the final balances of a given operation.
+    /// Invariants:
+    ///         (1) active + inactive balances must be == 0 or >= minimum stake
+    ///         (2) active + inactive + retired balances must be == 0 or >= minimum stake
+    fn is_min_stake_enforced(
```
styppo:
```suggestion
    fn is_min_stake_satisfied(
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -637,6 +768,105 @@ impl StakingContract {
...
+        active_balance: Coin,
+        inactive_balance: Coin,
+        retired_balance: Coin,
+    ) -> bool {
+        // Check invariant 1: that total non-retired balances (active+inactive) do not violate the minimum stake.
+        let non_retired_balances = active_balance + inactive_balance;
```
styppo:
```suggestion
        let non_retired_balance = active_balance + inactive_balance;
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -500,96 +549,175 @@ impl StakingContract {
...
+                "Tried to retire stake that would violate the minimum stake"
+            );
+            return Err(AccountError::InvalidCoinValue);
+        }
+
+        // Fail if the the funds are locked or jailed.
+        if !self.is_inactive_stake_released(store, &staker, block_number) {
+            debug!(
+                ?staker_address,
+                "Tried to retire stake with inactive balance locked or jail"
```
styppo:
```suggestion
                "Tried to retire stake with inactive balance locked or jailed"
```

test-utils/src/transactions.rs
```diff
@@ -212,25 +215,51 @@ impl<R: Rng + CryptoRng> TransactionsGenerator<R> {
...
+                    value += Coin::from_u64_unchecked(1);
+                }
             }
+            _ => {}
         }
+
         // Ensure that the value is the validator deposit if we are creating a validator.
-        if matches!(recipient, IncomingType::CreateValidator) {
-            value = Coin::from_u64_unchecked(Policy::VALIDATOR_DEPOSIT);
+        // Ensure that the value is the minimum stake is respected in the staker operations.
```
styppo:
```suggestion
        // Ensure that the minimum stake is respected in staker operations.
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -35,17 +37,31 @@ use crate::{
...
 ///                        (b) the active balance is zero and the inactive balance has been released (*).
-/// 4. Unstake:          Removes coins from a staker's balance to outside the staking contract.
-///                      Only inactive stake that has been released can be withdrawn (*).
+/// 4. RetireStake:      Permanently marks the given balance for future withdrawal. Only inactive funds can be retired.
+///                      This action is only possible if:
+///                        (a) the resulting non-retired funds respect the invariant 1 - minimum stake for non-retired funds.
+///                        (b) the inactive funds are released (*).
+/// 5. RemoveStake:      Removes the retired balance from a staker to outside the staking contract.
+///                      The retired balance can always be withdrawn as long as it removes all retired stake.
+///                      In case of total sake = 0 results in the deletion of the staker.
```
styppo:
```suggestion
///                      If the staker's total balance drops to 0, the staker is deleted.
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -35,17 +37,31 @@ use crate::{
...
+///             (invariant 2) Min stake for total staker balances:
+///                           active + inactive + retired balance must be == 0 or >= minimum stake
+///             (invariant 3) Inactivation block height is always associated with some inactive balance:
+///                          `inactive_from` is Some(_) && `inactive_balance` > 0
+///             (invariant 4) When there are no inactive funds, there is no inactivation block height:
+///                          `inactive_from` is None && `inactive_balance` == 0
 ///
-/// Create, Stake, SetActiveStake, and Update are incoming transactions to the staking contract.
-/// Unstake is an outgoing transaction from the staking contract.
+/// Create, Stake, SetActiveStake, Update and RetireStake are incoming transactions to the staking contract.
```
styppo:
```suggestion
/// Create, AddStake, SetActiveStake, Update and RetireStake are incoming transactions to the staking contract.
```

primitives/account/src/account/staking_contract/traits.rs
```diff
@@ -459,35 +479,62 @@ impl AccountTransactionInteraction for StakingContract {
...
+                        )?
+                        .into(),
                     )
-                    .map(|receipt| Some(receipt.into()))?;
+                } else {
+                    // Update the balances of the validator and staking contract.
+                    staker.retired_balance = new_retired_balance;
+                    self.balance -= transaction.fee;
+
+                    // Update the validator entry.
```
styppo:
```suggestion
                    // Update the staker entry.
```

primitives/account/src/account/staking_contract/traits.rs
```diff
@@ -459,35 +479,62 @@ impl AccountTransactionInteraction for StakingContract {
...
+                            &mut store,
+                            &staker_address,
+                            transaction.fee,
+                            &mut TransactionLog::empty(),
+                        )?
+                        .into(),
                     )
-                    .map(|receipt| Some(receipt.into()))?;
+                } else {
+                    // Update the balances of the validator and staking contract.
```
styppo:
```suggestion
                    // Update the balances of the staker and staking contract.
```

primitives/account/src/account/staking_contract/traits.rs
```diff
@@ -459,35 +479,62 @@ impl AccountTransactionInteraction for StakingContract {
...
-                        block_state.number,
-                        &mut TransactionLog::empty(),
+                let new_retired_balance = staker.retired_balance.safe_sub(transaction.fee)?;
+                self.enforce_min_stake(
+                    staker.balance,
+                    staker.inactive_balance,
+                    new_retired_balance,
+                )?;
+
+                let receipt = if new_retired_balance.is_zero() {
```
styppo: This must check if the **total** balance is zero.

What was the reason to change this from the previous implementation that simply called through to `remove_stake` in all cases and did not need this `if` at all? Generating the `DeleteStaker` event by checking if the staker has been removed from the store seems a bit simpler and with less duplication.

primitives/account/src/account/staking_contract/traits.rs
```diff
@@ -459,35 +479,62 @@ impl AccountTransactionInteraction for StakingContract {
...
+                    staker.retired_balance = new_retired_balance;
+                    self.balance -= transaction.fee;
+
+                    // Update the validator entry.
+                    store.put_staker(&staker_address, staker);
+
+                    None
+                };
 
                 tx_logger.push_log(Log::StakerFeeDeduction {
```
styppo, line 511: I think it would be more consistent to log `StakerFeeDeduction` *before* `DeleteStaker`. The fee is deducted before the staker is deleted.

primitives/account/src/account/staking_contract/traits.rs
```diff
@@ -553,21 +596,47 @@ impl AccountTransactionInteraction for StakingContract {
...
+                        return Err(AccountError::InvalidReceipt);
+                    }
+                };
+
+                // Update the balances of the validator and staking contract.
+                staker.retired_balance += transaction.fee;
+                self.balance += transaction.fee;
+
+                // Update the validator entry.
+                store.put_staker(&staker_address, staker);
```
styppo: See comment above regarding simply calling through to `revert_remove_stake` here as it was before.

primitives/account/src/account/staking_contract/traits.rs
```diff
@@ -553,21 +596,47 @@ impl AccountTransactionInteraction for StakingContract {
...
+
+                if store.get_staker(&staker_address).is_none() {}
```
styppo: Unnecessary line

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -500,96 +538,162 @@ impl StakingContract {
...
 
-        // Update the staker's balance.
-        staker.inactive_balance = new_balance;
-        if staker.inactive_balance.is_zero() {
-            staker.inactive_from = None;
-        }
-
-        // Update our balance.
+        // Update balances.
+        staker.retired_balance = Coin::ZERO;
```
styppo: It would be more consistent to subtract instead of setting to zero.

PR #2132, opened 2024-01-18
network: Add verification and selection of DHT records

### LINKED ISSUE #1531: Review current DHT implementation
As noted in #1530, the DHT is using a quorum of 1 to inform the application of a result.
This issue is to revisit this implementation and potentially increment the value of the quorum to report the application when a result for a query is obtained. Note that to increment this quorum value, the network must introduce a method for verifying and selecting the best result as also noted in the [`kad` specification](https://github.com/libp2p/specs/blob/master/kad-dht/README.md#value-retrieval).

jsdanielh:
- network: Implement selection of DHT (kad) records
  - Change the `network-interface` to receive parameters that implements the `PartialOrd` trait such that the network can compare them and be able to select the best one.
  - Change the `network-libp2p` implementation to expect a quorum of 3 answers before accepting a DHT record and add logic to replicate the best answer to outdated peers.
    This closes #1531.
- network: Change DHT interface to validate records
  - Change the `dht_get` and `dht_put` methods to add additional parameters and additional traits for existing parameters to be able to implement record validation.
    This partially addresses #1531.
- utils: Rename `data` field of `TaggedSignature` to `signature`
  - Rename the `data` field of `TaggedSignature` to `signature` in the `utils` subcrate.
- Implement tagged signing traits for Schnorr and BLS keys
  - Implement the tagged signing traits `TaggedPublicKey` and `TaggedKeyPair` for `BLS` and `Ed25519`  keypairs/keys.
- keys: Remove warnings
  - Remove compilation warnings when the `serde-derive` feature is not enabled.

## Pull request checklist


bls/src/tagged_signing.rs
```diff
@@ -0,0 +1,27 @@
...
+use nimiq_serde::Deserialize;
+use nimiq_utils::tagged_signing::{TaggedKeyPair, TaggedPublicKey};
+
+use crate::{CompressedSignature, KeyPair, PublicKey};
+
+impl TaggedKeyPair for KeyPair {
+    type PublicKey = PublicKey;
+
+    fn sign(&self, message: &[u8]) -> Vec<u8> {
+        self.sign(&message.to_vec()).compress().as_ref().to_vec()
```
hrxi: The first `.to_vec()` seems like an unnecessary allocation. Can it be removed?

bls/src/tagged_signing.rs
```diff
@@ -0,0 +1,27 @@
...
+impl TaggedPublicKey for PublicKey {
+    fn verify(&self, msg: &[u8], sig: &[u8]) -> bool {
+        let signature = match CompressedSignature::deserialize_from_vec(sig) {
+            Ok(compressed_signature) => match compressed_signature.uncompress() {
+                Ok(signature) => signature,
+                Err(_) => return false,
+            },
+            Err(_) => return false,
+        };
+        self.verify(&msg.to_vec(), &signature)
```
hrxi: Same here, `.to_vec()` sounds like an unnecessary allocation.

network-interface/src/network.rs
```diff
@@ -157,16 +158,18 @@ pub trait Network: Send + Sync + Unpin + 'static {
...
+    async fn dht_get<K, V, T>(&self, k: &K) -> Result<Option<V>, Self::Error>
     where
         K: AsRef<[u8]> + Send + Sync,
-        V: Deserialize + Send + Sync;
+        V: Deserialize + Send + Sync + TaggedSignable + PartialOrd,
+        T: TaggedKeyPair + Send + Sync + Serialize;
 
     /// Puts a value to the distributed hash table
-    async fn dht_put<K, V>(&self, k: &K, v: &V) -> Result<(), Self::Error>
+    async fn dht_put<K, V, T>(&self, k: &K, v: &V, kp: &T) -> Result<(), Self::Error>
```
hrxi:
```suggestion
    async fn dht_put<K, V, T>(&self, k: &K, v: &V, keypair: &T) -> Result<(), Self::Error>
```

network-interface/src/network.rs
```diff
@@ -157,16 +158,18 @@ pub trait Network: Send + Sync + Unpin + 'static {
         T: Topic + Sync;
 
     /// Gets a value from the distributed hash table
-    async fn dht_get<K, V>(&self, k: &K) -> Result<Option<V>, Self::Error>
+    async fn dht_get<K, V, T>(&self, k: &K) -> Result<Option<V>, Self::Error>
     where
         K: AsRef<[u8]> + Send + Sync,
-        V: Deserialize + Send + Sync;
+        V: Deserialize + Send + Sync + TaggedSignable + PartialOrd,
+        T: TaggedKeyPair + Send + Sync + Serialize;
```
hrxi: This type parameter is a bit annoying because you shouldn't really change it from call to call, otherwise you'd get weird effects. I don't have a proposal how we could fix it. Do you know any alternatives?

validator-network/src/network_impl.rs
```diff
@@ -305,9 +302,12 @@ where
         secret_key: &SecretKey,
     ) -> Result<(), Self::Error> {
         let peer_id = self.network.get_local_peer_id();
-        let record = ValidatorRecord::new(peer_id);
+        let record = ValidatorRecord::new(
+            peer_id,
+            OffsetDateTime::now_utc().unix_timestamp() as u64 * 1000,
```
hrxi:
```suggestion
            OffsetDateTime::now_utc().unix_timestamp_nanos() / 1_000_000 as u64,
```
Is there a reason for choosing milliseconds in particular?

utils/src/tagged_signing.rs
```diff
@@ -122,16 +129,123 @@ pub trait TaggedKeyPair: Sized {
...
+        formatter.write_str("struct TaggedSigned")
+    }
+
+    fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+    where
+        A: SeqAccess<'de>,
+    {
+        let _tag: u8 = seq
+            .next_element()?
+            .ok_or_else(|| A::Error::invalid_length(0, &self))?;
```
hrxi: Tag should be checked.

utils/src/tagged_signing.rs
```diff
@@ -122,16 +129,123 @@ pub trait TaggedKeyPair: Sized {
...
 pub trait TaggedPublicKey {
     fn verify(&self, msg: &[u8], sig: &[u8]) -> bool;
 }
 
+#[derive(Clone)]
+pub struct TaggedSigned<TSignable, TScheme>
+where
+    TSignable: TaggedSignable,
+    TScheme: TaggedKeyPair,
+{
```
hrxi, line 140: `Deserialize`, `Serialize` could probably be derived we add a `Tag<TSignable>` field that always serializes to a `u8`, would probably be easier and more future-proof.

network-interface/src/network.rs
```diff
@@ -157,16 +158,18 @@ pub trait Network: Send + Sync + Unpin + 'static {
         T: Topic + Sync;
 
     /// Gets a value from the distributed hash table
-    async fn dht_get<K, V>(&self, k: &K) -> Result<Option<V>, Self::Error>
+    async fn dht_get<K, V, T>(&self, k: &K) -> Result<Option<V>, Self::Error>
     where
         K: AsRef<[u8]> + Send + Sync,
-        V: Deserialize + Send + Sync;
+        V: Deserialize + Send + Sync + TaggedSignable + PartialOrd,
```
hrxi:
```suggestion
        V: Deserialize + Send + Sync + TaggedSignable + Ord,
```
`Ord` would be more appropriate because it implies stuff like `a == a` and `a <= b && b <= c` implying `a <= c`. See https://doc.rust-lang.org/1.75.0/std/cmp/trait.Ord.html#corollaries.

validator-network/src/validator_record.rs
```diff
@@ -1,59 +1,40 @@
...
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
 #[serde(bound = "TPeerId: Serialize + Deserialize")]
 pub struct ValidatorRecord<TPeerId>
 where
     TPeerId: Serialize + Deserialize,
 {
+    /// Validator Peer ID
     pub peer_id: TPeerId,
-    // TODO: other info, like public key?
+    /// Record timestamp
```
hrxi:
```suggestion
    /// Record timestamp in milliseconds since 1970-01-01 00:00:00 UTC, excluding leap seconds (Unix time)
```

utils/src/tagged_signing.rs
jsdanielh: I added a `u8` field initialized with `TSignable::TAG` in the `new` function. Let me know if you were thinking on a better solution

utils/src/tagged_signing.rs
jsdanielh: Moved it to derive `Serialize` and `Deserialize`

network-interface/src/network.rs
jsdanielh: In principle I wanted to keep the `DHT` interface generic in terms of the data that we're putting/getting including what kind of keys you use for signing messages. So yeah, not sure how to achieve both things.

validator-network/src/network_impl.rs
jsdanielh: Left it as it is since `nanos` returns a `i128` and I believe it's more natural to convert from a `i64`. No specific reason to use millis, I used the same timestamp resolution that we use in blocks.

hrxi: Converting from seconds means the lowest three digits will always be 000. It seems it's more correct to get it from nanos so that the lowest three digits actually carry information.

jsdanielh: Ah yes, makes sense. Changed it to use `nanos`

hrxi:
Rebasing for merge.

hrxi (approved):

PR #2151, opened 2024-01-26
Document `pubsub_id` param of `request_missing_blocks`

hrxi:


consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -272,6 +272,10 @@ impl<N: Network> BlockQueue<N> {
     /// Requests missing blocks.
     /// If a block locator is given, it is inserted at the beginning of the full block locator list.
     /// The list contains all blocks from the head until the last macro block.
+    ///
+    /// `pubsub_id` specifies the gossipsub ID that of the gossip that led to
```
jsdanielh:
```suggestion
    /// `pubsub_id` specifies the gossipsub ID that of the peer that led to
```

hrxi: Maybe the "gossipsub peer ID of the gossip message that led to…"? It's the message leading to the request, not the peer.

jsdanielh: Ah, that makes sense

hrxi: Fixed.

jsdanielh (approved):

sisou:
Alternatively, you could have renamed the argument to `pubsub_peer_id`, putting the "peer" meaning of it in the name and removing the need to document it so detailed.

hrxi:
That would still not be clear enough to me. "Why do I need a pubsub peer ID for requesting missing blocks? What happens if I set it to `None`?"

PR #2153, opened 2024-01-26
Add `BlockchainEvent::Stored(Block)` for forked and inferior chain blocks

nibhar:
Also streamline the event firings in general. Any `Finalized` and `EpochFinalized` event does *not* supersede an `Extended` event. After this commit these will be emitted consistently.

Note that the validator still must `on_blockchain_extended` for `EpochFinalized` as well as `Finalized` as the ordering of emitted events would otherwise matter since the validator set needs adjustment before the block producer is initialized.

### DIFF
--- a/blockchain-interface/src/error.rs
+++ b/blockchain-interface/src/error.rs
@@ -15,6 +15,9 @@ pub enum BlockchainEvent {
     Extended(Blake2bHash),
     HistoryAdopted(Blake2bHash),
     Rebranched(Vec<(Blake2bHash, Block)>, Vec<(Blake2bHash, Block)>),
+    /// Given Block was stored in the chain store but was not adopted as new head block.
+    /// I.e. forked blocks and inferior chain blocks.
+    Stored(Block),
     Finalized(Blake2bHash),
     EpochFinalized(Blake2bHash),
 }
--- a/blockchain/src/blockchain/push.rs
+++ b/blockchain/src/blockchain/push.rs
@@ -161,6 +161,12 @@ impl Blockchain {
         }
         txn.commit();
 
+        // Fork and inferior chain block fire a Stored Event.
+        // They can never fire a Finalized or EpochFinalized as then they would not be inferior/forked.
+        this.notifier
+            .send(BlockchainEvent::Stored(chain_info.head))
+            .ok();
+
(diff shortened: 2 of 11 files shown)


blockchain-interface/src/error.rs
```diff
@@ -15,6 +15,9 @@ pub enum BlockchainEvent {
     Extended(Blake2bHash),
     HistoryAdopted(Blake2bHash),
     Rebranched(Vec<(Blake2bHash, Block)>, Vec<(Blake2bHash, Block)>),
+    /// Given Block was stored in the chain store but was not adopted as new head block.
+    /// I.e Forked blocks and inferior chain blocks.
```
hrxi:
```suggestion
    /// I.e. forked blocks and inferior chain blocks.
```

blockchain/src/blockchain/push.rs
```diff
@@ -161,6 +161,12 @@ impl Blockchain {
         }
         txn.commit();
 
+        // Fork and inferior chain block fire a Stored Event.
+        // They can never fire a Finalized or EpochFinalized as then the would not be inferior/forked.
```
hrxi:
```suggestion
        // They can never fire a Finalized or EpochFinalized as then they would not be inferior/forked.
```

consensus/src/consensus/remote_event_dispatcher.rs
```diff
@@ -253,12 +253,17 @@ impl<N: Network> Future for RemoteEventDispatcher<N> {
...
-                    // We dont't notify about reverted block, only adopted blocks
+                    // We don't notify about reverted block, only adopted blocks
                     new_blocks.extend(adopted_blocks.into_iter().map(|(_, block)| block));
                 }
                 BlockchainEvent::HistoryAdopted(_) => {
                     // In the future we might be interested in other events
                 }
+                BlockchainEvent::Stored(_block) => {
+                    // Stored events are not reported as they are not on the main chain.
+                    // If they ever become main chain blocks they will be reported then with the respective
```
hrxi:
```suggestion
                    // If they ever become main chain blocks, they will be reported then with the respective
```
helps me understand the sentence.

hrxi (commented):
Just some typos. PR itself looks good.

PR #2154, opened 2024-01-26
Add minor fixes in several subcrates

hrxi:


validator-network/src/network_impl.rs
```diff
@@ -90,14 +90,14 @@ where
     /// Look up the peer ID for a validator ID.
```
jsdanielh, line 89: Could you address the clippy warning above?

hrxi: Done.

jsdanielh (approved):

PR #2163, opened 2024-01-29
Run expensive tests regularly, but not on every pull request

hrxi:
This should cut down our PR CI times by 10 minutes or so.

### DIFF
--- a/.github/workflows/build+test.yml
+++ b/.github/workflows/build+test.yml
@@ -29,17 +29,8 @@ jobs:
     - name: Free Disk Space (Ubuntu)
       uses: jlumbroso/free-disk-space@main
       with:
-        # this might remove tools that are actually needed,
-        # if set to "true" but frees about 6 GB
         tool-cache: true
-        
-        # all of these default to true, but feel free to set to
-        # "false" if necessary for your workflow
-        android: true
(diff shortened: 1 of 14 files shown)


.github/workflows/build+test.yml
```diff
@@ -29,17 +29,8 @@ jobs:
...
       with:
-        # this might remove tools that are actually needed,
-        # if set to "true" but frees about 6 GB
         tool-cache: true
-        
-        # all of these default to true, but feel free to set to
-        # "false" if necessary for your workflow
-        android: true
-        dotnet: true
-        haskell: true
```
jsdanielh, line 40: Why did we remove them?

.github/workflows/code_coverage.yml
```diff
@@ -47,11 +47,7 @@ jobs:
       uses: jlumbroso/free-disk-space@main
       with:
           tool-cache: true
-          android: true
-          dotnet: true
-          haskell: true
```
jsdanielh, line 52: Why did we remove them?

.github/workflows/build+test.yml
hrxi: These are defaults.

PR #2172, opened 2024-01-31
validator-network: Separate state lock in different locks

jsdanielh:
Change the `validator-network` implementation to separate the lock around the state into different locks for the different subcomponents of the network state such when modifying a part of the state doesn't require to block reads to the other subcomponents of the state. Since they are independent from each other, they should be allowed to be modified independently as well.

## Pull request checklist

### DIFF
--- a/validator-network/src/network_impl.rs
+++ b/validator-network/src/network_impl.rs
@@ -15,17 +15,6 @@ use time::OffsetDateTime;
 use super::{MessageStream, NetworkError, PubsubId, ValidatorNetwork};
 use crate::validator_record::ValidatorRecord;
 
-/// Validator Network state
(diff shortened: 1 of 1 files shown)


hrxi (approved):
Looks good to me.

nibhar (approved):

sisou:
Could this improve the behavior of validator full nodes getting hung up on macro blocks? Or is this something unrelated to any reported node hang ups?

jsdanielh:
> Could this improve the behavior of validator full nodes getting hung up on macro blocks? Or is this something unrelated to any reported node hang ups?

Unrelated as of now. The full nodes hanging up in election blocks is still under investigation

PR #2179, opened 2024-02-01
Increase max chunk requests

viquezclaudio:
Increase the maximum number of trie chunk requests. This is needed for cases where we have a high number of txns per epoch

## Pull request checklist

### DIFF
--- a/consensus/src/sync/live/state_queue/mod.rs
+++ b/consensus/src/sync/live/state_queue/mod.rs
@@ -34,7 +34,7 @@ use super::{
 use crate::sync::live::diff_queue::{DiffQueue, QueuedDiff};
 
 /// The max number of chunk requests per peer.
-pub const MAX_REQUEST_RESPONSE_CHUNKS: u32 = 100;
+pub const MAX_REQUEST_RESPONSE_CHUNKS: u32 = 5000;
 
 /// The request of a trie chunk.
 /// The request specifies a limit on the number of nodes in the chunk.


jsdanielh (approved):
LGTM

sisou:
As you say this is needed for more txns per epoch, is this number now able to cover the theoretical maximum of txns in an epoch, i.e. if every micro block were full?

viquezclaudio:
This was tested with up to 500 txns per block, which is not full capacity (~700 txn/block)
However, this particular variable is only used for state sync (full nodes), when we request trie chunks to build our accounts trie. 
If the epoch contains a big number of txns, then we have to do lots of these requests (which might be rate limited by other peers if this variable is too low)
This does not have anything to do with increasing the theoretical capacity or limit of our blockchain performance.

sisou:
I don't understand how you say this needs to be higher for a high number of transactions, but also say it's not related to the theoretical capacity or limits.

So either the theoretical limits are irrelevant to practical limits (which shouldn't be the case), or the relation between chunk requests and transactions is not linear.

Also, how are transaction counts related to the accounts trie? That should be two separate kind of tries/stores?

Is this variable a rate limit _per minute_, maybe? Then it would make sense to have a maximum unrelated to the theoretical tx/s limit, as nodes can only process chunks so fast. But if it's a rate limit per epoch or per batch, then it must be as high as the maximum number of chunks that can happen in that epoch or batch, no?

If I don't understand correctly, then maybe the naming of the variable or the comment is misleading?

sisou:
I see that the `TIME_WINDOW` is set to this default, which is 10 seconds:
https://github.com/nimiq/core-rs-albatross/blob/761e6e9a60fb82993b2fa17087a22b9232553cff/network-interface/src/request/mod.rs#L10

So it's a rate limit per 10s? Then it makes sense to not be related to the tx/s limit. And I guess `5000` is high enough? Is 5000 based on any real-life observation, or just a guestimate? Because a rate limit should be just as high as necessary, not much higher. And 5000 per 10s seems like a lot. 500 per second? Sounds like a DoS attack.

viquezclaudio:
Yes, this is a variable related to peer rate limiting, and as you noted, we currently have a 10s rate limiting window. 
This number is based on a real testing scenario of 7 validators and a spammer node, where one of them is trying to sync.
With the previous limit (100) then by spamming more than 100tx/block we can easily get rate limited (and I observed this happening a lot)
With the limit that I'm proposing in this PR I was able to test up to 500tx/block without getting rate limited, but I do get rate limited a little bit by spamming 700tx/block.
However in a healthy network, a given node should not spam a single node with requests, and it should distribute the requests among multiple nodes, thus, this number might be a bit high, however, during a core meeting it was decided that we wanted to start by setting this kind of limits high (as they can easily be changed by any user later) and optimize accordingly. 

When I was referring to our theoretical tx limit, I was referring to the transaction capacity per block (currently ~700txs per block) and our block production rate (a target of 1 block per second), these two variables would affect the overall blockchain performance that we can get, and this PR is not changing or affecting any of these two variables.

sisou:
Thanks for the clarification! 👍

PR #2183, opened 2024-02-02
Add `NetworkId`s to block headers and Tendermint votes

### LINKED ISSUE #2175: Equivocation proofs are valid across blockchains, e.g. testnet and mainnet
This should be fixed by including a network ID in the relevant places.

hrxi:
This allows to distinguish between different chains for the purposes of equivocation proofs.

Fixes #2175.


sisou:
Please also add the new field to the RPC block type:
https://github.com/nimiq/core-rs-albatross/blob/8abdaf2d5c13cbeb4cb04056c48f5e2901ebce1f/rpc-interface/src/types.rs#L161
https://github.com/nimiq/core-rs-albatross/blob/8abdaf2d5c13cbeb4cb04056c48f5e2901ebce1f/rpc-interface/src/types.rs#L214

 and the web-client plain block:
https://github.com/nimiq/core-rs-albatross/blob/8abdaf2d5c13cbeb4cb04056c48f5e2901ebce1f/web-client/src/client/block.rs#L11

hrxi:
Thanks for reminding me of the rpc-interface and web-client types, I completely forgot about those.

I'm wondering whether it makes sense to include the network here. The rpc or web client already know which network they're connected to, and the block is going to tell them the same. Is there a use case for this field on the rpc or web client?

sisou:
Is there a use case for the `version` field? IMO it should be as complete as possible, we can't imagine all use cases right now. Maybe someone wants to experiment with a migration between network IDs, and needs it 🤷

hrxi:
> Please also add the new field to the RPC block type:

Done. Still need to fix a couple of tests.

web-client/src/client/block.rs
```diff
@@ -22,6 +22,8 @@ pub struct PlainBlockCommonFields {
     /// The timestamp of the block. It follows the Unix time and has millisecond precision.
     pub timestamp: u64,
 
+    /// The network that this block is valid for.
+    pub network: &'static str,
```
sisou, line 26: @hrxi You are using `&'static str` instead of the `String` type I have been using everywhere else. Is there an advantage to using either?

genesis-builder/src/lib.rs
```diff
@@ -126,6 +129,11 @@ impl GenesisBuilder {
         self
     }
 
+    pub fn with_network(&mut self, network: NetworkId) -> &mut Self {
```
jsdanielh, line 132: Could you add some `rustdoc`?

primitives/block/tests/verify.rs
```diff
@@ -463,5 +553,7 @@ fn test_verify_election_macro_body() {
     });
 
     // Skipping the verification of the PK tree root should make the verify function to pass
-    assert_eq!(block.verify(), Ok(()));
+    assert_eq!(block.verify(NetworkId::UnitAlbatross), Ok(()));
 }
+
+// TODO: network ID test
```
jsdanielh, line 559: Is this still a `TODO`?

primitives/src/tendermint.rs
```diff
@@ -36,7 +41,7 @@ pub struct TendermintIdentifier {
 
 impl TendermintIdentifier {
     /// Maximum size in bytes for a `TendermintIdentifier` in binary serialization.
-    pub const MAX_SIZE: usize = 2 * nimiq_serde::U32_MAX_SIZE + TendermintStep::SIZE;
+    pub const MAX_SIZE: usize = nimiq_serde::U8_SIZE + 2 * 4 + TendermintStep::SIZE;
```
jsdanielh, line 44: Why not `U32_SIZE`?

jsdanielh:
Rebasing to merge it

web-client/src/client/block.rs
hrxi: `String` is the more general-purpose type. `&'static str` only works for strings already known at compile time. Since we know all the possible networks, `&'static str` makes it a bit more efficient since it doesn't need an allocation.

primitives/src/tendermint.rs
hrxi: Because we explicitly serialize it as four bytes and not as the postcard serialization of a `u32`.

primitives/block/tests/verify.rs
hrxi: Nope, will do a follow-up PR.

genesis-builder/src/lib.rs
hrxi: Yes, will do a follow-up PR.

PR #2189, opened 2024-02-05
History store also stores transaction results

### LINKED ISSUE #1979: Basic `HistoricTransaction`s do not verify the execution result
Since the execution result is not included in the hash, a history node could lie about the execution result of a basic transaction.

https://github.com/nimiq/core-rs-albatross/blob/70b70dca9e45b9049d9e6e11236004adb5f6ea1d/primitives/transaction/src/lib.rs#L185-L190

ii-cruz:
Introduces the executed transaction hash that includes both the transaction result (Ok or Err) and the transaction itself.
Changes history store to indexes to go from raw transaction to executed transaction hash and address to raw transaction hash.

## Pull request checklist

### DIFF
--- a/blockchain/src/blockchain/verify.rs
+++ b/blockchain/src/blockchain/verify.rs
@@ -5,6 +5,7 @@ use nimiq_database::{
     traits::{ReadTransaction, WriteTransaction},
     TransactionProxy as DBTransaction, WriteTransactionProxy,
 };
+use nimiq_hash::Hash;
 use nimiq_primitives::policy::Policy;
 
 use crate::{blockchain_state::BlockchainState, BlockProducer, Blockchain};
@@ -102,8 +103,9 @@ impl Blockchain {
     fn verify_transactions(&self, block: &Block) -> Result<(), BlockError> {
         if let Some(transactions) = block.transactions() {
             for transaction in transactions {
+                let transaction = transaction.get_raw_transaction();
                 if !self.tx_verification_cache.is_known(&transaction.hash()) {
-                    transaction.get_raw_transaction().verify(self.network_id)?;
+                    transaction.verify(self.network_id)?;
                 }
             }
         }
--- a/blockchain/src/history/history_store.rs
+++ b/blockchain/src/history/history_store.rs
@@ -44,12 +44,13 @@ pub struct HistoryStore {
     /// A database of all historic transactions indexed by their hash (= leaf hash in the history
     /// tree).
     hist_tx_table: TableProxy,
-    /// A database of all leaf hashes and indexes indexed by the hash of the transaction. This way we
-    /// can start with a transaction hash and find it in the MMR.
+    /// A database of all leaf hashes and indexes indexed by the hash of the (raw) transaction. This way we
+    /// can start with a raw transaction hash and find it in the MMR.
+    /// Mapping of raw tx to leaf hash  (= executed tx hash + index).
     tx_hash_table: TableProxy,
     /// A database of the last leaf index for each block number.
     last_leaf_table: TableProxy,
-    /// A database of all transaction (and reward inherent) hashes indexed by their sender and
+    /// A database of all raw transaction (and reward inherent) hashes indexed by their sender and
     /// recipient addresses.
     address_table: TableProxy,
 }
@@ -223,7 +224,6 @@ impl HistoryStore {
 
             // Remove it from the transaction hash database.
             let tx_hash = hist_tx.tx_hash();
-
             txn.remove_item(
                 &self.tx_hash_table,
                 &tx_hash,
@@ -277,7 +277,7 @@ impl HistoryStore {
             let mut duplicate = cursor.first_duplicate::<OrderedHash>();
 
             while let Some(v) = duplicate {
-                if !removed_txs.contains(&v.hash) {
+                if !removed_txs.contains(&v.hash.into()) {
                     break;
                 }
                 cursor.remove();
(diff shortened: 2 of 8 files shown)


blockchain/src/history/history_store.rs
```diff
@@ -44,12 +44,13 @@ pub struct HistoryStore {
     /// A database of all historic transactions indexed by their hash (= leaf hash in the history
     /// tree).
     hist_tx_table: TableProxy,
-    /// A database of all leaf hashes and indexes indexed by the hash of the transaction. This way we
-    /// can start with a transaction hash and find it in the MMR.
+    /// A database of all leaf hashes and indexes indexed by the hash of the (raw) transaction. This way we
+    /// can start with a raw transaction hash and find it in the MMR.
+    /// Mapping of raw tx to leaf hash  (= executed tx hash + index).
```
paberr, line 49: There seems to be a mismatch between here and what we actually put in (as it sometimes seems to be the ext. tx hash).

blockchain/src/history/history_store.rs
```diff
@@ -934,24 +934,25 @@ impl HistoryStore {
...
         match &hist_tx.data {
             HistoricTransactionData::Basic(tx) => {
                 let tx = tx.get_raw_transaction();
+                // The raw tx hash corresponds to the hash without the execution result.
+                // Thus for basic historic transactions we want discoverability for the raw transaction.
+                let raw_tx_hash: Blake2bHash = tx.hash();
+
+                txn.put(
+                    &self.tx_hash_table,
+                    &raw_tx_hash,
```
paberr, line 948: Here, we put the raw hash as the comment above indicates.

blockchain/src/history/history_store.rs
```diff
@@ -981,21 +982,38 @@ impl HistoryStore {
                 let index_tx_recipient =
                     self.get_last_tx_index_for_address(&ev.reward_address, Some(txn)) + 1;
 
+                txn.put(
+                    &self.tx_hash_table,
+                    &executed_tx_hash,
```
paberr: Here, it's the executed tx hash.

blockchain/src/history/history_store.rs
```diff
@@ -981,21 +982,38 @@ impl HistoryStore {
...
-            HistoricTransactionData::Equivocation(_) => {}
-            // Do not index punishment events, since I do not see a use case
-            // for this at the time.
-            HistoricTransactionData::Penalize(_) | HistoricTransactionData::Jail(_) => {}
+            HistoricTransactionData::Equivocation(_)
+            | HistoricTransactionData::Penalize(_)
+            | HistoricTransactionData::Jail(_) => {
+                txn.put(
+                    &self.tx_hash_table,
+                    &executed_tx_hash,
```
paberr: Executed again.

paberr (commented):
Also, this code is used for discovering duplicate transactions, right?
What is the expected behaviour? Should it be possible for a failed transaction to be re-included?

blockchain/src/history/history_store.rs
ii-cruz: The executed tx hash and the raw tx hash are the same for these cases. it only defers on the HistoricTransactionData::Basic(tx).

ii-cruz:
> Also, this code is used for discovering duplicate transactions, right? What is the expected behaviour? Should it be possible for a failed transaction to be re-included?

The behavior we had before was to not allow it. So that behavior is kept with these changes.

primitives/transaction/src/historic_transaction.rs
```diff
@@ -13,6 +18,84 @@ use crate::{
     Transaction as BlockchainTransaction,
 };
 
+/// The raw transaction hash is a type wrapper.
+/// This corresponds to the hash of the transaction without the execution result.
+/// This hash is the external facing hash.
+#[derive(Clone, Debug, Eq, PartialEq, Hash)]
+pub struct RawTransactionHash {
+    hash: Blake2bHash,
+}
```
jsdanielh: Why not using a `tuple struct`? 
```suggestion
pub struct RawTransactionHash(Blake2bHash)
```

primitives/transaction/src/historic_transaction.rs
```diff
@@ -13,6 +18,84 @@ use crate::{
...
+
+/// The executed transaction hash is a type wrapper.
+/// This corresponds to the hash of the transaction with the execution result.
+/// This hash is used in the history store so that we have a record of the transaction
+/// and its execution result. This is an internal to it.
+#[derive(Clone, Debug)]
+struct ExecutedTransactionHash {
+    hash: Blake2bHash,
+}
```
jsdanielh: Why not using a `tuple struct`?
```suggestion
struct ExecutedTransactionHash (Blake2bHash)
```

primitives/transaction/src/lib.rs
```diff
@@ -183,11 +183,23 @@ impl ExecutedTransaction {
         }
     }
 
-    pub fn hash(&self) -> Blake2bHash {
+    pub fn serialize_content(&self) -> Vec<u8> {
```
jsdanielh, line 187: Could you add some `rustdoc`?

blockchain/src/blockchain/verify.rs
```diff
@@ -5,6 +5,7 @@ use nimiq_database::{
     traits::{ReadTransaction, WriteTransaction},
     TransactionProxy as DBTransaction, WriteTransactionProxy,
 };
+use nimiq_hash::Hash;
```
jsdanielh, line 8: Mmm, why do we need it?

ii-cruz: This is needed because we hash a transaction on the verify function in line 106. 
However, you made me notice that we were hashing the executed transaction and checking if it has been already checked in the mempool. Since the mempool only knows raw_tx_hash it would always be false and we would always re-verify every transaction. 

This was a nice catch.

jsdanielh (approved):

jsdanielh:
Rebasing to merge it

PR #2206, opened 2024-02-09
Fix set active stake and revert update staker

ii-cruz:
The set active stake was rebalancing between the total balance instead of the non-retired balance. This added the retired balance to the inactive stake increasing the total staker funds.
It also fixes the same issue on the revert update staker.
This also fixes the tests to take these scenarios into account.

## Pull request checklist

### DIFF
--- a/primitives/account/src/account/staking_contract/staker.rs
+++ b/primitives/account/src/account/staking_contract/staker.rs
@@ -514,11 +514,11 @@ impl StakingContract {
         let mut staker = store.expect_staker(staker_address)?;
 
         // Fail if staker does not have sufficient funds.
-        let total_balance = staker.total_balance();
-        if total_balance < new_active_balance {
+        let non_retired_balance = staker.none_retired_balance();
+        if non_retired_balance < new_active_balance {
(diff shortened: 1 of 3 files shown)


primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -484,7 +484,7 @@ impl StakingContract {
         }
 
         // Restore the previous balances and values.
-        let total_balance = staker.total_balance();
+        let total_balance = staker.none_retired_balance();
```
styppo:
```suggestion
        let non_retired_balance = staker.non_retired_balance();
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -484,7 +484,7 @@ impl StakingContract {
...
         staker.active_balance = receipt.active_balance;
         staker.inactive_balance = total_balance - staker.active_balance;
```
styppo:
```suggestion
        staker.inactive_balance = non_retired_balance - staker.active_balance;
```

PR #2216, opened 2024-02-15
Add state to the validator-network and fix some network issues

jsdanielh:
- lib: Go back to use `3` as the default network quorum value
  - Change the client code to use `3` as the default network quorum value.
    Also make the `devnet` scripts use this value for configuring the
    nodes.
- validator-network: Do less validator `PeerId` cache clears    
  - Change the function that clears the validator `PeerId` to receive the request error such that it doesn't clear the cache if     `InboundRequestError::NoReceiver` error was received. This error is an indicative that the requested peer has no receiver for such request which could happen when the peer is no running the specified aggregation and this is not an indicative of the `PeerId` cache entry being wrong. Thus in case of receiving this error, the network won't invalidate the cache entry for this validator.
- network-libp2p: Fix mishandling of `FinishedWithNoAdditionalRecord`
  - Fix the mishandling of `FinishedWithNoAdditionalRecord` since we could get such event even if the `query.finish` isn't called (which the network is calling once `Quorum` records have been received).
    This mishandling was causing some of the queries to hang indefinitely since no error nor result was being sent to the application layer.
    Now the network always pushes the result to the application layer once the `FinishedWithNoAdditionalRecord` is received.
- validator-network: Add some state to the `PeerId` cache
  - Add some state to the validator `PeerId` cache to make less queries as possible to the DHT.
    This also makes the `validator-network` able to quickly return if the DHT query is still in progress allowing upper layers to continue their process without blocking them until the query finishes or errors out.

## Pull request checklist

### DIFF
--- a/network-libp2p/src/network.rs
+++ b/network-libp2p/src/network.rs
@@ -890,38 +890,6 @@ impl Network {
                                                         .query_mut(&id)
                                                         .unwrap()
                                                         .finish();
-                                                    let signed_best_record = results
-                                                        .best_value
-                                                        .clone()
-                                                        .get_signed_record();
-                                                    // Send the best result to the application layer ASAP since we already know it
-                                                    if let Some(output) = state.dht_gets.remove(&id)
-                                                    {
-                                                        if output
-                                                            .send(Ok(signed_best_record
-                                                                .clone()
-                                                                .value))
-                                                            .is_err()
-                                                        {
-                                                            error!(query_id = ?id, error = "receiver hung up", "could not send get record query result to channel");
-                                                        }
-                                                    } else {
-                                                        warn!(query_id = ?id, ?step, "GetRecord query result for unknown query ID");
-                                                    }
-                                                    if !results.outdated_values.is_empty() {
-                                                        // Now push the best value to the outdated peers
-                                                        let outdated_peers = results
-                                                            .outdated_values
-                                                            .iter()
-                                                            .map(|dht_record| {
-                                                                dht_record.get_peer_id()
-                                                            });
-                                                        swarm.behaviour_mut().dht.put_record_to(
-                                                            signed_best_record,
-                                                            outdated_peers,
-                                                            kad::Quorum::One,
-                                                        );
-                                                    }
                                                 }
                                             } else {
                                                 log::error!(query_id = ?id, "DHT inconsistent state");
@@ -937,11 +905,36 @@ impl Network {
                                             cache_candidates,
                                         },
                                     )) => {
-                                        // Remove the query and push the best result to the cache candidates
+                                        // Remove the query, send the best result to the application layer
+                                        // and push the best result to the cache candidates
                                         if let Some(results) = state.dht_get_results.remove(&id) {
+                                            let signed_best_record =
+                                                results.best_value.clone().get_signed_record();
+                                            // Send the best result to the application layer
+                                            if let Some(output) = state.dht_gets.remove(&id) {
+                                                if output
+                                                    .send(Ok(signed_best_record.clone().value))
+                                                    .is_err()
+                                                {
+                                                    error!(query_id = ?id, error = "receiver hung up", "could not send get record query result to channel");
+                                                }
+                                            } else {
+                                                warn!(query_id = ?id, ?step, "GetRecord query result for unknown query ID");
+                                            }
+                                            if !results.outdated_values.is_empty() {
+                                                // Now push the best value to the outdated peers
+                                                let outdated_peers = results
+                                                    .outdated_values
+                                                    .iter()
+                                                    .map(|dht_record| dht_record.get_peer_id());
+                                                swarm.behaviour_mut().dht.put_record_to(
+                                                    signed_best_record.clone(),
+                                                    outdated_peers,
+                                                    kad::Quorum::One,
+                                                );
+                                            }
+                                            // Push the best result to the cache candidates
                                             if !cache_candidates.is_empty() {
-                                                let signed_best_record =
-                                                    results.best_value.get_signed_record();
                                                 let peers = cache_candidates
                                                     .iter()
                                                     .map(|(_, &peer_id)| peer_id);
@@ -951,6 +944,8 @@ impl Network {
                                                     kad::Quorum::One,
                                                 );
                                             }
+                                        } else {
+                                            panic!("DHT inconsistent state, query_id: {:?}", id);
                                         }
                                     }
                                     QueryResult::GetRecord(Err(error)) => {
--- a/validator-network/src/network_impl.rs
+++ b/validator-network/src/network_impl.rs
@@ -6,7 +6,7 @@ use log::warn;
 use nimiq_bls::{lazy::LazyPublicKey, CompressedPublicKey, KeyPair, SecretKey};
 use nimiq_network_interface::{
     network::{CloseReason, MsgAcceptance, Network, SubscribeEvents, Topic},
-    request::{Message, Request, RequestCommon},
+    request::{InboundRequestError, Message, Request, RequestCommon, RequestError},
 };
 use nimiq_serde::{Deserialize, Serialize};
 use parking_lot::RwLock;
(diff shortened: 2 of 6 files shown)


validator-network/src/network_impl.rs
```diff
@@ -57,7 +67,9 @@ where
     }
 
     /// Returns the local validator ID, if elected, `Err(NotElected)` otherwise.
-    fn local_validator_id<T: Error + 'static>(&self) -> Result<u16, NetworkError<T>> {
+    fn local_validator_id<T: Error + std::marker::Sync + 'static>(
```
hrxi:
```suggestion
    fn local_validator_id<T: Error + Sync + 'static>(
```

validator-network/src/network_impl.rs
```diff
@@ -76,50 +88,111 @@ where
...
+                self.validator_peer_id_cache
+                    .write()
+                    .insert(public_key.compressed().clone(), CacheState::InProgress);
+            }
+            let self_ = self.arc_clone();
+            tokio::spawn(async move {
+                Self::update_peer_id_cache(&self_, validator_id, &public_key).await;
+            });
         }
+        Err(NetworkError::UnknownValidator(validator_id))
```
hrxi, line 184: If I understand this code correctly, it changes the behavior for validators whose peer IDs are not yet known. Previously, the peer ID would be looked up and then the message would be sent. Now, the peer ID will be looked up, but the message will be reported as failed. Is that intentional?

> This also makes the `validator-network` able to quickly return if the DHT query is still in progress allowing upper layers to continue their process without blocking them until the query finishes or errors out.

I guess that sounds like "yes".

validator-network/src/network_impl.rs
```diff
@@ -76,50 +88,111 @@ where
...
-                "Could not find peer ID for validator in DHT: public_key = {:?}",
-                public_key
-            );
-            Err(NetworkError::UnknownValidator(validator_id))
+        if update_peer_cache {
+            // Cache is empty for this validator ID, query the entry
+            {
+                self.validator_peer_id_cache
+                    .write()
+                    .insert(public_key.compressed().clone(), CacheState::InProgress);
```
hrxi: In order not to run into concurrency issues, I think the `match cache_state` from above needs to be repeated here, with the write lock present. Otherwise, another future might have inserted a `CacheState::InProgress` or even a `CacheState::Resolved` here, and we'd start a second future or throw away the result.

network-libp2p/src/network.rs
```diff
@@ -951,6 +944,8 @@ impl Network {
                                                     kad::Quorum::One,
                                                 );
                                             }
+                                        } else {
+                                            log::error!(query_id = ?id, "DHT inconsistent state");
```
hrxi: Should this be unreachable? If so, a panic would probably be more appropriate.

validator-network/src/network_impl.rs
```diff
@@ -76,50 +88,111 @@ where
...
 
-    /// Clears the validator->peer_id cache.
+    /// Clears the validator->peer_id cache on a `RequestError`.
     /// The cached entry should be cleared when the peer id might have changed.
-    fn clear_validator_peer_id_cache(&self, validator_id: u16) {
-        if let Some(validator_key) = self.validator_keys.read().get(usize::from(validator_id)) {
-            self.validator_peer_id_cache
-                .write()
-                .remove(validator_key.compressed());
+    fn clear_validator_peer_id_cache_on_error(&self, validator_id: u16, error: RequestError) {
```
hrxi:
```suggestion
    fn clear_validator_peer_id_cache_on_error(&self, validator_id: u16, error: &RequestError) {
```
reduces necessary cloning.

hrxi (commented):
I found one potential issue, but the whole thing looks very good. :)

validator-network/src/network_impl.rs
```diff
@@ -76,50 +88,111 @@ where
         }
     }
 
+    // Looks up the peer ID for a validator public key in the DHT and updates the internal cache
+    async fn update_peer_id_cache(&self, validator_id: u16, public_key: &LazyPublicKey) {
+        self.validator_peer_id_cache
+            .write()
+            .insert(public_key.compressed().clone(), CacheState::InProgress);
```
nibhar: Might be redundant with L168-170.

validator-network/src/network_impl.rs
```diff
@@ -76,50 +88,111 @@ where
...
-        let (peer_id, public_key) = {
-            let public_key = self
-                .validator_keys
-                .read()
-                .get(usize::from(validator_id))
-                .ok_or(NetworkError::UnknownValidator(validator_id))?
-                .clone();
+    ) -> Result<N::PeerId, NetworkError<N::Error>>
+    where
+        <N as Network>::Error: Sync + Send,
```
nibhar: Could go to the parent `impl<N> ValidatorNetworkImpl<N> ...`

validator-network/src/network_impl.rs
```diff
@@ -76,50 +88,111 @@ where
...
+                        log::debug!(validator_id, "Record resolution is in progress");
+                        false
+                    }
+                }
+            } else {
+                log::error!(
+                    ?public_key,
+                    validator_id,
+                    "Could not find peer ID for validator in DHT",
+                );
```
nibhar: Just out of curiosity, feel free to ignore this.
`error!` might be a bit high to me as this seemingly would be hit during normal execution. For example whenever a node starts, or when a new validator gets elected. Since it would be expected behaviour then maybe debug is more appropriate.

validator-network/src/network_impl.rs
nibhar: Yes, this should take an upgradable read in the beginning and keep it. Upgrade here then.

validator-network/src/network_impl.rs
```diff
@@ -76,50 +88,111 @@ where
...
+            // Cache is empty for this validator ID, query the entry
+            {
+                self.validator_peer_id_cache
+                    .write()
+                    .insert(public_key.compressed().clone(), CacheState::InProgress);
+            }
+            let self_ = self.arc_clone();
+            tokio::spawn(async move {
+                Self::update_peer_id_cache(&self_, validator_id, &public_key).await;
+            });
```
nibhar, line 182: This function is `async`, it could be called without the `tokio::spawn`. If there is good reason not to, then this function does not need to be `async`. 
As I understand it this is supposed to early return thus the spawn, so I would remove the async.

validator-network/src/network_impl.rs
hrxi: Not necessarily upgradable read, but maybe rather read, drop, then write, and then re-do the same steps with the write lock, for more performance. Otherwise the upgradable read locks out all other threads in the same code path.

validator-network/src/network_impl.rs
jsdanielh: yes

validator-network/src/network_impl.rs
```diff
@@ -76,50 +87,116 @@ where
...
+        if update_peer_cache {
+            // Cache is empty for this validator ID, query the entry
+            {
+                // Re-check the validator Peer ID cache with the write lock taken and update it if necessary
+                let mut validator_peer_id_cache = self.validator_peer_id_cache.write();
+                if let Some(cache_state) =
+                    validator_peer_id_cache.get(&public_key.compressed().clone())
+                {
+                    match cache_state {
+                        CacheState::Resolved(_) => {}
```
hrxi:
```suggestion
                        CacheState::Resolved(peer_id) => return Ok(*peer_id),
```
In this case, another thread has already resolved it. We can return the result immediately.

validator-network/src/network_impl.rs
```diff
@@ -76,50 +87,116 @@ where
...
+                        CacheState::Error(_) => {
+                            validator_peer_id_cache
+                                .insert(public_key.compressed().clone(), CacheState::InProgress);
+                        }
+                        CacheState::InProgress => {}
```
hrxi:
```suggestion
                        CacheState::InProgress => {
                            log::debug!(validator_id, "Record resolution is in progress");
                            return Err(NetworkError::UnknownValidator(validator_id));
                        }
```
In this case, another thread is already working on the resolution. We shouldn't start a second one.

PR #2227, opened 2024-02-19
Use a lightweight JSON-parser for parsing Webauthn clientData

sisou:
Also handle errors correctly, fixing all FIXMEs.

Instead of hand-parsing external clientDataJSON for Webauthn signatures, this PR introduces a small library to parse it.

This also fixes a few FIXMEs left in the section from the comment. These FIXMEs could also be fixed without the parser.

Unfortunately, even though the JSON-parser crate should only be 6.4kb according to crates.io, this PR increases the main-thread "primitives" WASM file by 33.5kb, which is 3.7% of the previous 904.9kb. No idea why.

## Pull request checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4412,6 +4412,7 @@ dependencies = [
  "nimiq-test-log",
  "nimiq-test-utils",
  "nimiq-utils",
+ "rust_json",
  "serde",
  "serde_json",
  "strum_macros",
@@ -5752,6 +5753,12 @@ dependencies = [
  "blake2b_simd",
  "constant_time_eq 0.3.0",
 ]
+
+[[package]]
+name = "rust_json"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "540c69f089252cacc268b7f273cfcd4ddeeae2d961936e43baa1311ab1c17388"
 
 [[package]]
 name = "rustc-demangle"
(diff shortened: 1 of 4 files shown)


hrxi:
Could this potentially be done using `serde_json`, which we already depend on?

sisou:
I did the change to `serde_json`, and the increase in size is still 23.7kb, or 2.6%. Better than `rust_json`, but not great. I think the reason is that JSON parsing is not used anywhere else in the code, so the logic for it has to be included only for this method.

sisou:
Instead of using a JSON-parser, we can also keep hand-parsing it (I think it's enough) and just fix the FIXMEs:
https://github.com/nimiq/core-rs-albatross/commit/39d15e06e0a51f5a156764564cf9ccc694cb54a5

hrxi:
Hmmm. I've taken a closer look at the `try_from_webauthn` function. Its purpose seems to be to decompose the JSON in a way that can be reversed later. This seems brittle, both in the old and in the new variant. Is the field order guaranteed by the standard? JSON strings can be encoded in multiple ways, whitespace can appear between fields, are those guaranteed not to exist by the standard?

Perhaps we should rather store the whole JSON string, as any kind of decompose-and-restore function is hard to get right?

sisou:
Storing the whole JSON blob takes a lot of bytes on the network and on-chain (in the proof), that's why I'm trying to only store the minimal information to recreate it.

The structure is indeed standardized, have a look at the description of #1867 for the relevant links. I had to however add flags for non-standard behavior already (see also that PR), which can be extended should we observe others in the future.

hrxi:
> Instead of using a JSON-parser, we can also keep hand-parsing it (I think it's enough)

I want to highlight that there's even a suggestion on how to hand-parse the JSON in the standard, so it's not too bad IMO.

hrxi:
Superseded by #2275.

PR #2229, opened 2024-02-19
Add config for release notes generation

sisou:
https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuring-automatically-generated-release-notes

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,15 @@
+changelog:
+  categories:
+    - title: Protocol Breaking Changes
+      labels:
+        - testnet-restart
+    - title: What's changed
+      labels:
+        - '*'
+      exclude:
+        labels:
+          - testnet-restart
+          - dependency
+    - title: Dependencies
+      labels:
+        - dependency


.github/release.yml
```diff
@@ -0,0 +1,15 @@
+changelog:
+  categories:
+    - title: Breaking Changes
```
jsdanielh:
```suggestion
    - title: Protocol Breaking Changes
```

.github/release.yml
```diff
@@ -0,0 +1,15 @@
...
+      labels:
+        - testnet-restart
+    - title: Updates
```
jsdanielh: Maybe, _Other Changes_?

sisou: What if there a no protocol breaking changes? This is the default category, which in the future should be the main category once there are no more breaking changes.

jsdanielh: Yeah, not sure. To me _Updates_ sounds more like dependency updates but this is not that

sisou: The default without a config file is "What's changed". We can use that. The documentation suggests "Features" as the title.

jsdanielh: Yeah, I'll be ok with any of those 2

jsdanielh: @sisou Will change it to "What's changed" to merge this PR. wdyt?

sisou: Sure!

jsdanielh (approved):

PR #2241, opened 2024-02-21
Fix typos

rex4539:


rex4539:
@sisou Could you please take a look?

scripts/devnet/node.py
```diff
@@ -319,7 +319,7 @@ def poll(self):
 
         :return: The return code of the process if it ended, None if it hasn't
             finished
-        :rtype: int ot None
+        :rtype: int to None
```
jsdanielh:
```suggestion
        :rtype: int or None
```

jsdanielh (changes requested):
Thanks for the PR. I only noted a wrong fix. Could you fix it?

scripts/devnet/node.py
rex4539: Fixed and rebased.

jsdanielh (approved):

jsdanielh:
@rex4539 Will push an additional commit since you just uncovered a compilation error that was hidden under the `fature` typo.

PR #2248, opened 2024-02-23
RPC schema generator

Eligioo:
This pull request introduces a new tool called `nimiq-rpc-schema`. It generates a JSON object written to stdout based on the trait and struct definitions in the `rpc-interface`-crate by parsing the code into an [AST](https://en.wikipedia.org/wiki/Abstract_syntax_tree) and turning it into an [OpenRPC specification](https://playground.open-rpc.org/?schemaUrl=https://raw.githubusercontent.com/nimiq/core-rs-albatross/stefan/openrpc-gen/tools/src/rpc-schema/schema.json). This schema is meant as the base for generating a consistent API documentation or the generation of RPC clients in various programming languages.

## Pull request checklist

### DIFF
--- a/rpc-interface/src/types.rs
+++ b/rpc-interface/src/types.rs
@@ -14,7 +14,7 @@ use nimiq_blockchain_proxy::BlockchainReadProxy;
 use nimiq_bls::CompressedPublicKey;
 use nimiq_collections::BitSet;
 use nimiq_hash::{Blake2bHash, Blake2sHash, Hash};
-use nimiq_keys::{Address, Ed25519PublicKey};
+use nimiq_keys::{Address, Ed25519PublicKey, Ed25519Signature, PrivateKey};
 use nimiq_primitives::{
     coin::Coin, networks::NetworkId, policy::Policy, slots_allocation::Validators,
 };
@@ -683,6 +683,28 @@ pub struct Account {
 
     #[serde(flatten)]
     pub account_additional_fields: AccountAdditionalFields,
+}
+
+/// A Ed25519 signature containing the actual signature and the corresponding public key.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ReturnSignature {
+    /// The public key of the keypair that signed the signature.
+    pub public_key: Ed25519PublicKey,
+    /// The output of the signing process. The signature can be used to verify the authenticity and integrity of the message using the corresponding public key.
+    pub signature: Ed25519Signature,
+}
+
+/// A wallet account.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ReturnAccount {
+    /// The address of the wallet account.
+    pub address: Address,
+    /// The public key of the account keypair.
+    pub public_key: Ed25519PublicKey,
+    /// The private key of the account keypair.
+    pub private_key: PrivateKey,
 }
 
 #[serde_as]
--- a/tools/Cargo.toml
+++ b/tools/Cargo.toml
@@ -25,12 +25,22 @@ path = "src/address/main.rs"
 name = "nimiq-signtx"
 path = "src/signtx/main.rs"
 
+[[bin]]
+name = "nimiq-rpc-schema"
+path = "src/rpc-schema/main.rs"
+
 [dependencies]
 anyhow = "1.0"
 clap = { version = "4.5", features = ["cargo"] }
+convert_case = "0.6"
 hex = "0.4"
 log = { package = "tracing", version = "0.1", features = ["log"] }
+quote = "1.0"
 rand = "0.8"
+schemars = "0.8"
+serde = "1.0"
+serde_json = "1.0"
+syn = { version = "2.0", features = ["full"] }
 thiserror = "1.0"
 
 nimiq-bls = { workspace = true }
--- a/tools/src/rpc-schema/main.rs
+++ b/tools/src/rpc-schema/main.rs
@@ -1,0 +1,113 @@
+mod openrpc;
+mod parser;
+
+use std::{env, fs};
+
+use anyhow::Error;
+use clap::{crate_authors, crate_description, crate_version, Arg, Command};
+use syn::parse_file;
+use thiserror::Error;
+
+use crate::openrpc::OpenRpcBuilder;
+
+fn main() -> Result<(), Error> {
+    let matches = Command::new("RPC schema generator")
+        .version(crate_version!())
+        .author(crate_authors!())
+        .about(crate_description!())
+        .arg(
+            Arg::new("openrpc_version")
+                .short('o')
+                .long("openrpc-version")
+                .value_name("OPENRPC_VERSION")
+                .help("Specify the OpenRPC version of the document. Usually you want to match this with your release version number."),
+        )
+        .arg(
+            Arg::new("openrpc_title")
+                .short('t')
+                .long("openrpc-title")
+                .value_name("OPENRPC_TITLE")
+                .default_value("Nimiq JSON-RPC Specification")
+                .help("Specify the OpenRPC title of the document."),
+        )
+        .arg(
+            Arg::new("source")
+            .short('s')
+            .long("source")
+            .value_name("SOURCE")
+            .default_value("rpc-interface/src")
+            .help("The folder that contains the source of the RPC interface traits and structs.")
+        )
+        .get_matches();
+
+    let mut builder = OpenRpcBuilder::builder()
+        .version(
+            matches
+                .get_one::<String>("openrpc_version")
+                .ok_or(AppError::OpenRpcArgumentMissing("version".to_string()))?
+                .to_string(),
+        )
+        .title(
+            matches
+                .get_one::<String>("openrpc_title")
+                .ok_or(AppError::OpenRpcArgumentMissing("title".to_string()))?
+                .to_string(),
+        );
+
+    match fs::read_dir(
+        matches
+            .get_one::<String>("source")
+            .ok_or(AppError::SourceCode)?,
+    ) {
+        Ok(entries) => {
+            for entry in entries.flatten() {
+                let path = entry.path();
+                if path.is_file() {
+                    if let Ok(contents) = fs::read_to_string(&path) {
+                        match parse_file(&contents) {
+                            Ok(ast) => {
+                                let structs = parser::extract_structs_from_ast(&ast);
+                                let fns = parser::extract_fns_from_ast(&ast);
+                                for index in 0..structs.len() {
+                                    builder = builder.with_schema(structs.get(index).unwrap());
+                                }
+                                for index in 0..fns.len() {
+                                    builder = builder.with_method(
+                                        fns.get(index).unwrap(),
+                                        path.file_name()
+                                            .unwrap()
+                                            .to_string_lossy()
+                                            .replace(".rs", ""),
+                                    );
+                                }
+                            }
+                            Err(err) => {
+                                return Err(AppError::ParsingError(err).into());
+                            }
+                        }
+                    }
+                }
+            }
+        }
+        Err(err) => return Err(AppError::LoadDirectoryError(err).into()),
+    }
+
+    print!(
+        "{}",
+        serde_json::to_string_pretty(&builder.build()).expect("Failed to serialize OpenRPC spec")
+    );
+
+    Ok(())
+}
+
+#[derive(Debug, Error)]
+enum AppError {
+    #[error("Unable to load specified directory: {0}")]
+    LoadDirectoryError(#[from] std::io::Error),
+    #[error("Failed to parse Rust source file: {0}")]
+    ParsingError(#[from] syn::Error),
+    #[error("OpenRPC {0} argument is missing")]
+    OpenRpcArgumentMissing(String),
+    #[error("The source argument is missing")]
+    SourceCode,
+}
--- a/tools/src/rpc-schema/openrpc/mod.rs
+++ b/tools/src/rpc-schema/openrpc/mod.rs
@@ -1,0 +1,126 @@
+use std::collections::BTreeMap;
+
+use serde_json::{Map, Value};
+
+use self::document::{
+    Components, ContactObject, InfoObject, LicenseObject, MethodObject, Openrpc, OpenrpcDocument,
+    TagObject, TagOrReference,
+};
+use crate::parser::{ParsedItemStruct, ParsedTraitItemFn};
+
+#[allow(unused)]
+pub mod document;
+
+/// A builder for constructing an OpenRPC document.
+#[derive(Clone)]
+pub struct OpenRpcBuilder {
+    open_rpc_doc: OpenrpcDocument,
+    structs: Vec<ParsedItemStruct>,
+    methods: Vec<(ParsedTraitItemFn, String)>,
+}
+
+impl OpenRpcBuilder {
+    /// Adds a schema based on a Rust struct to the OpenRPC document.
+    pub fn with_schema(mut self, item_struct: &ParsedItemStruct) -> OpenRpcBuilder {
+        self.structs.push(item_struct.clone());
+        self
+    }
+
+    /// Adds a method based on a Rust trait method to the OpenRPC document.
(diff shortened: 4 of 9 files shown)


hrxi:
Could you add a CI job to check that the generated schema matches?

Eligioo:
@hrxi I have thought about this or try to solve this in a way in the form of a test that validates the schema. However the type for `document.schema` is a `HashMap<String, Option<Value>>`. So every time you run the generator, it inserts schemas into the hashmap but the order they are placed in differs each run meaning that generated schema is different.

Open for suggestions though to make it more consistent

hrxi:
> However the type for `document.schema` is a `HashMap<String, Option<Value>>`.

Do you control this type? If so, you could replace it by a `BTreeMap<String, Option<Value>>`, that would sort it automatically. If not, could you link me to where it is defined?

rpc-interface/src/types.rs
```diff
@@ -685,6 +685,21 @@ pub struct Account {
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ReturnSignature {
+    pub public_key: Ed25519PublicKey,
+    pub signature: Ed25519Signature,
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ReturnAccount {
```
jsdanielh, line 701: Could you add `rustdoc` to the struct and its members?

rpc-interface/src/types.rs
```diff
@@ -685,6 +685,21 @@ pub struct Account {
     pub account_additional_fields: AccountAdditionalFields,
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ReturnSignature {
```
jsdanielh, line 691: Could you add `rustdoc` to the struct and its members?

tools/Cargo.toml
```diff
@@ -25,12 +25,21 @@ path = "src/address/main.rs"
...
 path = "src/signtx/main.rs"
 
+[[bin]]
+name = "nimiq-rpc-schema"
+path = "src/rpc-schema/main.rs"
+
 [dependencies]
 anyhow = "1.0"
+convert_case = "0.6"
 clap = { version = "4.5", features = ["cargo"] }
```
jsdanielh:
```suggestion
clap = { version = "4.5", features = ["cargo"] }
convert_case = "0.6"
```

tools/src/rpc-schema/main.rs
```diff
@@ -0,0 +1,55 @@
...
+                }
+            }
+        }
+    } else {
+        return Err("Unable to load directory".into());
+    }
+
+    let json_spec =
+        serde_json::to_string_pretty(&builder.build()).expect("Failed to serialize OpenRPC spec");
+    let mut file = File::create("tools/src/rpc-schema/schema.json").expect("Failed to create file");
```
jsdanielh: Mmm, I'd say that this shouldn't land in `tools/src`. I'd say `doc` would be more suitable

tools/src/rpc-schema/main.rs
```diff
@@ -0,0 +1,55 @@
...
+    fs::{self, File},
+    io::Write,
+};
+
+use syn::parse_file;
+
+use crate::openrpc::OpenRpcBuilder;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    let directory = "./rpc-interface/src";
```
jsdanielh: Mmm, this tool is assuming that we're going to run it in the root of the repository. I'd rather accept a command line argument

tools/src/rpc-schema/openrpc.rs
```diff
@@ -0,0 +1,89 @@
...
+
+use open_rpc_schema::document::{
+    Components, ContactObject, InfoObject, LicenseObject, MethodObject, Openrpc, OpenrpcDocument,
+};
+use serde_json::{Map, Value};
+
+use crate::parser::{ParsedItemStruct, ParsedTraitItemFn};
+
+#[derive(Clone)]
+pub struct OpenRpcBuilder {
```
jsdanielh: Could you add some `rustdoc`?

tools/src/rpc-schema/openrpc.rs
```diff
@@ -0,0 +1,89 @@
...
+    open_rpc_doc: OpenrpcDocument,
+    structs: Vec<ParsedItemStruct>,
+    methods: Vec<ParsedTraitItemFn>,
+}
+
+impl OpenRpcBuilder {
+    pub fn with_schema(mut self, item_struct: &ParsedItemStruct) -> OpenRpcBuilder {
```
jsdanielh: Could you add some `rustdoc`?

tools/src/rpc-schema/openrpc.rs
```diff
@@ -0,0 +1,89 @@
...
+        self.structs.push(item_struct.clone());
+        self
+    }
+
+    pub fn with_method(mut self, item_fn: &ParsedTraitItemFn) -> OpenRpcBuilder {
```
jsdanielh: Could you add some `rustdoc`?

tools/src/rpc-schema/openrpc.rs
```diff
@@ -0,0 +1,89 @@
...
+        self.methods.push(item_fn.clone());
+        self
+    }
+
+    pub fn builder() -> OpenRpcBuilder {
```
jsdanielh: Could you add some `rustdoc`?

tools/src/rpc-schema/openrpc.rs
```diff
@@ -0,0 +1,89 @@
...
+                examples: None,
+                example_pairings: Some(HashMap::new()),
+                content_descriptors: None,
+                tags: None,
+            }),
+            ..Default::default()
+        }, structs: vec![], methods: vec![]}
+    }
+
+    pub fn build(self) -> OpenrpcDocument {
```
jsdanielh: Could you add some `rustdoc`?

tools/src/rpc-schema/parser.rs
```diff
@@ -0,0 +1,443 @@
...
+};
+use quote::ToTokens;
+use serde_json::{Map, Value};
+use syn::{
+    Field, File, GenericArgument, Ident, ItemStruct, Pat, PatIdent, Path, PathArguments,
+    PathSegment, ReturnType, TraitItem, TraitItemFn, Type,
+};
+
+#[derive(Debug, Clone)]
+pub struct ParsedItemStruct(ItemStruct);
```
jsdanielh, line 16: Could you add some `rustdoc`?

tools/src/rpc-schema/parser.rs
```diff
@@ -0,0 +1,443 @@
...
+#[derive(Debug, Clone)]
+pub struct ParsedTraitItemFn(TraitItemFn);
```
jsdanielh, line 19: Could you add some `rustdoc`?

tools/src/rpc-schema/parser.rs
```diff
@@ -0,0 +1,443 @@
...
+
+impl ParsedItemStruct {
+    #[inline]
+    pub fn title(&self) -> String {
```
jsdanielh, line 24: Could you add some `rustdoc`?

tools/src/rpc-schema/parser.rs
```diff
@@ -0,0 +1,443 @@
...
+        self.0.ident.to_string()
+    }
+
+    #[inline]
+    pub fn description(&self) -> String {
```
jsdanielh, line 30: Could you add some `rustdoc`?

tools/src/rpc-schema/parser.rs
```diff
@@ -0,0 +1,443 @@
...
+        "".into()
+    }
+
+    pub fn properties(&self, structs: &[ParsedItemStruct]) -> Value {
```
jsdanielh, line 35: Could you add some `rustdoc`?

tools/src/rpc-schema/parser.rs
```diff
@@ -0,0 +1,443 @@
...
+            | "BitSet"
+            | "S"
+            | "T"
+            | "BlockAdditionalFields"
+            | "MultiSignature" => Value::String("object".into()),
+            _ => panic!("{:?}", inner_ident),
+        }
+    }
+
+    pub fn required_fields(&self) -> Vec<Value> {
```
jsdanielh, line 144: Could you add some `rustdoc`?

tools/src/rpc-schema/parser.rs
```diff
@@ -0,0 +1,443 @@
...
+                }
+            }
+            _ => (path, ident),
+        }
+    }
+}
+
+impl ParsedTraitItemFn {
+    #[inline]
+    pub fn title(&self) -> String {
```
jsdanielh, line 201: Could you add some `rustdoc`?

tools/src/rpc-schema/parser.rs
```diff
@@ -0,0 +1,443 @@
...
+        self.0.sig.ident.to_string().to_case(Case::Camel)
+    }
+
+    #[inline]
+    pub fn description(&self) -> String {
```
jsdanielh, line 207: Could you add some `rustdoc`?

tools/src/rpc-schema/parser.rs
```diff
@@ -0,0 +1,443 @@
...
+        "".into()
+    }
+
+    pub fn params(&self, structs: &[ParsedItemStruct]) -> Vec<ContentDescriptorOrReference> {
```
jsdanielh, line 224: Could you add some `rustdoc`?

tools/src/rpc-schema/parser.rs
```diff
@@ -0,0 +1,443 @@
...
+
+    fn param_required(_ty: &Type) -> bool {
+        // At the moment, all params are required even if the type is wrapped in an Option.
+        // if ty.to_token_stream().to_string().contains("Option") {
+        //     return false;
+        // }
+        true
+    }
+
+    pub fn return_type(&self, structs: &[ParsedItemStruct]) -> ContentDescriptorOrReference {
```
jsdanielh, line 280: Could you add some `rustdoc`?

tools/src/rpc-schema/parser.rs
```diff
@@ -0,0 +1,443 @@
...
+    }
+}
+
+impl From<TraitItemFn> for ParsedTraitItemFn {
+    fn from(val: TraitItemFn) -> Self {
+        ParsedTraitItemFn(val)
+    }
+}
+
+pub fn extract_structs_from_ast(file: &File) -> Vec<ParsedItemStruct> {
```
jsdanielh, line 453: Could you add some `rustdoc`?

tools/src/rpc-schema/parser.rs
```diff
@@ -0,0 +1,443 @@
...
+    file.items
+        .iter()
+        .filter_map(|item| match item {
+            syn::Item::Struct(item_struct) => Some(item_struct.clone().into()),
+            _ => None,
+        })
+        .collect()
+}
+
+pub fn extract_fns_from_ast(file: &File) -> Vec<ParsedTraitItemFn> {
```
jsdanielh, line 464: Could you add some `rustdoc`?

tools/src/rpc-schema/schema.json
```diff
@@ -0,0 +1,4420 @@
+{
```
jsdanielh: Why to have this file versioned? To me looks like the product of this tool and it might change. Sounds like we need to have a CI that on release publish this file somewhere

jsdanielh (commented):
Found some suggestions

hrxi:
If we can't control the output, we should probably post-process it to make it reproducible. But I still think we can control the output better. ^^

tools/src/rpc-schema/main.rs
Eligioo: I will rework this and it will just write it to stdout. Then let the CI call this tool and add it as a release attachment to the latest release when one is created.

tools/src/rpc-schema/schema.json
Eligioo: I will rework this and it will just write it to stdout. Then let the CI call this tool and add it as a release attachment to the latest release when one is created.

tools/src/rpc-schema/openrpc/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /// Creates a new instance of OpenRpcBuilder with default settings and returns it.
+    pub fn builder() -> OpenRpcBuilder {
+        OpenRpcBuilder{ open_rpc_doc: OpenrpcDocument {
+            openrpc: Openrpc::V26,
+            info: InfoObject {
+                title: "Nimiq JSON-RPC Specification".to_string(),
+                description: Some("Through the use of JSON-RPC, Nimiq nodes expose a set of standardized methods and endpoints that allow external applications and tools to interact, stream and control the behavior of the nodes. This includes functionalities such as retrieving information about the blockchain state, submitting transactions, managing accounts, and configuring node settings.".to_string()),
+                version: "Testnet 0.20.0".to_string(),
```
jsdanielh: We might need to avoid hardcoding this in the future since this will be changing a lot

tools/src/rpc-schema/openrpc/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+                terms_of_service: None,
+                contact: Some(ContactObject { name: Some("The Nimiq Foundation".to_string()), email: Some("info@nimiq.com".to_string()), url: Some("https://nimiq.com".to_string()) }),
```
jsdanielh: Same as contact and license. Wondering if we can get them from`Cargo.toml`

tools/src/rpc-schema/openrpc/mod.rs
```diff
@@ -0,0 +1,103 @@
...
+                license: Some(LicenseObject{ name: Some("Apache License, Version 2.0".to_string()), url: Some("http://www.apache.org/licenses/LICENSE-2.0".to_string()) }),
```
jsdanielh: Same as contact and license. Wondering if we can get them from`Cargo.toml`

jsdanielh (commented):
Could you also rebase and squash the commits?

tools/src/rpc-schema/openrpc/mod.rs
Eligioo: Contact and licence are now extracted from the `Cargo.toml`.

tools/src/rpc-schema/openrpc/mod.rs
Eligioo: Contact and licence are now extracted from the `Cargo.toml`.

tools/src/rpc-schema/openrpc/mod.rs
Eligioo: `Title` has been turned into a `clap` argument with a default value.
`Version` was an oversight because the `InfoObject` didn't implement the `Default` trait at first but is a required `clap` argument so that will always be overwritten. Thus has been removed here.
`Description` can stay like this in my opinion because the meaning of RPC and this document won't change.

Eligioo:
@jsdanielh Squashed and rebased

onmax:
Hey, I am doing some scripts in JS and I found a problem.

The current JSON has a chunk with the following code:

```json
  		{
          "name": "logTypes",
          "schema": {
            "type": "array",
            "items": {
              "type": "array"
            }
          },
          "required": true
        }
```


The problem with this is that a `type: "array"` must have an `items` key. Otherwise tools [like this](https://github.com/json-schema-tools/titleizer/blob/master/src/index.ts#L60) will throw an error.

Is there a way of changing @Eligioo? Nothing fancy but at least `items: {type: object}` maybe?


> I use [`@json-schema-tools` libraries](https://github.com/json-schema-tools)

Eligioo:
@onmax thanks for letting me know. I fixed this and you should be able to fetch a new document from the latest release.

It is generated now like the following which is correct for `logTypes`

```
{
  "name": "logTypes",
  "schema": {
    "type": "array",
    "items": {
      "type": "string"
    }
  },
  "required": true
}
```

jsdanielh (approved):

PR #2156, opened 2024-01-26
Improve performance of TrieNode serialization

jsdanielh:
Improve performance of `TrieNode` serialization.

## Pull request checklist

- [X] All tests pass. The project builds and runs.
- [X] I have resolved any merge conflicts.
- [X] I have resolved all `clippy` and `rustfmt` warnings.

PR #2249, opened 2024-02-23
Improve trienode serialization performance by vendoring postcard

hrxi:
The PR to postcard is unlikely to get merged anytime soon. We could instead vendor postcard and get the performance this way.

Alternative to #2156.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2200,7 +2200,7 @@ dependencies = [
  "httpdate",
  "itoa",
  "pin-project-lite",
- "socket2 0.4.9",
+ "socket2 0.5.5",
  "tokio",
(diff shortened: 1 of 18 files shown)


sisou:
With other forked dependencies we use(d) forked Github repos as the dependency location in the Cargo.toml files, instead of putting the whole code into this repo.

jsdanielh:
> With other forked dependencies we use(d) forked Github repos as the dependency location in the Cargo.toml files, instead of putting the whole code into this repo.

Yeah, I'd second this option instead

jsdanielh (commented):
LGTM

jsdanielh (approved):

PR #2274, opened 2024-03-01
validator-network: Fix multiple DHT queries for the same validator ID

jsdanielh:
- Fix the `get_validator_peer_id` function to properly handle the case where no cache entry is present once it obtains the write lock. This avoids multiple DHT queries for the same validator ID when there are multiple requests being made and no cache entry is present.
- Only clear the `PeerId` cache on errors if the faced error is for the same `PeerId` that is store in the cache. This avoids unnecessary clears for already properly resolved `PeerId`s.

## Pull request checklist

### DIFF
--- a/validator-network/src/network_impl.rs
+++ b/validator-network/src/network_impl.rs
@@ -147,11 +147,6 @@ where
                     }
                 }
             } else {
-                log::debug!(
-                    ?public_key,
-                    validator_id,
-                    "Could not find peer ID for validator in DHT. Performing a DHT query",
-                );
                 true
             }
         };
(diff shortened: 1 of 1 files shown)


validator-network/src/network_impl.rs
```diff
@@ -195,9 +195,10 @@ where
                 if let Some(validator_key) =
                     self.validator_keys.read().get(usize::from(validator_id))
                 {
-                    self.validator_peer_id_cache
-                        .write()
-                        .remove(validator_key.compressed());
+                    self.validator_peer_id_cache.write().insert(
+                        validator_key.compressed().clone(),
+                        CacheState::Error(NetworkError::UnknownValidator(validator_id)),
+                    );
```
hrxi: This should probably only be done if this invalidates the cache entry that led to the error. Would that be possible?

jsdanielh: Not sure if I'm following you, this is trying to invalidate the cache entry that led to the error

hrxi: I'm imagining the following scenario:

The current cache entry of validator 1 is invalid. Two requests to that are currently in flight. The first one fails and clears the cache entry. Another request is sent, causing the cache to be updated to the correct value. The second initial request fails, and now clears the cache value even though it's already been updated.

jsdanielh: Yeah, thanks! Updated PR to incorporate the change we discussed

hrxi (approved):
Looks good to me. :)

PR #2262, opened 2024-02-28
Store the whole clientDataJSON, but compressed

hrxi:
This makes reconstruction trivial. Previously, reconstruction of the original JSON was a delicate process, even more so since Android Chrome's behavior violates the standard: https://issues.chromium.org/issues/40191627.

Compressing is done based on expected data, this allows to save a lot of space for the redundant challenge information and the general JSON structure.

This is an alternative to #2227, also fixing https://github.com/nimiq/core-rs-albatross/pull/1867#discussion_r1470116959.

PR #2275, opened 2024-03-01
Improve hand-parsing of clientDataJSON

hrxi:
Try to be very precise while doing so. This allows us to work around Android Chrome's standard-violating behavior: https://issues.chromium.org/issues/40191627.

Alternative to #2262 and #2227.

### DIFF
--- a/primitives/transaction/Cargo.toml
+++ b/primitives/transaction/Cargo.toml
@@ -21,10 +21,11 @@ base64 = "0.22"
 bitflags = { version = "2.4", features = ["serde"] }
 log = { package = "tracing", version = "0.1", features = ["log"] }
 serde = "1.0"
+serde_json = "1.0"
 strum_macros = "0.26"
 thiserror = "1.0"
 tsify = { git = "https://github.com/sisou/tsify", branch = "sisou/comments", default-features = false, features = ["js"], optional = true }
-url = "2.4"
+url = { version = "2.4", features = ["serde"] }
 wasm-bindgen = { version = "0.2", optional = true }
 
 nimiq-bls = { workspace = true, features = ["serde-derive"] }
--- a/primitives/transaction/src/signature_proof.rs
+++ b/primitives/transaction/src/signature_proof.rs
@@ -1,8 +1,8 @@
-use std::{cmp::Ord, error::Error, fmt};
+use std::{cmp::Ord, error::Error, fmt, str};
 
-use base64::Engine;
+use base64::prelude::{Engine, BASE64_URL_SAFE_NO_PAD};
 use bitflags::bitflags;
-use nimiq_hash::{Blake2bHasher, Hasher, Sha256Hasher};
+use nimiq_hash::{Blake2bHash, Hash, HashOutput, Sha256Hash};
 use nimiq_keys::{Address, Ed25519PublicKey, Ed25519Signature, PublicKey, Signature};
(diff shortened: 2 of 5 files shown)


web-client/src/primitives/signature_proof.rs
```diff
@@ -33,7 +33,7 @@ impl SignatureProof {
         public_key: &PublicKeyUnion,
         signature: &SignatureUnion,
         authenticator_data: &[u8],
-        client_data_json: &[u8],
+        client_data_json: &str,
```
sisou: The reason this was typed as `&[u8]` and not string is that the browser credentials API provides it as a Uint8Array. So with an u8 slice, it can be directly handed in without userland conversion.

Please keep that if possible.

hrxi: Makes sense, will do so.

hrxi: Done.

primitives/transaction/Cargo.toml
```diff
@@ -21,10 +21,11 @@ base64 = "0.21"
 bitflags = { version = "2.4", features = ["serde"] }
 log = { package = "tracing", version = "0.1", features = ["log"] }
 serde = "1.0"
+serde_json = "1.0"
 strum_macros = "0.26"
 thiserror = "1.0"
 tsify = { git = "https://github.com/sisou/tsify", branch = "sisou/comments", default-features = false, features = ["js"], optional = true }
-url = "2.4"
+url = { version = "2.4", features = ["serde"] }
```
jsdanielh, line 28: Didn't this update the `Cargo.lock`?

primitives/transaction/src/signature_proof.rs
```diff
@@ -332,10 +222,105 @@ bitflags! {
 
 #[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub struct WebauthnExtraFields {
```
jsdanielh, line 225: Could you add rustdoc to the `struct` and its members?

primitives/transaction/src/signature_proof.rs
```diff
@@ -154,87 +97,34 @@ impl SignatureProof {
...
-            r":\/\/"
-        } else {
-            "://"
+        let challenge: Blake2bHash = message.hash();
+
+        // 2. The RP ID is the SHA256 hash of the hostname
+        let rp_id = match webauthn_fields.rp_id() {
+            Ok(rp_id) => rp_id,
+            Err(e) => {
+                debug!("Failed to extract PR ID: {e}");
```
jsdanielh:
```suggestion
                debug!(error=e, "Failed to extract PR ID");
```

sisou: And it's `RP` ID, as in Relaying Party.

primitives/transaction/Cargo.toml
hrxi: No.

primitives/transaction/src/signature_proof.rs
hrxi: Done.

primitives/transaction/src/signature_proof.rs
hrxi: Both fixed.

jsdanielh (approved):
LGTM

jsdanielh:
Rebasing it to merge it

PR #2277, opened 2024-03-04
Set default desired peer count to 12

Eligioo:

Set the default value for the network desired number of peers to 12.
In the devnet scenarios the value is set to `min_peers` to keep previous behavior for testing to avoid several dial attempts when a node is taken down.

## Pull request checklist

### DIFF
--- a/lib/src/client.rs
+++ b/lib/src/client.rs
@@ -297,10 +297,7 @@ impl ClientInner {
             false,
             required_services,
             tls_config,
-            config
-                .network
-                .desired_peer_count
-                .unwrap_or(config.consensus.min_peers),
+            config.network.desired_peer_count,
             config.network.autonat_allow_non_global_ips,
             config.network.only_secure_ws_connections,
             config.network.allow_loopback_addresses,
--- a/lib/src/config/config.rs
+++ b/lib/src/config/config.rs
@@ -122,8 +122,8 @@ pub struct NetworkConfig {
 
     /// Optional desired number of peers for the network to connect to.
     /// The network will always try to maintain this number of connections.
-    #[builder(default)]
-    pub desired_peer_count: Option<usize>,
+    #[builder(default = "12")]
+    pub desired_peer_count: usize,
 
     /// Optional setting to allow network autonat to use non global IPs
     #[builder(default)]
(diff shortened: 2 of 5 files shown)


web-client/src/client/lib.rs
```diff
@@ -157,6 +157,7 @@ impl Client {
         config.network.seeds = seed_nodes;
         config.network.only_secure_ws_connections = true;
         config.network_id = web_config.network_id;
+        config.network.desired_peer_count = Some(8);
```
sisou: Where does `8` come from? Just a random number?

Eligioo: For the most part yes. I've tested with various numbers and noticed that at some point, you won't notices a difference anymore. Since the amount of data a light client needs is relatively low, around 8 is a nice number whereas 10+ doesn't seem necessary. Going lower than 5, I had encounters where the consensus never got reached and the client kept syncing but always blocks behind. Also for Nimiq 1.0, the client roughly connects with the same amount of peers.

sisou: This answer is to my satisfaction :+1:

sisou (approved):

sisou:
Here is a question to @jsdanielh though: Why does `desired_peer_count` fall back to `min_peer_count` in the first place. Those two have nothing in common in meaning. I think that is the root cause for this failure, and also affects all other node types (I think it even already came up in the Telegram dev channel).

sisou:
The default should be 12 when not specified otherwise, so why is this number not used for the light-client?
https://github.com/nimiq/core-rs-albatross/pull/2038/files#diff-19108d6db09124d07b1f465ebe19a763861d5fc69d7b490aaa23c3e963f83b28R78

jsdanielh:
> The default should be 12 when not specified otherwise, so why is this number not used for the light-client? https://github.com/nimiq/core-rs-albatross/pull/2038/files#diff-19108d6db09124d07b1f465ebe19a763861d5fc69d7b490aaa23c3e963f83b28R78

This is a different default that is defined internally in the network. The final network configuration is defined by the `Client`

jsdanielh:
Per our discussion today, we're going to fix this differently.

Eligioo (commented):
@jsdanielh LGTM

Issue #2289, opened 2024-03-06
Light client - rejects block on live sync push event


Eligioo:
## New issue checklist

- [x] I have read all of the [`README`](https://github.com/nimiq-network/core/blob/master/README.md) 
- [x] I have searched [existing issues](https://github.com/nimiq-network/core/issues?q=is%3Aissue+sort%3Acreated-desc) and **this is not a duplicate**.

### General information

- Library version(s): 0.20
- Reproducible in the testnet? (Yes/No): Yes

## Bug report
It has been observed that running a light client for some time, causes the node to reject a block at some point. This prevents the node from accepting new blocks and thus falls behind. It is noteworthy that all the rejected blocks are **macro** blocks so far.

The node doesn't lose consensus as it keeps receiving gossipsubs. Eventually the block queue notices that the gossipsubs are to too far ahead (head_height + window_ahead_max) and starts a macro sync. Which brings the node back to the actual head of the chain and can from there on accept new blocks properly. Until the node rejects a block again and then this cycle repeats itself.

Although the blocks are rejected, they actually end up on the blockchain. Here are some example hashes which can be found in the attached log file:
```
d9da6da5af60ec28c25e36f3d56e31f51121e0818ffc8aa2ca0bf82e9128d6b4
f2b1aca4f451d61c87e8df38ca8a986c512896db2d8ef0f6037a5dc2d53c298f
dd317a6905bfefc758c8a6108e1d66f7f4e58c8c16b1741f251e51a62dcad284
```

#### Expected behavior
Be able to push and accept newly received blocks

#### Actual behavior
Rejecting some blocks.
Reason: `Head push operation failed because of Invalid block: Missing body`

#### Steps to reproduce
Set the sync_mode to light in the client.toml and run the client.

#### Crash log? Screenshots? Videos? Sample project?

Search the log file for `Rejected block`
[light-client-rejects-live-push-event.log](https://github.com/nimiq/core-rs-albatross/files/14508410/light-client-rejects-live-push-event.log)

sisou:
Unfortunately no reason is given in the WARN message. Could you enable TRACE logs and see if any reason is logged?

Eligioo:
Head push operation failed because of Invalid block: Missing body

sisou:
[...]

PR #2291, opened 2024-03-06
Check that the peer sends us a body exactly when we ask for it

hrxi:
Probably helps with #2289.

Also validate gossipsub messages for blocks we already know.

### DIFF
--- a/consensus/src/sync/live/block_queue/mod.rs
+++ b/consensus/src/sync/live/block_queue/mod.rs
@@ -178,7 +178,15 @@ impl<N: Network> BlockQueue<N> {
         let head_height = blockchain.block_number();
 
         // Ignore blocks that we already know.
-        if blockchain.contains(&block.hash(), true) {
+        if let Ok(info) = blockchain.get_chain_info(&block.hash(), false) {
+            self.report_validation_result(
+                pubsub_id,
+                if info.on_main_chain {
+                    MsgAcceptance::Accept
+                } else {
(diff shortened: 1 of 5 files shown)


consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -162,13 +162,30 @@ impl<N: Network> BlockQueue<N> {
...
         let block_number = block.block_number();
         let head_height = blockchain.block_number();
 
         // Ignore blocks that we already know.
-        if blockchain.contains(&block.hash(), true) {
+        if let Ok(info) = blockchain.get_chain_info(&block.hash(), false) {
+            self.report_validation_result(
+                pubsub_id,
+                if info.on_main_chain {
+                    MsgAcceptance::Accept
```
jsdanielh, line 185: To me if the block is already known, we should use `Ignore`: Gossipsub is not supposed to send duplicate messages and an `Accept` will forward the message.

hrxi: I think gossipsub deduplicates the message already. Since this is the first time that we see this message in gossipsub, we should probably forward it, even if we got it by other means. … I'd think, at least.

jsdanielh: Ok, I see your point. Let's use accept then

jsdanielh (approved):
Could you rebase the branch?

PR #2292, opened 2024-03-06
Move validity target as soon as we have a new macro

viquezclaudio:
## Pull request checklist


jsdanielh (commented):
Mmm, given this logic I see that:
We could enter into `if latest_macro_head_number > verifier_block_number`, then have `epoch_complete` and not have an election in window and then we would be done?

jsdanielh:
Per our discussion we will be merging this as it is since it addresses the majority of the cases. We will do a patch release with this change and #2293 documents the missing work on this issue.

Issue #2301, opened 2024-03-08
Upstreaming libp2p broke NodeJS support of the web-client


sisou:
## General information

- Library version(s): >= 0.19.0
- Browser version(s): NodeJS
- Devices/Simulators/Machine affected: all
- Reproducible in the testnet? (Yes/No): yes
- Related issues: https://github.com/libp2p/rust-libp2p/pull/4889, #2047

#### Expected behavior

NodeJS can run the client in a `worker_thread`, for which I created a [special entry point](https://github.com/nimiq/core-rs-albatross/blob/albatross/web-client/dist/nodejs/worker.js).

#### Actual behavior

The client in a NodeJS worker thread panics with `'to have a window or worker context'`, which is the exact panic added in https://github.com/libp2p/rust-libp2p/pull/4889.

#### Steps to reproduce

Run the NodeJS example in the `web-client/example/node`. Make sure to follow the instructions in `web-client/example/README.md`.

#### Crash log? Screenshots? Videos? Sample project?

![image](https://github.com/nimiq/core-rs-albatross/assets/1828163/fe105202-3dc4-4a23-824a-40bc1ce7a27d)

```
2024-03-08T20:11:56.006000128Z ERROR panic                | thread '<unnamed>' panicked at 'to have a window or worker context': /home/soeren/.cargo/registry/src/index.crates.io-6f17d22bba15001f/libp2p-websocket-websys-0.3.1/src/lib.rs:308 
```

## Question or Feature Request

Add support for [NodeJS Worker Threads](https://nodejs.org/api/worker_threads.html) to libp2p's websocket-websys transport?

I need to find out how it worked before libp2p was upstreamed in #2047, as it _did_ work before, when I added NodeJS support and the example.

sisou:
Before the change to `WindowOrWorkerContext`, the websocket-websys transport was patched to bind to global `setTimeout` and `clearTimeout` directly, without having to have `web-sys`'s `Window` or `WorkerContext`: https://github.com/jsdanielh/rust-libp2p/commit/34f65f1c2d670be02a572412bf90a093a04835af

sisou:
I'm able to pass the first initialization check of the client by mocking `WorkerGlobalScope` in the NodeJS worker thread:

```js
global.WorkerGlobalScope = global;
```

[...]

PR #2302, opened 2024-03-11
Fix Nodejs support for web-client

Eligioo:
This PR is a workaround of the upstream libp2p that broke the support for NodeJS: #2301.

- it sets the `global.WorkerGlobalScope` as a reference to itself which libp2p requires.
- it cleans up performance measurements periodically emitted by the logs that currently cause the memory to grow indefinitely.
- It catches the exception of a failing websocket connetion without exiting the worker thread.

## Pull request checklist

### DIFF
--- a/web-client/dist/nodejs/worker.js
+++ b/web-client/dist/nodejs/worker.js
@@ -9,6 +9,24 @@ const { Client } = require('./worker-wasm/index.js');
 
 // Provide a global WebSocket implementation, which is expected by the WASM code built for browsers.
 global.WebSocket = w3cwebsocket;
+// Workaround for Node.js as it currently lacks support for Web Workers by pretending there is
+// a WorkerGlobalScope object available which is checked within the libp2p's websocket-websys transport.
+global.WorkerGlobalScope = global;
+
+// Clean up performance measurements created by tracing.
+// TODO: find a solution to disable creating these measurements.
+setInterval(() => {
+    global.performance.clearMarks();
+    global.performance.clearMeasures();
+    global.performance.clearResourceTimings();
+}, 300 * 1000);
+
+// Prevent NodeJS exiting on an uncaught exception that we currently expect,
+// until it is fixed in upstream libp2p websocket-websys transport.
+process.on('uncaughtException', error => {
+    if (error.message.includes('closure invoked recursively')) return;
+    throw error;
+});
 
 // Defined both here and in main thread exports.js
 Comlink.transferHandlers.set('function', {
--- a/web-client/dist/nodejs/worker.mjs
+++ b/web-client/dist/nodejs/worker.mjs
@@ -9,6 +9,24 @@ import wasm from './worker-wasm/index.js';
 
 // Provide a global WebSocket implementation, which is expected by the WASM code built for browsers.
 global.WebSocket = websocket.w3cwebsocket;
+// Workaround for Node.js as it currently lacks support for Web Workers by pretending there is
+// a WorkerGlobalScope object available which is checked within the libp2p's websocket-websys transport.
+global.WorkerGlobalScope = global;
+
+// Clean up performance measurements created by tracing.
+// TODO: find a solution to disable creating these measurements.
+setInterval(() => {
+    global.performance.clearMarks();
+    global.performance.clearMeasures();
+    global.performance.clearResourceTimings();
+}, 300 * 1000);
+
+// Prevent NodeJS exiting on an uncaught exception that we currently expect,
+// until it is fixed in upstream libp2p websocket-websys transport.
+process.on('uncaughtException', error => {
+    if (error.message.includes('closure invoked recursively')) return;
+    throw error;
+});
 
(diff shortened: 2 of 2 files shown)


web-client/dist/nodejs/worker.js
```diff
@@ -9,6 +9,26 @@ const { Client } = require('./worker-wasm/index.js');
 
 // Provide a global WebSocket implementation, which is expected by the WASM code built for browsers.
 global.WebSocket = w3cwebsocket;
+// Provide a reference to the global object necessary for the libp2p2 websocket transport.
```
sisou:
```suggestion
// Workaround for libp2p's websocket-websys transport's worker detection,
// which checks for web workers' `WorkerGlobalScope` interface.
```

web-client/dist/nodejs/worker.js
```diff
@@ -9,6 +9,26 @@ const { Client } = require('./worker-wasm/index.js');
...
+global.WorkerGlobalScope = global;
+
+setInterval(() => {
+    // Clean up performance measurements created by logging.
+    // TODO: find a solution to disable creating these measurements.
+    global.performance.clearMarks();
```
sisou:
```suggestion

// Clean up performance measurements created by tracing.
// TODO: find a solution to disable creating these measurements.
setInterval(() => {
    global.performance.clearMarks();
```

web-client/dist/nodejs/worker.js
```diff
@@ -9,6 +9,26 @@ const { Client } = require('./worker-wasm/index.js');
...
+
+process.on('uncaughtException', reason => {
+    // Temporarily catch this error to prevent an exit of the worker.
+    // TODO: properly catch this exception.
+    if (reason.message.includes('closure invoked recursively')) {
+        console.error('Error connecting with peer.');
+        return;
+    }
+    throw reason;
+});
```
sisou:
```suggestion

// Prevent NodeJS exiting on an uncaught exception that we currently expect,
// until it is fixed in upstream libp2p websocket-websys transport.
process.on('uncaughtException', error => {
    if (error.message.includes('closure invoked recursively')) return;
    throw error;
});
```

sisou (changes requested):
Please also make the same additions to `worker.mjs` in the same folder.

sisou (approved):

hrxi:
Do we have an upstream fix as well?

sisou:
We do not, we don't even have an issue yet. We want to create a minimum reproduction first. This is a brute-force fix: https://github.com/sisou/rust-libp2p/commit/0886623409f553cd2a2da5b6fe6e252d10abc788, which could be a guiding light for a real fix.

web-client/dist/nodejs/worker.js
```diff
@@ -9,6 +9,24 @@ const { Client } = require('./worker-wasm/index.js');
 
 // Provide a global WebSocket implementation, which is expected by the WASM code built for browsers.
 global.WebSocket = w3cwebsocket;
+// Workaround for libp2p's websocket-websys transport's worker detection,
+// which checks for web workers' `WorkerGlobalScope` interface.
```
jsdanielh: Per our discussion it seems like this is more a workaround to `nodejs` not supporting the `web-api` rather than a workaround to libp2p's websocket-websys transport's worker detection.

web-client/dist/nodejs/worker.mjs
```diff
@@ -9,6 +9,24 @@ import wasm from './worker-wasm/index.js';
 
 // Provide a global WebSocket implementation, which is expected by the WASM code built for browsers.
 global.WebSocket = websocket.w3cwebsocket;
+// Workaround for libp2p's websocket-websys transport's worker detection,
+// which checks for web workers' `WorkerGlobalScope` interface.
```
jsdanielh, line 13: Per our discussion it seems like this is more a workaround to `nodejs` not supporting the `web-api` rather than a workaround to libp2p's websocket-websys transport's worker detection.

web-client/dist/nodejs/worker.js
sisou: As an alternative to the `WorkerGlobalScope` workaround, we could also look into using https://github.com/developit/web-worker to polyfill NodeJS's `worker_threads` to become Web Workers.

sisou: But if the workaround necessary for libp2p is _only_ this assignment, then a whole other dependency seems like overkill.

But it seems your review comment is only about the describing code comment, not the mocking of `WorkerGlobalScope` itself, right?

jsdanielh: > But if the workaround necessary for libp2p is _only_ this assignment, then a whole other dependency seems like overkill.
> 
> But it seems your review comment is only about the describing code comment, not the mocking of `WorkerGlobalScope` itself, right?

Exactly

jsdanielh (approved):

jsdanielh:
Rebasing branch to merge it

Issue #2294, opened 2024-03-07
WARN & ERROR in network_impl and tendermint


joeleol:
## New issue checklist

- [x] I have read all of the [`README`](https://github.com/nimiq-network/core/blob/master/README.md) 
- [x] I have searched [existing issues](https://github.com/nimiq-network/core/issues?q=is%3Aissue+sort%3Acreated-desc) and **this is not a duplicate**.

### General information

- Library version(s): v0.20.3
- Browser version(s): n/a
- Devices/Simulators/Machine affected: my validator
- Reproducible in the testnet? (Yes/No): yes
- Related issues:

#### Crash log? Screenshots? Videos? Sample project?

<img width="1058" alt="Screenshot 2024-03-07 at 09 12 17" src="https://github.com/nimiq/core-rs-albatross/assets/64198137/14c0a729-78e7-46e2-9dd7-d25998059478">

## Question or Feature Request

> Is this normal? I did not see it before in my logs...

hrxi:
Does this happen in the beginning only? Or does it happen for a prolonged period of time?

joeleol:
This is happening since the start of the service, it's still happening now, but slightly reduced.

PR #2310, opened 2024-03-14
Don't forget validator IDs immediately when a message fails to send

hrxi:
Touches upon #2294 but does not address the root cause.

### DIFF
--- a/validator-network/src/network_impl.rs
+++ b/validator-network/src/network_impl.rs
@@ -16,13 +16,37 @@ use super::{MessageStream, NetworkError, PubsubId, ValidatorNetwork};
 use crate::validator_record::ValidatorRecord;
 
 /// Validator `PeerId` cache state
-enum CacheState<TPeerId, TNetworkError> {
+#[derive(Clone, Copy)]
+enum CacheState<TPeerId> {
     /// Cache entry has been resolved with the peer ID
     Resolved(TPeerId),
-    /// Cache entry could not have been resolved
-    Error(TNetworkError),
-    /// Cache entry resolution is in progress (and result is yet unknown)
-    InProgress,
+    /// Cache entry could not have been resolved.
+    ///
+    /// We might know a previous peer ID.
+    Error(Option<TPeerId>),
+    /// Cache entry resolution is in progress (and result is yet unknown).
+    ///
+    /// We might know a previous peer ID.
+    InProgress(Option<TPeerId>),
+    /// No cached peer ID, but a previous one is known.
+    Empty(TPeerId),
+}
+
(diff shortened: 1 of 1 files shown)


validator-network/src/network_impl.rs
```diff
@@ -90,102 +113,105 @@ where
...
+            .get_mut(public_key.compressed())
+        {
+            Some(cache_entry) => {
+                if let CacheState::InProgress(prev_peer_id) = *cache_entry {
+                    *cache_entry = match cache_value {
+                        Ok(peer_id) => CacheState::Resolved(peer_id),
+                        Err(()) => CacheState::Error(prev_peer_id),
+                    };
+                } else {
+                    unreachable!("cache state must be \"in progress\"");
```
jsdanielh, line 156: I'd add a comment that this function assumes that the cache entry is updated with an `InProgress` entry before.

validator-network/src/network_impl.rs
```diff
@@ -90,102 +113,105 @@ where
...
-                        .insert(public_key.compressed().clone(), CacheState::InProgress);
-                    log::debug!(
-                        ?public_key,
-                        validator_id,
-                        "Could not find peer ID for validator in DHT. Performing a DHT query",
-                    );
-                }
+                    CacheState::Empty(prev_peer_id) => CacheState::InProgress(Some(prev_peer_id)),
+                };
+                *cache_state = new_cache_state.clone();
```
jsdanielh: Clippy warning:
```suggestion
                *cache_state = new_cache_state;
```

validator-network/src/network_impl.rs
```diff
@@ -90,102 +113,105 @@ where
...
+            } else {
+                new_cache_state = CacheState::InProgress(None);
+                // No cache entry for this validator ID: we are going to perform the DHT query
+                validator_peer_id_cache
+                    .insert(public_key.compressed().clone(), new_cache_state.clone());
```
jsdanielh: Clippy warning:
```suggestion
                    .insert(public_key.compressed().clone(), new_cache_state);
```

validator-network/src/network_impl.rs
hrxi: Done.

validator-network/src/network_impl.rs
hrxi: Fixed.

validator-network/src/network_impl.rs
hrxi: Fixed.

jsdanielh (approved):

PR #2317, opened 2024-03-15
Move Rustdocs from rpc-server to rpc-interface for trait methods

Eligioo:
Move all the Rustdocs from the `rpc-server` methods tot he `rpc-interface` trait methods. 

## Pull request checklist

### DIFF
--- a/rpc-interface/src/consensus.rs
+++ b/rpc-interface/src/consensus.rs
@@ -11,20 +11,24 @@ use crate::types::{RPCResult, Transaction, ValidityStartHeight};
 pub trait ConsensusInterface {
     type Error;
 
+    /// Returns a boolean specifying if we have established consensus with the network.
     // `nimiq_jsonrpc_derive::proxy` requires the receiver type to be a mutable reference.
     #[allow(clippy::wrong_self_convention)]
     async fn is_consensus_established(&mut self) -> RPCResult<bool, (), Self::Error>;
 
+    /// Given a serialized transaction, it will return the corresponding transaction struct.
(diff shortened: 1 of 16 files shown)


rpc-interface/src/consensus.rs
```diff
@@ -359,6 +435,13 @@ pub trait ConsensusInterface {
         validity_start_height: ValidityStartHeight,
     ) -> RPCResult<String, (), Self::Error>;
 
+    /// Sends a `update_validator` transaction to the network. You need to provide the address of a basic
+    /// account (the sender wallet) to pay the transaction fee.
+    ///  Since JSON doesn't have a primitive for Option (it just has the null primitive), we can't
```
jsdanielh:
```suggestion
    /// Since JSON doesn't have a primitive for Option (it just has the null primitive), we can't
```

rpc-interface/src/consensus.rs
```diff
@@ -359,6 +435,13 @@ pub trait ConsensusInterface {
...
+    /// have a double Option. So we use the following work-around for the signal data:
+    ///  null = No change in the signal data field.
+    ///  "" = Change the signal data field to None.
+    ///  "0x29a4b..." = Change the signal data field to Some(0x29a4b...).
```
jsdanielh:
```suggestion
    /// null = No change in the signal data field.
    /// "" = Change the signal data field to None.
    /// "0x29a4b..." = Change the signal data field to Some(0x29a4b...).
```

jsdanielh (approved):

PR #2318, opened 2024-03-16
Add own abstraction layer over tokio::time::Interval

sisou:
Using `gloo-timers` for WASM. This change gets rid of the unmaintained `wasm-timer` crate, of which we were already using a fork with a fix from me anyway.

The resulting abstraction layer is very thin and since it exposes the exact same interface for both native and WASM targets, chunks of code that concerned themselves with the differentiation can be removed.

## Pull request checklist

### DIFF
--- a/handel/Cargo.toml
+++ b/handel/Cargo.toml
@@ -30,9 +30,7 @@ nimiq-collections = { workspace = true }
 nimiq-hash = { workspace = true }
 nimiq-serde = { workspace = true }
 nimiq-time = { workspace = true }
-nimiq-utils = { workspace = true, features = [
-    "math",
-] }
+nimiq-utils = { workspace = true, features = ["math"] }
 
 [dev-dependencies]
 nimiq-network-interface = { workspace = true }
@@ -40,3 +38,4 @@ nimiq-network-mock = { workspace = true }
 nimiq-test-log = { workspace = true }
 
 tokio = { version = "1.37", features = ["rt", "time", "macros"] }
+nimiq-utils = { workspace = true, features = ["math"] }
--- a/handel/Cargo.toml
+++ b/handel/Cargo.toml
@@ -17,6 +17,7 @@ workspace = true
 async-trait = "0.1"
 futures = { workspace = true }
 log = { workspace = true }
+instant = { version = "0.1", features = ["wasm-bindgen"] }
 parking_lot = "0.12"
 rand = "0.8"
 serde = "1.0"
@@ -28,6 +29,7 @@ nimiq-bls = { workspace = true }
 nimiq-collections = { workspace = true }
 nimiq-hash = { workspace = true }
 nimiq-serde = { workspace = true }
+nimiq-time = { workspace = true }
 nimiq-utils = { workspace = true, features = [
     "math",
 ] }
--- a/handel/tests/mod.rs
+++ b/handel/tests/mod.rs
@@ -286,11 +286,7 @@ async fn it_can_aggregate() {
     );
 
     // aggregating should not take more than 300 ms per each 7 contributors
-    let timeout_ms = 300u64 * (contributor_num / 7 + 1) as u64;
-
-    let deadline = tokio::time::Instant::now()
-        .checked_add(Duration::from_millis(timeout_ms))
-        .unwrap();
+    let timeout = Duration::from_millis(300u64 * (contributor_num / 7 + 1) as u64);
 
     // The final value needs to be the sum of all contributions.
     // For instance for `contributor_num = 7: 8 + 7 + 6 + 5 + 4 + 3 + 2 + 1 = 36`
@@ -300,7 +296,7 @@ async fn it_can_aggregate() {
     }
 
     loop {
-        match tokio::time::timeout_at(deadline, aggregation.next()).await {
+        match nimiq_time::timeout(timeout, aggregation.next()).await {
             Ok(Some(aggregate)) => {
                 if aggregate.num_contributors() == contributor_num + 1
                     && aggregate.value == exp_agg_value
@@ -318,7 +314,7 @@ async fn it_can_aggregate() {
     net.disconnect();
 
     // give the other aggregations time to complete themselves
-    tokio::time::sleep(Duration::from_millis(100)).await;
+    nimiq_time::sleep(Duration::from_millis(100)).await;
 
     // after we have the final aggregate create a new instance and have it (without any other instances)
     // return a fully aggregated contribution and terminate.
--- a/handel/tests/mod.rs
+++ b/handel/tests/mod.rs
@@ -5,6 +5,7 @@ use futures::{
     future::{BoxFuture, FutureExt},
     stream::StreamExt,
 };
+use instant::Instant;
 use nimiq_bls::PublicKey;
 use nimiq_collections::bitset::BitSet;
 use nimiq_handel::{
@@ -286,7 +287,11 @@ async fn it_can_aggregate() {
     );
 
     // aggregating should not take more than 300 ms per each 7 contributors
-    let timeout = Duration::from_millis(300u64 * (contributor_num / 7 + 1) as u64);
+    let timeout_ms = 300u64 * (contributor_num / 7 + 1) as u64;
+
+    let deadline = Instant::now()
+        .checked_add(Duration::from_millis(timeout_ms))
+        .unwrap();
 
     // The final value needs to be the sum of all contributions.
     // For instance for `contributor_num = 7: 8 + 7 + 6 + 5 + 4 + 3 + 2 + 1 = 36`
@@ -296,7 +301,12 @@ async fn it_can_aggregate() {
     }
 
     loop {
-        match nimiq_time::timeout(timeout, aggregation.next()).await {
+        match nimiq_time::timeout(
+            deadline.saturating_duration_since(Instant::now()),
+            aggregation.next(),
+        )
+        .await
+        {
             Ok(Some(aggregate)) => {
                 if aggregate.num_contributors() == contributor_num + 1
                     && aggregate.value == exp_agg_value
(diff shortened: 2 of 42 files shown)


sisou:
I looked at the failing test `it_can_limit_requests_rate`: Before my force-push, which fixed most intervals to the previous behavior of having their first tick only after the first duration, it failed at the first limit reset already with a `ConnectionClosed` error. After, it only fails at the third limit reset. However, it's still not failing with limit-reset related problem, but because of a disconnecting peer.

I did not yet investigate the other failing test `it_can_reset_requests_rate_with_reconnections`.

styppo (approved):

sisou:
@styppo you are ignoring two request-response tests in your commits here. Is there no way to enable them again? Or will you create a tracking issue for them?

handel/tests/mod.rs
```diff
@@ -300,7 +296,7 @@ async fn it_can_aggregate() {
     }
 
     loop {
-        match tokio::time::timeout_at(deadline, aggregation.next()).await {
+        match nimiq_time::timeout(timeout, aggregation.next()).await {
```
hrxi: This changes the timeout logic from 300 ms \* contributors / 7 for the entire thing to 300 ms \* contributors / 7 per loop.

network-libp2p/src/network.rs
```diff
@@ -1717,15 +1635,28 @@ impl Network {
...
                 }
+            };
+
+            // let timeout = Req::TIME_WINDOW.mul_f32(1.5f32);
+            // let future = response_rx.map_err(std::io::Error::other);
+            //match wasm_timer::TryFutureExt::timeout(future, timeout).await {
+            match result {
+                Err(_) => Err(RequestError::OutboundRequest(
+                    OutboundRequestError::SenderFutureDropped,
+                )),
```
hrxi: Could you add a `warn!` here so the condition could be more easily noticed in logs?

time/src/gloo.rs
```diff
@@ -0,0 +1,43 @@
+use std::{convert::TryInto, future::Future, pin::pin, time::Duration};
+
+use futures::future::{select, Either};
+use gloo_timers::future::{IntervalStream, TimeoutFuture};
+use send_wrapper::SendWrapper;
+
+pub type Interval = SendWrapper<IntervalStream>;
```
hrxi, line 7: From the [`send_wrapper` docs](https://docs.rs/send_wrapper/0.6.0/send_wrapper/index.html):
> This Rust library implements a wrapper type called `SendWrapper` which allows you to move around non-`Send` types between threads, as long as you access the contained value only from within the original thread.

This means we're locking our wasm implementation to one thread. I guess that's okay.

time/src/tokio.rs
```diff
@@ -0,0 +1,22 @@
...
+use tokio::time::{
+    interval_at, sleep as tokio_sleep, timeout as tokio_timeout, Instant, Sleep, Timeout,
+};
+pub use tokio_stream::wrappers::IntervalStream as Interval;
+
+pub fn interval(period: Duration) -> Interval {
+    assert!(
+        period.as_millis() <= u32::MAX as u128,
+        "Period as millis must fit in u32"
+    );
```
hrxi, line 14: A comment why this is done would be good. I assume it's due to the restriction on [`gloo`'s `sleep` function](https://docs.rs/gloo/0.11.0/gloo/timers/future/fn.sleep.html):

> This function will panic if the specified `Duration` cannot be casted into a `u32` in milliseconds.

web-client/tests/wasm.rs
```diff
@@ -66,7 +66,9 @@ pub async fn it_can_initialize_with_mock_network() {
         }),
     );
 
-    spawn_local(consensus);
```
hrxi, line 70: Why was this removed?

hrxi (commented):
Found a small bug in test code. Plus, as @sisou mentions, tests seem to be ignored.

web-client/tests/wasm.rs
sisou: Because the NodeJS test runner was not exiting after the test completed successfully, because of pending timeouts or intervals. We tried to work around it by not spawning the consensus, but only polling it once (the added lines after this removed line), but that didn't work really either.

I have now changed the WASM tests to run in headless Chrome and there the test runner exits correctly, so I have added this back in.

handel/tests/mod.rs
sisou: Do you mean the variable was copied wrong (`deadline` => `timeout`), or the new `timeout` method behaves differently than `timeout_at` did?

hrxi: I think I mean neither.

Previously, the `timeout_at` function made sure that the entire loop did not take longer than `timeout_ms`. Now, the usage of the `timeout` function only makes sure that each loop *iteration* takes no longer than `timeout`.

I.e. previously the whole loop was limited in time, now each *iteration* can take as long as the whole loop previously.

Not sure if that made it clearer…

sisou: Yes, I understand now.

OK, so since we don't have a `timeout_at` anymore, we need to make the `timeout` duration dynamic. Calculate it from the difference of `deadline - <now>`. Would that be right?

hrxi: Yes, that would work.

sisou: Done, check out https://github.com/nimiq/core-rs-albatross/pull/2318/commits/a3fb6060120bcb67e01b4e5943f3b471be1462cb

handel/Cargo.toml
```diff
@@ -17,6 +17,7 @@ workspace = true
 async-trait = "0.1"
 futures = { package = "futures-util", version = "0.3", features = ["sink"] }
 log = { package = "tracing", version = "0.1", features = ["log"] }
+instant = { version = "0.1", features = [ "wasm-bindgen" ] }
```
jsdanielh: I'd add the `wasm-bindgen` feature as a feature of this crate:
```
[features]
wasm-bindgen = ["instant/wasm-bindgen"]

File: network-libp2p/src/network.rs [hunk 7]
```diff
@@ -1719,15 +1637,28 @@ impl Network {
...
+                    return Err(RequestError::OutboundRequest(
                         OutboundRequestError::SenderFutureDropped,
-                    ))
+                    ));
                 }
+            };
+
+            // let timeout = Req::TIME_WINDOW.mul_f32(1.5f32);
+            // let future = response_rx.map_err(std::io::Error::other);
+            //match wasm_timer::TryFutureExt::timeout(future, timeout).await {
```
jsdanielh:
```suggestion
```

network-libp2p/Cargo.toml
```diff
@@ -41,14 +41,14 @@ tokio = { version = "1.37", features = ["macros", "rt", "tracing"] }
...
 nimiq-bls = { workspace = true }
 nimiq-macros = { workspace = true }
 nimiq-network-interface = { workspace = true }
 nimiq-primitives = { workspace = true, features = ["policy"] }
 nimiq-hash = { workspace = true }
 nimiq-serde = { workspace = true }
+nimiq-time = { workspace = true }
 nimiq-utils = { workspace = true, features = [
     "tagged-signing",
     "libp2p",
```
jsdanielh, line 53: Looks like we could remove the `tokio-time` feature.

network-libp2p/tests/request_response.rs
```diff
@@ -605,6 +604,7 @@ async fn send_n_request_to_fail(net1: &Arc<Network>, net2: &Arc<Network>, n: u32
 
 #[cfg(feature = "tokio-time")]
```
jsdanielh: Is this still needed?

network-libp2p/tests/request_response.rs
```diff
@@ -727,6 +727,7 @@ async fn it_can_limit_requests_rate_after_reconnection() {
...
```
jsdanielh: Is this still needed?

time/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+#[cfg(target_family = "wasm")]
+mod gloo;
+#[cfg(not(target_family = "wasm"))]
+mod tokio;
+
+#[cfg(target_family = "wasm")]
+pub use gloo::*;
+#[cfg(not(target_family = "wasm"))]
```
jsdanielh, line 8: I'd do it differently by putting this into the `Cargo.toml` such that these dependencies get included only depending on the target. For this to work, the `Cargo.toml` should look like this:
```
[dependencies]
[target.'cfg(target_arch = "wasm32")'.dependencies]
gloo-timers = { version = "0.2", features = ["futures"], optional = true }
[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
tokio = { version = "1.36", features = ["time"], optional = true }
[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
tokio-stream = { version = "0.1", features = ["time"], optional = true }
```
Then the code here could just use the features `tokio`, `tokio-stream` and `gloo` for enabling code here.

validator/Cargo.toml
```diff
@@ -21,6 +21,7 @@ workspace = true
 async-trait = "0.1"
 byteorder = "1.5"
 futures = { package = "futures-util", version = "0.3" }
+instant = { version = "0.1", features = [ "wasm-bindgen" ] }
```
jsdanielh: I'd add a feature `wasm-bindgen` that selects this:
```
[features]
wasm-bindgen = ["instant/wasm-bindgen"]
```

sisou:
@styppo you missed a3fb606 when rebasing.

sisou:
Are the wasm test now exiting correctly? Because I also added this commit on the previous branch, which is now no longer here: https://github.com/nimiq/core-rs-albatross/commit/09a9d76d8a169f2fe7e040f24f9993f9f7a90ee3

sisou:
Although it also then needs a change in the workflow:

https://github.com/nimiq/core-rs-albatross/blob/6c05d3779c93c31304061e856c71d2ee1467265c/.github/workflows/build%2Btest.yml#L112

```diff
- wasm-pack test --node
+ wasm-pack test --chrome --headless
```

web-client/tests/wasm.rs
sisou: After the last rebase of this PR by @styppo the commit reverting this is gone. The relevant changes to the test runner are now in PR #2513. Once that is merged, this PR can be rebased on top and this change here reverted.

PR #2343, opened 2024-04-03
As Non light client: broadcast BlockHeaders

### LINKED ISSUE #2288: Light client - not receiving all block headers over gossipsub
## New issue checklist

- [x] I have read all of the [`README`](https://github.com/nimiq-network/core/blob/master/README.md) 
- [x] I have searched [existing issues](https://github.com/nimiq-network/core/issues?q=is%3Aissue+sort%3Acreated-desc) and **this is not a duplicate**.

### General information

- Library version(s): 0.20

## Bug report
With version 0.20 we are currently observing that light clients don't receive all block header messages over gossipsub. This prevents having a smooth way of accepting newly produced blocks as the `block_queue` has to request missing blocks constantly. The client basically isn't aware the chain has moved on until it receives a new gossipsub so it can request missing blocks again.

#### Expected behavior
A light client to receive all the blocker header messages from the its connected peers.

#### Actual behavior
Not receiving all messages

#### Steps to reproduce

Set the `sync_mode` to `light` in the `client.toml` and run the client.

This log contains only log messages of received gossipsub messages

- note that there are gaps
- gossipsubs for the same block is received multiple times by different peers
- old, non-relevant gossipsubs are received when the head of the client is higher

[missing-gossipsubs-filtered.log](https://github.com/nimiq/core-rs-albatross/files/14500367/missing-gossipsubs-filtered.log)
```
2024-03-05T09:21:13.670000128Z DEBUG block_queue | Received block via gossipsub block=#17869370:MI:242edbb2bd peer_id=12D3KooWF4pZSd4wPcTqwhWVCBkV3T4qWdzdQSKpyHE342c64xeA
2024-03-05T09:21:21.659000064Z DEBUG block_queue | Received block via gossipsub block=#17869378:MI:df283ae507 peer_id=12D3KooWF4pZSd4wPcTqwhWVCBkV3T4qWdzdQSKpyHE342c64xeA
2024-03-05T09:21:22.380000000Z DEBUG block_queue | Received block via gossipsub block=#17869378:MI:df283ae507 peer_id=12D3KooWAjftHreEDsNMK3qPBSe63LZcQorv6q8H3FR7CWg57CxV
2024-03-05T09:21:23.656999936Z DEBUG block_queue | Received block via gossipsub block=#17869380:MI:62d7fee408 peer_id=12D3KooWF4pZSd4wPcTqwhWVCBkV3T4qWdzdQSKpyHE342c64xeA
2024-03-05T09:21:31.662000128Z DEBUG block_queue | Received block via gossipsub block=#17869388:MI:32a75bb061 peer_id=12D3KooWF4pZSd4wPcTqwhWVCBkV3T4qWdzdQSKpyHE342c64xeA
2024-03-05T09:21:32.361999872Z DEBUG block_queue | Received block via gossipsub block=#17869388:MI:32a75bb061 peer_id=12D3KooWAjftHreEDsNMK3qPBSe63LZcQorv6q8H3FR7CWg57CxV
2024-03-05T09:21:32.656999936Z DEBUG block_queue | Received block via gossipsub block=#17869389:MI:4efd2076c9 peer_id=12D3KooWF4pZSd4wPcTqwhWVCBkV3T4qWdzdQSKpyHE342c64xeA
2024-03-05T09:21:41.656000000Z DEBUG block_queue | Received block via gossipsub block=#17869398:MI:901c17ad0d peer_id=12D3KooWF4pZSd4wPcTqwhWVCBkV3T4qWdzdQSKpyHE342c64xeA
2024-03-05T09:21:42.379000064Z DEBUG block_queue | Received block via gossipsub block=#17869398:MI:901c17ad0d peer_id=12D3KooWAjftHreEDsNMK3qPBSe63LZcQorv6q8H3FR7CWg57CxV
2024-03-05T09:21:42.648999936Z DEBUG block_queue | Received block via gossipsub block=#17869399:MI:514cdbb604 peer_id=12D3KooWF4pZSd4wPcTqwhWVCBkV3T4qWdzdQSKpyHE342c64xeA
2024-03-05T09:21:47.651000064Z DEBUG block_queue | Received block via gossipsub block=#17869404:MI:edf8016961 peer_id=12D3KooWF4pZSd4wPcTqwhWVCBkV3T4qWdzdQSKpyHE342c64xeA
```

nibhar:
This makes every non light client rebroadcast blocks they get on the block header topic. It is not ideal as it may lead to duplication depending on how pubsup messages are de-duped. 
To my knowledge in this case the messages actual bytes are the distinguishing factor. If that is the case they should all be the same for micro blocks and only differ with macro blocks which have different proofs.

Resolves #2288

### DIFF
--- a/consensus/src/sync/live/block_queue/mod.rs
+++ b/consensus/src/sync/live/block_queue/mod.rs
@@ -4,6 +4,7 @@ use std::{
         hash_map::{Entry as HashMapEntry, HashMap},
         BTreeSet, HashSet,
     },
+    error::Error,
     pin::Pin,
     sync::Arc,
     task::{Context, Poll},
@@ -441,14 +442,57 @@ impl<N: Network> BlockQueue<N> {
         Some(QueuedBlock::Missing(blocks))
     }
 
+    /// Publishes a given Block on the BlockHeaderTopic. It will strip the micro bodies from the blocks if they are present.
+    ///
+    /// ## Panic
+    /// For macro blocks the body must be present and if it is not this function will panic!
+    fn publish_block_header(&self, mut block: Block) {
+        let network = Arc::clone(&self.network);
+
+        tokio::spawn(async move {
+            let block_id = format!("{}", block);
+            log::debug!(block = block_id, "Broadcasting on BlockHeaderTopic",);
+
+            match block {
+                // Remove the body from micro blocks before publishing to the block header topic.
+                Block::Micro(ref mut micro_block) => micro_block.body = None,
+                // Macro blocks must be always sent with body.
+                Block::Macro(ref macro_block) => {
+                    if macro_block.body.is_none() {
+                        panic!("Macro block bodies must exists in order to publish on BlockHeaderTopic")
+                    }
+                }
+            }
+
+            if let Err(e) = network.publish::<BlockHeaderTopic>(block).await {
+                debug!(
+                    block = block_id,
+                    error = &e as &dyn Error,
+                    "Failed to publish block header"
+                );
+            }
+        });
+    }
+
     /// Fetches the relevant blocks for any given `BlockchainEvent`
     fn get_new_blocks_from_blockchain_event(&self, event: BlockchainEvent) -> Vec<Block> {
         // Collect block numbers and hashes of newly added blocks first.
         let mut block_infos = vec![];
 
         match event {
-            BlockchainEvent::Extended(block_hash)
-            | BlockchainEvent::HistoryAdopted(block_hash)
+            BlockchainEvent::Extended(block_hash) => {
+                // A bit hacky: include_micro_bodies is used here to determine whether or not this node is a light node.
+                // Non light clients should republish all blocks to the BlockHeaderTopic.
+                if self.config.include_micro_bodies {
+                    if let Ok(block) = self.blockchain.read().get_block(&block_hash, true) {
+                        self.publish_block_header(block.clone());
+                        block_infos.push(block);
+                    }
+                } else if let Ok(block) = self.blockchain.read().get_block(&block_hash, false) {
+                    block_infos.push(block);
+                }
+            }
+            BlockchainEvent::HistoryAdopted(block_hash)
             | BlockchainEvent::Finalized(block_hash)
             | BlockchainEvent::EpochFinalized(block_hash) => {
                 if let Ok(block) = self.blockchain.read().get_block(&block_hash, false) {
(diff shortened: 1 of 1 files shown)


consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -441,26 +442,93 @@ impl<N: Network> BlockQueue<N> {
...
     /// Fetches the relevant blocks for any given `BlockchainEvent`
     fn get_new_blocks_from_blockchain_event(&self, event: BlockchainEvent) -> Vec<Block> {
         // Collect block numbers and hashes of newly added blocks first.
         let mut block_infos = vec![];
 
         match event {
-            BlockchainEvent::Extended(block_hash)
-            | BlockchainEvent::HistoryAdopted(block_hash)
+            BlockchainEvent::Extended(block_hash) => {
+                // A bit hacky: include_micro_bodies is used here to determine wether or not this node is a light node.
```
Eligioo:
```suggestion
                // A bit hacky: include_micro_bodies is used here to determine whether or not this node is a light node.
```

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -441,26 +442,93 @@ impl<N: Network> BlockQueue<N> {
...
+            }
+            BlockchainEvent::HistoryAdopted(block_hash)
             | BlockchainEvent::Finalized(block_hash)
             | BlockchainEvent::EpochFinalized(block_hash) => {
                 if let Ok(block) = self.blockchain.read().get_block(&block_hash, false) {
                     block_infos.push(block);
                 }
             }
             BlockchainEvent::Rebranched(_, new_blocks) => {
+                // A bit hacky: include_micro_bodies is used here to determine wether or not this node is a light node.
```
Eligioo:
```suggestion
                // A bit hacky: include_micro_bodies is used here to determine whether or not this node is a light node.
```

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -441,26 +442,93 @@ impl<N: Network> BlockQueue<N> {
...
+                        self.publish_block_header(block);
+                    }
+                }
+
                 for (_block_hash, block) in new_blocks {
                     block_infos.push(block);
                 }
             }
             BlockchainEvent::Stored(block) => {
+                // A bit hacky: include_micro_bodies is used here to determine wether or not this node is a light node.
```
Eligioo:
```suggestion
                // A bit hacky: include_micro_bodies is used here to determine whether or not this node is a light node.
```

sisou:
@nibhar how can I test if the messages are correctly de-duped with a local devnet? What log messages do I have to look for, or add?

sisou:
Running it without any changes, I only see one log message `Received block via gossipsub` per block in the web client, both micro and macro blocks. But I think that message is only logged after the `BlockStream` already deduped the blocks? Or maybe not? I cannot find any deduping logic for `BlockStream`, so I think it works!

nibhar:
I think what _should_ happen is that the pubsub implementation should create a unique identifier for messages. Important is that the messages we publish here are identical no matter who they are published by in terms of that identifier. 
Once the pubsub implementation then starts to send the IHAVE's and IWANT's they will not be requested by any peer which already has that specific message. 

So the desired behaviour would be that no traffic is produced for the additional publishes in excess of what a single publish would produce.

By the way, we do have a similar problem with macro blocks. They are assembled locally and for the asynchronous nature of the distributed system multiple nodes can produce a block at the same time, publishing it once each. However macro blocks are not unique as they can have a variety of different valid proofs. Thus they do (and should) create different identifiers and therefore will be IWANTed by other peers. Just in case this helps to see differences about this behaviour.

jsdanielh (approved):

sisou:
Just experienced this while updating my node and it syncing up with the chain again after I rebuilt with the latest tag:

It's also broadcasting old blocks during (live?) sync.

sisou:
I looked at the websocket connections to my peers, and I receive a message of 440B (bytes I think) every second from _every_ full/history peer. I'm connected to 13 peers, that's 5.7kB/s. Quite a bit of traffic.

PR #2360, opened 2024-04-11
Fix VALIDATOR_DEPOSIT value for migration tool

Eligioo:
The validator deposit is missing a proper value.

### DIFF
--- a/pow-migration/src/state/mod.rs
+++ b/pow-migration/src/state/mod.rs
@@ -7,7 +7,7 @@ use nimiq_genesis_builder::config::{
     GenesisAccount, GenesisHTLC, GenesisStaker, GenesisVestingContract,
 };
 use nimiq_keys::{Address, Ed25519PublicKey as SchnorrPublicKey};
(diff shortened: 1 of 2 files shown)


pow-migration/src/state/mod.rs
```diff
@@ -23,7 +23,7 @@ use crate::state::types::{Error, GenesisAccounts, GenesisValidator};
 // POW estimated block time in milliseconds
 const POW_BLOCK_TIME_MS: u64 = 60 * 1000; // 1 min
                                           // PoS validator deposit
-const VALIDATOR_DEPOSIT: u64 = 10; // FixMe: This should match the actual PoS validator deposit.
+const VALIDATOR_DEPOSIT: u64 = 10_000_000_000; // 100_000 coins represented in the smallest unit
```
jsdanielh: The correct way would be to use the value defined in `policy`

jsdanielh (changes requested):

jsdanielh (approved):

PR #2364, opened 2024-04-12
pow-migration: Allow stakers to increase stake and switch validators

### LINKED ISSUE #2352: Migration: Support increase stake
Add the ability to increase the stake to some specific validator during the pre-staking window.

### LINKED ISSUE #2353: Migration:  Support changing validators during pre-stake
Add the ability to move some staked funds from Validator A to Validator B during the pre-stake phase

### LINKED ISSUE #2362: Migration: assign staker to at most 1 validator
The migration tool currently allows you to register as a staker for multiple validators by sending multiple pre-staking transactions.

jsdanielh:
- Check that the registered stakers and stakers created from validator deposits that exceed the validator minimum deposit are always greater than the minimum stake. Otherwise keep them as burnt coins.
- Use the total balance used to register stakers and validators to decrease it from the burn address balance in the PoS genesis in order to preserve the total supply from the PoW chain.
- Allow stakers to increase stake and switch validators by ordering the transactions per sender, then keeping track of stakers being added and then increasing the staker and/or updating the delegation if there were previous valid transactions.
  This closes #2352, closes #2353 and closes #2362.
- Sort the possible validator registration transactions by timestamp before parsing them.

## Pull request checklist

### DIFF
--- a/pow-migration/src/genesis/mod.rs
+++ b/pow-migration/src/genesis/mod.rs
@@ -6,7 +6,7 @@ use nimiq_database::DatabaseProxy;
 use nimiq_genesis_builder::config::GenesisConfig;
 use nimiq_hash::Blake2bHash;
 use nimiq_keys::{KeyPair, SecureGenerate};
-use nimiq_primitives::networks::NetworkId;
+use nimiq_primitives::{coin::Coin, networks::NetworkId};
 use nimiq_rpc::Client;
 use nimiq_vrf::VrfSeed;
 use rand::{rngs::StdRng, SeedableRng};
@@ -88,7 +88,6 @@ pub async fn get_pos_genesis(
     let vrf_seed = VrfSeed::default().sign_next_with_rng(&KeyPair::generate(&mut rng), &mut rng);
 
     log::info!("Getting PoW account state");
-    let genesis_accounts = get_accounts(client, &final_block, pos_genesis_ts).await?;
 
     let (genesis_stakers, genesis_validators) =
         if let Some(registered_agents) = pos_registered_agents {
@@ -109,6 +108,22 @@ pub async fn get_pos_genesis(
             )
             .await?
         };
+
+    // Calculate how much stake was burnt into registering validators and stakers
+    let mut burnt_registration_balance = genesis_stakers
+        .iter()
+        .fold(Coin::ZERO, |acc, staker| acc + staker.balance);
+    burnt_registration_balance += genesis_validators
+        .iter()
+        .fold(Coin::ZERO, |acc, validator| acc + validator.balance);
+
+    let genesis_accounts = get_accounts(
+        client,
+        &final_block,
+        pos_genesis_ts,
+        burnt_registration_balance,
+    )
+    .await?;
 
     Ok(GenesisConfig {
         network: network_id,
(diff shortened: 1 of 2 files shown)


pow-migration/src/state/mod.rs
```diff
@@ -315,13 +325,18 @@ pub async fn get_stakers(
...
         );
         if validator.balance > Coin::from_u64_unchecked(Policy::VALIDATOR_DEPOSIT) {
-            stakers.push(GenesisStaker {
-                staker_address: validator.validator.validator_address.clone(),
-                balance: validator.balance - Coin::from_u64_unchecked(Policy::VALIDATOR_DEPOSIT),
-                delegation: validator.validator.validator_address.clone(),
-                inactive_balance: Coin::ZERO,
-                inactive_from: None,
-            })
+            let staker_address = validator.validator.validator_address.clone();
```
sisou, line 328:
```suggestion
            let delegation_address = validator.validator.validator_address.clone();
```

sisou: Having a struct `validator` inside a struct `validator` seems like bad naming.

pow-migration/src/state/mod.rs
```diff
@@ -315,13 +325,18 @@ pub async fn get_stakers(
...
+            stakers.insert(
+                staker_address.clone(),
+                GenesisStaker {
+                    staker_address,
+                    balance: validator.balance
+                        - Coin::from_u64_unchecked(Policy::VALIDATOR_DEPOSIT),
+                    delegation: validator.validator.validator_address.clone(),
+                    inactive_balance: Coin::ZERO,
+                    inactive_from: None,
```
sisou, line 340: Why even have the fields `inactive_balance` and `inactive_from` in `GenesisStaker`? Aren't they always ZERO and None? You also don't have the `retired_*` fields.

pow-migration/src/state/mod.rs
```diff
@@ -337,7 +352,9 @@ pub async fn get_stakers(
     }
 
     // Now look for the commit transaction
```
sisou: This comment seems wrong. It's not looking for the validator commit transaction here, but for pre-stake transactions.

sisou (commented):
Does this script check for the minimum stake when creating a staker? I think it should, and stakers below the threshold must be skipped and burnt.

pow-migration/src/genesis/mod.rs
```diff
@@ -110,6 +109,21 @@ pub async fn get_pos_genesis(
             .await?
         };
 
+    // Calculate how much stake was burnt into registering validators and stakers
+    let mut burnt_registration_balance = genesis_stakers.iter().map(|staker| staker.balance).sum();
+    burnt_registration_balance += genesis_validators
+        .iter()
+        .map(|validator| validator.balance)
```
Eligioo: Using the `fold` combinator here is the equivalent of `map().sum()` without mapping the struct first.

jsdanielh:
> Does this script check for the minimum stake when creating a staker? I think it should, and stakers below the threshold must be skipped and burnt.

You're right. Added this check as well

pow-migration/src/state/mod.rs
jsdanielh: These are fields required for building the genesis. In our case they are always zero but they come from the genesis crate which exposes them. The retired fields are not part of a staker definition.

viquezclaudio (approved):
LGTM

PR #2370, opened 2024-04-15
Light History Store for full nodes

### LINKED ISSUE #1704: Full node unable to apply election block
Currently fully synced full nodes are having the issue that after some event it loses the ability to store new incoming blocks and therefore can't keep up.

Multiple incidents have been observed and so far they all show the same pattern just before it can't continue to store blocks:
```
2023-07-13T13:04:40.637552192Z INFO  micro                | Skip block pushed block_number=233038
2023-07-13T13:04:40.663131683Z DEBUG validator            | Initializing block producer next_block_number=233039
2023-07-13T13:04:40.672803841Z DEBUG micro                | Not our turn, waiting for micro block #233039 block_number=233039 slot_band=4
2023-07-13T13:04:40.674427687Z DEBUG block_queue          | Received block via gossipsub block=#233039:MI:5633601d56
2023-07-13T13:04:40.674597462Z DEBUG block_queue          | Received block via gossipsub block=#233038:MI:0d572aaf9d
2023-07-13T13:04:40.992743659Z INFO  chain_ordering       | New block in Unknown chain fork_block_number=233037 current_block_number=233038 new_block_number=233038
2023-07-13T13:04:40.992827717Z DEBUG push                 | Storing block - on fork block=#233038:MI:74af9b8a35
2023-07-13T13:04:42.066183858Z INFO  chain_ordering       | New block in Inferior chain fork_block_number=233037 current_block_number=233038 new_block_number=233038
2023-07-13T13:04:42.066278922Z DEBUG push                 | Storing block - on inferior chain block=#233038:MI:0d572aaf9d
2023-07-13T13:04:42.681963371Z DEBUG block_queue          | Received block via gossipsub block=#233040:MA:686b97ed50
```

After this only three log messages are displayed and interestingly the receiving blocks via `gossipsub` is one of them. Which gives the feeling that the connections with other nodes aren't actually dropped but the node is unable to process them. Deadlock?
```
2023-07-13T13:04:43.736174015Z ERROR network_impl         | could not send request peer_id=12D3KooWAjftHreEDsNMK3qPBSe63LZcQorv6q8H3FR7CWg57CxV error=Inbound error: No receiver for request
2023-07-13T13:04:43.889421145Z DEBUG consensus            | Initiating head requests (to 11 peers)
2023-07-13T13:04:45.422861040Z DEBUG block_queue          | Received block via gossipsub block=#233040:MA:686b97ed50
```

Note that the consensus head hangs at #233038 but new `gossipsub` messages are coming in. This looks similar to #1692 .
```
2023-07-13T13:06:18.681751953Z DEBUG block_queue          | Received block via gossipsub block=#233127:MI:2373e55674
2023-07-13T13:06:09.389811588Z INFO  nimiq_client         | Consensus status: true - Head: #233038- 5506df7b33945b78f7985aa0b20a876b55ee43e90575349aa0480dc53303c1a2 block_number=233038 num_peers=14 status=true
2023-07-13T13:06:29.675350336Z DEBUG block_queue          | Received block via gossipsub block=#233138:MI:0ac26a9b0e
2023-07-13T13:06:29.389776842Z INFO  nimiq_client         | Consensus status: true - Head: #233038- 5506df7b33945b78f7985aa0b20a876b55ee43e90575349aa0480dc53303c1a2 block_number=233038 num_peers=14 status=true
2023-07-13T13:06:48.813880555Z DEBUG block_queue          | Received block via gossipsub block=#233157:MI:379d8d4a25
2023-07-13T13:06:49.389738268Z INFO  nimiq_client         | Consensus status: true - Head: #233038- 5506df7b33945b78f7985aa0b20a876b55ee43e90575349aa0480dc53303c1a2 block_number=233038 num_peers=14 status=true
```

At some point the node falls so far behind that verification for mempool txs start to fail
```
2023-07-13T13:05:52.071047472Z DEBUG verify               | Mempool-verify tx invalid at this block height block_number=233039 validity_start_height=233100
```
[full-node-cant-keep-up.log](https://github.com/nimiq/core-rs-albatross/files/12039916/full-node-cant-keep-up.log)

viquezclaudio:
Use the light history store for full node clients
Necessary changes to use the history interface
Several changes to the usage of the history store

...
## Pull request checklist

### DIFF
--- a/blockchain/src/blockchain/blockchain.rs
+++ b/blockchain/src/blockchain/blockchain.rs
@@ -19,7 +19,8 @@ use tokio::sync::broadcast::{channel as broadcast, Sender as BroadcastSender};
 use crate::chain_metrics::BlockchainMetrics;
 use crate::{
     blockchain_state::BlockchainState, chain_store::ChainStore, history::HistoryStore,
-    interface::HistoryInterface, reward::genesis_parameters,
+    interface::HistoryInterface, light_history_store::LightHistoryStore,
+    reward::genesis_parameters,
 };
 
 const BROADCAST_MAX_CAPACITY: usize = 256;
@@ -69,6 +70,9 @@ pub struct BlockchainConfig {
     /// Maximum number of epochs (other than the current one) that the ChainStore will store fully.
     /// Epochs older than this number will be pruned.
     pub max_epochs_stored: u32,
+    /// The history store that is used by the full blockchain.
+    /// If this is set to true, the light history store is used.
+    pub light_history_store: bool,
 }
 
 impl Default for BlockchainConfig {
@@ -76,6 +80,7 @@ impl Default for BlockchainConfig {
         Self {
             keep_history: true,
             max_epochs_stored: Policy::MIN_EPOCHS_STORED,
+            light_history_store: false,
         }
     }
 }
@@ -133,14 +138,12 @@ impl Blockchain {
         }
 
         let chain_store = ChainStore::new(env.clone());
-        let history_store = HistoryStore::new(env.clone());
 
         Ok(match chain_store.get_head(None) {
             Some(head_hash) => Blockchain::load(
                 env,
                 config,
                 chain_store,
-                history_store,
                 time,
                 network_id,
                 genesis_block,
@@ -150,7 +153,6 @@ impl Blockchain {
                 env,
                 config,
                 chain_store,
-                history_store,
                 time,
                 network_id,
                 genesis_block,
@@ -164,7 +166,7 @@ impl Blockchain {
         env: DatabaseProxy,
         config: BlockchainConfig,
         chain_store: ChainStore,
-        history_store: HistoryStore,
+
         time: Arc<OffsetTime>,
         network_id: NetworkId,
         genesis_block: Block,
@@ -265,6 +267,14 @@ impl Blockchain {
         let (tx_fork, _rx_fork) = broadcast(BROADCAST_MAX_CAPACITY);
         let (tx_log, _rx_log) = broadcast(BROADCAST_MAX_CAPACITY);
 
+        let history_store = if config.light_history_store {
+            Box::new(LightHistoryStore::new(env.clone(), network_id))
+                as Box<dyn HistoryInterface + Sync + Send>
+        } else {
+            Box::new(HistoryStore::new(env.clone(), network_id))
+                as Box<dyn HistoryInterface + Sync + Send>
+        };
+
         Ok(Blockchain {
             env,
             config,
@@ -274,7 +284,7 @@ impl Blockchain {
             fork_notifier: tx_fork,
             log_notifier: tx_log,
             chain_store,
-            history_store: Box::new(history_store) as Box<dyn HistoryInterface + Sync + Send>,
+            history_store,
             state: BlockchainState {
                 accounts,
                 main_chain,
@@ -301,7 +311,7 @@ impl Blockchain {
         env: DatabaseProxy,
         config: BlockchainConfig,
         chain_store: ChainStore,
-        history_store: HistoryStore,
+
         time: Arc<OffsetTime>,
         network_id: NetworkId,
         genesis_block: Block,
@@ -332,6 +342,14 @@ impl Blockchain {
         let (tx_fork, _rx_fork) = broadcast(BROADCAST_MAX_CAPACITY);
         let (tx_log, _rx_log) = broadcast(BROADCAST_MAX_CAPACITY);
 
+        let history_store = if config.light_history_store {
+            Box::new(LightHistoryStore::new(env.clone(), network_id))
+                as Box<dyn HistoryInterface + Sync + Send>
+        } else {
+            Box::new(HistoryStore::new(env.clone(), network_id))
+                as Box<dyn HistoryInterface + Sync + Send>
+        };
+
         Ok(Blockchain {
             env,
             config,
@@ -341,7 +359,7 @@ impl Blockchain {
             fork_notifier: tx_fork,
             log_notifier: tx_log,
             chain_store,
-            history_store: Box::new(history_store) as Box<dyn HistoryInterface + Sync + Send>,
+            history_store,
             state: BlockchainState {
                 accounts,
                 macro_info: main_chain.clone(),
--- a/blockchain/src/blockchain/history_sync.rs
+++ b/blockchain/src/blockchain/history_sync.rs
@@ -1,5 +1,6 @@
-use std::error::Error;
+use std::{collections::HashMap, error::Error};
 
+use itertools::Itertools;
 use nimiq_account::{BlockLogger, BlockState};
 use nimiq_block::{Block, BlockError};
 use nimiq_blockchain_interface::{
@@ -96,24 +97,41 @@ impl Blockchain {
     /// Extends the current chain with transactions from the validity sync process
     pub fn extend_validity_sync(
         this: RwLockUpgradableReadGuard<Blockchain>,
-        epoch: u32,
         history: &[HistoricTransaction],
     ) -> Option<Blake2bHash> {
         let mut txn = this.write_transaction();
 
-        // Store the new historic transactions into the History tree.
-        if let Some((root, _)) = this.history_store.add_to_history(&mut txn, epoch, history) {
-            txn.commit();
+        let mut txns_per_block: HashMap<u32, Vec<HistoricTransaction>> = HashMap::new();
+
+        for txn in history {
+            let bn = txn.block_number;
+
+            if let Some(txns) = txns_per_block.get_mut(&bn) {
+                txns.push(txn.clone());
+            } else {
+                let new_txns = vec![txn.clone()];
+                txns_per_block.insert(bn, new_txns);
+            }
+        }
+
+        let mut root = Blake2bHash::default();
+        for bn in txns_per_block.keys().sorted() {
+            let hist_txs = txns_per_block.get(bn).unwrap();
+
+            root = this
+                .history_store
+                .add_to_history(&mut txn, *bn, hist_txs)
+                .expect("Pushing txns to the history store failed")
+                .0;
 
             debug!(
-                num_transactions = history.len(),
+                num_transactions = hist_txs.len(),
+                block = bn,
                 "Pushed txns to the history store during validity sync"
             );
-            Some(root)
-        } else {
-            txn.abort();
-            None
         }
+        txn.commit();
+        Some(root)
     }
 
     /// Extends the current chain with a macro block (election or checkpoint) during history sync.
@@ -347,7 +365,7 @@ impl Blockchain {
         // Store the new historic transactions into the History tree.
         this.history_store.add_to_history(
             &mut txn,
-            block.epoch_number(),
+            block.block_number(),
             &history[first_new_hist_tx..],
         );
 
@@ -370,7 +388,7 @@ impl Blockchain {
 
         let wanted_history_root = this
             .history_store
-            .get_history_tree_root(block.epoch_number(), Some(&txn))
+            .get_history_tree_root(block.block_number(), Some(&txn))
             .ok_or(PushError::InvalidBlock(BlockError::InvalidHistoryRoot))?;
 
         if *block.history_root() != wanted_history_root {
--- a/blockchain/src/blockchain/history_sync.rs
+++ b/blockchain/src/blockchain/history_sync.rs
@@ -1,6 +1,5 @@
-use std::{collections::HashMap, error::Error};
+use std::{collections::BTreeMap, error::Error};
 
-use itertools::Itertools;
 use nimiq_account::{BlockLogger, BlockState};
 use nimiq_block::{Block, BlockError};
 use nimiq_blockchain_interface::{
@@ -101,26 +100,23 @@ impl Blockchain {
     ) -> Option<Blake2bHash> {
         let mut txn = this.write_transaction();
 
-        let mut txns_per_block: HashMap<u32, Vec<HistoricTransaction>> = HashMap::new();
+        //let mut txns_per_block: HashMap<u32, Vec<HistoricTransaction>> = HashMap::new();
+
+        let mut txns_per_block: BTreeMap<u32, Vec<HistoricTransaction>> = BTreeMap::new();
 
         for txn in history {
-            let bn = txn.block_number;
-
-            if let Some(txns) = txns_per_block.get_mut(&bn) {
-                txns.push(txn.clone());
-            } else {
-                let new_txns = vec![txn.clone()];
-                txns_per_block.insert(bn, new_txns);
-            }
+            txns_per_block
+                .entry(txn.block_number)
+                .or_default()
+                .push(txn.clone());
         }
 
         let mut root = Blake2bHash::default();
-        for bn in txns_per_block.keys().sorted() {
-            let hist_txs = txns_per_block.get(bn).unwrap();
 
+        for (bn, hist_txs) in txns_per_block {
             root = this
                 .history_store
-                .add_to_history(&mut txn, *bn, hist_txs)
+                .add_to_history(&mut txn, bn, &hist_txs)
                 .expect("Pushing txns to the history store failed")
                 .0;
 
@@ -130,6 +126,7 @@ impl Blockchain {
                 "Pushed txns to the history store during validity sync"
             );
         }
+
         txn.commit();
         Some(root)
     }
--- a/blockchain/src/blockchain/history_sync.rs
+++ b/blockchain/src/blockchain/history_sync.rs
@@ -97,11 +97,8 @@ impl Blockchain {
     pub fn extend_validity_sync(
         this: RwLockUpgradableReadGuard<Blockchain>,
         history: &[HistoricTransaction],
-    ) -> Option<Blake2bHash> {
+    ) -> Blake2bHash {
         let mut txn = this.write_transaction();
-
-        //let mut txns_per_block: HashMap<u32, Vec<HistoricTransaction>> = HashMap::new();
-
         let mut txns_per_block: BTreeMap<u32, Vec<HistoricTransaction>> = BTreeMap::new();
 
         for txn in history {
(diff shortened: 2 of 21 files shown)


sisou:
Does this fix #1704?

viquezclaudio:
Correct, this is the proper fix for 1704.
However this still needs more testing.

sisou:
:+1: I updated the PR description with that issue number.

blockchain/src/blockchain/history_sync.rs
```diff
@@ -96,24 +97,41 @@ impl Blockchain {
...
+
+        for txn in history {
+            let bn = txn.block_number;
+
+            if let Some(txns) = txns_per_block.get_mut(&bn) {
+                txns.push(txn.clone());
+            } else {
+                let new_txns = vec![txn.clone()];
+                txns_per_block.insert(bn, new_txns);
+            }
```
paberr:
```suggestion
            txns_per_block.entry(bn).or_default().push(txn.clone());
```

blockchain/src/blockchain/history_sync.rs
```diff
@@ -96,24 +97,41 @@ impl Blockchain {
...
+        }
+
+        let mut root = Blake2bHash::default();
+        for bn in txns_per_block.keys().sorted() {
+            let hist_txs = txns_per_block.get(bn).unwrap();
```
paberr: Using a `BTreeMap`, the keys would already be sorted.
Then, this could be simplified to `for (bn, hist_txs) in txns_per_block`.
Not sure whether it's worth it though.

Also, there's an inconsistency in naming: `txs` vs `txns`

blockchain/src/history/history_store.rs
```diff
@@ -648,13 +655,38 @@ impl HistoryInterface for HistoryStore {
...
+        let ext_hash_vec = self.get_hist_tx_by_hash(tx_hash, txn_opt);
+
+        // If the vector is empty then we have never seen a transaction with this hash.
+        if ext_hash_vec.is_empty() {
+            return false;
+        }
+
+        for hist_tx in ext_hash_vec {
+            // If the transaction is inside the validity window, return true.
+            if hist_tx.block_number >= validity_window_start {
```
paberr: Currently, we only check whether the transaction occurred *after* the validity window started.
However, it could potentially still fall outside the window if it occurred after it finished.
The second boundary check seems missing.

blockchain/src/history/mmr_store.rs
```diff
@@ -190,6 +191,14 @@ impl<'a, 'env> LightMMRStore<'a, 'env> {
         block_number: u32,
     ) -> Self {
         let size = get_size(hist_tree_table, tx, block_number);
+
+        // If size is 0 we need to copy the entries from the previous block number to the new one
+        if size == 0 {
+            init_mmr_from_block_number(hist_tree_table, tx, block_number);
+        }
```
paberr: I don't understand what this is about

blockchain/src/history/mmr_store.rs
```diff
@@ -232,6 +241,72 @@ pub fn remove_block_from_store(
...
+
+        txn.put(hist_tree_table, &new_key, &value);
+
+        let (next_key, _) = match cursor.next::<Vec<u8>, Blake2bHash>() {
+            Some(v) => v,
+            None => return,
+        };
+
+        // Deconstruct the key into a block number and a node index.
+        let (new_block_number, _) = key_to_index(next_key).unwrap();
```
paberr, line 332: `key_to_index` returns an `epoch_number` instead of a `block_number` (according to the variable names in the function; so this needs updating)

blockchain/src/history/mmr_store.rs
```diff
@@ -232,6 +241,72 @@ pub fn remove_block_from_store(
     }
 }
 
+/// Copies the mmr entries of the previous block number
+/// This is used to initialize the tree for the given block number, using the previous tree as a baseline
+pub fn init_mmr_from_block_number(
```
paberr, line 271: What is the advantage of copying the contents over accessing the previous ones when needed?
Less complexity when accessing?

blockchain/src/blockchain/history_sync.rs
viquezclaudio: Yes, I considered using a BTreeMap, but I wasn't sure if it was worth to add the dependency and use it vs just sorting the keys

blockchain/src/history/history_store.rs
viquezclaudio: Here I'm not doing anything different than what we were doing before, i.e: I just moved this code: https://github.com/nimiq/core-rs-albatross/blob/albatross/blockchain/src/blockchain/wrappers.rs#L196
to the history store file.

Do you suggest that we used to have a bug? 
What do you mean by "if it occurred after it finished." ?

blockchain/src/history/mmr_store.rs
viquezclaudio: For the light history store, what we are using are "peaks-only MMRs" 
However, this means that if we have to revert blocks, we don't have enough information to just remove txns and go back to a previous state, as we do with the regular MMR.

So what we are doing is to store "peaks only mmrs" per block, i.e. essentially we have "snapshots" of how the peaks only MMR looked at each block. 

However, since history nodes construct a big MMR with all the transactions of the given epoch, and this is the root that we need to verify against, then it means the starting point for block N is the peaks only mmr of block N-1, which is what this line is doing:
when we start pushing transactions for a block that we didn't have before, we first need to start with the peaks only mmr of the previous block and continue from there.

Let me know if this clarifies just question

blockchain/src/history/mmr_store.rs
viquezclaudio: See my comment above.

blockchain/src/history/mmr_store.rs
viquezclaudio: I'll update the comment, because we are using the same functions that the regular history store uses (which constructs keys based on {epoch_number, index} , but for the peaks only mmr, we construct keys based on {block_number, index}

blockchain/src/blockchain/history_sync.rs
hrxi: > if it was worth to add the dependency

It's in the standard library, so would be no additional dependency: [`std::collections::BTreeMap`](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html).

blockchain/src/blockchain/blockchain.rs
```diff
@@ -69,13 +70,17 @@ pub struct BlockchainConfig {
     /// Maximum number of epochs (other than the current one) that the ChainStore will store fully.
     /// Epochs older than this number will be pruned.
     pub max_epochs_stored: u32,
+    /// The history store that is used by the full blockchain.
+    /// If this is set to true, the light history store is used.
+    pub light_history_store: bool,
```
nibhar, line 75: Is this not already somewhat encoded in `keep_history` flag? Is there a usecase for `keep_history & light_history_store`? Similarly is there a usecase for `!use_history & !light_history_store`?

blockchain/src/blockchain/history_sync.rs
```diff
@@ -96,24 +96,39 @@ impl Blockchain {
...
         this: RwLockUpgradableReadGuard<Blockchain>,
-        epoch: u32,
         history: &[HistoricTransaction],
     ) -> Option<Blake2bHash> {
         let mut txn = this.write_transaction();
 
-        // Store the new historic transactions into the History tree.
-        if let Some((root, _)) = this.history_store.add_to_history(&mut txn, epoch, history) {
-            txn.commit();
+        //let mut txns_per_block: HashMap<u32, Vec<HistoricTransaction>> = HashMap::new();
```
nibhar: Leftover?

blockchain/src/blockchain/history_sync.rs
```diff
@@ -96,24 +96,39 @@ impl Blockchain {
     /// Extends the current chain with transactions from the validity sync process
     pub fn extend_validity_sync(
         this: RwLockUpgradableReadGuard<Blockchain>,
-        epoch: u32,
         history: &[HistoricTransaction],
     ) -> Option<Blake2bHash> {
```
nibhar: Looks to me as if the changes made this no longer fail for any reason, but panic instead.
```suggestion
    ) -> Blake2bHash {
```

blockchain/src/history/history_store.rs
nibhar: I don't know what that constant is, but that would be the check. This does not make any assumptions on where the chain is so the validity window that starts at `validity_window_start` could be a long time ago. Thus there are transactions after this window which would be valid.

i.e.
```suggestion
            if hist_tx.block_number >= validity_window_start && hist_tx.block_number <= validity_window_start + VALIDITY_WINDOW {
```

Depending on how it was used this would potentially not be buggy, but checking both boundaries is certainly more sound.

blockchain/src/history/history_store.rs
```diff
@@ -1156,15 +1188,100 @@ impl HistoryInterface for HistoryStore {
...
-        _block: &nimiq_block::Block,
-        _inherents: Vec<Inherent>,
+        txn: &mut WriteTransactionProxy,
+        block: &nimiq_block::Block,
+        inherents: Vec<Inherent>,
     ) -> Option<(Blake2bHash, u64)> {
-        todo!()
+        match block {
+            nimiq_block::Block::Macro(macro_block) => {
+                // Store the transactions and the inherents into the History tree.
```
nibhar: The transactions Vec of the following call is empty. Is this a misleading comment or are transactions missing here?

blockchain/src/history/light_history_store.rs
```diff
@@ -85,25 +93,44 @@ impl HistoryInterface for LightHistoryStore {
...
+        txn: &mut WriteTransactionProxy,
+        block: &MicroBlock,
+        _inherents: Vec<Inherent>,
+    ) -> Option<u64> {
+        remove_block_from_store(&self.hist_tree_table, txn, block.block_number());
+
+        self.validity_store
+            .delete_block_transactions(txn, block.block_number());
+
+        // TODO: We dont keep track of the size of the txns that we removed, is this necessary?
```
nibhar, line 112: TODO

`Some(0)` also seems to be incorrect, so I would suggest to either change the return type, or to actually return the correct return value.

blockchain/src/history/light_history_store.rs
```diff
@@ -147,28 +203,50 @@ impl HistoryInterface for LightHistoryStore {
...
+                    Policy::first_block_of(epoch_number + 1).unwrap() - 1,
+                ));
+
+                // Get the Merkle tree length
+                tree.num_leaves()
+            } else {
+                log::debug!(" validity out of bounds!");
+                0
+            }
+        };
```
nibhar: The comments here are not helpful to me, but I would have appreciated some helpful comments.

I don't know how much it matters and if I understand correctly but when epochs can be incomplete in the `hist_tree_table` then neither of these return the actual size of the epoch, but only the known size.

blockchain/src/history/light_history_store.rs
```diff
@@ -147,28 +203,50 @@ impl HistoryInterface for LightHistoryStore {
...
+
+        length
```
nibhar: Could be returned directly from the conditionals. Conditional can be collapsed.

blockchain/src/history/light_history_store.rs
```diff
@@ -177,6 +255,13 @@ impl HistoryInterface for LightHistoryStore {
...
 
+        for tx in hist_txs {
+            self.validity_store
+                .add_transaction(txn, block_number, tx.tx_hash().into())
+        }
+
+        self.validity_store.update_validity_store(txn, block_number);
+
         // Return the history root.
         Some((root, 0))
```
nibhar, line 270: That `0` also seems to be incorrect. Either change the return type, or return the actual correct value.

blockchain/src/history/mmr_store.rs
```diff
@@ -232,6 +262,75 @@ pub fn remove_block_from_store(
...
```
nibhar, line 271: Not entirely sure why the MMR of a single block preceding the current one is the way to go when the MMR actually spans an entire epoch. A comment explaining this would be nice as it certainly threw me of and I cannot see mention as to why in this PR.

consensus/src/sync/light/validity_window.rs
```diff
@@ -51,17 +51,71 @@ impl<TNetwork: Network> LightMacroSync<TNetwork> {
...
-            // In this case we are going to start a new request
-            let epoch_number = Policy::epoch_at(verifier_block_number);
-            let chunk_index = 0;
+            // By default we set the parameters assuming we are starting from stratch
+            let mut epoch_number = Policy::epoch_at(verifier_block_number);
+            let mut chunk_index = 0;
+            let mut root_hash = expected_root.clone();
+            let mut last_chunk_items: Option<usize> = None;
+            let mut verifier_block_number = verifier_block_number;
+            let mut election_in_window = election_in_window;
```
nibhar: These could just be `mut` in the arguments of the function already.

consensus/src/sync/light/validity_window.rs
```diff
@@ -51,17 +51,71 @@ impl<TNetwork: Network> LightMacroSync<TNetwork> {
...
+
+            // First we need to check if we already have items in the history store, and request only the remaining portion (if any)
+            match &self.blockchain {
+                BlockchainProxy::Full(blockchain) => {
+                    let blockchain_wr = blockchain.read();
+
+                    let (first_bn, last_bn) = blockchain_wr.history_store.history_store_range(None);
+
+                    // This means we already know the first epoch so we need to request the missing items from the current epoch
```
nibhar: What does the first epoch have to do with that? I assume you mean the epoch in which the start of the window is. Reading this and trying to understand it I do not necessarily know yet there is at most 2 relevant epochs.

consensus/src/sync/light/validity_window.rs
```diff
@@ -51,17 +51,71 @@ impl<TNetwork: Network> LightMacroSync<TNetwork> {
...
+                            current_items = current_length,
+                            "We already have items in our history store, moving to the current epoch",
+                        );
+
+                        chunk_index = (current_length / CHUNK_SIZE) as u32;
+                        last_chunk_items = Some(current_length % CHUNK_SIZE);
+                        root_hash = blockchain_wr.macro_head().header.history_root.clone();
+                        verifier_block_number = blockchain_wr.macro_head().block_number();
+                        election_in_window = false;
+                    } else if last_bn >= validity_window_start {
```
nibhar, line 88:
```suggestion
                    } else if last_bn >= validity_window_start {
                    // The window is not empty but there are missing items in the epoch the window starts in.
```
Or similar.

network-interface/src/peer_info.rs
```diff
@@ -66,9 +66,7 @@ impl Services {
                     | Services::TRANSACTION_INDEX
             }
             NodeType::Light => Services::empty(),
-            NodeType::Full => {
-                Services::ACCOUNTS_PROOF | Services::FULL_BLOCKS | Services::ACCOUNTS_CHUNKS
-            }
+            NodeType::Full => Services::ACCOUNTS_PROOF | Services::FULL_BLOCKS,
```
nibhar: Interesting. So accounts chunks can now only be received by History nodes?

nibhar (commented):
Very high level I am not a big fan of `bn` as `block_number`. I would much prefer to not abbreviate that to stay consistent throughout the codebase.

Generally looks mostly good to me. Left a couple of comments which I think are worth addressing. I personally would also appreciate some more documentation as to why things are done the way they are done, but that is just me.

A couple of clippy warnings need fixing too.

blockchain/src/blockchain/blockchain.rs
viquezclaudio: AFAIK, `keep_history` is used to determine if we want to prune the history store,  keeping only up to "max_epochs_stored" epochs, whereas "light_history_store" specifies the type of History Store that is being used, and these two are independent.
i.e.: you could potentially have a node using a light story store where keep_history is true  (not sure if there is a  use case for this)
Also, with this implementation, it is possible to have full nodes with the regular history store and full nodes with the light history store.

However, if we decide that:

- Full nodes only use the light story store and they only keep up to max_epochs_stored 
 
Then this can indeed be simplified, as the different combinations mentioned above would no longer be supported

nibhar: Fair enough. More flexibility is good, if it does not lead to configurations which are bogus. Thanks for explaining.

blockchain/src/history/history_store.rs
viquezclaudio: I will update the comment, we are only storing inherent for this particular case

blockchain/src/history/light_history_store.rs
viquezclaudio: The problem is that this comes from the regular history store, where we have actual transactions and the pruning process is different.
For the light history store, internally, we just keep peaks only MMRs so we don't have the information to calculate this, so not sure what we want to do here

blockchain/src/history/light_history_store.rs
viquezclaudio: I will update the comments

blockchain/src/history/light_history_store.rs
viquezclaudio: Same problem as before, we dont keep track of the sizes in the light history store

blockchain/src/history/mmr_store.rs
viquezclaudio: I will update the comment

consensus/src/sync/light/validity_window.rs
viquezclaudio: Yes, the comment is misleading. I will update it

network-interface/src/peer_info.rs
viquezclaudio: Accounts chunks can only be provided by history nodes.
They can no longer be provided by full nodes. 

This struct represents what each node type can provide

viquezclaudio: As per our discussion, I created a new issue for this #2592

blockchain/src/history/light_history_store.rs
viquezclaudio: Created #2593 to track this issue

nibhar (approved):
Good to go from my end even though I still don't like the `bn` abbreviations. I guess @paberr should take another look.

PR #2373, opened 2024-04-17
Variable-sized Coin serialization

sisou:

This PR makes `Coin` serialize to a varint in transactions and accounts, both over the wire and their internal representations (history store). 

### Reasoning

No transaction will likely ever make use of the full u64 range for its `value`, and especially not for `fee`. So most zero bytes for Coins in most transactions are wasted. This is especially true for signaling transactions, which have 16 zero-bytes to represent 0 value and (mostly) 0 fee. This PR reduces them to two zero bytes.

### Drawbacks

- A transaction could increase in size when adding or increasing the fee, requiring to increase the fee _again_ if targeting a specific fee-per-byte.

### Exceptions

There are two exceptions for backwards-compatibility with PoW transactions:
1. Coins in vesting contract creation data are still serialized with `fixint::be`
2. For determining a transaction's hash, `tx.serialize_content()` serializes `value` and `fee` as before with `fixint::be`

## Pull request checklist


hrxi (approved):
Looks like a good improvement to me.

hrxi:
I think we decided not to do that because it's a breaking change. Is that correct?

PR #2383, opened 2024-04-21
pow-migration: Fix burnt registration balance calculation and fix accounts migration

jsdanielh:
- Fix accounts migration by setting the `Policy.NUM_SNAPSHOTS_MAX` constant that allows to get accounts tree chunks for blocks that are this value below the head of the chain.
- Fix math for `burnt_registration_balance` since it was incorrectly summing the validator balance twice. Also rename the `balance` field to `total_stake` in the `GenesisValidator` struct to clarify that this balance includes the total stake of this validator (validator deposit and staker's delegated stake).

## Pull request checklist

### DIFF
--- a/pow-migration/src/state/mod.rs
+++ b/pow-migration/src/state/mod.rs
@@ -20,8 +20,12 @@ use nimiq_transaction::account::htlc_contract::{AnyHash, AnyHash32, AnyHash64};
 
 use crate::state::types::{Error, GenesisAccounts, GenesisValidator};
 
-// POW estimated block time in milliseconds
+// PoW estimated block time in milliseconds
 const POW_BLOCK_TIME_MS: u64 = 60 * 1000; // 1 min
+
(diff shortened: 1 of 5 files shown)


pow-migration/src/state/mod.rs
```diff
@@ -125,6 +141,13 @@ pub async fn get_accounts(
         htlc_accounts: vec![],
     };
     let mut start_prefix = "".to_string();
+
+    // Check that the PoW client is already set up
+    if client.get_constant("Policy.NUM_SNAPSHOTS_MAX").await? != POW_MAX_SNAPSHOTS {
+        log::error!("RPC client is not set up for accounts migration. Call first `setup_pow_rpc_server` first");
```
sisou:
```suggestion
        log::error!("RPC client is not set up for accounts migration. Call `setup_pow_rpc_server` first");
```

viquezclaudio (approved):

PR #2392, opened 2024-04-23
pow-migration: Fix genesis block timestamp calculation

jsdanielh:
Fix genesis block timestamp calculation since it was assuming that PoW block timestamps where in milliseconds when they are actually in seconds.

## Pull request checklist

### DIFF
--- a/pow-migration/src/genesis/mod.rs
+++ b/pow-migration/src/genesis/mod.rs
@@ -15,11 +15,8 @@ use time::OffsetDateTime;
 use crate::{
     genesis::types::{Error, PoSRegisteredAgents, PoWRegistrationWindow},
     history::get_history_root,
-    state::{get_accounts, get_stakers, get_validators},
+    state::{get_accounts, get_stakers, get_validators, POW_BLOCK_TIME},
 };
-
-// POW estimated block time in milliseconds
-const POW_BLOCK_TIME_MS: u64 = 60 * 1000; // 1 min
 
 /// Gets the genesis config file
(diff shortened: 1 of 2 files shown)


viquezclaudio (approved):

pow-migration/src/genesis/mod.rs
```diff
@@ -75,7 +72,7 @@ pub async fn get_pos_genesis(
 
     // The PoS genesis timestamp is the cutting block timestamp plus a custom delay
     let pos_genesis_ts =
-        pow_reg_window.confirmations as u64 * POW_BLOCK_TIME_MS + final_block.timestamp as u64;
+        pow_reg_window.confirmations as u64 * POW_BLOCK_TIME + final_block.timestamp as u64;
```
sisou, line 75: But now you have a `pos_genesis_ts` in seconds, but for PoS you need it in milliseconds. So I think there's a `* 1000` missing here.

Because now you are passing down the genesis timestamp in seconds, but then in e.g. `pos_vesting_account_from_account` and `pos_htlc_account_from_account` this is multiplied with the blocktime in milliseconds.

sisou: On the other hand you are comparing it to `cutting_block.timestamp`, which is also in seconds, so it should stay in seconds? In any case, somewhere there needs to be a conversion, and I think you should encode the unit of the variable in its name everywhere, so call it `pos_genesis_ts_unix`, or `pos_genesis_ts_seconds` to prevent confusion.

PR #2395, opened 2024-04-24
Add command to list pre-stakers and optionally filter by validator

### LINKED ISSUE #2387: Migration: Create a command to dump the stake allocation
Create a command to dump the stake allocation once the pre-stake window is finished.

Eligioo:
Add the command `list-stakers`, similar to `list-validators`, to dump the stake allocation with optionally the choice to only get the allocations by providing a validator address.

## Pull request checklist

### DIFF
--- a/pow-migration/src/main.rs
+++ b/pow-migration/src/main.rs
@@ -6,6 +6,7 @@ use log::{info, level_filters::LevelFilter};
 use nimiq_keys::Address;
 use nimiq_lib::config::{config::ClientConfig, config_file::ConfigFile};
 use nimiq_pow_migration::{
+    exit_with_error,
     genesis::write_pos_genesis,
     get_block_windows,
     history::{get_history_store_height, migrate_history},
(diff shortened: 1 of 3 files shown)


pow-migration/src/main.rs
```diff
@@ -164,9 +173,56 @@ async fn main() {
...
-        println!("Registered validators:");
-        for validator in registered_validators {
-            println!("{}", validator.validator.validator_address);
+
+        if args.list_validators {
+            println!("Registered validators:");
+            for validator in registered_validators {
+                println!("{}", validator.validator.validator_address);
+            }
+        } else {
```
hrxi: This is making `--list-stakers` and `--list-validators` mutually exclusive. Is this intended? If so, the conflict should probably be added to the `Args` struct. If not, both should be checked here.

pow-migration/src/main.rs
```diff
@@ -164,9 +173,56 @@ async fn main() {
...
+            {
+                Ok((pre_stakers, _)) => match args.validator {
+                    Some(address) => {
+                        if let Ok(address) = Address::from_any_str(&address) {
+                            pre_stakers
+                                .into_iter()
+                                .filter(|pre_staker| pre_staker.delegation == address)
+                                .collect()
+                        } else {
+                            log::error!("Invalid address as --validator argument provided");
```
hrxi:
```suggestion
                            log::error!("Invalid address provided as --validator argument");
```

pow-migration/src/main.rs
```diff
@@ -164,9 +173,56 @@ async fn main() {
...
-        for validator in registered_validators {
-            println!("{}", validator.validator.validator_address);
+
+        if args.list_validators {
+            println!("Registered validators:");
+            for validator in registered_validators {
+                println!("{}", validator.validator.validator_address);
+            }
+        } else {
+            if pow_client.block_number().await.unwrap()
```
hrxi: You're using the "unwrap or process exit" pattern quite a lot. You could either use `.expect(<string>)` or it could be extracted into a function:
```rs
fn unwrap_or_exit<T, E>(message: &'static str, result: Result<T, E>) -> T {
    match result {
        Ok(t) => t,
        Err(error) => {
            log::error(?error, "{}", message);
            std::process::exit(1);
        }
    }
}
```

pow-migration/src/main.rs
```diff
@@ -21,8 +23,15 @@ struct Args {
     /// Path to the PoS configuration file
     #[arg(short, long)]
     config: String,
+    /// Output a list of registered validators
     #[arg(short, long)]
     list_validators: bool,
+    /// Output a list of pre-stakers
+    #[arg(short = 's', long)]
+    list_stakers: bool,
+    /// Only show the pre-stakers for a specific validator. This argument is only intended to be used in conjunction with list-stakers
```
jsdanielh: I believe we should make this explicit in the `clap` parser by using an nested struct. For instance the help argument wouldn't be able to document this.

pow-migration/src/main.rs
jsdanielh: Yeah, I guess we want them to be exclusive but we should change the parser to account for this.

pow-migration/src/main.rs
jsdanielh:
```suggestion
                            log::error!(%address, "Invalid address provided as argument ('--validator, -v')");
```

pow-migration/src/main.rs
```diff
@@ -164,9 +173,56 @@ async fn main() {
...
+                    std::process::exit(1);
+                }
+            };
+
+            println!("Pre-stakers:");
+            for pre_staker in pre_stakers {
+                println!(
+                    "{} has delegated stake to {} with a prestake of {} NIM",
+                    pre_staker.staker_address,
+                    pre_staker.delegation.to_user_friendly_address(),
```
jsdanielh: Display of an address should already use the user friendly representation
```suggestion
                    pre_staker.delegation,
```

pow-migration/src/main.rs
Eligioo: Addressed in 763cd1f.

pow-migration/src/main.rs
Eligioo: Fixed

jsdanielh (approved):
LGTM

PR #2403, opened 2024-04-25
pow-migration: Fix detection of switching validator in pre-stake

### LINKED ISSUE #2399: Migration: Validator change in pre-staking window not detected
This address https://test.nimiq.watch/#NQ95+FHVN+JAN4+8LKT+LCE2+LC3T+P861+NK5M+HVGS delegated 2.8 million NIM to `NQ65 ... H17C` in block `3027744`. Then it sent a 1 NIM transaction 3 minutes later in block `3027747` with data to delegate to `NQ26 ... XVNY`. That should have changed the delegation to the later validator.

However, in the genesis file this staker is delegating to `NQ65 ... H17C`, which I think is incorrect.

jsdanielh:
Fix the detection of switching validator in pre-stake as it was filtering always transactions that didn't have the minimum stake but for switching to another validator in pre-stake, the stake is already above the minimum since it was already counted to another validator.

This fixes #2399.

## Pull request checklist

### DIFF
--- a/pow-migration/src/state/mod.rs
+++ b/pow-migration/src/state/mod.rs
@@ -1,6 +1,11 @@
 pub mod types;
 
-use std::{collections::HashMap, ops::Range, str::FromStr, vec};
+use std::{
+    collections::{hash_map::Entry, HashMap},
+    ops::Range,
+    str::FromStr,
+    vec,
+};
 
 use nimiq_bls::PublicKey as BlsPublicKey;
 use nimiq_genesis_builder::config::{
@@ -336,7 +341,7 @@ pub async fn get_stakers(
         let extra_validator_deposit = registered_validator
             .total_stake
             .saturating_sub(Coin::from_u64_unchecked(Policy::VALIDATOR_DEPOSIT));
-        let balance_to_burn = if extra_validator_deposit > Coin::ZERO
+        if extra_validator_deposit > Coin::ZERO
             && extra_validator_deposit >= Coin::from_u64_unchecked(Policy::MINIMUM_STAKE)
         {
             let staker_address = registered_validator.validator.validator_address.clone();
@@ -351,14 +356,12 @@ pub async fn get_stakers(
                     inactive_from: None,
                 },
             );
-            Coin::ZERO
-        } else {
-            extra_validator_deposit
         };
 
         // Update the validator balance with the burnt balance
         let mut updated_registered_validator = registered_validator.clone();
-        updated_registered_validator.total_stake -= balance_to_burn;
+        updated_registered_validator.total_stake =
+            Coin::from_u64_unchecked(Policy::VALIDATOR_DEPOSIT);
         validators.insert(
             updated_registered_validator
                 .validator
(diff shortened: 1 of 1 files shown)


pow-migration/src/state/mod.rs
```diff
@@ -392,27 +397,38 @@ pub async fn get_stakers(
...
+                                        Entry::Occupied(mut entry) => {
+                                            // If we have an entry for this staker, treat this as a switch to another
+                                            // validator or an increase of the stake (if the validator isn't changed)
+                                            let staker = entry.get_mut();
+                                            staker.delegation =
+                                                validator.validator.validator_address.clone();
+                                            staker.balance += stake;
+
+                                            // Update the validator total stake
+                                            validator.total_stake += stake;
```
Eligioo: Shouldn't we also decrease the `total_stake` of the old validator if we switching between validators?

jsdanielh: Yup, you are right. Added logic to address this

Eligioo:
@jsdanielh I do think this is still prone to incorrect outcomes. I believe `decrease_validator_total_stake` should only be set to `Some` if the staker is actually switching between validators. It can very well be that the staker doesn't change validator and only simply increase his stake. In this scenario the `stakers.entry` will be `Occupied` as you process the top-up transaction after his initial pre-stake transaction. But right now this will result in `validator.total_stake` to reduced in the end, which shouldn't happen.

To make this easier to reason about, and since stake can jump around validators quite a lot, is that you compute the total stake of a validator in a second loop after you've done processing all the transactions in `for (_, txns) in txns_by_sender.iter_mut()`. Instead of trying to keep track of the total stake as you process the `txns_by_sender`.

jsdanielh:
> @jsdanielh I do think this is still prone to incorrect outcomes. I believe `decrease_validator_total_stake` should only be set to `Some` if the staker is actually switching between validators. It can very well be that the staker doesn't change validator and only simply increase his stake. In this scenario the `stakers.entry` will be `Occupied` as you process the top-up transaction after his initial pre-stake transaction. But right now this will result in `validator.total_stake` to reduced in the end, which shouldn't happen.
> 
> To make this easier to reason about, and since stake can jump around validators quite a lot, is that you compute the total stake of a validator in a second loop after you've done processing all the transactions in `for (_, txns) in txns_by_sender.iter_mut()`. Instead of trying to keep track of the total stake as you process the `txns_by_sender`.

Makes sense. Just made the change to implement this.

pow-migration/src/state/mod.rs
```diff
@@ -433,6 +445,14 @@ pub async fn get_stakers(
         }
     }
 
+    // Now that we have the stakers, update the validator's total stake
+    for staker in stakers.values() {
+        let validator = validators
+            .get_mut(&staker.delegation.to_string())
+            .expect("Delegated validator must exist");
+        validator.total_stake += staker.balance;
```
Eligioo: `validator.total_stake += staker.balance` would be called every time you call the `get_stakers` method thus compounding the `total_stake`.

pow-migration/src/state/mod.rs
```diff
@@ -433,6 +445,14 @@ pub async fn get_stakers(
...
+    }
```
Eligioo: A suggestion to make the `total_stake` consistent upon every time `get_stakers` being called.

```suggestion
    // Now that we have the stakers, recompute the validator's total stake
    for genesis_validator in validators.values_mut() {
        let stakers: Vec<&GenesisStaker> = stakers
            .iter()
            .filter_map(|staker| {
                if staker.1.delegation == genesis_validator.validator.validator_address {
                    return Some(staker.1);
                }
                None
            })
            .collect();

        genesis_validator.total_stake = Coin::from_u64_unchecked(Policy::VALIDATOR_DEPOSIT)
            + stakers
                .iter()
                .fold(Coin::ZERO, |acc, staker| acc + staker.balance);
    }
```

Eligioo:
LGTM

Eligioo (approved):

PR #2407, opened 2024-04-26
Remove unnecessary tokio::spawn

### LINKED ISSUE #2404: Browser light client cannot maintain connections
### General information

- Library version(s): v0.21.0
- Browser version(s): Chrome 123.0.6312.122
- Devices/Simulators/Machine affected: Desktop Linux
- Related issues: #

#### Expected behavior

Browser light clients can establish and hold connections with peers.

#### Actual behavior

Browser clients panic when connecting to the first peer (seed) with:

```
thread '<unnamed>' panicked at 'there is no reactor running, must be called from the context of a Tokio 1.x runtime': /core-rs-albatross/network-interface/src/request/mod.rs:223
```

and can then no longer maintain connections, as peers disconnect after a few seconds because of response timeouts on their side.

#### Steps to reproduce

Either
- Open https://wallet.pos.nimiq-testnet.com in your browser.

or

- Compile the browser client locally (`./web-client/scripts/build.sh --only web --skip-types --skip-wasm-opt`) and serve the `web-client` folder, then go to http://localhost/example. If `seed1` is not up, change the seed node in `web-client/example/module.js` to a node that is up.

viquezclaudio:
- Unnecessary tokio spawn
- This is not compatible with web clients (WASM)

This could explain the failure mentioned in #2404 

Before [this change](https://github.com/nimiq/core-rs-albatross/commit/e7a0319fafaa10fcc7256061b89c2ad7e086c8c9) regular nodes didn't try to connect to web clients, because they were filtered: web clients do not provide any service to regular nodes. This can be observed in this [code](https://github.com/nimiq/core-rs-albatross/commit/e7a0319fafaa10fcc7256061b89c2ad7e086c8c9#diff-26b6d741d633ef4dcc78ac3d521f89ea1ead7110cad61ef8a987e23c5a12012cL528).

This means that before, **web clients were not responding to any of the consensus handlers**.

However, after the _handle incompatible peers_ commit, regular nodes can now perform head requests to web clients, which causes web clients to execute a `tokio::spawn`, which is not allowed (a special task executor is necessary in web environments).

This `tokio::spawn` seems unnecessary or superfluous (at least) according to the logic around it.

## Pull request checklist

### DIFF
--- a/network-interface/src/request/mod.rs
+++ b/network-interface/src/request/mod.rs
@@ -253,16 +253,10 @@ pub fn message_handler<
             .for_each_concurrent(MAX_CONCURRENT_HANDLERS, |(msg, peer_id)| {
                 let req_environment = req_environment.clone();
                 async move {
-                    let req_environment = req_environment.clone();
+                    log::trace!("{:?} {:#?}", peer_id, msg);
 
-                    tokio::spawn(async move {
-                        log::trace!("{:?} {:#?}", peer_id, msg);
-
-                        // Messages do not have a response (so the response is ignored)
-                        msg.message_handle(peer_id, &req_environment);
-                    })
-                    .await
-                    .expect("Request handler panicked")
+                    // Messages do not have a response (so the response is ignored)
+                    msg.message_handle(peer_id, &req_environment);
                 }
             })
             .await
--- a/network-interface/src/request/mod.rs
+++ b/network-interface/src/request/mod.rs
@@ -217,27 +217,20 @@ pub fn request_handler<T: Send + Sync + Clone + 'static, Req: Handle<N, T>, N: N
                 let network = Arc::clone(&network);
                 let req_environment = req_environment.clone();
                 async move {
-                    let req_environment = req_environment.clone();
-                    let network = Arc::clone(&network);
+                    log::trace!("[{:?}] {:?} {:#?}", request_id, peer_id, msg);
 
-                    tokio::spawn(async move {
-                        log::trace!("[{:?}] {:?} {:#?}", request_id, peer_id, msg);
-
-                        // Try to send the response, logging to debug if it fails
-                        if let Err(err) = network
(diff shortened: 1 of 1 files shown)


network-interface/src/request/mod.rs
```diff
@@ -220,24 +220,20 @@ pub fn request_handler<T: Send + Sync + Clone + 'static, Req: Handle<N, T>, N: N
                     let req_environment = req_environment.clone();
                     let network = Arc::clone(&network);
 
-                    tokio::spawn(async move {
```
sisou, line 223: There's the same kind of spawn in the `message_handler` function below. Maybe that can be removed, too?

viquezclaudio: That one looks unnecessary  too

sisou (approved):
Works! It might even have fixed the "closure invoked recursively or after being dropped" problem with libp2p's websocket-websys transport :astonished:

paberr (approved):

network-interface/src/request/mod.rs
```diff
@@ -220,24 +220,20 @@ pub fn request_handler<T: Send + Sync + Clone + 'static, Req: Handle<N, T>, N: N
                     let req_environment = req_environment.clone();
                     let network = Arc::clone(&network);
```
hrxi: These clones are likely unneeded if the `tokio::spawn` is removed below.

hrxi (commented):
> This `tokio::spawn` seems unnecessary or superfluous (at least) according to the logic around it.

It doesn't seem entirely superfluous, it makes sure that each request cannot block the rest of the requests. If that's still wanted, then this change is incorrect.

jsdanielh:
> > This `tokio::spawn` seems unnecessary or superfluous (at least) according to the logic around it.
> 
> It doesn't seem entirely superfluous, it makes sure that each request cannot block the rest of the requests. If that's still wanted, then this change is incorrect.

We will address this in a separate issue (#2434).

PR #2417, opened 2024-05-01
GH action for publishing OpenRPC document on new version

Eligioo:
This PR contains a new Github Action that with every release will publish a new OpenRPC document to the latest release tag.

## Pull request checklist

### DIFF
--- a/.github/workflows/github_release_docker.yml
+++ b/.github/workflows/github_release_docker.yml
@@ -12,10 +12,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
-      - uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: --release
+      - name: Build binaries in release mode
+        run: |
+          cargo build --release
       - name: Archive production artifacts
         uses: actions/upload-artifact@v4
         with:
--- a/.github/workflows/publish_openrpc_document.yml
+++ b/.github/workflows/publish_openrpc_document.yml
@@ -1,0 +1,30 @@
+name: Publish OpenRPC document
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  build_release_binary_and_publish_document:
+    name: Build nimiq-rpc-schema with release flag
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+      - name: Build binary in release mode
+        run: |
+          cargo build --release --bin nimiq-rpc-schema
+      - name: Get release information
+        id: get_release
+        uses: bruceadams/get-release@v1.3.2
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Run nimiq-rpc-schema and save JSON output
+        run: |
+          cargo run --release --bin nimiq-rpc-schema -- -o ${{ steps.get_release.outputs.tag_name }} > openrpc-document.json
+      - name: Upload OpenRPC document to latest release
+        uses: JasonEtco/upload-to-release@v0.1.1
+        with:
+          args: openrpc-document.json application/json
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/npm_publish.yml
+++ b/.github/workflows/npm_publish.yml
@@ -14,7 +14,6 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: stable
           targets: wasm32-unknown-unknown
       - uses: Swatinem/rust-cache@v2
       - uses: jetli/wasm-bindgen-action@v0.2.0


jsdanielh (approved):

.github/workflows/publish_openrpc_document.yml
```diff
@@ -0,0 +1,40 @@
...
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install protoc
+        run: |
+          sudo apt-get update && sudo apt-get install -y protobuf-compiler
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+      - uses: actions-rs/cargo@v1
```
jsdanielh: This action is no longer maintained, could you switch to a simple `run`?

.github/workflows/publish_openrpc_document.yml
```diff
@@ -0,0 +1,40 @@
...
+jobs:
+  build_release_binary_and_publish_document:
+    name: Build nimiq-rpc-schema with release flag
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install protoc
+        run: |
+          sudo apt-get update && sudo apt-get install -y protobuf-compiler
+      - uses: actions-rs/toolchain@v1
```
jsdanielh: This action is no longer maintained, could you switch to `dtolnay/rust-toolchain@stable`?

jsdanielh (changes requested):

.github/workflows/github_release_docker.yml
```diff
@@ -14,13 +14,10 @@ jobs:
...
-        with:
-          toolchain: stable
-      - uses: actions-rs/cargo@v1
-        with:
-          command: build
-          args: --release
+      - uses: dtolnay/rust-toolchain@stable
+      - name: Build binaries in release mode
+        run: |
+          cargo build --release
```
jsdanielh, line 17: Ah nice. Thanks!

jsdanielh (approved):

jsdanielh:
Rebasing branch to merge it

.github/workflows/publish_openrpc_document.yml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install protoc
+        run: |
+          sudo apt-get update && sudo apt-get install -y protobuf-compiler
+      - uses: dtolnay/rust-toolchain@stable
+      - name: Build binary in release mode
+        run: |
+          cargo build --release --bin nimiq-rpc-schema
+      - name: Get latest release
```
jeffesquivels: I think `bruceadams/get-release@v1.3.2` [gets the release associated with the tag that triggered this action](https://github.com/bruceadams/get-release/blob/main/src/get-release.js#L10-L14), not the latest release.

Which makes wonder how would this behave if a tag was created without a release associated to it (because I'm not sure if in our release process we create the the tag first, thus triggering this action, and then create the release or both things happen at the same time).

.github/workflows/publish_openrpc_document.yml
```diff
@@ -0,0 +1,37 @@
...
+        id: get_release
+        uses: bruceadams/get-release@v1.3.2
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Run nimiq-rpc-schema and save JSON output
+        run: |
+          ./target/release/nimiq-rpc-schema -o ${{ steps.get_release.outputs.tag_name }} > openrpc-document.json
+      - name: Upload OpenRPC document to latest release
```
jeffesquivels, line 25: Same comment as above.

.github/workflows/publish_openrpc_document.yml
sisou: I agree, if there's a tag but no release, this will break, or even worse, upload it's generated RPC spec to an older release. Can the trigger for this action be changed to trigger on release, not on tag?

Eligioo: I set the trigger to `on release created`

.github/workflows/publish_openrpc_document.yml
Eligioo: Resolved

Eligioo:
@sisou as per the [documentation](https://github.com/marketplace/actions/rustup-toolchain-install#inputs) of `dtolnay/rust-toolchain`, I removed the `toolchain` input in your NPM Publish Action: `Important: the default is to match the @rev as described above. When passing an explicit toolchain as an input instead of @rev, you'll want to use "dtolnay/rust-toolchain@master" as the revision of the action.`

EDIT: I wasn't aware that this basically happend for all the GH Actions we have so created a new PR addressing the rest.

.github/workflows/publish_openrpc_document.yml
```diff
@@ -0,0 +1,33 @@
...
+        run: |
+          cargo build --release --bin nimiq-rpc-schema
+      - name: Get release information
+        id: get_release
+        uses: bruceadams/get-release@v1.3.2
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Run nimiq-rpc-schema and save JSON output
+        run: |
+          ./target/release/nimiq-rpc-schema -o ${{ steps.get_release.outputs.tag_name }} > openrpc-document.json
```
hrxi:
```suggestion
          cargo run --bin=nimiq-rpc-schema --quiet -- -o ${{ steps.get_release.outputs.tag_name }} > openrpc-document.json
```

.github/workflows/publish_openrpc_document.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Upload OpenRPC document to latest release
+        uses: actions/upload-release-asset@v1
+        with:
+          upload_url: ${{ steps.get_release.outputs.upload_url }}
+          asset_path: ./openrpc-document.json
+          asset_name: openrpc-document.json
+          asset_content_type: application/json
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
hrxi: Missing newline.

.github/workflows/publish_openrpc_document.yml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Get release information
+        id: get_release
+        uses: bruceadams/get-release@v1.3.2
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Run nimiq-rpc-schema and save JSON output
+        run: |
+          ./target/release/nimiq-rpc-schema -o ${{ steps.get_release.outputs.tag_name }} > openrpc-document.json
+      - name: Upload OpenRPC document to latest release
+        uses: actions/upload-release-asset@v1
```
hrxi: https://github.com/actions/upload-release-asset seems to be archived. Does that mean anything for our usage?

.github/workflows/publish_openrpc_document.yml
hrxi: It'd also work without the `--quiet`, actually, Cargo's own output goes to stderr.

.github/workflows/publish_openrpc_document.yml
Eligioo: It still uses an old NodeJS version with is deprecated and can potentially break things in the future. 

Replaced it with [JasonEtco/upload-to-release](https://github.com/marketplace/actions/upload-to-release)

hrxi (approved):
Looks good to me now. :)

jsdanielh (approved):

PR #2420, opened 2024-05-02
Genesis block inclusion proofs

### LINKED ISSUE #2421: Deadlock requesting block inclusion proofs for the genesis block
Requesting block inclusion proofs generates a deadlock since the handle for such request loops indefinitely with the blockchain read lock taken. This makes the client to hang without printing any log entry.

viquezclaudio:
This prevents an issue that was observed in the PoS testnet, after the PoW testnet state was migrated.

This was caused by requests for block inclusion proofs of the genesis block, which caused the `get_interlink_hop` function to get into an infinite loop while holding the blockchain read lock.

This fixes #2421.

## Pull request checklist

### DIFF
--- a/consensus/src/consensus/consensus_proxy.rs
+++ b/consensus/src/consensus/consensus_proxy.rs
@@ -354,6 +354,7 @@ impl<N: Network> ConsensusProxy<N> {
                             let mut already_proven = false;
                             if election_head.hash() == block_hash
                                 || election_head.header.parent_election_hash == block_hash
+                                || response.block.block_number() == Policy::genesis_block_number()
                             {
                                 already_proven = true;
                             } else if let Some(ref interlink) = election_head.header.interlink {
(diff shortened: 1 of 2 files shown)


jsdanielh (approved):

sisou:
Could the interlink function be fixed instead? Why should we not allow history proofs for epoch 0?

viquezclaudio:
We still allow history proofs (transaction inclusion proofs) for epoch 0, what we do not allow (we don't need) is a block inclusion proof for the genesis, i.e: the genesis block doesn't need to be proved that it is part of the chain

sisou:
Ah, OK. That was an oversight then when these proofs were introduced or worked on last time. I remember us working on it February last year. Weird we didn't think of the genesis block.

In any case, can a check be added to the interlink generator, too?

sisou:
It feels like it's a simple matter of returning an empty vec in case the `target_number` is 0:
https://github.com/nimiq/core-rs-albatross/blob/ba0179f1db9b243ab33809685ea1763fa5154afb/primitives/block/src/block_proof.rs#L19

PR #2424, opened 2024-05-02
Fix merkle path generation for multisig signature proofs in web-client

sisou:
In #1884 I introduced a bug that lead to the multisig signature proof methods to always generate an empty merkle path, which then led to invalid transaction proofs, as the computed signer of the signature proof does not match the sender of the transaction.

## Pull request checklist

### DIFF
--- a/web-client/src/primitives/signature_proof.rs
+++ b/web-client/src/primitives/signature_proof.rs
@@ -38,24 +38,17 @@ impl SignatureProof {
         public_keys: &PublicKeyUnionArray,
         signature: &Signature,
     ) -> Result<SignatureProof, JsError> {
-        let mut public_keys: Vec<_> = SignatureProof::unpack_public_keys(public_keys)?
-            .into_iter()
-            .map(|k| match k {
-                nimiq_keys::PublicKey::Ed25519(ref public_key) => public_key.serialize_to_vec(),
-                nimiq_keys::PublicKey::ES256(ref public_key) => public_key.serialize_to_vec(),
-            })
-            .collect();
-        public_keys.sort_unstable();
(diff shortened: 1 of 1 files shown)


web-client/src/primitives/signature_proof.rs
```diff
@@ -123,7 +123,10 @@ impl SignatureProof {
 
         nimiq_utils::merkle::Blake2bMerklePath::new::<nimiq_hash::Blake2bHasher, Vec<u8>>(
             &public_keys,
-            &leaf_value.serialize_to_vec(),
+            &match leaf_value {
+                nimiq_keys::PublicKey::Ed25519(public_key) => public_key.serialize_to_vec(),
+                nimiq_keys::PublicKey::ES256(public_key) => public_key.serialize_to_vec(),
+            },
```
hrxi, line 129: This seems to imply that we do not distinguish between Ed25519/ES256 keys in the merkle tree. I know that this isn't the pull request introducing that, but something like that shouldn't be done in this cryptographic context.

hrxi (approved):
The pull request itself looks fine. Nice that you wrote a failing test first!

web-client/src/primitives/signature_proof.rs
sisou: We don't. If we'd do, we'd break all existing multisig as they would now hash to a different merkle root and thus address.

ES256 keys are 33 bytes long if that helps.

hrxi: That helps.

jsdanielh (approved):

jsdanielh:
Rebasing branch to merge it

PR #2425, opened 2024-05-02
Dont compute interlink hop for genesis

viquezclaudio:
## Pull request checklist

### DIFF
--- a/consensus/src/consensus/consensus_proxy.rs
+++ b/consensus/src/consensus/consensus_proxy.rs
@@ -352,9 +352,25 @@ impl<N: Network> ConsensusProxy<N> {
                         if response.block.block_number() <= election_head.block_number() {
                             let block_hash = response.block.hash();
                             let mut already_proven = false;
+
+                            if response.block.block_number() == Policy::genesis_block_number() {
+                                let genesis_hash = self
+                                    .blockchain
+                                    .read()
+                                    .get_block_at(Policy::genesis_block_number(), false)
+                                    .unwrap()
+                                    .hash();
+
+                                if genesis_hash == response.block.hash() {
+                                    already_proven = true;
+                                } else {
+                                    log::warn!(peer = %peer_id, "The genesis hash from the peer does not match our own");
+                                    continue;
+                                }
+                            }
+
                             if election_head.hash() == block_hash
                                 || election_head.header.parent_election_hash == block_hash
-                                || response.block.block_number() == Policy::genesis_block_number()
                             {
                                 already_proven = true;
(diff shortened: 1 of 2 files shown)


jsdanielh (approved):

consensus/src/consensus/consensus_proxy.rs
```diff
@@ -352,9 +352,25 @@ impl<N: Network> ConsensusProxy<N> {
...
+                                    .blockchain
+                                    .read()
+                                    .get_block_at(Policy::genesis_block_number(), false)
+                                    .unwrap()
+                                    .hash();
+
+                                if genesis_hash == response.block.hash() {
+                                    already_proven = true;
+                                } else {
+                                    log::warn!(peer = %peer_id, "The genesis hash from the peer does not match our own");
```
sisou, line 367: I believe a connection between peers with different genesis blocks can't even get past the initial handshake phase, so this situation _should_ be impossible.

viquezclaudio: Correct, but a malicious peer could establish a connection using the appropriate genesis but then, it could reply to this request using a different genesis to try to prove some txns that are not part of the chain.

Or maybe I'm being very paranoid?

sisou: And the idea here is then in this scenario, this continues but then fails validation somewhere else?

Could you add a test for this malicious behavior to make sure this actually correctly fails somewhere then?

primitives/block/src/block_proof.rs
```diff
@@ -19,6 +19,11 @@ impl BlockInclusionProof {
         let target_number = Policy::epoch_at(target);
         let latest_number = Policy::epoch_at(latest_election_number);
 
+        // This means the target is the genesis block.
+        if target_number == 0 {
+            return Vec::new();
+        }
```
hrxi, line 25: Why does this need to be special-cased? What's the motivation for this PR in general?

consensus/src/consensus/consensus_proxy.rs
hrxi: The "failed validation" is probably supposed to be the missing `already_proven = true;`. However, it seems that it'd go on to request block proofs for the genesis block in case of mismatch, that doesn't sound particularly useful.

primitives/block/src/block_proof.rs
sisou: @viquezclaudio found that computing the interlink for the genesis block gets stuck in an endless loop in #2420

PR #2427, opened 2024-05-03
Don't explicitly define rust toolchain input when using stable/nightly revs

Eligioo:
As stated in the [documentation](https://github.com/marketplace/actions/rustup-toolchain-install#inputs) of dtolnay/rust-toolchain, you should only define the `toolchain` input when your GH Action rev is set to `master`. This PR removes those inputs and set the correct revs.

Fixing this for the `npm_publish` action happens in #2417 

## Pull request checklist

### DIFF
--- a/.github/workflows/build+test.yml
+++ b/.github/workflows/build+test.yml
@@ -16,9 +16,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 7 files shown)


hrxi (approved):
LGTM, thanks!

PR #2442, opened 2024-05-08
WASM: add abstraction for `tokio::spawn`

### LINKED ISSUE #2434: Unify task executors for web based and not web clients
- Unify task executors for web based and not web clients.
- Re-add the task spawns for the message and request handlers in the network-interface that got removed in #2407.

styppo:
- Add a utility function `spawn` which calls `wasm_bindgen_futures::spawn_local` for WASM targets and `tokio::spawn` otherwise.
- Remove `TaskExecutor`.

Closes #2434.


sisou (approved):
Approved from a functional standpoint: web-client works

hrxi (approved):
Looks good to me. And it even has a negative diff!

utils/Cargo.toml
```diff
@@ -30,6 +30,8 @@ rand = { version = "0.8", optional = true }
 rand_core = { version = "0.6", optional = true }
 serde = "1.0"
 thiserror = { version = "1.0", optional = true }
+tokio = { version = "1.37.0", optional = true }
+wasm-bindgen-futures = { version = "0.4", optional = true }
```
nibhar: would it be possible to do something along the lines of:
```toml
[target.'cfg(target_family = "wasm")'.dependencies]
wasm-binggen-futures = { version = "0.4", optional = true }

[target.'cfg(not(target_family = "wasm"))'.dependencies]
tokio = { version = "1.37.0", optional = true }
```
as those dependencies are unnecessary otherwise. The exact syntax to facilitate this is not quite known to me though.

consensus/Cargo.toml
```diff
@@ -23,7 +23,7 @@ workspace = true
 async-trait = "0.1"
 futures = { package = "futures-util", version = "0.3" }
 futures-executor = { version = "0.3" }
-instant = { version = "0.1", features = [ "wasm-bindgen" ] }
+instant = { version = "0.1", features = ["wasm-bindgen"] }
```
jsdanielh, line 26: I think we would want to introduce a feature called `wasm_bindgen` that enables this other feature such that it doesn't always get pulled. Something like:
```
[features]
wasm-bindgen = ["instant/wasm-bindgen"]

File: utils/src/spawn/mod.rs [hunk 3]
```diff
@@ -0,0 +1,9 @@
+#[cfg(not(target_family = "wasm"))]
```
jsdanielh, line 1: I'd just use the `tokio` feature instead.

utils/src/spawn/mod.rs
```diff
@@ -0,0 +1,9 @@
...
+mod tokio;
+#[cfg(target_family = "wasm")]
+mod wasm;
+
+#[cfg(not(target_family = "wasm"))]
```
jsdanielh, line 6: I'd just use the `tokio` feature instead.

utils/src/spawn/mod.rs
```diff
@@ -0,0 +1,9 @@
...
+mod tokio;
+#[cfg(target_family = "wasm")]
```
jsdanielh, line 3: I'd just use the `wasm-bindgen-futures` feature instead.

utils/src/spawn/mod.rs
```diff
@@ -0,0 +1,9 @@
...
+mod wasm;
+
+#[cfg(not(target_family = "wasm"))]
+pub use tokio::*;
+#[cfg(target_family = "wasm")]
```
jsdanielh, line 8: I'd just use the `wasm-bindgen-futures` feature instead.

utils/Cargo.toml
jsdanielh: Yeah, and then I'd use the `wasm-bindgen-futures` or `tokio` features in the rest of the code to select what gets enabled.

styppo: @nibhar Including the dependencies based on the target works fine 👍 I added it in db345de.

styppo: @jsdanielh I don't think that using features instead of the configured target makes sense.

Generally, we want the codebase to be generic (i.e. platform-independent) as much as possible. This means that we want to keep the platform-specific code to a minimum and ideally keep it local as well. By using the target to select the correct implementation, we are encapsulating the platform-specific behaviour, containing it to this crate. The rest of the codebase can simply use the crate without having to worry about selecting the right features for the right target, keeping it generic. Using features would propagate the platform dependency to every consumer of this crate which is not what we want. The only advantage of using features would be that we get more control over which implementation is selected. However, there is not reasonable use case where we would want the tokio implementation for a wasm target or vice versa, so it would just result in unnecessary configuration overhead and more places to get wrong.

web-client/src/client/lib.rs
```diff
@@ -163,14 +163,9 @@ impl Client {
...
-            }),
-        )
-        .await
-        .expect("Client initialization failed");
+        let mut client = nimiq::client::Client::from_config(config)
+            .await
+            .expect("Client initialization failed");
         log::info!("Web client initialized");
 
         // Start consensus.
```
Eligioo, line 173: @styppo the `web-client` crate has a bunch of `wasm_bindgen::spawn_local`s. Those can get replaced with the new `spawn` abstraction too.

nibhar: In that case it is guaranteed for it to not be a `tokio::spawn` though as the web client will be compiled for the wasm target always. It could be done, but it would not add much I guess.

Eligioo: That's true but for consistency across the codebase using the abstraction as much makes sense. For example if you want to alter the impl behind the abstraction maybe later in the far future.

PR #2444, opened 2024-05-08
Remove unused `Argon2dHash` and `Argon2dHasher`

### LINKED ISSUE #2455: `Argon2dHasher` uses a predictable salt
Link:		https://hackerone.com/reports/2491434  
Date:		2024-05-06 06:41:10 UTC  
By:		ryanrb  
Weakness:	Use of a One-Way Hash with a Predictable Salt  
  
Details:  
## Summary
The `hash` lib uses a hardcoded, predictable salt of `nimiqrocks!` which is leveraged in `Argon2dHasher`. `Argon2dHasher` is then used in `web-client`.

The [original commit](https://github.com/nimiq/core-rs-albatross/commit/df77717d8cea4ba48795fa8c5d6a25e5bab696a6#) did not make any mention of why this approach was chosen, which may be able to add further context. In absence of that, it is assumed that this should be adjusted to a more secure approach.

Project: `core-rs-albatross`
File reference: `hash/src/lib.rs`
Line: 269
```

// Argon2d

const ARGON2D_LENGTH: usize = 32;
const NIMIQ_ARGON2_SALT: &str = "nimiqrocks!";
```

[source: Rust argon2 docs](https://docs.rs/argon2/latest/argon2/)

Ensure that the salt is a sufficiently random value. The docs for `argon2` make use of `let salt = SaltString::generate(&mut OsRng);` to generate a secure salt.

## References:
- [Rust argon2 docs](https://docs.rs/argon2/latest/argon2/)
- [CWE-760](https://cwe.mitre.org/data/definitions/760.html)

## Impact

Per [CWE-760](https://cwe.mitre.org/data/definitions/760.html):
> This makes it easier for attackers to pre-compute the hash value using dictionary attack techniques such as rainbow tables, effectively disabling the protection that an unpredictable salt would provide.

As was noted above, the original commit occurred 6 years ago on Jun 24, 2018. Given the open-source nature of this project, any potential weaknesses like this are amplified by the fact that they are freely discoverable on the open internet.

This implementation is public in `web-client`, but doesn't appear to be used, so the direct impact is difficult to assess.  However, current or future usages of it are susceptible to the risks laid out in CWE-760.

hrxi:
They were only exposed to the web-client, but I couldn't find any reference to this function.

The Argon2d/Argon2id implementation in `nimiq_hash::argon2kdf` remain.

Fixes #2455 by removing the unused code.


hrxi:
@sisou Am I correct that this function is also not used in the web client? I looked at https://github.com/nimiq/wallet, in the `albatross-testnet` branch.

sisou:
I exposed it for backward-compatibility with the PoW client, which also had that method. It's not used in our apps, but doesn't mean someone else doesn't use it. But it's probably a low-usage API. 
I think it's fine to remove as a breaking change with the Albatross client.

sisou:
Can you maybe expose an Argon2id method on the web-client API as a replacement?

hrxi:
> Can you maybe expose an Argon2id method on the web-client API as a replacement?

Done.

web-client/src/primitives/hash.rs
```diff
@@ -30,4 +25,36 @@ impl Hash {
         nimiq_hash::Hasher::digest(nimiq_hash::sha512::Sha512Hasher::default(), data)
             .serialize_to_vec()
     }
+
+    #[wasm_bindgen(js_name = computeNimiqArgon2d)]
+    pub fn compute_nimiq_argon2d(
```
sisou, line 48:
```suggestion
    pub fn compute_argon2d_kdf(
```

web-client/src/primitives/hash.rs
```diff
@@ -30,4 +25,36 @@ impl Hash {
...
+            password,
+            salt,
+            iterations,
+            derived_key_length,
+            Argon2Variant::Argon2d,
+        )?)
+    }
+
+    #[wasm_bindgen(js_name = computeNimiqArgon2d)]
+    pub fn compute_nimiq_argon2id(
```
sisou, line 74:
```suggestion
    pub fn compute_argon2id_kdf(
```

sisou (changes requested):
Since these are KDF functions, I'd name them as such.

Also, are you able to add rustdocs to the method, explaining their arguments? I wouldn't know what a `derived_key_length` is for example.

web-client/src/primitives/hash.rs
hrxi: Since we pass nimiq-specific parameters, and this is not the general-purpose argon2d, I think it's important to have `nimiq` in the name.

hrxi:
> Since these are KDF functions, I'd name them as such.

[Argon2](https://en.wikipedia.org/wiki/Argon2) is always a key derivation function, a special kind of hash function. I'm not explicitly mentioning that sha256 is a cryptographic hash function, so intuitively I wouldn't do it for Argon2 either.

> Also, are you able to add rustdocs to the method, explaining their arguments? I wouldn't know what a `derived_key_length` is for example.

I don't know it yet either, but I'll try to find out and document it.

sisou (approved):

jsdanielh (approved):
LGTM

PR #2450, opened 2024-05-09
Remove performance marks in web-client

### LINKED ISSUE #2443: Investigate increasing memory usage of the `web-client` in browsers
## New issue
I created this issue as a reference for the ever-growing memory usage of the `web-client` in browsers. I have only verified that in Chrome this issue is showing but I think it's safe to say that for other browsers this happens too.

After having the Nimiq Wallet tab open for an hour, 4.9 GB of memory usage was observed on my system by simply have it open and let it do nothing.

Screenshot 2024-05-08 at 14 33 16

Eligioo:

- Remove unbounded [Performance](https://developer.mozilla.org/en-US/docs/Web/API/Performance) measurements being created when every tracing span is entered or exited. This caused the memory consumption to grow rapidly as they never get cleaned up by default.
- Replace deprecated `tracing_web::MakeConsoleWriter` in favour of `tracing_web::MakeWebConsoleWriter`.
- Remove periodic clean-up of `Performance` measurements in NodeJS environment.

## Pull request checklist

### DIFF
--- a/lib/src/extras/web_logging.rs
+++ b/lib/src/extras/web_logging.rs
@@ -1,9 +1,7 @@
 use log::{level_filters::LevelFilter, Level};
 use nimiq_log::{Formatting, MaybeSystemTime, TargetsExt};
-use tracing_subscriber::{
-    filter::Targets, fmt::format::Pretty, layer::SubscriberExt, util::SubscriberInitExt, Layer,
-};
-use tracing_web::{performance_layer, MakeConsoleWriter};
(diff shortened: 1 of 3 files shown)


sisou (approved):

nibhar (approved):

sisou:
It's a crazy difference. I'm running the demo and it's staying around 90MB memory for over 20 minutes now, not endlessly growing! 🚀

sisou:
After over an hour, still stable at 90MB for the tab! :tada:

jsdanielh (approved):

PR #2461, opened 2024-05-09
network-libp2p: Add fixes and improvements to discovery protocol

### LINKED ISSUE #2458: DoS in Discovery Handler
Link:		https://hackerone.com/reports/2463319  
Date:		2024-04-15 08:36:39 UTC  
By:		d0nut  
Weakness:	HTTP DoS  
  
Details:  
## Summary
A malicious peer could wait until a victim peer entered the `HandlerState::ReceiveHandshakeAck` (or a few other handler states) and stop responding, leaving the client in a state where it perpetually waits for the next part of the handshake. This is because there is no timeout on connections during the handshake, leading to a Denial of Service scenario. 

(Ignore the selected weakness - no standard "Denial of Service" weakness existed for some reason).

File reference: `network-libp2p/src/discovery/handler.rs`
Line: 477

## Recommendation
Add a timeout on the discovery handler handshake that fails the handshake after a period of time.

## Impact

A malicious peer could intentionally stall out victim peers

jsdanielh:

- Add verification of peer contacts received from the different messages that can transmit a list of peer contacts in the discovery protocol. The verification includes checks for appropriate lengths, signatures and number of addresses.
- Add timeout for the state transition in the discovery handler such that a peer don't hold us in a state indefinitely if it doesn't send us back a message we expect.
  This fixes #2458.
- Do not use the observed addresses sent through the discovery protocol since these addresses are not authenticated and it's been observed that they are not adding any useful address to a peer and instead were adding addresses with incorrect TCP/UDP ports.
- Reduce the reported observed addresses to a single address within the discovery protocol. This avoids DoS attacks from a peer sending several observed addresses for a specific connection. In terms of functionality, it stays the same since we were only reporting a single observed address (the address that got us a connection). Also reduced the interaction between the handler and the behaviour by setting handler attributes on the constructor instead of using events to the handler.
- Add limits to the peer contact addresses and add the ability to verify that these limits are followed when a peer contact is received from the network (discovery protocol).

## Pull request checklist

### DIFF
--- a/network-libp2p/src/discovery/behaviour.rs
+++ b/network-libp2p/src/discovery/behaviour.rs
@@ -11,8 +11,7 @@ use libp2p::{
     identity::Keypair,
     swarm::{
         behaviour::{ConnectionClosed, ConnectionEstablished},
-        CloseConnection, ConnectionDenied, ConnectionId, FromSwarm, NetworkBehaviour,
-        NotifyHandler, ToSwarm,
+        CloseConnection, ConnectionDenied, ConnectionId, FromSwarm, NetworkBehaviour, ToSwarm,
     },
     Multiaddr, PeerId,
 };
@@ -22,7 +21,7 @@ use nimiq_time::{interval, Interval};
 use parking_lot::RwLock;
 
 use super::{
-    handler::{Handler, HandlerInEvent, HandlerOutEvent},
+    handler::{Handler, HandlerOutEvent},
     peer_contacts::{PeerContact, PeerContactBook},
 };
 
@@ -87,7 +86,7 @@ pub enum Event {
     Update,
 }
 
-type DiscoveryToSwarm = ToSwarm<Event, HandlerInEvent>;
+type DiscoveryToSwarm = ToSwarm<Event, ()>;
 
 /// Network behaviour for peer exchange.
 ///
@@ -170,13 +169,14 @@ impl NetworkBehaviour for Behaviour {
         _connection_id: ConnectionId,
         peer: PeerId,
         _local_addr: &Multiaddr,
-        _remote_addr: &Multiaddr,
+        remote_addr: &Multiaddr,
     ) -> Result<Handler, ConnectionDenied> {
         Ok(Handler::new(
             peer,
             self.config.clone(),
             self.keypair.clone(),
             self.peer_contact_book(),
+            remote_addr.clone(),
         ))
     }
 
@@ -184,7 +184,7 @@ impl NetworkBehaviour for Behaviour {
         &mut self,
         _connection_id: ConnectionId,
         peer: PeerId,
-        _addr: &Multiaddr,
+        addr: &Multiaddr,
         _role_override: Endpoint,
     ) -> Result<Handler, ConnectionDenied> {
         Ok(Handler::new(
@@ -192,6 +192,7 @@ impl NetworkBehaviour for Behaviour {
             self.config.clone(),
             self.keypair.clone(),
             self.peer_contact_book(),
+            addr.clone(),
         ))
     }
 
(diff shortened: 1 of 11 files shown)


network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -107,6 +125,15 @@ impl PeerContact {
         self.addresses
             .retain(|addr| !to_remove_addresses.contains(addr));
     }
+
+    /// Verifies wether the lengths of the advertised addresses are within
```
Eligioo:
```suggestion
    /// Verifies whether the lengths of the advertised addresses are within
```

network-libp2p/src/discovery/handler.rs
nibhar, line 1: I cannot  comment on it directly, but in the match arm for `HandlerState::ReceiveHandshakeAck` it begins with a `peer_contact` and `peer_contacts` as well. The `peer_contact` is `verify`ed, but for the `peer_contacts` I cannot find it. They could have too many addresses as well as I believe they are also received over the network and thus are not correctly sized by construction. I very well may be missing something here though.

jsdanielh: Good point. Added verification of the size and signatures in another commit.

network-libp2p/src/discovery/handler.rs
```diff
@@ -169,21 +165,23 @@ pub struct Handler {
 }
 
 impl Handler {
+    const STATE_TRANSITION_TIMEOUT_MS: u64 = 750;
```
hrxi: This is quite a short timeout, some packet loss might already reach that. Not sure what would be a good value.

```suggestion
    const STATE_TRANSITION_TIMEOUT: Duration = Duration::from_millis(5000);
```

network-libp2p/src/discovery/handler.rs
```diff
@@ -374,6 +376,10 @@ impl ConnectionHandler for Handler {
                     }
 
                     self.state = HandlerState::ReceiveHandshake;
+                    self.state_timeout
+                        .as_mut()
+                        .expect("state timeout must exist")
+                        .reset(Duration::from_millis(Self::STATE_TRANSITION_TIMEOUT_MS));
```
hrxi: Just another option without an `expect`. If you want to make sure there was a timeout for the previous state, too, then your version is better.
```suggestion
                    self.state_timeout = Some(Delay::new(Duration::from_millis(
                        Self::STATE_TRANSITION_TIMEOUT_MS,
                    )));
```

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -97,8 +111,12 @@ impl PeerContact {
...
 
     /// Adds a set of addresses
-    pub fn add_addresses(&mut self, addresses: Vec<Multiaddr>) {
-        self.addresses.extend(addresses)
+    pub fn add_addresses(&mut self, addresses: Vec<Multiaddr>) -> Result<(), PeerContactError> {
+        if addresses.len() + self.addresses.len() > Self::MAX_ADDRESSES {
+            return Err(PeerContactError::AdvertisedAddressesExceeded);
+        }
+        self.addresses.extend(addresses);
+        Ok(())
```
hrxi: Maybe this could add addresses until it overflows, so that at least some are recorded?

network-libp2p/tests/network.rs
```diff
@@ -352,9 +352,9 @@ async fn create_network_with_n_peers(n_peers: usize) -> Vec<Network> {
     networks
 }
 
-#[test(tokio::test)]
+#[test(tokio::test(flavor = "multi_thread", worker_threads = 10))]
 async fn connections_stress_and_reconnect() {
-    let peers: usize = 15;
+    let peers: usize = 10;
```
hrxi, line 321: Is there a reason for this change?

hrxi (commented):
Looks good to me, left some comments.

network-libp2p/tests/network.rs
jsdanielh: No reason. Changed it back.

network-libp2p/src/discovery/behaviour.rs
```diff
@@ -318,11 +304,9 @@ impl NetworkBehaviour for Behaviour {
...
-                for address in observed_addresses {
-                    self.events
-                        .push_back(ToSwarm::NewExternalAddrCandidate(address));
-                }
+            HandlerOutEvent::ObservedAddress { observed_address } => {
+                self.events
+                    .push_back(ToSwarm::NewExternalAddrCandidate(observed_address));
             }
             HandlerOutEvent::Update => self.events.push_back(ToSwarm::GenerateEvent(Event::Update)),
             HandlerOutEvent::Error(_) => self.events.push_back(ToSwarm::CloseConnection {
```
nibhar, line 312: If I understand correctly this is where the `StateTransitionTimeout` would end up. I may have missed it, but I think some logging would be nice here.

nibhar (approved):
LGTM

PR #2463, opened 2024-05-10
lib: Hide sensitive information for some structs

jsdanielh:
Hide sensitive information that was being leaked deriving `Debug` for the following structs in the `lib` subcrate:
- `Credentials`
- `NetworkSettings`
- `RpcServerSettings`
- `MetricsServerSettings`

This fixes #2459.

## Pull request checklist

- [X] All tests pass. The project builds and runs.
- [X] I have resolved any merge conflicts.
- [X] I have resolved all `clippy` and `rustfmt` warnings.

PR #2464, opened 2024-05-10
Introduce `Sensitive` which can be used for hiding `Debug` impls

### LINKED ISSUE #2459: Sensitive Information Disclosure via `Debug` implementation
Link:		https://hackerone.com/reports/2447137  
Date:		2024-04-04 13:24:20 UTC  
By:		d0nut  
Weakness:	Information Disclosure  
  
Details:  
## Summary
The `Credentials` struct in `lib/src/config/config.rs` implements `Debug`. If this struct is logged as-is, the `password` field will be logged as well. A pattern found in the codebase is to either implement a custom `Debug` implementation to replace any sensitive information with `"***"` instead.

The `Credentials` struct is used in a couple of structs that also implement `Debug`, but with custom `Debug` implementations to mitigate this risk.

The `NetworkSettings`, `RpcServerSettings`, and `MetricsServerSettings` structs all have a similar problem where they leak sensitive keys via `Debug` implementation. Unlike `Credentials`, though, they _do_ leak it via a `log::trace` line at `client/src/main.rs:L28`

File reference: `lib/src/config/config.rs`
Line: 524
```
#[derive(Clone, Debug, PartialEq, Eq)]
pub struct Credentials {
    /// Username
    pub username: String,
    /// Password
    pub password: String,
}
```

## Recommendation
Create `Debug` implementations for `Credentials`, `NetworkSettings`, `RpcServerSettings`, and `MetricsServerSettings` that obfuscates the sensitive information.

## Impact

The impact is that this struct is not safe by default from logging sensitive information. If it were added to a struct with a `Debug` implementation, it would gladly leak the password into the logs.

In the case of the `NetworkSettings`, etc structs, it _does_ log this information.

hrxi:
Use it everywhere where we used to manually implement `Debug` on types to hide information.

Alternative to #2463.

Fixes #2459.

### DIFF
--- a/lib/src/config/config_file/mod.rs
+++ b/lib/src/config/config_file/mod.rs
@@ -12,6 +12,7 @@ use nimiq_mempool::{
 use nimiq_network_interface::Multiaddr;
 use nimiq_primitives::{coin::Coin, networks::NetworkId};
 use nimiq_serde::Deserialize;
+use nimiq_utils::Sensitive;
 use thiserror::Error;
 use url::Url;
 
@@ -124,7 +125,7 @@ impl FromStr for ConfigFile {
 #[serde(deny_unknown_fields)]
 pub struct NetworkSettings {
     pub peer_key_file: Option<String>,
-    pub peer_key: Option<String>,
+    pub peer_key: Option<Sensitive<String>>,
 
     #[serde(default)]
     pub listen_addresses: Vec<String>,
@@ -241,7 +242,7 @@ pub struct RpcServerSettings {
     #[serde(default)]
     pub methods: Vec<String>,
     pub username: Option<String>,
-    pub password: Option<String>,
+    pub password: Option<Sensitive<String>>,
 }
 
 #[derive(Clone, Debug, Deserialize, Default)]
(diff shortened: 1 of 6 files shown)


lib/src/config/config_file/mod.rs
```diff
@@ -252,7 +253,7 @@ pub struct MetricsServerSettings {
     pub bind: Option<String>,
     pub port: Option<u16>,
     pub username: Option<String>,
```
sisou, line 255: If RPC username is sensitive, then metrics server username should, too?

hrxi:
Yea, exactly what @sisou says. I'm interested in whether we want to hide usernames or not. Previously, we'd hide them.

lib/src/config/config_file/mod.rs
hrxi: Done, but unsure if usernames should be hidden or not, see below.

hrxi:
After a quick discussion, we're showing all usernames now.

jsdanielh (approved):
LGTM

PR #2469, opened 2024-05-13
Add function to obtain genesis hash

### LINKED ISSUE #2466: Light client panics at `unwrap()` on an `Err` value
### General information

- Library version(s): v0.21.1
- Browser version(s): Chrome
- Devices/Simulators/Machine affected: Light client (WASM)
- Related issues: 8f11c5b25eaf9973aa56b1de812a222e64d79d26

#### Expected behavior

Light client does not panic.

#### Actual behavior

```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: BlockNotFound':
    core-rs-albatross/consensus/src/consensus/consensus_proxy.rs:361 
```

#### Steps to reproduce

Log into the PoS Testnet wallet with an account that had transactions in the PoW testnet, so the PoS Wallet queries transactions from before the genesis block.

image

viquezclaudio:
The genesis hash is used for various checks, it is now stored as part of the blockchain 
This fixes #2466

## Pull request checklist

### DIFF
--- a/blockchain-interface/src/abstract_blockchain.rs
+++ b/blockchain-interface/src/abstract_blockchain.rs
@@ -89,6 +89,9 @@ pub trait AbstractBlockchain {
     /// Fetches a given block, by its block number.
     fn get_block_at(&self, height: u32, include_body: bool) -> Result<Block, BlockchainError>;
 
+    /// Obtains the hash associated with the genesis block.
+    fn get_genesis_hash(&self) -> Blake2bHash;
+
     /// Fetches a given block, by its hash.
     fn get_block(&self, hash: &Blake2bHash, include_body: bool) -> Result<Block, BlockchainError>;
 
--- a/light-blockchain/src/abstract_blockchain.rs
+++ b/light-blockchain/src/abstract_blockchain.rs
@@ -63,6 +63,10 @@ impl AbstractBlockchain for LightBlockchain {
             .map(|chain_info| chain_info.head.clone())
     }
 
+    fn get_genesis_hash(&self) -> Blake2bHash {
+        self.genesis_block.hash()
+    }
+
     fn get_blocks(
         &self,
         start_block_hash: &Blake2bHash,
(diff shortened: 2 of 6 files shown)


jsdanielh (approved):
LGTM

blockchain-interface/src/abstract_blockchain.rs
```diff
@@ -89,6 +89,8 @@ pub trait AbstractBlockchain {
     /// Fetches a given block, by its block number.
     fn get_block_at(&self, height: u32, include_body: bool) -> Result<Block, BlockchainError>;
 
+    fn get_genesis_hash(&self) -> Blake2bHash;
```
Eligioo, line 93: Rustdocs

light-blockchain/src/abstract_blockchain.rs
```diff
@@ -63,6 +63,10 @@ impl AbstractBlockchain for LightBlockchain {
             .map(|chain_info| chain_info.head.clone())
     }
 
+    fn get_genesis_hash(&self) -> Blake2bHash {
+        self.genesis_block.hash()
```
hrxi, line 67: Why is it cached for the (normal) blockchain but not for the light blockchain?

sisou: The light blockchain stores the whole genesis block in `self`, whereas the normal blockchain does not. For the normal blockchain to query the hash each time would be resource intensive, whereas the light blockchain already has the block loaded and just fetches it's hash.

hrxi: I see. Thanks!

blockchain-interface/src/abstract_blockchain.rs
```diff
@@ -89,6 +89,9 @@ pub trait AbstractBlockchain {
     /// Fetches a given block, by its block number.
     fn get_block_at(&self, height: u32, include_body: bool) -> Result<Block, BlockchainError>;
 
+    /// Obtains the hash associated to the genesis block.
```
sisou:
```suggestion
    /// Obtains the hash associated with the genesis block.
```

PR #2475, opened 2024-05-16
Fix reward inherent hashes

### LINKED ISSUE #2408: Rpc call doesn't work for Reward transaction
getTransactionByHash RPC call doesn't work on Reward transaction, for example for the reward tx
{
                "type": "reward",
                "blockNumber": 3119250,
                "blockTime": 1714164925985,
                "validatorAddress": "NQ57 2F6C X3GB Y9B7 04U5 2BVA 4BVC M2T0 ELRL",
                "target": "NQ57 2F6C X3GB Y9B7 04U5 2BVA 4BVC M2T0 ELRL",
                "value": 7077840,
                "hash": "22333611d8908e568da2b25f9ffe88272b64dcbdd02baa2d0256676e744862f5"
            },

getTransactionByHash using the above hash yield no result.

hrxi:
The wrong hash function was used, instead `tx_hash` is correct.

Fixes #2408.

### DIFF
--- a/rpc-interface/src/types.rs
+++ b/rpc-interface/src/types.rs
@@ -653,21 +653,20 @@ pub enum Inherent {
 
 impl Inherent {
     pub fn try_from(hist_tx: HistoricTransaction) -> Option<Self> {
-        let hash = hist_tx.data.hash();
         Some(match hist_tx.data {
             HistoricTransactionData::Basic(_) => return None,
             HistoricTransactionData::Equivocation(_) => return None,
             HistoricTransactionData::Reward(RewardEvent {
-                validator_address,
-                reward_address,
+                ref validator_address,
+                ref reward_address,
                 value,
             }) => Inherent::Reward {
                 block_number: hist_tx.block_number,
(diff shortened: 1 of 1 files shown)


nibhar (approved):

rpc-interface/src/types.rs
```diff
@@ -653,21 +653,20 @@ pub enum Inherent {
...
+                ref validator_address,
+                ref reward_address,
                 value,
             }) => Inherent::Reward {
                 block_number: hist_tx.block_number,
                 block_time: hist_tx.block_time,
-                validator_address,
-                target: reward_address,
+                validator_address: validator_address.clone(),
+                target: reward_address.clone(),
```
sisou, line 667: Why is cloning better than using `ref`?

hrxi: Wanted to move the `.tx_hash()` into the branch, but that requires the full object without moving stuff out. Since `Address` is just 20 bytes, cloning doesn't do much work here anyway.

jsdanielh (approved):

PR #2479, opened 2024-05-17
Add genesis config hash as extra data to migration ready transaction

Eligioo:
* Fix PoW to PoS chain migration where the candidate block was also migrated into the PoS history store
* Migration ready transactions now include a genesis config hash as extra data. This way validator ready transactions that have a different starting point are not counted towards the total validator readiness.

## Pull request checklist

### DIFF
--- a/pow-migration/src/lib.rs
+++ b/pow-migration/src/lib.rs
@@ -12,9 +12,11 @@ use std::{
 
 use nimiq_database::DatabaseProxy;
 use nimiq_genesis_builder::config::GenesisConfig;
+use nimiq_hash::{Blake2bHasher, Hasher};
 use nimiq_keys::Address;
 use nimiq_primitives::networks::NetworkId;
 use nimiq_rpc::Client;
+use nimiq_serde::Serialize;
 use thiserror::Error;
 use tokio::time::sleep;
 
@@ -237,7 +239,8 @@ pub async fn migrate(
     }
 
     let mut reported_ready = false;
-    let genesis_config;
+    let mut genesis_config;
+    let genesis_config_hash;
 
     loop {
         let current_height = pow_client.block_number().await.unwrap();
@@ -302,7 +305,24 @@ pub async fn migrate(
             )
             .await?;
 
-            log::info!("PoS Genesis generation is completed");
+            // Sort vectors for a consistent hash digest
+            genesis_config.validators.sort();
+            genesis_config.stakers.sort();
+            genesis_config.basic_accounts.sort();
+            genesis_config.vesting_accounts.sort();
+            genesis_config.htlc_accounts.sort();
+
+            let mut hasher = Blake2bHasher::new();
+            genesis_config
+                .serialize_to_writer(&mut hasher)
+                .unwrap_or_else(|error| {
+                    exit_with_error(error, "Failed to serialize genesis config")
+                });
+            genesis_config_hash = hasher.finish();
+            log::info!(
+                genesis_config_hash = %genesis_config_hash,
+                "PoS Genesis generation is completed"
+            );
 
             break;
         }
@@ -337,6 +357,7 @@ pub async fn migrate(
                 pow_client,
                 validator_address.to_user_friendly_address(),
                 candidate_block..next_candidate,
+                &genesis_config_hash,
             )
             .await;
 
@@ -347,7 +368,10 @@ pub async fn migrate(
                     "We didn't find a ready transaction from our validator in this window"
                 );
                 // Report we are ready to the Nimiq PoW chain:
-                let transaction = generate_ready_tx(validator_address.to_user_friendly_address());
+                let transaction = generate_ready_tx(
+                    validator_address.to_user_friendly_address(),
+                    &genesis_config_hash,
+                );
 
                 match send_tx(pow_client, transaction).await {
                     Ok(_) => reported_ready = true,
@@ -364,6 +388,7 @@ pub async fn migrate(
             pow_client,
             validators.clone(),
             candidate_block..next_candidate,
+            &genesis_config_hash,
         )
         .await;
         match validators_status {
@@ -418,12 +443,12 @@ pub fn launch_pos_client(
     // Check that we were able to launch the PoS client
     match child.try_wait() {
         Ok(Some(status)) => {
-            log::error!(%status, "Pos client unexpectedly exited");
+            log::error!(%status, "PoS client unexpectedly exited");
             Err(Error::PoSUnexpectedExit(status))
         }
         Ok(None) => {
             let pid = child.id();
-            log::info!(pid, "Pos client running");
+            log::info!(pid, "PoS client running");
             Ok(child.wait()?)
         }
         Err(error) => {
--- a/pow-migration/src/monitor/mod.rs
+++ b/pow-migration/src/monitor/mod.rs
@@ -3,6 +3,7 @@ pub mod types;
 use std::ops::Range;
 
 use log::{error, info};
+use nimiq_hash::Blake2bHash;
 use nimiq_keys::Address;
 use nimiq_primitives::coin::Coin;
 use nimiq_rpc::{
@@ -24,11 +25,12 @@ pub const READY_PERCENTAGE: u8 = 80;
 ///   Sender: Validator address
 ///   Recipient: Burn address
 ///   Value: 1 Luna
-///   Data: TBD
+///   Data: Hash of the generated `GenesisConfig`
 ///
-pub fn generate_ready_tx(validator: String) -> OutgoingTransaction {
+pub fn generate_ready_tx(validator: String, hash: &Blake2bHash) -> OutgoingTransaction {
     info!(
         validator_address = validator,
+        pos_genesis_hash = %hash,
         "Generating ready transaction"
     );
     OutgoingTransaction {
@@ -36,7 +38,7 @@ pub fn generate_ready_tx(validator: String) -> OutgoingTransaction {
         to: Address::burn_address().to_user_friendly_address(),
         value: 1, //Lunas
         fee: 0,
-        data: None,
+        data: Some(hash.to_hex()),
     }
 }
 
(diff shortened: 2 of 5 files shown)


pow-migration/src/lib.rs
```diff
@@ -302,7 +305,24 @@ pub async fn migrate(
...
+            genesis_config.vesting_accounts.sort();
+            genesis_config.htlc_accounts.sort();
+
+            let mut hasher = Blake2bHasher::new();
+            genesis_config
+                .serialize_to_writer(&mut hasher)
+                .unwrap_or_else(|error| {
+                    exit_with_error(error, "Failed to serialize genesis config")
+                });
+            genesis_block_hash = hasher.finish();
```
sisou: Is that actually the hash of the genesis block as it will be written to the database? Or just the hash of the genesis _config_?

Eligioo: My original goal was to compute the actual genesis block hash but it turned out that constructing it when it's not in the store yet isn't that straightforward and so I switched to simply hashing the `GenesisConfig`. Which achieves the same goal. But didn't properly update the naming.

So to answer your question yes you are right, it's the hash of the config and not of the actual block ;)

sisou: Hmm.. that's then not useful after the migration anymore. Would be nice to have the actual block hash, to actually have a connection between the two chains.

Eligioo: It's not designed to be used after the migration. But rather an extra filter for honest migrators to prevent that ready transactions that get send, by other validators, just before the activation window closes, accidentally fall into the next window.

This hash is by no means trustworthy in the sense that it doesn't actually have to mean that you've migrated. You can send this transaction regardless and simply mimic the hash that other validators are broadcasting.

What would be your use case to know the actual genesis block hash during the migration?

sisou: I get that for the purpose it serves (syncing between validators) this fits the bill perfectly. 

Just, from a bigger picture, for me it would have sentimental and "proof" value if we could write the genesis block of the PoS chain into the end of the PoW chain. It _connects_ the two and makes it more of a continuation.

sisou: ~~Unfortunately we cannot do it the other way round and write the transition block hash into the PoS chain, as the extra data field of the genesis block is already used to encode the genesis supply amount.~~

sisou: Ohhh! Actually, I just found that the genesis block's `parentHash` is the PoW chain's transition block! Cool! :sunglasses:

Eligioo: Yes it's an actual continuation. 
`parent_election_hash` = hash of the parent election block before the genesis block, aka PoW genesis block hash
`parent_hash` = hash of the parent block before the genesis block, aka the last PoW block hash
`history_root` = root of all the PoW transactions.

pow-migration/src/monitor/mod.rs
```diff
@@ -24,19 +25,20 @@ pub const READY_PERCENTAGE: u8 = 80;
...
 ///   Recipient: Burn address
 ///   Value: 1 Luna
-///   Data: TBD
+///   Data: Hash of the generated `GenesisConfig`
 ///
-pub fn generate_ready_tx(validator: String) -> OutgoingTransaction {
+pub fn generate_ready_tx(validator: String, hash: &Blake2bHash) -> OutgoingTransaction {
     info!(
         validator_address = validator,
+        pos_genesis_hash = hash.to_hex(),
```
jsdanielh: `Blake2bHash` implements `Display`
```suggestion
        pos_genesis_hash = %hash,
```

pow-migration/src/monitor/mod.rs
```diff
@@ -24,19 +25,20 @@ pub const READY_PERCENTAGE: u8 = 80;
...
         "Generating ready transaction"
     );
     OutgoingTransaction {
         from: validator,
         to: Address::burn_address().to_user_friendly_address(),
         value: 1, //Lunas
         fee: 0,
-        data: None,
+        data: Some(hash.to_hex()),
```
jsdanielh, line 41: Which is the reason why we want this in `hex` rather than the raw hash?

pow-migration/src/monitor/mod.rs
```diff
@@ -45,12 +47,24 @@ pub async fn get_ready_txns(
...
     if let Ok(transactions) = client.get_transactions_by_address(&validator, 10).await {
+        let genesis_config_hex = pos_genesis_config_hash.to_hex();
+
         let filtered_txns: Vec<TransactionDetails> = transactions
             .into_iter()
             .filter(|txn| {
-                // Here we filter by current epoch
+                // Check if the txn contains extra data and matches our genesis block hash
+                if let Some(txn_data) = &txn.data {
+                    if *txn_data != genesis_config_hex {
```
jsdanielh: Same here, why in `hex`?

pow-migration/src/monitor/mod.rs
```diff
@@ -111,7 +127,16 @@ pub async fn check_validators_ready(
             let filtered_txns: Vec<TransactionDetails> = transactions
                 .into_iter()
                 .filter(|txn| {
-                    // Here we filter by the readiness criteria, TBD
+                    // Check if the txn contains extra data and matches our genesis block hash
+                    if let Some(txn_data) = &txn.data {
+                        if *txn_data != genesis_config_hex {
```
jsdanielh: Same here, why in `hex`?

pow-migration/src/lib.rs
```diff
@@ -302,7 +305,24 @@ pub async fn migrate(
...
+
+            let mut hasher = Blake2bHasher::new();
+            genesis_config
+                .serialize_to_writer(&mut hasher)
+                .unwrap_or_else(|error| {
+                    exit_with_error(error, "Failed to serialize genesis config")
+                });
+            genesis_config_hash = hasher.finish();
+            log::info!(
+                genesis_config_hash = genesis_config_hash.to_hex(),
```
jsdanielh: `Blake2bHash` implements `Display`
```suggestion
                %genesis_config_hash,
```

pow-migration/src/monitor/mod.rs
```diff
@@ -45,12 +47,24 @@ pub async fn get_ready_txns(
...
 ) -> Vec<TransactionDetails> {
     if let Ok(transactions) = client.get_transactions_by_address(&validator, 10).await {
+        let genesis_config_hex = pos_genesis_config_hash.to_hex();
+
         let filtered_txns: Vec<TransactionDetails> = transactions
             .into_iter()
             .filter(|txn| {
-                // Here we filter by current epoch
+                // Check if the txn contains extra data and matches our genesis block hash
+                if let Some(txn_data) = &txn.data {
```
jsdanielh: I'd suggest to factorize the txn data check in a function to avoid code duplication

pow-migration/src/monitor/mod.rs
```diff
@@ -111,7 +127,16 @@ pub async fn check_validators_ready(
             let filtered_txns: Vec<TransactionDetails> = transactions
                 .into_iter()
                 .filter(|txn| {
-                    // Here we filter by the readiness criteria, TBD
+                    // Check if the txn contains extra data and matches our genesis block hash
+                    if let Some(txn_data) = &txn.data {
```
jsdanielh: I'd suggest to factorize the txn data check in a function to avoid code duplication

pow-migration/src/monitor/mod.rs
Eligioo: The PoW RPC specs expects `OutgoingTransaction.data` to be a `String`. A raw `Blake2bHash` doesn't implement that. I could have gone for `to_string` but `to_hex` is something we control and have defined which is a little more consistent.

Eligioo:
@jsdanielh addressed feedback and left a comment

pow-migration/src/monitor/mod.rs
```diff
@@ -24,19 +25,20 @@ pub const READY_PERCENTAGE: u8 = 80;
 ///   Sender: Validator address
 ///   Recipient: Burn address
 ///   Value: 1 Luna
-///   Data: TBD
+///   Data: Hash of the generated `GenesisConfig`
```
jsdanielh, line 28: @ines-toupeira We might need to update the documentation to reflect this.

pow-migration/src/monitor/mod.rs
```diff
@@ -45,23 +47,43 @@ pub async fn get_ready_txns(
...
         Vec::new()
     }
 }
 
+/// Checks if the provided transaction meets the criteria in order to be
+/// considered a valid ready-transaction
+fn is_valid_ready_txn(
+    txn: &TransactionDetails,
+    block_window: &Range<u32>,
+    genesis_config_hex: &String,
```
jsdanielh: I think I'd prefer to clarify this is a hash rather than clarifying the representation which I think we can assume it would be in `hex` given the type.
```suggestion
    genesis_config_hash: &String,
```

pow-migration/src/monitor/mod.rs
```diff
@@ -92,6 +115,7 @@ pub async fn check_validators_ready(
     log::debug!(registered_stake = %total_stake);
 
     let mut ready_validators = Vec::new();
+    let genesis_config_hex = pos_genesis_config_hash.to_hex();
```
jsdanielh:
```suggestion
    let genesis_config_hash_hex = pos_genesis_config_hash.to_hex();
```

pow-migration/src/monitor/mod.rs
```diff
@@ -45,23 +47,43 @@ pub async fn get_ready_txns(
     client: &Client,
     validator: String,
     block_window: Range<u32>,
+    pos_genesis_config_hash: &Blake2bHash,
 ) -> Vec<TransactionDetails> {
     if let Ok(transactions) = client.get_transactions_by_address(&validator, 10).await {
+        let genesis_config_hex = pos_genesis_config_hash.to_hex();
```
jsdanielh:
```suggestion
        let genesis_config_hash_hex = pos_genesis_config_hash.to_hex();
```

pow-migration/src/monitor/mod.rs
sisou: I think it would actually be better to take the hash type here, not a string. Then you don't need to cast to hex when calling this method, only inside this method when comparing to the tx data.

jsdanielh: > I think it would actually be better to take the hash type here, not a string. Then you don't need to cast to hex when calling this method, only inside this method when comparing to the tx data.

But then we would be calling `.to_hex()` every single time for every transaction we check

jsdanielh (approved):
LGTM

PR #2480, opened 2024-05-17
Unsubscribe from subscribed address when transaction sending fails

### LINKED ISSUE #2456: Sending a transaction in `web-client` may not always unsubscribe from address
Link:		https://hackerone.com/reports/2491408  
Date:		2024-05-06 06:08:42 UTC  
By:		ryanrb  
Weakness:	Improper Resource Shutdown or Release  
  
Details:  
## Summary
In `web-client`, the method `send_transaction` will conditionally subscribe to an address based on specific conditions. Later in the method, the code ensures that this is cleaned up by unsubscribing from the address.

However, the actual method for sending a transaction uses the `?` operator, which means the operation can fail and return at that point—potentially causing the method to return after subscribing but before cleaning up the subscription.

### `consensus.send_transaction` is invoked with `?`

Project: `core-rs-albatross`
File reference: `web-client/src/client/lib.rs`
Line: 555
```
// Actually send the transaction
consensus.send_transaction(tx.native()).await?;
```

### The address is not unsubscribed from until line 574

Project: `core-rs-albatross`
File reference: `web-client/src/client/lib.rs`
Line: 578
```
// Unsubscribe from any address we subscribed to, without caring about the result
if let Some(address) = subscribed_address {
    let owned_consensus = consensus.clone();
    spawn_local(async move {
        let _ = owned_consensus
            .unsubscribe_from_addresses(vec![address], 1)
            .await;
    });
}
```

## Recommendation
Ensure that the address is unsubscribed from regardless of the success or failure of the outcome of `send_transaction`.

## References:
- [CWE-404](https://cwe.mitre.org/data/definitions/404.html)

## Impact
Depending on the amount of resources that subscribing to an address takes up, enough failed transactions could lead to resource exhaustion over time if nothing else cleans up the subscription.

sisou:

Unsubscribe from a previously subscribed address during tx-sending in the web-client. The address is subscribed to to get notified of the inclusion of the transaction in the chain, then unsubscribed at the end of the function. However, `send_transaction` itself can fail, in which case the subscription was not cleaned up.

@hrxi This currently is simply a duplication of the code found at the end of the function. Do you know a better way that only defines the logic once and then calls it from both places? Or another means to achieve this same result?

## Pull request checklist

### DIFF
--- a/web-client/src/client/lib.rs
+++ b/web-client/src/client/lib.rs
@@ -532,8 +532,10 @@ impl Client {
         let consensus = self.inner.consensus_proxy();
 
         struct AddressSubscription {
-            address: nimiq_keys::Address,
-            consensus: ConsensusProxy,
+            // Use Options to be able to take the address and consensus out of the
+            // struct during drop, avoiding another clone.
+            address: Option<nimiq_keys::Address>,
+            consensus: Option<ConsensusProxy>,
         }
         impl AddressSubscription {
             pub async fn subscribe(
@@ -543,13 +545,16 @@ impl Client {
                 consensus
                     .subscribe_to_addresses(vec![address.clone()], 1, None)
                     .await?;
-                Ok(Self { address, consensus })
+                Ok(Self {
+                    address: Some(address),
+                    consensus: Some(consensus),
+                })
             }
         }
         impl Drop for AddressSubscription {
             fn drop(&mut self) {
-                let address = self.address.clone();
-                let consensus = self.consensus.clone();
+                let address = self.address.take().unwrap();
+                let consensus = self.consensus.take().unwrap();
                 // Unsubscribe from the address without caring about the result
                 spawn_local(async move {
                     let _ = consensus.unsubscribe_from_addresses(vec![address], 1).await;
--- a/web-client/src/client/lib.rs
+++ b/web-client/src/client/lib.rs
@@ -4,6 +4,7 @@ use std::{
         hash_map::{Entry, HashMap},
         HashSet,
     },
+    mem,
     rc::Rc,
     str::FromStr,
     time::Duration,
@@ -15,6 +16,7 @@ use futures::{
 };
 use js_sys::{global, Array, Function, JsString};
 use log::level_filters::LevelFilter;
+use nimiq::client::ConsensusProxy;
 pub use nimiq::{
     config::{
         config::ClientConfig,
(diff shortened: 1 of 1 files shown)


web-client/src/client/lib.rs
```diff
@@ -552,7 +552,21 @@ impl Client {
...
+                if let Some(address) = subscribed_address.clone() {
+                    let owned_consensus = consensus.clone();
+                    spawn_local(async move {
+                        let _ = owned_consensus
+                            .unsubscribe_from_addresses(vec![address], 1)
+                            .await;
+                    });
+                }
+                error
+            })?;
```
jsdanielh: Mmm, my $0.02 is that this is very error prone if we add more code between these lines and the place where we do the unsusbscribe below plus we are duplicating code. Maybe we could store this in a `struct` and implement `Drop` for this `struct`?

sisou: Oh, that's brilliant! 👍

hrxi: You can also see the pattern in 297a99a0b1531b703b6332132842410d51cf33f9.

sisou: Done.

web-client/src/client/lib.rs
```diff
@@ -526,23 +528,47 @@ impl Client {
...
+                consensus
+                    .subscribe_to_addresses(vec![address.clone()], 1, None)
+                    .await?;
+                Ok(Self { address, consensus })
+            }
+        }
+        impl Drop for AddressSubscription {
+            fn drop(&mut self) {
+                let address = self.address.clone();
+                let consensus = self.consensus.clone();
```
nibhar: If these are both options you can `.take()` instead of `.clone()`. After they are moved the struct is droped and the `None` vanish, so there is no real need for the clones. Sadly as far as i know there is otherwise only `unsafe` ways of taking ownership here.

web-client/src/client/lib.rs
```diff
@@ -569,15 +595,7 @@ impl Client {
...
-        // Unsubscribe from any address we subscribed to, without caring about the result
-        if let Some(address) = subscribed_address {
-            let owned_consensus = consensus.clone();
-            spawn_local(async move {
-                let _ = owned_consensus
-                    .unsubscribe_from_addresses(vec![address], 1)
-                    .await;
-            });
-        }
+        mem::drop(address_subscription);
```
nibhar, line 603: This will be implicitly executed as soon as the scope ends. Is there reason to do it earlier than that?

sisou: Otherwise Clippy complains the variable is never used and suggests to prefix it with an underscore. But then I fear it is compiled out?

web-client/src/client/lib.rs
sisou: So is your suggestion to make them options?

web-client/src/client/lib.rs
nibhar: The call to the drop implementation should make it not unused in the resulting representation. Thus even if underscored it should not get optimized away as it has observable behaviour attached to it (the drop impl). 

I am not entirely sure however if that is in fact true, so maybe keep it like it is.

web-client/src/client/lib.rs
nibhar: Yes I would do that. It is a very small change and makes it more clean in my opinion.

web-client/src/client/lib.rs
hrxi: > The call to the drop implementation should make it not unused in the resulting representation. Thus even if underscored it should not get optimized away as it has observable behaviour attached to it (the drop impl).
> 
> I am not entirely sure however if that is in fact true, so maybe keep it like it is.

Yes, if underscored, it doesn't get optimized away. Note that this is different from the name which only consists of an underscore `_` which means "immediately drop this".

I think the `mem::drop` is still fine, this way this PR doesn't change when the address is unsubscribed from in the happy case.

nibhar: Yeah I don't mind it either. More of a curiosity ask.

nibhar (approved):

jsdanielh (approved):
LGTM

PR #2218, opened 2024-02-15
Update web-client transaction-getter to handle known transactions like the PoW client did

sisou:
The PoW client by @mar-v-in did some [genius handling of known transactions](https://github.com/nimiq/core-js/blob/fad56cdf5a1cb13ca61f867b46048b0e83e85d94/src/main/generic/api/Client.js#L719-L839), which can be optionally passed to the client's `getTransactionsByAddress` method.

It, and now this client too, make sure all known transactions are included in the receipts, unconfirmed known transactions are queried again (and confirmed if they are, returned as expired if they have expired), pending or unsent transactions are re-broadcast, known transactions not in the receipts are also rechecked.

## Pull request checklist

- [x] All tests pass. The project builds and runs.
- [x] I have resolved any merge conflicts.
- [x] I have resolved all `clippy` and `rustfmt` warnings.

PR #2484, opened 2024-05-20
docs: added `known_transaction_details` clarification

onmax:

I have been playing with the web client and it has taken me longer than I would like to admit to work out what this parameter does.

If we add this clarification, developers will have better context when using this feature.

## Pull request checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -147,9 +147,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.83"
+version = "1.0.86"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "25bdb32cbbdce2b519a9cd7df3a678443100e265d5e25ca763b7572a5104f5f3"
+checksum = "b3d1d046238990b9cf5bcde22a3fb3584ee5cf65fb2765f454ed428c7a0063da"
 
 [[package]]
 name = "ark-crypto-primitives"


sisou:
There's more changes coming with https://github.com/nimiq/core-rs-albatross/pull/2218.

Inspiration for docs can also come from the description of the original implementation: 
https://github.com/nimiq/core-js/blob/fad56cdf5a1cb13ca61f867b46048b0e83e85d94/src/main/generic/api/Client.js#L719-L839

Maybe this can become part of #2218?

onmax:
Note apart: In the future we could use objects as parameters.

Is it hard to do? Maybe use function overriding? (not sure if it possible) 

I am thinking of something like this:

```ts
interface GetTransactionsByAddressOptions {
    sinceBlockHeight?: number,
    knownTransactionDetails?: PlainTransactionDetailsArrayType,
    limit? : number
    minPeers?: number
}
getTransactionsByAddress(address, options?: GetTransactionsByAddressOptions) {}
```

Cargo.lock
```diff
@@ -3118,12 +3118,9 @@ dependencies = [
...
 [[package]]
 name = "log"
-version = "0.4.17"
+version = "0.4.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "abb12e687cfb44aa40f41fc3978ef76448f9b6038cad6aef4259d3c095a2382e"
-dependencies = [
- "cfg-if",
-]
+checksum = "90ed8c1e510134f979dbc4f070f87d4313098b704861a105fe34231c70a3901c"
```
jsdanielh: This shouldn't be part of the PR since it's already in `albatross`. Did you miss to rebase `albatross`?

onmax: I rebased with github. I will fix it

onmax: Fixed

jsdanielh (approved):

hrxi (approved):
LGTM.

PR #2486, opened 2024-05-20
Make password comparisons constant-time

### LINKED ISSUE #2452: Possible Basic Auth Timing Attack in RPC server via `nimiq_jsonrpc_server`
Link:		https://hackerone.com/reports/2491447  
Date:		2024-05-06 07:18:39 UTC  
By:		ryanrb  
Weakness:	Information Exposure Through Timing Discrepancy  
  
Details:  
## Summary
Similar to [#2447113](https://hackerone.com/reports/2447113), the basic auth implementation used in `lib/src/extras/rpc_server.rs` leverages `nimiq_jsonrpc_server`, which does a simplistic string comparison of passwords.

Project: `core-rs-albatross`
File reference: `lib/src/extras/rpc_server.rs`
Line: 67
```
use nimiq_jsonrpc_server::{AllowListDispatcher, Config, ModularDispatcher, Server as _Server};
// ...
pub type Server = _Server<AllowListDispatcher<ModularDispatcher>>;
//...

Ok(Server::new(
    Config {
        bind_to: (config.bind_to.unwrap_or_else(default_bind), config.port).into(),
        enable_websocket: false,
        ip_whitelist: None,
        basic_auth,
    },
    AllowListDispatcher::new(dispatcher, allowed_methods),
))
```

File reference: nimiq_jsonrpc_server:server/lib.rs ([link](https://github.com/nimiq/jsonrpc/blob/ab641be80e8ec2080c7f8268a236d35e7739d217/server/src/lib.rs#L180))
Line: 180
```
if auth_header.0.username() == basic_auth.username
    && auth_header.0.password() == basic_auth.password
```
## Recommendation
Use a time-safe comparison when it comes to the password check to avoid leaking information.

## References:
- [#2447113](https://hackerone.com/reports/2447113)
- [Github: Nimiq jsonrpc](https://github.com/nimiq/jsonrpc/tree/master)

## Impact

The RPC server's basic auth password could potentially be compromised via timing analysis if the check in `nimiq_jsonrpc_server` is used as-is.

### LINKED ISSUE #2460: Timing Attack Possible in `Credentials::check` implementation
Link:		https://hackerone.com/reports/2447113  
Date:		2024-04-04 12:51:10 UTC  
By:		d0nut  
Weakness:	Information Exposure Through Timing Discrepancy  
  
Details:  
## Summary
In `lib/src/config/config.rs` exists a `Credential` struct which holds both a `username` and a `password` field. This struct has a `check` method which takes in a supplied `username` and `password`. The implementation does a simplistic string comparison. Raw string comparisons like this are vulnerable to timing attacks. 

For example if the actual password was `"hunter2"` and the supplied password was `"unknown"` then the equals-based comparison here will terminate on the first byte comparison between `h` and `u`. The invocation of the `check` function would terminate quicker as a result. However, if the supplied password was `"hunter3"`, then the check terminates on the final byte comparison between `2` and `3`, which takes longer as the first 6 bytes all passed the comparison.

Notably the timing difference here is _miniscule_. However there _is_ a measurable difference which, under sufficient testing, can leak information via statistically significant differences over a sufficient timeframe. 

Another aspect to note is that it didn't appear that `::check(..)` was used _currently_, but may be used in the future. I wanted to call this out regardless just in case it was used sometime in the future. 

File reference: `lib/src/config/config.rs`
Line: 539
```rust
pub fn check<U: AsRef<str>, P: AsRef<str>>(&self, username: U, password: P) -> bool {
    self.username == username.as_ref() && self.password == password.as_ref()
}
```
## Recommendation
Use a time-safe comparison when it comes to the password check to avoid leaking information.

## References:
https://security.stackexchange.com/questions/83660/simple-string-comparisons-not-secure-against-timing-attacks

## Impact

If this check function was used in the future it could provide a way to leak the password via a statistically-based timing attack

hrxi:
Do that by hashing the passwords using blake2b and then using `subtle` to do a constant-time comparison.

Fixes #2452.
Fixes #2460.

Requires https://github.com/nimiq/jsonrpc/pull/26.

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -36,6 +36,7 @@ serde = "1.0"
 serde_json = { version = "1.0", features = ["preserve_order"], optional = true }
 signal-hook = { version = "0.3", optional = true }
 strum_macros = "0.26"
+subtle = "2.5"
 thiserror = "1.0"
 time = { version = "0.3", optional = true }
 tokio = { version = "1.37", features = ["rt"], optional = true }
@@ -53,6 +54,7 @@ nimiq-bls = { workspace = true }
 nimiq-consensus = { workspace = true, default-features = false }
 nimiq-database = { workspace = true, optional = true }
 nimiq-genesis = { workspace = true, default-features = false }
+nimiq-hash = { workspace = true }
 nimiq-jsonrpc-core = { workspace = true, optional = true }
 nimiq-jsonrpc-server = { workspace = true, optional = true }
 nimiq-keys = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2211,7 +2211,7 @@ dependencies = [
  "httpdate",
  "itoa",
  "pin-project-lite",
- "socket2 0.5.5",
+ "socket2 0.4.9",
  "tokio",
  "tower-service",
  "tracing",
@@ -2230,10 +2230,13 @@ dependencies = [
  "h2 0.4.4",
  "http 1.1.0",
  "http-body 1.0.0",
+ "httparse",
  "httpdate",
+ "itoa",
  "pin-project-lite",
  "smallvec",
  "tokio",
+ "want",
 ]
 
 [[package]]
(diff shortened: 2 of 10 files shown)


lib/Cargo.toml
```diff
@@ -55,6 +56,7 @@ nimiq-database = { workspace = true, optional = true }
 nimiq-genesis = { workspace = true, default-features = false }
 nimiq-jsonrpc-core = { workspace = true, optional = true }
 nimiq-jsonrpc-server = { workspace = true, optional = true }
+nimiq-hash = { workspace = true }
```
jsdanielh: Let's keep the alphabetical order:
```suggestion
nimiq-genesis = { workspace = true, default-features = false }
nimiq-hash = { workspace = true }
nimiq-jsonrpc-core = { workspace = true, optional = true }
nimiq-jsonrpc-server = { workspace = true, optional = true }
```

jsdanielh (commented):
Just a small suggestion. LGTM otherwise

lib/Cargo.toml
hrxi: I just can't do the alphabet in my head. Or at least I made an error at that time. ^^

hrxi: Fixed.

jsdanielh (approved):

PR #2512, opened 2024-05-22
Calculate a transaction's related addresses and expose them via RPC

### LINKED ISSUE #2252: Feature request - expose an interface to retrieve all affected addresses from a transaction
Current transactions returned over RPC include the `from` and `to` fields but these fields do not always cover all affected addresses in the transaction.

For example:
* given pool X 
* having delegated staker A 
* pool X adds stake to staker A 
* the transaction `from` field is set to pool X and the recipient is set to the staking contract. The address of staker A has to be parsed from `recipientData`

This issue requests either of two things:
* include the `staker_address` in the transaction, so we always know which staker is affected by a staking contract transaction and / or 
* expose an RPC method to retrieve the affected addresses for a transaction. Apparently a transaction struct already has a function to retrieve affected addresses which may be something that can be exposed over RPC so other languages don't have to manually parse convoluted fields to retrieve the required information.

sisou:
## Pull request checklist

### DIFF
--- a/primitives/transaction/src/lib.rs
+++ b/primitives/transaction/src/lib.rs
@@ -3,9 +3,9 @@ extern crate log;
 
 use std::{
     cmp::{Ord, Ordering},
+    collections::BTreeSet,
     convert::TryFrom,
-    io,
-    io::Write,
+    io::{self, Write},
     sync::Arc,
 };
 
@@ -23,7 +23,12 @@ use nimiq_utils::merkle::Blake2bMerkleProof;
 pub use signature_proof::*;
 use thiserror::Error;
 
-use crate::account::AccountTransactionVerification;
+use crate::account::{
+    htlc_contract::{CreationTransactionData as HtlcCreationData, OutgoingHTLCTransactionProof},
+    staking_contract::IncomingStakingTransactionData,
+    vesting_contract::CreationTransactionData as VestingCreationData,
+    AccountTransactionVerification,
+};
 
 mod equivocation_locator;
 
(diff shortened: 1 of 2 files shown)


primitives/transaction/src/lib.rs
```diff
@@ -432,6 +437,133 @@ impl Transaction {
...
+    pub fn related_addresses(&self) -> HashSet<Address> {
+        let mut set = HashSet::new();
+        set.insert(self.sender.clone());
+        set.insert(self.recipient.clone());
+
+        if let Ok(proof) = SignatureProof::deserialize_from_vec(&self.proof) {
+            set.insert(proof.compute_signer());
+        }
+
+        if self.recipient_type == AccountType::Vesting {
```
hrxi: An exhaustive `match` over the account types might be better because it makes sure we cover all cases (if we add some in the future).

rpc-interface/src/types.rs
```diff
@@ -555,6 +564,7 @@ pub struct Transaction {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub confirmations: Option<u32>,
     pub size: usize,
+    pub related_addresses: Vec<Address>,
```
hrxi:
```suggestion
    pub related_addresses: BTreeSet<Address>,
```
Duplicates aren't meaningful, I think.

hrxi (commented):
Looks good to me, left two comments.

rpc-interface/src/types.rs
sisou: The method that computes them uses a `HashSet` already. But yeah, it could be used here, too. I didn't check if it still outputs as an array in JSON, or a weird object representation.

sisou: updated :+1:

hrxi (approved):
LGTM, thanks!

PR #2513, opened 2024-05-22
Run wasm tests in headless Chrome instead of NodeJS

sisou:
The wasm-pack test instrumentation doesn't play nice with `wasm-timer` on NodeJS, failing on the return value of `setTimeout` or `setInterval`. Browsers return a `number`, which the test instrumentation expects, whereas NodeJS returns an `object`.

Running the tests in a browser means we can re-enable a test that was disabled for the above reason.

## Pull request checklist


jsdanielh (approved):
LGTM

PR #2515, opened 2024-05-23
Get rid of `HistoricTransaction::into_transaction`

### LINKED ISSUE #2514: For reward inherents, converting historic transaction to executed transaction produces invalid transaction hash
The method

https://github.com/nimiq/core-rs-albatross/blob/de902e5ea9f2e6fe6180988977c5081b70d7f35d/primitives/transaction/src/historic_transaction.rs#L300

produces an invalid transaction hash for reward inherents. This method is used in RPC in

- `get_transaction_by_hash`
- `get_transactions_by_block_number`
- `get_transactions_by_batch_number`
- `get_transactions_by_address`

and also used in the web client.

All of these return invalid transaction hashes for reward transactions. This is even visible when getting a reward transaction by its correct hash, it returns an object with another hash:

```bash
arpl > tx:get 484d65d8b1514ce3566a51687681b780df2631ea8f61b80445d590f5e7299f14
{
  hash: '6ed6bbc5c76e69766d3335027cc8a89f3994827020ae765bdf73e819b6c819ee',
  from: 'NQ81 C01N BASE 0000 0000 0000 0000 0000 0000',
  # ...
}
```

The right way to do it is to create a transaction object with a custom hash, like in https://github.com/nimiq/core-rs-albatross/pull/2470.

hrxi:
This method was lossy, among others, it lost the actual transaction hash.

Instead, move the generation of transactions to the places where the hash can also be captured.

Fixes #2514.

### DIFF
--- a/rpc-server/src/dispatchers/blockchain.rs
+++ b/rpc-server/src/dispatchers/blockchain.rs
@@ -159,11 +159,12 @@ impl BlockchainInterface for BlockchainDispatcher {
             let mut hist_txs = blockchain.history_store.get_hist_tx_by_hash(&hash, None);
 
             // Unpack the transaction or raise an error.
-            let hist_tx = match (hist_txs.pop(), hist_txs.len()) {
-                (Some(hist_tx), 0) => hist_tx,
-                (Some(_), _) => return Err(Error::MultipleTransactionsFound(hash)),
-                (None, _) => return Err(Error::TransactionNotFound(hash)),
-            };
+            if hist_txs.len() > 1 {
+                return Err(Error::MultipleTransactionsFound(hash));
+            }
+            let hist_tx = hist_txs
+                .pop()
+                .ok_or_else(|| Error::TransactionNotFound(hash.clone()))?;
 
             // Convert the historic transaction into a regular transaction. This will also convert
             // reward inherents.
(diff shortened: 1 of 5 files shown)


sisou:
Did the method lose any other details, too?

sisou (commented):
I cannot review from code-style point-of-view, but I have tested these changes and the hashes returned are now correct.

rpc-server/src/dispatchers/blockchain.rs
```diff
@@ -156,36 +156,25 @@ impl BlockchainInterface for BlockchainDispatcher {
...
             // Unpack the transaction or raise an error.
-            let historic_tx = match historic_tx_vec.len() {
-                0 => {
-                    return Err(Error::TransactionNotFound(hash));
-                }
-                1 => historic_tx_vec.pop().unwrap(),
-                _ => {
-                    return Err(Error::MultipleTransactionsFound(hash));
-                }
+            let hist_tx = match (hist_txs.pop(), hist_txs.len()) {
```
ii-cruz:
```suggestion
                        if hist_txs.len() > 1 {
                return Err(Error::MultipleTransactionsFound(hash));
            }
            let hist_tx = hist_txs
                .pop()
                .map_or(Err(Error::TransactionNotFound(hash)), |x| Ok(x))?;
```

This or something resembling this suggestion looks easier to understand than the match with the tuple. At least to me.

ii-cruz (approved):

rpc-server/src/dispatchers/blockchain.rs
hrxi: Thanks! I couldn't figure out something without `unwrap` and easily understandable. I like your variant. :)

hrxi: Changed.

PR #2518, opened 2024-05-27
Change serialization order of KeyPair to have the private key first

sisou:
This is the same order of serialization that core-js used. This makes (de)serializing KeyPairs backwards-compatible.

It's also the same order that BLS KeyPairs have:
https://github.com/nimiq/core-rs-albatross/blob/de902e5ea9f2e6fe6180988977c5081b70d7f35d/bls/src/types/keypair.rs#L8-L11

### DIFF
--- a/keys/src/key_pair.rs
+++ b/keys/src/key_pair.rs
@@ -9,8 +9,8 @@ use crate::{Ed25519PublicKey, Ed25519Signature, PrivateKey};
 #[derive(Default, Clone, Debug, PartialEq, Eq, Hash)]
 #[cfg_attr(feature = "serde-derive", derive(Serialize, Deserialize))]
 pub struct KeyPair {
-    pub public: Ed25519PublicKey,
     pub private: PrivateKey,
+    pub public: Ed25519PublicKey,
 }
 
 impl KeyPair {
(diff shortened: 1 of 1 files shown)


sisou:
I actually don't think this requires a testnet restart. As keypairs include a private key, they are never sent over the wire AFAIK. They are not even returned from RPC. They are only used in the web-client library API to Javascript.

hrxi:
> This is the same order of serialization that core-js used. This makes (de)serializing KeyPairs backwards-compatible.

Where are they serialized/deserialized? Could we serialize the private key only there? The public key could be derived.

Otherwise we should probably check that they correspond to each other in the deserialization.

sisou:
They are (de)serialized when calling `serialize_to_vec` in the web client JS API when calling `.serialize()` or `.to_hex()` on the wasm wrapper. And deserialized with the equivalent methods.

Nowhere else. Never as part of the protocol.

That was the same case in the PoW code. It's an export/import feature for wallets (which are basically keypairs).

Of course we can add a deserialization check, but it wasn't done before either.

hrxi (approved):

hrxi:
Rebasing.

PR #2519, opened 2024-05-27
Enable `unused_qualifications` lint

hrxi:


sisou (commented):
Reduce clutter, nice 👍

hash/benches/hash_serialize.rs
```diff
@@ -1,23 +0,0 @@
-use criterion::{black_box, criterion_group, criterion_main, Criterion};
```
jsdanielh, line 1: Why are we removing this benchmark?

jsdanielh (approved):

PR #2520, opened 2024-05-27
Switch to wasm-bindgen's newly included TryFromJsValue trait

sisou:
I don't know what's better: one extra dependency less (with this PR), or not having to clone (`.to_owned()`) JsValues (without this PR)?

## Pull request checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4904,7 +4904,6 @@ dependencies = [
  "tracing",
  "tsify",
  "wasm-bindgen",
- "wasm-bindgen-derive",
  "wasm-bindgen-futures",
  "wasm-bindgen-test",
  "web-sys",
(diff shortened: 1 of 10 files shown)


hrxi:
I've asked why the `JsValue` is not taken by reference: https://github.com/rustwasm/wasm-bindgen/issues/3974.

sisou:
@hrxi As mentioned in https://github.com/rustwasm/wasm-bindgen/discussions/4022#discussioncomment-10178019, the `wasm-bindgen-derive` create, which this PR replaces, called `clone()` on all values internally anyway. So in my opinion, switching to the (internal) trait provided by `wasm-bindgen` is still good, as it removes more lines of code than it adds. Wdyt? If you disagree, feel free to close this PR instead.

sisou:
No idea why the web-client tests fail. For some reason, the tests in Chrome don't exit anymore with this PR, whereas before they ran just fine. I don't see how this PR changes anything relevant :confused:

sisou:
Closed because of unexplainable failing tests and the fact that the trait is internal und thus unstable.

PR #2532, opened 2024-05-27
Add log message to state sync

viquezclaudio:
Add a small log message to know if state sync is progressing. 

## Pull request checklist

### DIFF
--- a/consensus/src/sync/live/state_queue/mod.rs
+++ b/consensus/src/sync/live/state_queue/mod.rs
@@ -523,6 +523,16 @@ impl<N: Network> Stream for StateQueue<N> {
         loop {
             match self.chunk_request_component.poll_next_unpin(cx) {
                 Poll::Ready(Some((chunk, start_key, peer_id))) => {
+                    let key = u32::from_str_radix(&format!("{}00000000", start_key)[..8], 16)
+                        .unwrap_or(0);
+
+                    let percentage = (key as f32 / u32::MAX as f32) * 100.0;
+                    log::debug!(
+                        ?start_key,
+                        "Received state sync chunk, ~{}% complete",
+                        percentage,
+                    );
+
                     if let Some(state_chunks) = self.on_chunk_received(chunk, start_key, peer_id) {
                         return Poll::Ready(Some(state_chunks));
                     }


consensus/src/sync/live/state_queue/mod.rs
```diff
@@ -523,6 +523,9 @@ impl<N: Network> Stream for StateQueue<N> {
         loop {
             match self.chunk_request_component.poll_next_unpin(cx) {
                 Poll::Ready(Some((chunk, start_key, peer_id))) => {
+                    log::debug!(
+                        start_key=?start_key, "Received state sync chunk",
```
jsdanielh:
```suggestion
                        ?start_key, "Received state sync chunk",
```

sisou:
What do you think about adding a percentage estimate to the log message, based on how much the number made from the first 2 or 4 bytes of the start hash is out of `0xffff` or `0xffffffff` respectively?

sisou:
Also, is there a "State sync complete" log message?

viquezclaudio:
@sisou  The percentage estimate sounds good to me.
And yes, we already have a state sync complete message:
https://github.com/nimiq/core-rs-albatross/blob/albatross/consensus/src/sync/live/state_queue/mod.rs#L494

consensus/src/sync/live/state_queue/mod.rs
```diff
@@ -523,6 +523,34 @@ impl<N: Network> Stream for StateQueue<N> {
...
+                        let max: u32 = 0xffff;
+                        let percentage = (key as f32 / max as f32) * 100.0;
+
+                        log::debug!(
+                            ?start_key,
+                            "Received state sync chunk, we are {} % completed",
+                            percentage
+                        );
+                    } else {
+                        // For smaller keys we just log it, usually this corresponds to the beggining.
```
sisou:
```suggestion
                        // For smaller keys we just log it, usually this corresponds to the beginning.
```

sisou:
See also the comments I accidentally put on the commit itself: https://github.com/nimiq/core-rs-albatross/commit/63098abb757d6dd6b61c24b60e33e88eb4286df6

consensus/src/sync/live/state_queue/mod.rs
```diff
@@ -523,6 +523,33 @@ impl<N: Network> Stream for StateQueue<N> {
         loop {
             match self.chunk_request_component.poll_next_unpin(cx) {
                 Poll::Ready(Some((chunk, start_key, peer_id))) => {
+                    if start_key.len() >= 8 {
+                        // Extract the first 4 bytes (8 nibles ) of the key nibbles
+                        let key =
+                            i32::from_str_radix(&start_key.slice(0, 8).to_string(), 16).unwrap();
```
hrxi: Can probably replace the whole `if-else` thing with `let key = u32::from_str_radix(&format!("{}00000000", start_key)[..8].unwrap()`

sisou:
For the percentage, I think you should dedupe the log line. You can do

```rust
let percentage = if start_key.len() >= 8 {
    // ...
    (key as f32 / u32::MAX as f32) * 100.0
} else if start_key.len() >= 4 {
    // ...
} else {
    // ...
}
```

And then log once.

consensus/src/sync/live/state_queue/mod.rs
sisou: You removed the double `g`, but forgot to add the other `n`.

consensus/src/sync/live/state_queue/mod.rs
viquezclaudio: I really like this!
Just made a small change to take into consideration when the start key is "ε"

sisou (approved):

jsdanielh (approved):

PR #2533, opened 2024-05-27
State sync not required for live sync instead of  macro sync

### LINKED ISSUE #2347: History root doesn't match real history root
This was observed when a full node was trying to synchronize to the testnet.
It happens after a second validity window synchronization process is started and the existing and some existing history is removed

[log4.log](https://github.com/nimiq/core-rs-albatross/files/14872535/log4.log)

viquezclaudio:
Remove the accounts complete requisite when determining that a peer is sufficiently close for live sync instead of macro sync.

## Pull request checklist


nibhar (approved):
Looks correct to me to remove that check.

@styppo maybe chime in here though.

PR #2541, opened 2024-05-29
SyncQueue: fix manipulation of `next_outgoing_index` on error path

### LINKED ISSUE #2495: RequestMissingBlocks sometimes fails

styppo:
When the final request for an id failed (after retries), an error was returned to the caller without incrementing `next_outgoing_index`. This resulted in subsequent requests getting stuck without their results ever being returned from `SyncQueue`. Additionally, failed requests did not respect their queuing order when reporting the error to the caller.

Closes #2495.


jsdanielh (approved):
LGTM

PR #2547, opened 2024-05-30
Fix history sync to use additional non prunable checkpoints

### LINKED ISSUE #2502: Additional checkpoints to keep history sync size small are not returned

jsdanielh:
- Fix implementation of `get_epoch_chunks` since it was only returning a single block and not really looking for all non-prunable macro blocks in the same epoch as it should have.
  This fixes #2502.
- Fix the computation of history chunk indexes when making `HistoryChunkRequest` requests in history sync.
- Use election macro head for validating batch sets instead of using the macro head. This is because we could ask for a batch set of an epoch containing checkpoint macro blocks we already know and in this case, the verification of batch sets would fail since it may contain checkpoint macro blocks we already know and that has a block number lower than our macro head.

## Pull request checklist

### DIFF
--- a/blockchain/src/chain_store.rs
+++ b/blockchain/src/chain_store.rs
@@ -177,6 +177,10 @@ impl ChainStore {
         txn.put(&self.height_idx, &height, hash);
     }
 
+    /// Gets the set of macro block hashes that delimits the epoch chunks.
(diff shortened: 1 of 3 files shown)


nibhar (approved):
LGTM

PR #2549, opened 2024-05-31
Remove `BlockHeader` enum

hrxi:
It was essentially unused except for the single `BlockHeader::verify` function which was instead moved to `Block::verify_header`.

`BlockComponents` was also unused and removed.

Also rename `MacroBlock::is_election_block` to `is_election`. This unifies the naming with `Block::is_election` and a couple of other functions.

### DIFF
--- a/blockchain/src/blockchain/verify.rs
+++ b/blockchain/src/blockchain/verify.rs
@@ -46,7 +46,7 @@ impl Blockchain {
 
         // Verify the interlink (or its absence)
         if let Block::Macro(macro_) = &block {
-            if macro_.is_election_block() {
+            if macro_.is_election() {
                 if let Some(interlink) = &macro_.header.interlink {
                     let expected_interlink = self.election_head().get_next_interlink().unwrap();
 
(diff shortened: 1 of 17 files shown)


styppo (approved):

blockchain/src/blockchain/verify.rs
```diff
@@ -45,9 +45,9 @@ impl Blockchain {
         }
 
         // Verify the interlink (or its absence)
-        if let BlockHeader::Macro(macro_header) = block.header() {
-            if block.is_election() {
-                if let Some(interlink) = &macro_header.interlink {
+        if let Block::Macro(macro_) = &block {
```
Eligioo: Did the `macro_`s accidentally happen from refactoring?

hrxi: No. I wanted to call them `macro`, but that's not allowed because `macro` is a keyword. AFAIK it's common to just append an underscore in this case.

styppo: In the rest of the codebase, we use `r#macro`.

sisou: Or you can use `macro_block`.

hrxi: > Or you can use `macro_block`.

Done.

jsdanielh (approved):

PR #2559, opened 2024-06-05
Test the gossipsub message size limit

### LINKED ISSUE #2529: 1mb max message size is too small for macroblocks
Network libp2p. Config. 1mb max message size for gossip sub can easily be exceeded for macro blocks due to reward txs.

https://github.com/nimiq/core-rs-albatross/blob/1cf1c3f4e503d199d5ff0c123d13660b23215d3c/network-libp2p/src/config.rs#L57

hrxi:
Make sure it's well above the size of a macro block.

Fixes #2529.

### DIFF
--- a/network-interface/src/network.rs
+++ b/network-interface/src/network.rs
@@ -37,6 +37,9 @@ pub trait Topic {
     const VALIDATE: bool;
 }
 
+/// Network implementations have to at least support messages of this size.
+pub const MIN_SUPPORTED_MSG_SIZE: usize = 1024 * 1024;
+
 #[derive(Clone, Debug)]
 pub enum MsgAcceptance {
     Accept,
--- a/primitives/block/src/multisig.rs
+++ b/primitives/block/src/multisig.rs
@@ -1,6 +1,7 @@
 use nimiq_bls::AggregateSignature;
 use nimiq_collections::bitset::BitSet;
 use nimiq_handel::contribution::{AggregatableContribution, ContributionError};
+use nimiq_primitives::policy::Policy;
 use serde::{Deserialize, Serialize};
 
 /*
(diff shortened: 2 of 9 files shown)


hrxi:
The current maximum size of a macro block is 201_832 bytes, so roughly a fifth of the current maximum message size.

network-interface/src/network.rs
```diff
@@ -37,6 +37,8 @@ pub trait Topic {
     const VALIDATE: bool;
 }
 
+pub const MIN_SUPPORTED_MSG_SIZE: usize = 1024 * 1024;
```
sisou, line 41: Do you mean `MAX`?

hrxi: I had a hard time naming this constant.

I want to say that this is the minimum maximum message size of the underlying network. I.e. the network has to at least support messages of this size. I don't want to tell the underlying network that it has to limit the sizes to exactly this, just that it needs to support messages of this size.

sisou: That's great. I'd suggest putting a rustdoc comment on it, explaining it exactly like that:

```rust
/// Network implementations have to at least support messages of this size
pub const MIN_SUPPORTED_MSG_SIZE: usize = 1024 * 1024;
```

hrxi: Thanks for the concise description. Added.

primitives/block/src/multisig.rs
```diff
@@ -15,6 +16,10 @@ pub struct MultiSignature {
 }
 
 impl MultiSignature {
```
jsdanielh, line 18:
```suggestion
impl MultiSignature {
 #[allow(clippy::identity_op)]
```

hrxi: Done.

jsdanielh (approved):

PR #2565, opened 2024-06-05
Increase uptime in `FourValidatorsReconnectSpammer` DevNet scenario

styppo:
With the transaction spammer running, re-syncing in debug mode is really slow. The previous uptime setting of 100 seconds was not always sufficient for a node that had been killed to fully re-sync before another node would get killed, resulting in the chain to stall and a failed scenario run.

This PR increases the node uptime for the spammer scenario. It also reduces the downtime to make it easier for nodes to catch up.


hrxi (approved):
LGTM.

hrxi:
Rebasing to get it merged.

PR #2568, opened 2024-06-05
BlockQueue: Restrict buffering of blocks per peer per height

### LINKED ISSUE #2510: DOS vector Unbounded nested map.
DOS vector Unbounded nested map.

check the announced block. Round bound per peer basis.

https://github.com/nimiq/core-rs-albatross/blob/3e08cee265122eb58eebfe0f84090a272f674355/consensus/src/sync/live/block_queue/mod.rs#L234

nibhar:
Changes the `insert_block_into_buffer` such that it checks if the propagation source of the block is already present twice for that block height. If so the block will be dropped as only malicious peers would propagate more than 2 blocks (original block and potential skip block superseding it).

If no `pubsub_id` is given as for example a `MissingBlockRequest` would result in no additional checks are introduced. 

Also changed the semantics of the functions return value to reflect if the block was added instead of the formerly used block was known. Removed an unnecessary check in the process as well.

This PR fixes #2510.

### DIFF
--- a/consensus/src/sync/live/block_queue/mod.rs
+++ b/consensus/src/sync/live/block_queue/mod.rs
@@ -100,6 +100,8 @@ pub struct BlockQueue<N: Network> {
 }
 
 impl<N: Network> BlockQueue<N> {
+    const MAX_BUFFERED_PER_PEER_PER_HEIGHT: usize = 5;
+
     pub async fn new(network: Arc<N>, blockchain: BlockchainProxy, config: QueueConfig) -> Self {
         let block_stream = if config.include_micro_bodies {
             network.subscribe::<BlockTopic>().await.unwrap().boxed()
@@ -259,11 +261,14 @@ impl<N: Network> BlockQueue<N> {
         let mut parent_block_number = block_number - 1;
 
         // Insert block into buffer. If we already know the block, we're done.
-        let block_known = self.insert_block_into_buffer(block, pubsub_id.clone());
-        log::trace!("Buffering block #{}, known={}", block_number, block_known);
-        if block_known {
+        if !self.insert_block_into_buffer(block, pubsub_id.clone()) {
+            log::trace!(
+                block_number,
+                "Not buffering block - already known or exceeded the per peer limit",
+            );
             return;
         }
+        log::trace!(block_number, "Buffering block");
 
         // If the parent of this block is already in the buffer, follow the chain to see whether there are still blocks missing.
         while let Some(parent_block_hash) =
@@ -296,12 +301,76 @@ impl<N: Network> BlockQueue<N> {
         );
     }
 
+    /// Attempts to insert a block into the buffer.
+    /// This may fail due to the peer who propagated the block as given in the optional `pubsub_id` parameter
+    /// is already having too many blocks buffered. Blocks without a pubsub_id will always be buffered.
+    ///
+    /// At most `MAX_BUFFERED_PER_PEER_PER_HEIGHT` block per peer per block height will be buffered leaving enough
+    /// room for a bock and the skip block making it obsolete. For benign peers that is sufficient, as they will
+    /// more than that.
+    ///
+    /// ## Returns
+    /// * `true` if the block was added to the buffer
+    /// * `false` otherwise
     fn insert_block_into_buffer(&mut self, block: Block, pubsub_id: Option<N::PubsubId>) -> bool {
-        self.buffer
-            .entry(block.block_number())
-            .or_default()
-            .insert(block.hash(), (block, pubsub_id))
-            .is_some()
+        let peer_id = match pubsub_id {
+            // Blocks without a pubsub_id are requested and will always be buffered.
+            None => {
+                let map = self.buffer.entry(block.block_number()).or_default();
+                if map.contains_key(&block.hash()) {
+                    return false;
+                }
+                assert!(map.insert(block.hash(), (block, None)).is_none());
+                return true;
+            }
+            // For blocks with a pubsub id the propagation source is the peer which can
+            // only have MAX_BUFFERED_PER_PEER_PER_HEIGHT items buffered.
+            Some(ref pubsub_id) => pubsub_id.propagation_source(),
+        };
+
+        // Get the entry for the block number if it exists.
+        // Otherwise add the entry and add the block to it and return.
+        let map = match self.buffer.entry(block.block_number()) {
+            BTreeMapEntry::Occupied(occupied_entry) => occupied_entry.into_mut(),
+            BTreeMapEntry::Vacant(vacant_entry) => {
+                // Trivially okay to buffer as nothing is buffered yet.
+                vacant_entry.insert(HashMap::from([(block.hash(), (block, pubsub_id))]));
+                return true;
+            }
+        };
+
+        let mut blocks_by_peer = 0;
+        for (hash, (_block, pubsub)) in map.iter() {
+            if *hash == block.hash() {
+                // Block is already buffered. Ignore it here.
+                return false;
+            }
+
+            // Extract the propagation source or move on to the next item
+            let propagation_source = match pubsub {
+                Some(pubsub_id) => pubsub_id.propagation_source(),
+                None => continue,
+            };
+
+            // Check if the source is the current peer
+            if propagation_source == peer_id {
+                // Increase the counter
+                blocks_by_peer += 1;
+                // Check if the count exceeds the maximum buffer count
+                if blocks_by_peer == Self::MAX_BUFFERED_PER_PEER_PER_HEIGHT {
+                    log::debug!(
+                        ?peer_id, block_number=block.block_number(), dropped_block=?block,
+                        "Peer buffer exceeded limit, dropping block",
+                    );
+                    // If so return, as the peer cannot buffer any additional items.
+                    return false;
+                }
+            }
+        }
+        // Insert the block. Strictly speaking the .is_none() is redundant with previous checks
+        // and could be replaced with return true;
+        assert!(map.insert(block.hash(), (block, pubsub_id)).is_none());
+        true
     }
 
     fn get_parent_from_buffer(&self, block_number: u32, hash: &Blake2bHash) -> Option<Blake2bHash> {
--- a/consensus/src/sync/live/block_queue/mod.rs
+++ b/consensus/src/sync/live/block_queue/mod.rs
@@ -231,14 +231,6 @@ impl<N: Network> BlockQueue<N> {
             if self.request_component.take_peer(&peer_id).is_some() {
                 return Some(QueuedBlock::TooFarAhead(block, peer_id));
             }
-        } else if self.buffer.len() >= self.config.buffer_max {
-            // TODO: This does not account for the nested map
-            log::warn!(
-                "Discarding block {} - buffer full (max {})",
-                block,
-                self.buffer.len(),
-            );
-            self.report_validation_result(pubsub_id, MsgAcceptance::Ignore);
         } else if block_number <= macro_height {
             // Block is from a previous batch/epoch, discard it.
             log::warn!(


sisou:
> as only malicious peers would propagate more than 2 blocks (original block and potential skip block superseding it)

Can a second skip block be proposed and then added to the chain, if the first skip block doesn't get enough votes? E.g. can multiple skip blocks exist in proposed state per height?

nibhar:
> > as only malicious peers would propagate more than 2 blocks (original block and potential skip block superseding it)
> 
> Can a second skip block be proposed and then added to the chain, if the first skip block doesn't get enough votes? E.g. can multiple skip blocks exist in proposed state per height?

Any skip block that does not get enough votes will not be published.

Thinking about it though there may be multiple chains going at the same time with skip blocks prior to this block height and that might pose a problem.

i.e:
```
0.0 --> 1.0 --> 2.0
    \-> 1.1 --> 2.1
            \-> 2.2
```

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -296,12 +299,75 @@ impl<N: Network> BlockQueue<N> {
...
+    /// already having too many blocks buffered. Blocks without a pubsub_id will always be buffered.
+    ///
+    /// At most two block per peer per block height will be buffered leaving enough room for a bock and the
+    /// skip block making it obsolete. For benign peers that is sufficient, as they will never propagate
+    /// more than that.
+    ///
+    /// ## Returns
+    /// * `true` if the block was added to the buffer
+    /// * `false` otherwise
     fn insert_block_into_buffer(&mut self, block: Block, pubsub_id: Option<N::PubsubId>) -> bool {
```
hrxi, line 315:
```suggestion
    fn insert_block_into_buffer(&mut self, block: Block, pubsub_id: Option<N::PubsubId>) -> Result<(), ()> {
```
`Result<(), ()>` is perfect for this, but clippy will be mad at you… Perhaps just `#[allow(clippy::result_unit_err)]`?

nibhar:
After thinking about my previous comment it does make sense to have a higher limit than the 2 blocks which reasonably can exist on a single branch. Ultimately these two reasonable blocks exist on every previous branch which by the same logic is 2^n with n being the number of blocks in the batch. That is obviously unfeasible. 

I decided to introduce a constant for this instead and set it to a low 5 for the time being. Situations where more than 5 blocks are gossipped by the same peer for the same height should be exceedingly rare, however maybe it does not hurt to set it to a somewhat higher value like 10.

I log a debug message if a message is not buffered because the limit is exceeded to get an idea of the value in future ci and Testnet runs.

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -100,6 +100,8 @@ pub struct BlockQueue<N: Network> {
 }
 
 impl<N: Network> BlockQueue<N> {
+    const MAX_BUFFERED_PER_PEER_PER_HEIGHT: usize = 5usize;
```
hrxi:
```suggestion
    const MAX_BUFFERED_PER_PEER_PER_HEIGHT: usize = 5;
```

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -304,12 +301,81 @@ impl<N: Network> BlockQueue<N> {
...
-            .is_some()
+        let peer_id = match pubsub_id {
+            // Blocks without a pubsub_id are requested and will always be buffered.
+            None => {
+                return self
+                    .buffer
+                    .entry(block.block_number())
+                    .or_default()
+                    .insert(block.hash(), (block, pubsub_id))
+                    .is_none()
```
hrxi: The `insert` replaces the value stored in the hashmap. Is that intentional?

This means that `false` is returned if block was already present, but the pubsub id was replaced, at least this part might not be intentional?

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -304,12 +301,81 @@ impl<N: Network> BlockQueue<N> {
...
+            }
+            // For blocks with a pubsub id the propagation source is the peer which can
+            // only have 2 items buffered.
+            Some(ref pubsub_id) => pubsub_id.propagation_source(),
+        };
+
+        // Get the entry for the block number if it exists.
+        // Otherwise add the entry and add the block to it and return.
+        let mut entry = match self.buffer.entry(block.block_number()) {
+            BTreeMapEntry::Occupied(occupied_entry) => occupied_entry,
```
hrxi:
```suggestion
            BTreeMapEntry::Occupied(occupied_entry) => occupied_entry.into_mut(),
```

Using `into_mut()` instead of `get_mut()` below should get rid of the borrow checker problems.

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -304,12 +301,81 @@ impl<N: Network> BlockQueue<N> {
...
+            BTreeMapEntry::Vacant(vacant_entry) => {
+                // Trivially okay to buffer as nothing is buffered yet.
+                vacant_entry.insert(HashMap::from([(block.hash(), (block, pubsub_id))]));
+                return true;
+            }
+        };
+
+        // Access the entry.
+        // Necessary for borrow rules to only get the map here as the entry must also still live.
+        let map = entry.get_mut();
```
hrxi: See above.

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -304,12 +301,81 @@ impl<N: Network> BlockQueue<N> {
...
-            .insert(block.hash(), (block, pubsub_id))
-            .is_some()
+        let peer_id = match pubsub_id {
+            // Blocks without a pubsub_id are requested and will always be buffered.
+            None => {
+                return self
+                    .buffer
+                    .entry(block.block_number())
+                    .or_default()
+                    .insert(block.hash(), (block, pubsub_id))
```
hrxi:
```suggestion
                    .insert(block.hash(), (block, None))
```
(optional, made it clearer to me).

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -304,12 +301,81 @@ impl<N: Network> BlockQueue<N> {
...
+                        block,
+                    );
+                    // If so return, as the peer cannot buffer any additional items.
+                    return false;
+                }
+            }
+        }
+        // Insert the block. Strictly speaking the .is_none() is redundant with previous checks
+        // and could be replaced with return true;
+        map.insert(block.hash(), (block, pubsub_id)).is_none()
```
hrxi:
```suggestion
        assert!(map.insert(block.hash(), (block, pubsub_id)).is_none());
        true
```

hrxi (commented):
Looks good to me.

Mostly just style comments, but one actual question.

The block counting is currently linear in the amount of amount of blocks buffered at a given height. This is likely not a problem.

consensus/src/sync/live/block_queue/mod.rs
nibhar: I was considering `Result<bool, ()>` to 
`Ok(true)` if the block was added, 
`Ok(false)` if the block was already known 
`Err(())` if the block was not added due to buffering limitations
I decided that that level of granularity was not needed caller side, so I kept the `bool`. 
Why do you prefer `Result<(), ()>` over `bool`?

hrxi: I prefer `Result<(), ()>` over `bool` here, because I won't have to look at the documentation to know that `Ok(())` means that something succeeded, but I don't know if `true` or `false` means that the block was added. But it's more a question of taste, so it's not important to me to change this.

consensus/src/sync/live/block_queue/mod.rs
nibhar: As we only buffer any given block once it does not exactly matter whose pubsub id will be used here. It will of course be reflected once this peer or the just removed peer sends us another block at that height, but the result will be similar as once of them potentially does not get to buffer whereas the other does. Who does and who does not could be inverted, but I kept the original behaviour here. I do agree though that as I changed the semantics of the return, this is not strictly correct and I changed it. Thank you for the remark!

hrxi (approved):
LGTM, thanks!

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -304,12 +301,78 @@ impl<N: Network> BlockQueue<N> {
         );
     }
 
+    /// Attempts to insert a block into the buffer.
+    /// This may fail due to the peer who propagated the block as given in the optional `pubsub_id` parameter
+    /// already having too many blocks buffered. Blocks without a pubsub_id will always be buffered.
```
jsdanielh:
```suggestion
    /// is already having too many blocks buffered. Blocks without a pubsub_id will always be buffered.
```

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -304,12 +301,78 @@ impl<N: Network> BlockQueue<N> {
...
+            // Blocks without a pubsub_id are requested and will always be buffered.
+            None => {
+                let map = self.buffer.entry(block.block_number()).or_default();
+                if map.contains_key(&block.hash()) {
+                    return false;
+                }
+                assert!(map.insert(block.hash(), (block, None)).is_none());
+                return true;
+            }
+            // For blocks with a pubsub id the propagation source is the peer which can
```
jsdanielh, line 326:
```suggestion
            // For blocks with a pubsub id in the propagation source is the peer which can
```
```suggestion
            // For blocks with a pubsub id the propagation source is the peer which can
```

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -304,12 +301,78 @@ impl<N: Network> BlockQueue<N> {
         );
     }
 
+    /// Attempts to insert a block into the buffer.
+    /// This may fail due to the peer who propagated the block as given in the optional `pubsub_id` parameter
+    /// already having too many blocks buffered. Blocks without a pubsub_id will always be buffered.
+    ///
+    /// At most two block per peer per block height will be buffered leaving enough room for a bock and the
```
jsdanielh:
```suggestion
    /// At most `MAX_BUFFERED_PER_PEER_PER_HEIGHT` block per peer per block height will be buffered leaving enough room for a bock and the
```

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -304,12 +301,78 @@ impl<N: Network> BlockQueue<N> {
...
+            None => {
+                let map = self.buffer.entry(block.block_number()).or_default();
+                if map.contains_key(&block.hash()) {
+                    return false;
+                }
+                assert!(map.insert(block.hash(), (block, None)).is_none());
+                return true;
+            }
+            // For blocks with a pubsub id the propagation source is the peer which can
+            // only have 2 items buffered.
```
jsdanielh:
```suggestion
            // only have MAX_BUFFERED_PER_PEER_PER_HEIGHT items buffered.
```

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -304,12 +301,78 @@ impl<N: Network> BlockQueue<N> {
...
+            let propagation_source = match pubsub {
+                Some(pubsub_id) => pubsub_id.propagation_source(),
+                None => continue,
+            };
+
+            // Check if the source is the current peer
+            if propagation_source == peer_id {
+                // Increase the counter
+                blocks_by_peer += 1;
+                // Check if that is the count exceeds the maximum buffer count
```
jsdanielh:
```suggestion
                // Check if the exceeds the maximum buffer count
```

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -304,12 +301,78 @@ impl<N: Network> BlockQueue<N> {
...
+                if blocks_by_peer == Self::MAX_BUFFERED_PER_PEER_PER_HEIGHT {
+                    log::debug!(
+                        "Peer ({:?}) buffer for height {} exceeded limit, dropping block {}",
+                        peer_id,
+                        block.block_number(),
+                        block,
```
jsdanielh: I'd suggest to use tracing style:
```suggestion
                        "Peer ({:?}) buffer for height {} exceeded limit, dropping block {}",
                        peer_id,
                        block.block_number(),
                        block,
```
```suggestion
                        ?peer_id, block_number=block.block_number(), dropped_block=block
                        "Peer buffer exceeded limit, dropping block",
```

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -267,11 +261,14 @@ impl<N: Network> BlockQueue<N> {
...
-        log::trace!("Buffering block #{}, known={}", block_number, block_known);
-        if block_known {
+        if !self.insert_block_into_buffer(block, pubsub_id.clone()) {
+            log::trace!(
+                "Not buffering block #{} - already known or exceeded the per peer limit",
+                block_number
+            );
             return;
         }
+        log::trace!("Buffering block #{}", block_number);
```
jsdanielh: Use `tracing` style:
```suggestion
        log::trace!(block_number "Buffering block");
```

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -267,11 +261,14 @@ impl<N: Network> BlockQueue<N> {
         let mut parent_block_number = block_number - 1;
 
         // Insert block into buffer. If we already know the block, we're done.
-        let block_known = self.insert_block_into_buffer(block, pubsub_id.clone());
-        log::trace!("Buffering block #{}, known={}", block_number, block_known);
-        if block_known {
+        if !self.insert_block_into_buffer(block, pubsub_id.clone()) {
+            log::trace!(
+                "Not buffering block #{} - already known or exceeded the per peer limit",
+                block_number
```
jsdanielh: Use `tracing` style:
```suggestion
                block_number
                "Not buffering block - already known or exceeded the per peer limit",
```

jsdanielh (approved):

PR #2584, opened 2024-06-11
Macro sync: Check if peer is still connected after request

### LINKED ISSUE #2508: FIXME Check if the peer is still connected
Check the strategy that is used when a peer is disconnected during the poll epoch ids 
Make sure internal data structures are properly cleaned.

fiaxh:
Closes #2508


jsdanielh (approved):
LGTM

viquezclaudio (approved):

PR #2587, opened 2024-06-11
TransactionBuilder: Return Transaction instead of Result if creation can't fail

### LINKED ISSUE #2544: Handle failing creation of reactivation transaction
https://github.com/nimiq/core-rs-albatross/blob/3e08cee265122eb58eebfe0f84090a272f674355/validator/src/validator.rs#L765-L773

fiaxh:
Closes #2544

I don't think this error needs to be handled:

`new_reactivate_validator` only returns an error if `builder.generate()` returns an error. `builder.generate()` returns an error iff:
- Certain fields are not set: sender, recipient, value, validity_start_height, network_id
- It is a signaling transaction with a non-zero ~fee~ value

However, in `new_reactivate_validator`
- All those mandatory fields are set
- A signaling transaction is created, but the ~fee~ value is set to zero and can't be specified via parameters

Thus, the function call can never return an error and we don't have to handle it.

I addressed this by unwrapping the result of `builder.generate()` and changing the return type of `new_reactivate_validator` to `Transaction` instead of `Result`. Same for `new_update_validator`, `new_deactivate_validator` and `new_retire_validator`.

### DIFF
--- a/blockchain/tests/block_production.rs
+++ b/blockchain/tests/block_production.rs
@@ -246,8 +246,7 @@ fn it_can_produce_macro_block_after_punishment() {
         100.try_into().unwrap(),
         1 + Policy::genesis_block_number(),
         NetworkId::UnitAlbatross,
-    )
-    .unwrap();
+    );
     producer.next_block_with_txs(vec![], false, vec![reactivate_tx]);
 
     // Produce next checkpoint block.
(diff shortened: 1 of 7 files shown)


sisou:
I think you mean non-zero **value**. Fee can be non-zero, but signalling transaction cannot have a _value_:

https://github.com/nimiq/core-rs-albatross/blob/7775f0581c2999a53e5aeb6ddd1ea1e2f1c33e44/transaction-builder/src/lib.rs#L372-L374

The `TransactionBuilder::new_reactivate_validator` method does not take a `value` parameter.

fiaxh:
You are right, I meant value but wrote fee. With that change, the argument should work, though.

nibhar (approved):

jsdanielh (approved):

hrxi (approved):
LGTM.

PR #2588, opened 2024-06-11
Move rate limiting to swarm to fix the unprotected buffer

### LINKED ISSUE #2531: DoS vector on network
TODO Make buffer size configurable. unprotected buffer

Change rate limiting is to be applied before that, in a layer above to prevent this.

ii-cruz:
Moving rate limiting to swarm to fix the unprotected buffer of the network when handling a request.
Rate limit refactoring

## Pull request checklist

### DIFF
--- a/network-interface/src/request/mod.rs
+++ b/network-interface/src/request/mod.rs
@@ -143,7 +143,7 @@ pub trait RequestCommon:
 pub trait RequestSerialize: RequestCommon {
     /// Serializes a request.
     /// A serialized request is composed of:
-    /// - A variant for the Type ID of the request
+    /// - A variable sized integer for the Type ID of the request
     /// - Serialized content of the inner type.
     fn serialize_request(&self) -> Vec<u8> {
         let mut data = Vec::with_capacity(self.serialized_request_size());
@@ -167,7 +167,7 @@ pub trait RequestSerialize: RequestCommon {
 
     /// Deserializes a request
     /// A serialized request is composed of:
-    /// - A variant for the Type ID of the request
+    /// - A variable sized integer for the Type ID of the request
     /// - Serialized content of the inner type.
     fn deserialize_request(buffer: &[u8]) -> Result<Self, DeserializeError> {
         // Check for correct type.
--- a/network-interface/src/request/mod.rs
+++ b/network-interface/src/request/mod.rs
@@ -143,7 +143,7 @@ pub trait RequestCommon:
 pub trait RequestSerialize: RequestCommon {
     /// Serializes a request.
     /// A serialized request is composed of:
-    /// - A varint for the Type ID of the request
+    /// - A variant for the Type ID of the request
     /// - Serialized content of the inner type.
     fn serialize_request(&self) -> Vec<u8> {
         let mut data = Vec::with_capacity(self.serialized_request_size());
--- a/network-interface/src/request/mod.rs
+++ b/network-interface/src/request/mod.rs
@@ -26,17 +26,17 @@ impl fmt::Display for RequestType {
 }
 
 impl RequestType {
-    const fn new(type_id: u16, requires_response: bool) -> RequestType {
-        RequestType((type_id << 1) | requires_response as u16)
+    const fn new(type_id: u16, requires_response: bool) -> Self {
+        Self((type_id << 1) | requires_response as u16)
     }
-    pub fn from_request<R: RequestCommon>() -> RequestType {
-        RequestType::new(R::TYPE_ID, R::Kind::EXPECT_RESPONSE)
+    pub fn from_request<R: RequestCommon>() -> Self {
+        Self::new(R::TYPE_ID, R::Kind::EXPECT_RESPONSE)
     }
-    pub const fn request(type_id: u16) -> RequestType {
-        RequestType::new(type_id, true)
+    pub const fn request(type_id: u16) -> Self {
+        Self::new(type_id, true)
     }
-    pub const fn message(type_id: u16) -> RequestType {
-        RequestType::new(type_id, false)
+    pub const fn message(type_id: u16) -> Self {
+        Self::new(type_id, false)
     }
     pub const fn type_id(self) -> u16 {
         self.0 >> 1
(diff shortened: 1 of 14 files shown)


network-libp2p/src/network_types.rs
```diff
@@ -248,8 +250,14 @@ pub(crate) struct TaskState {
...
         HashMap<InboundRequestId, ResponseChannel<Option<OutgoingResponse>>>,
     /// Senders for replying to requests per `RequestType` for request-response
-    pub(crate) receive_requests:
-        HashMap<RequestType, mpsc::Sender<(Bytes, InboundRequestId, PeerId)>>,
+    pub(crate) receive_requests: HashMap<
+        RequestType,
+        (
+            mpsc::Sender<(Bytes, InboundRequestId, PeerId)>,
+            u32,
+            Duration,
```
nibhar: I would prefer these be encapsulated into a named struct. It is used in a couple more places than just here. Not tagging all of them but just here.

network-libp2p/src/rate_limiting.rs
```diff
@@ -142,10 +147,13 @@ impl RateLimit {
 
 // Network helpers for rate limiting
 
-pub(crate) fn is_under_the_rate_limits<Req: RequestCommon>(
-    peer_request_limits: Arc<Mutex<HashMap<PeerId, HashMap<u16, RateLimit>>>>,
+pub(crate) fn is_under_the_rate_limits(
```
nibhar: Personally i would always go for `exceeds_rate_limit()`. Obviously that would inverse the return. however you already inverse the return from `increment_and_is_allowed` and then when calling this you invert again, so it should be fine,.

That being said doing all of this caller side would remove some of the clutter. The function has a bunch of parameters which are just used to init or retrieve the `RateLimit`. I would probably inline this making the function as a whole obsolete. Specifically as it is only used in a single spot.

nibhar (approved):
Generally I would like this to be more encapsulated. It looks correct to me and if necessary could go in like this, but I think it would be worth our while.

network-libp2p/src/network_types.rs
```diff
@@ -248,8 +250,13 @@ pub(crate) struct TaskState {
     pub(crate) response_channels:
         HashMap<InboundRequestId, ResponseChannel<Option<OutgoingResponse>>>,
     /// Senders for replying to requests per `RequestType` for request-response
```
nibhar: This should probably reflect the update made.

network-libp2p/src/rate_limiting.rs
```diff
@@ -141,114 +163,120 @@ impl RateLimit {
...
+                request_id,
+                peer_id,
+                request_type,
+                request_rate_limit_data.max_requests,
+                request_rate_limit_data.time_window,
+            );
+            return true;
+        }
+
+        false
```
nibhar: I suggest keeping the debug log only. it is more informativ. Also I'd suggest to use tracing notation. Also if you log this callerside this can be replaced by 
```suggestion
        !requests_limit.increment_and_is_allowed(1)
```

nibhar (commented):
Smaller notes but nothing major. I did not mention this in the review itself but add some rustdoc to functions and fields please.

network-interface/src/request/mod.rs
```diff
@@ -143,7 +143,7 @@ pub trait RequestCommon:
 pub trait RequestSerialize: RequestCommon {
     /// Serializes a request.
     /// A serialized request is composed of:
-    /// - A varint for the Type ID of the request
+    /// - A variant for the Type ID of the request
```
hrxi:
```suggestion
    /// - A variable sized integer for the Type ID of the request
```

`varint` as in "variable sized integer".

network-libp2p/src/swarm.rs
```diff
@@ -702,53 +700,64 @@ fn handle_event(
...
-                                            None
-                                        }
-                                        None => None,
-                                    };
+
+                                    // Filter off sender if not alive.
+                                    let sender_data = state
+                                        .receive_requests
+                                        .get(&type_id)
+                                        .filter(|(sender, ..)| !sender.is_closed());
```
hrxi, line 708: This changes behavior. Previously, closed channels were removed from the map.

Is the behavior change intentional?

network-libp2p/src/swarm.rs
```diff
@@ -1204,6 +1222,29 @@ pub(crate) fn verify_record(record: &Record) -> Option<DhtRecord> {
...
+        .is_err()
+    {
+        error!(
+            %request_id,
+            %peer_id,
+            %type_id,
+            "Could not send response {:?}", response
+        );
+    }
+}
```
hrxi: I feel like this function isn't carrying its weight. What was the reason for extracting the code into a function? Logging a message on error?

network-interface/src/request/mod.rs
```diff
@@ -26,17 +26,17 @@ impl fmt::Display for RequestType {
...
     }
-    pub const fn request(type_id: u16) -> RequestType {
-        RequestType::new(type_id, true)
+    pub const fn request(type_id: u16) -> Self {
+        Self::new(type_id, true)
     }
-    pub const fn message(type_id: u16) -> RequestType {
-        RequestType::new(type_id, false)
+    pub const fn message(type_id: u16) -> Self {
+        Self::new(type_id, false)
```
hrxi, line 39: This seems unrelated to the PR. Was it done unintentionally using automated refactoring?

hrxi (commented):
Found a couple of potential issues.

network-interface/src/request/mod.rs
ii-cruz: I saw this issue on some other structs related to the rate-limiting and spotted it here too so I changed it as well

network-libp2p/src/swarm.rs
ii-cruz: They are being removed later yes. I discussed it with JD. The code becomes more readable like this and the removal still happens later.

network-libp2p/src/swarm.rs
ii-cruz: it's repeated in two places close by and simplified a bit the code there because there are so many If, else closes and nested scopes. Can change it back if you prefer, either way is fine imo

network-interface/src/request/mod.rs
ii-cruz: Thanks this was autocorrect mix-fixes

network-libp2p/src/swarm.rs
ii-cruz: I removed the function and the response and respective error handling are done in the original places

hrxi:
Rebasing to merge.

Issue #2295, opened 2024-03-07
Light client - verification failed for requesting missing blocks


Eligioo:
## New issue checklist

- [x] I have read all of the [`README`](https://github.com/nimiq-network/core/blob/master/README.md) 
- [x] I have searched [existing issues](https://github.com/nimiq-network/core/issues?q=is%3Aissue+sort%3Acreated-desc) and **this is not a duplicate**.

### General information

- Library version(s):0.20
- Devices/Simulators/Machine affected: Light clients
- Reproducible in the testnet? (Yes/No): yes

## Bug report

When a light client has to request missing blocks, there is a chance that the verification of this request fails. It could be that this issue exists for quite a while now but never really got noticed until in recent versions light clients had a low peer count (#2277) and not all gossipsub block header messages are received (#2288). Because of this combination, the node has to perform more missing block requests, probably causing the chance of this issue occurring to increase. When the verification fails the node is unable to continue accepting new blocks and can't recover from this state. Even when the macro sync kicks in.

It appears that blocks that already have been accepted, are requested as missing blocks. Also it tries to push it again from the missing block response. 

Note block 17778523 and 17778525 in the following snippet. And gossipsubs of 17778524 and 17778526 to be out of order.

```
2024-03-04T08:28:31.900441000Z DEBUG push                 | Accepted block block=#17778522:MI:00949eb694 kind="extend"
2024-03-04T08:28:31.915955000Z DEBUG push                 | Accepted block block=#17778523:MI:a7cd50e34d kind="extend"
2024-03-04T08:28:33.573134000Z DEBUG block_queue          | Received block via gossipsub block=#17778526:MI:5711919193 peer_id=12D3KooWGbKfQvV4r59BCYHoRzGbARTRFsxbx5SpUQb8C3ciQSiM
2024-03-04T08:28:33.574778000Z DEBUG block_queue          | Requesting missing blocks block_number=17778525 block_hash=6b1e5d0e649d17e7112279614277c8dcb07aafb56807ee08fac9d4cb9733ec01 head_hash=a7cd50e34dcf087199386aac6a26a8753af7df72003ce1fa9c2f028a3651e314 macro_height=17778464
2024-03-04T08:28:33.586606000Z DEBUG block_queue          | Received block via gossipsub block=#17778524:MA:ea5dd050b2 peer_id=12D3KooWGbKfQvV4r59BCYHoRzGbARTRFsxbx5SpUQb8C3ciQSiM
[...]

PR #2591, opened 2024-06-12
Make sure a received missing block is still relevant

nibhar:
As described in the discussion of #2295 a block could be requested even though it is no longer necessary to do so. Those requests would als be impossible to answer as they use a block in the future of the target as locator. This PR fixes that by simply not requesting these blocks anymore.


Eligioo (approved):
LGTM

jsdanielh (approved):

PR #2596, opened 2024-06-12
Increase mempool tests wait time

viquezclaudio:
Increase the time we wait to send and process txns in the mempool for mempool's tests. This is needed because sometimes, specially in slow machines (CI), the mempool tests fails because it has not finished processing all txns

### DIFF
--- a/mempool/tests/mod.rs
+++ b/mempool/tests/mod.rs
@@ -97,7 +97,7 @@ async fn send_txn_to_mempool(
     .await
     .expect("Send failed");
 
-    sleep(Duration::from_secs(1)).await;
+    sleep(Duration::from_secs(2)).await;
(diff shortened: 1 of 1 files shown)


hrxi (commented):
This is a fundamentally brittle approach. It would be cool if we could hear back from the mempool once the transaction has been processed.

But this works as a stop-gap solution.

viquezclaudio:
Thats what I discussed with @nibhar , the best way to do this Is to have some sort of signal from the mempool executor that we could await so we are sure all txns were processed, but currently we don't have such mechanism

nibhar (approved):
Exactly as @viquezclaudio said. Ideally we could just await something from the mempool here, but sadly it does not exist currently.

PR #2599, opened 2024-06-12
Peer provides the necessary services for syncing

viquezclaudio:
Make sure peers provide the requires services for the different stages of syncing

## Pull request checklist

### DIFF
--- a/consensus/src/sync/syncer.rs
+++ b/consensus/src/sync/syncer.rs
@@ -12,7 +12,10 @@ use nimiq_block::Block;
 use nimiq_blockchain_interface::AbstractBlockchain;
 use nimiq_blockchain_proxy::BlockchainProxy;
 use nimiq_hash::Blake2bHash;
-use nimiq_network_interface::network::{CloseReason, Network};
+use nimiq_network_interface::{
+    network::{CloseReason, Network},
+    peer_info::Services,
+};
 use nimiq_time::{interval, Interval};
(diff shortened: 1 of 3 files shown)


consensus/src/sync/syncer.rs
```diff
@@ -215,6 +218,12 @@ impl<N: Network, M: MacroSync<N::PeerId>, L: LiveSync<N>> Syncer<N, M, L> {
         let blockchain = self.blockchain.clone();
         let network = Arc::clone(&self.network);
 
+        // Light nodes do not provide accounts chunks, so they are classified as incompatible.
```
viquezclaudio, line 221: Maybe move this to the "move_peer_into_live_sync" function, so that light nodes are not moved into live sync?

nibhar:
I am not entirely familiar with this code so maybe this is an oversight on my end.

What about full nodes which are flagged incompatible by Full nodes macro sync as they don't provide History. Once the node transitions into live sync they are only required to provide AccountsChunks and FullBlocks which full nodes do provide. They need to be able to get out of the incompatible peer set now, as they are compatible with live sync.

Eligioo (approved):

viquezclaudio:
This PR might no longer be necessary in favor of re-using the history store for full nodes according to #2692

viquezclaudio:
Closing in favor of #2692

Issue #2186, opened 2024-02-05
Better insight into known addresses


sisou:
This is specifically relevant to the in-browser web-client.

Currently, users can only observe connected peers from the web-client. There is no way to see all the known addresses, even those not currently connected. The PoW client for Nimiq 1.x has this feature so that we can populate the Wallet network map with known but unconnected peers. 

This is a **feature request** to add a way to query the node's internal address book. Additionally, it would be nice to be notified about changes to it via emitted events (only nice-to-have).

sisou:
Closed by https://github.com/nimiq/core-rs-albatross/pull/2603.

PR #2603, opened 2024-06-13
Make the address book available to web clients

nibhar:
This PR adds the address book to the network in order to make it available for web client to query. 

Ultimately it would potentially also make sense to add that functionality on RPC.

Completes #2186.

### DIFF
--- a/web-client/example/module.js
+++ b/web-client/example/module.js
@@ -46,7 +46,7 @@ init().then(async () => {
 
     document.querySelector('#address-book').addEventListener("click", async () => {
         let contacts = await client.getAddressBook();
-        console.log(contacts);
+        console.table(contacts);
     });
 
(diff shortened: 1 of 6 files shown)


web-client/example/module.js
```diff
@@ -44,6 +44,11 @@ init().then(async () => {
         document.querySelector('#peers').textContent = numPeers;
     });
 
+    document.querySelector('#address-book').addEventListener("click", async () => {
+        let contacts = await client.getAddressBook();
+        console.log(contacts);
```
Eligioo:
```suggestion
        console.table(contacts);
```

Utilising `console.table` here makes sense as it's a list.

Eligioo:
LGTM. Just one minor comment.

PR #2619, opened 2024-06-14
Improve performance of history tree

paberr:
Keys we insert into the history tree are always sorted.
We can thus take advantage of the `append` functionality of mdbx which adds the key to the end of the database without a linear search.
Also add optimization for `all_ones` function since a profiling of the code revealed that `usize::count_zeros` is expensive and changing the operation to only rely on `usize::leading_zeros` and a mask is cheaper.

## Pull request checklist

### DIFF
--- a/blockchain/src/history/mmr_store.rs
+++ b/blockchain/src/history/mmr_store.rs
@@ -145,7 +145,7 @@ impl<'a, 'env> Store<Blake2bHash> for MMRStore<'a, 'env> {
     fn push(&mut self, elem: Blake2bHash) {
         if let Tx::Write(ref mut tx) = self.tx {
             let key = index_to_key(self.epoch_number, self.size);
-            tx.put(self.hist_tree_table, &key, &elem);
+            tx.append(self.hist_tree_table, &key, &elem);
             self.size += 1;
         }
     }
--- a/blockchain/src/history/mmr_store.rs
+++ b/blockchain/src/history/mmr_store.rs
@@ -1,4 +1,4 @@
-use std::{cmp, convert::TryInto};
+use std::{cmp, convert::TryInto, fmt};
 
 use nimiq_database::{
     traits::{ReadCursor, ReadTransaction, WriteCursor, WriteTransaction},
@@ -7,6 +7,9 @@ use nimiq_database::{
 use nimiq_hash::Blake2bHash;
 use nimiq_mmr::store::{LightStore, Store};
 use nimiq_primitives::policy::Policy;
+
+type WriteCursorProxy<'env> =
+    <WriteTransactionProxy<'env> as WriteTransaction<'env>>::WriteCursor<'env>;
 
 #[derive(Debug)]
 enum Tx<'a, 'env> {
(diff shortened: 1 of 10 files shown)


jsdanielh (approved):
LGTM

primitives/mmr/src/mmr/position.rs
```diff
@@ -143,7 +143,15 @@ impl Ord for Position {
...
+    if num == 0 {
+        return false;
+    }
+    // Find the highest set bit
+    let highest_set_bit = usize::BITS - num.leading_zeros();
+    // Create a mask with all 1s from the highest set bit to the least significant bit
+    let mask = (1 << highest_set_bit) - 1;
+    // Check if the number matches the mask
+    num == mask
 }
```
hrxi, line 150: I checked a couple of implementations on godbolt.org, and it seems `(num + 1).is_power_of_two()` generates the fewest instructions (just 4).

blockchain/src/history/mmr_store.rs
```diff
@@ -143,29 +159,33 @@ fn key_to_index(key: Vec<u8>) -> Option<(u32, usize)> {
...
     fn push(&mut self, elem: Blake2bHash) {
-        if let Tx::Write(ref mut tx) = self.tx {
+        // This function assumes that there is no higher epoch.
+        // Otherwise the append method will fail.
+        if let Some(ref mut cursor) = self.cursor {
             let key = index_to_key(self.epoch_number, self.size);
-            tx.put(self.hist_tree_table, &key, &elem);
+            cursor.append(&key, &elem);
             self.size += 1;
         }
```
hrxi, line 170: This should panic otherwise, as you must not push an element on a read-only store.

blockchain/src/history/mmr_store.rs
```diff
@@ -143,29 +159,33 @@ fn key_to_index(key: Vec<u8>) -> Option<(u32, usize)> {
...
+            let key = index_to_key(self.epoch_number, self.size - 1);
+            cursor.seek_key::<Vec<u8>, Blake2bHash>(&key).unwrap();
             for _ in 0..cmp::min(num_elems, self.size) {
-                let key = index_to_key(self.epoch_number, self.size - 1);
-                tx.remove(self.hist_tree_table, &key);
+                cursor.remove();
+                cursor.prev::<Vec<u8>, Blake2bHash>();
                 self.size -= 1;
             }
         }
```
hrxi, line 187: Same here, this should panic.

database/src/proxy/transaction.rs
```diff
@@ -111,6 +111,17 @@ impl<'db> WriteTransaction<'db> for WriteTransactionProxy<'db> {
...
+    where
+        K: nimiq_database_value::AsDatabaseBytes + ?Sized,
+        V: nimiq_database_value::AsDatabaseBytes + ?Sized,
+    {
+        match self.txn {
+            TransactionProxy::ReadTransaction(_) => unreachable!(),
+            TransactionProxy::WriteTransaction(ref mut txn) => txn.append(table, key, value),
+        }
+    }
+
```
hrxi, line 124: This seems to be dead code. If it's not used, it should probably be removed.

hrxi (commented):
Looks good to me, there's some unneeded code and I found some improvements.

primitives/mmr/src/mmr/position.rs
paberr: Amazing!

PR #2621, opened 2024-06-15
Catch if peer sends an invalid epoch_number instead of panicking

### LINKED ISSUE #2509: FIXME: Ban peer if it sends an invalid epoch_number instead of panicking.

Eligioo:
This fixes #2509 .

## Pull request checklist

### DIFF
--- a/consensus/src/sync/history/sync_clustering.rs
+++ b/consensus/src/sync/history/sync_clustering.rs
@@ -52,7 +52,7 @@ impl<TNetwork: Network> HistoryMacroSync<TNetwork> {
             Arc::clone(&network),
             peer_id,
             locators,
-            1000, // TODO: Use other value
+            Self::MAX_REQUEST_EPOCHS,
         )
         .await;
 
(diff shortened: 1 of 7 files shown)


styppo (approved):

consensus/src/sync/history/sync_clustering.rs
```diff
@@ -48,13 +48,10 @@ impl<TNetwork: Network> HistoryMacroSync<TNetwork> {
...
         };
 
-        let result = Self::request_macro_chain(
-            Arc::clone(&network),
-            peer_id,
-            locators,
-            1000, // TODO: Use other value
-        )
-        .await;
+        let max_epochs = 1000; // TODO: Use other value
```
nibhar: Could be a `const` on the type. 

As it is used by both implementors of the `MacroSync` trait it could also be moved there to be declared by every implementor if that generalization is plausible.

consensus/src/sync/light/sync_stream.rs
```diff
@@ -245,12 +244,11 @@ impl<TNetwork: Network> LightMacroSync<TNetwork> {
...
                             }
-                            // FIXME: Ban peer if it sends an invalid epoch_number instead of panicking.
                             Policy::first_block_of(epoch_ids.checkpoint_epoch_number() as u32)
-                                .expect("The supplied epoch number is out of bounds")
+                                .unwrap_or(u32::MAX)
                         })
-                        + Policy::blocks_per_batch()
-                        - 1;
+                        .saturating_add(Policy::blocks_per_batch())
+                        .saturating_sub(1);
```
nibhar: As blocks per batch is a positive non zero number the 1 could be moved into the saturating add, but it does not really matter.

Also looks like all of this would also be `macro_block_after()` and if so I would prefer to use it instead.

nibhar (approved):
Just some minor comments.

hrxi (approved):
LGTM.

hrxi:
Rebased to merge.

PR #2627, opened 2024-06-17
Only accept control transactions over the control transaction topic

### LINKED ISSUE #2614: Normal transactions can be sent via the control transaction topic
The difference is currently only checked on the client side:

https://github.com/nimiq/core-rs-albatross/blob/2abf923bd73877a7e45f239756195646d60da9f9/consensus/src/consensus/consensus_proxy.rs#L65-L68

hrxi:
Do this by using a wrapper that only accepts control transactions.

Fixes #2614.

### DIFF
--- a/primitives/transaction/src/control_transaction.rs
+++ b/primitives/transaction/src/control_transaction.rs
@@ -63,3 +63,78 @@ impl<'de> Deserialize<'de> for ControlTransaction {
         ControlTransaction::try_from(tx).map_err(D::Error::custom)
     }
 }
+
+#[cfg(test)]
+mod test {
+    use nimiq_keys::Address;
+    use nimiq_primitives::{account::AccountType, coin::Coin, networks::NetworkId, policy::Policy};
+    use nimiq_serde::{Deserialize, Serialize};
+
+    use super::ControlTransaction;
+    use crate::{account::staking_contract::IncomingStakingTransactionData, Transaction};
+
+    fn normal_tx() -> Transaction {
+        Transaction::new_basic(
+            Address::default(),
+            Address::default(),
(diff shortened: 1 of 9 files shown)


primitives/transaction/src/control_transaction.rs
```diff
@@ -0,0 +1,65 @@
...
+    fn from(tx: ControlTransactionError) -> Transaction {
+        tx.0
+    }
+}
+
+impl TryFrom<Transaction> for ControlTransaction {
+    type Error = ControlTransactionError;
+
+    fn try_from(tx: Transaction) -> Result<ControlTransaction, ControlTransactionError> {
+        if tx.sender_type != AccountType::Staking && tx.recipient_type != AccountType::Staking {
```
Eligioo: Should we also check if the value is 0 here?

hrxi: We discussed this shortly, and the answer was yes. However, since that is a breaking change, it cannot go into the next (compatible) testnet release.

Eligioo (approved):

jsdanielh (approved):

primitives/transaction/src/control_transaction.rs
sisou: Signaling transactions also have `TransactionFlags::SIGNALING` set, but AFAIK the combination of that flag and zero value is already checked in the transaction's internal validity and/or parser.

The parsed `IncomingStakingTransactionData` has a `.is_signaling()` method:
https://github.com/nimiq/core-rs-albatross/blob/21d91b3187e234c3319d9e3d774512fb04eb3c35/primitives/transaction/src/account/staking_contract/structs.rs#L90

PR #2629, opened 2024-06-17
Use faster byte array serialization for `Address`

hrxi:
It was missed because the serialization wasn't derived.


jsdanielh (approved):

hrxi:
Will push it to #2630 instead.

PR #2634, opened 2024-06-17
Verify sender account type in mempool

### LINKED ISSUE #2605: Staking transactions are not enforced to be to / from the staking contract
`verify` of `Transaction` currently checks that the `Staking` sender / recipient type must be set, if the the sender / recipient address is the staking contract, but does not enforce that the sender / recipient must be the staking contract, if the sender / recipient sender type is `Staking`.
Verification of the sender / recipient data, and potentially later processing of the data, are solely based on the sender / recipient type though.
Thus, it might be possible to, for example, send a transaction that is interpreted as a staking transaction, via the `Staking` sender / recipient type, which is not in fact to / from the staking contract.

https://github.com/nimiq/core-rs-albatross/blob/bf6bab5b938ef158f58fae4347276f2581d5102a/primitives/transaction/src/lib.rs#L337-L349

PS: I did not confirm the issue on-chain, just filing this issue based on reading the code.

styppo:
The account type on the sender side is now properly checked in the mempool via `reserve_balance()`.

The intrinsic staking contract account type check is removed entirely from the transaction verification. A mismatch on the recipient side will now result in a failed transaction, which brings it in line with other account type mismatches.

Closes #2605

### DIFF
--- a/blockchain/src/blockchain/wrappers.rs
+++ b/blockchain/src/blockchain/wrappers.rs
@@ -160,11 +160,11 @@ impl Blockchain {
         if let Ok(account) = self.state.accounts.get(address, None) {
             Some(account)
         } else {
-            warn!(%address, "Could not get account for address");
             None
         }
     }
 
+    /// The given account must correspond to the sender of the given transaction.
     pub fn reserve_balance(
         &self,
         account: &Account,
@@ -181,6 +181,7 @@ impl Blockchain {
         )
     }
 
+    /// The given account must correspond to the sender of the given transaction.
     pub fn release_balance(
         &self,
         account: &Account,
--- a/mempool/src/mempool_state.rs
+++ b/mempool/src/mempool_state.rs
@@ -74,15 +74,11 @@ impl MempoolState {
 
         if let Some(sender_state) = self.state_by_sender.get_mut(&tx.sender) {
             let reserved_balance = &mut sender_state.reserved_balance;
-            blockchain
-                .reserve_balance(&sender_account, tx, reserved_balance)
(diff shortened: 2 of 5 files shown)


jsdanielh (approved):

sisou:
Is `reserve_balance()`, which now has the check, also called by validators that receive the produced block without having seen the tx in their mempool?I.e. does this protect against a malicious validator that includes such a tx in a block?

styppo:
No, `reserve_balance()` is only used by the mempool. The sender account type is checked during commit [here](https://github.com/nimiq/core-rs-albatross/blob/2abf923bd73877a7e45f239756195646d60da9f9/primitives/account/src/accounts.rs#L384) and [here](https://github.com/nimiq/core-rs-albatross/blob/2abf923bd73877a7e45f239756195646d60da9f9/primitives/account/src/accounts.rs#L497), i.e. a transaction with an incorrect sender account type cannot be included in a block. The mempool now rejects these transactions such that validators don't attempt to produce invalid blocks.

blockchain/src/blockchain/wrappers.rs
```diff
@@ -160,11 +160,11 @@ impl Blockchain {
         if let Ok(account) = self.state.accounts.get(address, None) {
             Some(account)
         } else {
-            warn!(%address, "Could not get account for address");
             None
         }
     }
 
+    /// The given account must correspond to the sender of the given transaction.
```
nibhar, line 167: Out of curiosity: what happens if it isn't?

nibhar (approved):

PR #2638, opened 2024-06-18
Add metrics to the database

paberr:
Adds detailed metrics to the database crate – similarly to how `reth` is doing it.
This currently uses the `metrics` crate, which is not compatible with our `prometheus_client` crate.
I implemented a connection between these two but it is not really nice. Thus, I'll also open an issue to improve this situation at a later point.

Open TODOs:
- [x] Initialise database with metrics enabled.

## Pull request checklist

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -1,7 +1,6 @@
 use std::time::Duration;
 
 use log::info;
-use nimiq::prover::prover_main;
 pub use nimiq::{
     client::Client,
     config::{command_line::CommandLine, config::ClientConfig, config_file::ConfigFile},
@@ -13,6 +12,7 @@ pub use nimiq::{
         signal_handling::initialize_signal_handler,
     },
 };
+use nimiq::{extras::metrics_server::install_metrics, prover::prover_main};
 
 async fn main_inner() -> Result<(), Error> {
     // Keep for potential future reactivation
@@ -39,7 +39,7 @@ async fn main_inner() -> Result<(), Error> {
     // Initialize panic hook.
     initialize_panic_reporting();
 
-    // Initialize signal handler
+    // Initialize signal handler.
     initialize_signal_handler();
 
     // Early return in case of a proving process.
@@ -62,6 +62,12 @@ async fn main_inner() -> Result<(), Error> {
     let rpc_config = config.rpc_server.clone();
     let metrics_config = config.metrics_server.clone();
     let metrics_enabled = metrics_config.is_some();
+
+    // Initialize database logging.
+    let mut metrics_collector = None;
+    if metrics_enabled {
+        metrics_collector = Some(install_metrics());
+    }
 
     // Create client from config.
     let mut client: Client = Client::from_config(config).await?;
@@ -128,6 +134,7 @@ async fn main_inner() -> Result<(), Error> {
             client.consensus_proxy(),
             client.network(),
             &nimiq_task_metric,
+            metrics_collector.unwrap(),
         )
     }
 
--- a/database/Cargo.toml
+++ b/database/Cargo.toml
@@ -24,10 +24,15 @@ workspace = true
 
 [dependencies]
 bitflags = "2.6"
-libmdbx = "0.5.0"
+libmdbx = { git = "https://github.com/paberr/libmdbx-rs" }
 log = { workspace = true }
 tempfile = "3"
 thiserror = "1.0"
+metrics = "0.23"
+parking_lot = "0.12"
+rustc-hash = "1.1"
+strum = "0.26"
+strum_macros = "0.26"
 
 nimiq-database-value = { workspace = true }
 
--- a/database/src/lib.rs
+++ b/database/src/lib.rs
@@ -2,6 +2,7 @@ use bitflags::bitflags;
 
 mod error;
 pub mod mdbx;
+mod metrics;
 /// Database implementation that can handle volatile and persistent storage.
 pub mod proxy;
 /// Abstraction for methods related to the database.
--- a/database/src/mdbx/database.rs
+++ b/database/src/mdbx/database.rs
@@ -1,18 +1,27 @@
-use std::{borrow::Cow, fs, path::Path, sync::Arc};
+use std::{borrow::Cow, fmt, fs, path::Path, sync::Arc};
 
 use libmdbx::NoWriteMap;
 use log::info;
 
 use super::{MdbxReadTransaction, MdbxWriteTransaction};
-use crate::{traits::Database, DatabaseProxy, Error, TableFlags};
+use crate::{metrics::DatabaseEnvMetrics, traits::Database, DatabaseProxy, Error, TableFlags};
 
 pub(super) type DbKvPair<'a> = (Cow<'a, [u8]>, Cow<'a, [u8]>);
 
 /// Wrapper around the mdbx database handle.
 /// A database can hold multiple tables.
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct MdbxDatabase {
     pub(super) db: Arc<libmdbx::Database<NoWriteMap>>,
+    pub(super) metrics: Option<Arc<DatabaseEnvMetrics>>,
+}
+
+impl fmt::Debug for MdbxDatabase {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("MdbxDatabase")
+            .field("db", &self.db)
+            .finish()
+    }
 }
 
 impl Database for MdbxDatabase {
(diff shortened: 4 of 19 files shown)


Cargo.toml
```diff
@@ -162,8 +162,8 @@ categories = ["cryptography::cryptocurrencies"]
 keywords = ["nimiq", "cryptocurrency", "blockchain"]
 
 [workspace.lints]
-clippy.assigning_clones = "allow" # false positives: https://github.com/rust-lang/rust-clippy/issues/12709
-clippy.empty_docs = "allow" # false positives: https://github.com/rust-lang/rust-clippy/issues/12377
+clippy.assigning_clones = "allow"   # false positives: https://github.com/rust-lang/rust-clippy/issues/12709
+clippy.empty_docs = "allow"         # false positives: https://github.com/rust-lang/rust-clippy/issues/12377
```
hrxi: Formatting should probably go into a different commit, but I can do that before the PR is merged.

client/src/main.rs
```diff
@@ -128,6 +134,7 @@ async fn main_inner() -> Result<(), Error> {
             client.consensus_proxy(),
             client.network(),
             &nimiq_task_metric,
+            metrics_collector.unwrap(),
```
hrxi, line 137: The `unwrap` is a bit annoying, but I couldn't immediately see a better way.

database/Cargo.toml
```diff
@@ -21,10 +24,15 @@ workspace = true
 
 [dependencies]
 bitflags = "2.5"
-libmdbx = "0.5.0"
+libmdbx = { git = "https://github.com/paberr/libmdbx-rs" }
```
hrxi, line 27: Upstream PR: https://github.com/vorot93/libmdbx-rs/pull/31.

database/src/lib.rs
```diff
@@ -2,6 +2,7 @@ use bitflags::bitflags;
 
 mod error;
 pub mod mdbx;
+pub(crate) mod metrics;
```
hrxi:
```suggestion
mod metrics;
```
Is there a reason for all these `pub(crate)` annotations? E.g. here, it could simply be dropped.

database/src/mdbx/database.rs
```diff
@@ -49,15 +50,25 @@ impl Database for MdbxDatabase {
...
+        }
+
         MdbxTable { name }
     }
 
     fn read_transaction(&self) -> Self::ReadTransaction<'_> {
-        MdbxReadTransaction::new(self.db.begin_ro_txn().unwrap())
+        MdbxReadTransaction::new(
+            self.db.begin_ro_txn().unwrap(),
+            self.metrics.as_ref().cloned(),
```
hrxi:
```suggestion
            self.metrics.clone(),
```

database/src/metrics.rs
```diff
@@ -0,0 +1,423 @@
...
+    collections::HashMap,
+    hash::BuildHasherDefault,
+    time::{Duration, Instant},
+};
+
+use libmdbx::CommitLatency;
+use log::warn;
+use metrics::{counter, gauge, histogram, Counter, Gauge, Histogram, IntoLabels};
+use parking_lot::RwLock;
+use rustc_hash::{FxHashMap, FxHasher};
```
hrxi, line 15: Why use this over the hashmap from `std`?

database/src/metrics.rs
```diff
@@ -0,0 +1,423 @@
+// This file has been ported and adapted from reth (https://github.com/paradigmxyz/reth).
```
hrxi, line 1: Should probably get license information and the commit it's from.

The original license is MIT OR Apache-2.0, so we can use it in our code.

database/src/metrics.rs
```diff
@@ -0,0 +1,423 @@
...
+                OperationMetrics::new_with_labels(&[
+                    (Labels::Table.as_str(), table.to_string()),
+                    (Labels::Operation.as_str(), operation.as_str().to_string()),
+                ]),
+            );
+        }
+    }
+
+    /// Record a metric for database operation executed in `f`.
+    /// Panics if a metric recorder is not found for the given table and operation.
```
hrxi: Documentation seems to be wrong, I see no panic.

metrics-server/src/metrics.rs
```diff
@@ -0,0 +1,267 @@
...
+fn convert_unit_to_prometheus(unit: Unit) -> PrometheusUnit {
+    match unit {
+        Unit::Count => PrometheusUnit::Other("count".to_string()),
+        Unit::Percent => PrometheusUnit::Other("percent".to_string()),
+        Unit::Seconds => PrometheusUnit::Other("seconds".to_string()),
+        Unit::Milliseconds => PrometheusUnit::Other("milliseconds".to_string()),
+        Unit::Microseconds => PrometheusUnit::Other("microseconds".to_string()),
+        Unit::Nanoseconds => PrometheusUnit::Other("nanoseconds".to_string()),
+        Unit::Tebibytes => PrometheusUnit::Other("tebibytes".to_string()),
+        Unit::Gigibytes => PrometheusUnit::Other("gigibytes".to_string()),
```
hrxi:
```suggestion
        Unit::Gigibytes => PrometheusUnit::Other("gibibytes".to_string()),
```

metrics-server/src/metrics.rs
```diff
@@ -0,0 +1,267 @@
...
+}
+
+impl CounterFn for MetricsCounter {
+    fn increment(&self, value: u64) {
+        self.inner.inc_by(value);
+    }
+
+    fn absolute(&self, value: u64) {
+        // This technically violates the assumptions of having a monotonically increasing counter.
+        self.inner.inner().store(value, Ordering::SeqCst);
```
hrxi:
```suggestion
        self.inner.inner().fetch_max(value, Ordering::Relaxed);
```

This way it's not violating the assumption (and this is also how it's implemented upstream).

database/src/mdbx/metrics_handler.rs
```diff
@@ -0,0 +1,132 @@
...
+    },
+    time::{Duration, Instant},
+};
+
+use libmdbx::{CommitLatency, TransactionKind};
+use log::{trace, warn};
+
+use crate::metrics::{DatabaseEnvMetrics, TransactionMode, TransactionOutcome};
+
+/// Du
```
hrxi:
```suggestion
```

database/src/metrics.rs
```diff
@@ -0,0 +1,423 @@
...
+use strum_macros::{EnumCount as EnumCountMacro, EnumIter};
+
+const LARGE_VALUE_THRESHOLD_BYTES: usize = 4096;
+
+/// Caches metric handles for database environment to make sure handles are not re-created
+/// on every operation.
+///
+/// Requires a metric recorder to be registered before creating an instance of this struct.
+/// Otherwise, metric recording will no-op.
+#[derive(Debug)]
```
hrxi: No need to `derive(Debug)` here, I think, and then you can also remove the no-op debug implementation from below.

database/src/mdbx/metrics_handler.rs
```diff
@@ -0,0 +1,132 @@
...
+    /// to do anything on [`Drop::drop`].
+    close_recorded: bool,
+    /// If `true`, the backtrace of transaction will be recorded and logged.
+    /// See [`MetricsHandler::log_backtrace_on_long_read_transaction`].
+    record_backtrace: bool,
+    /// If `true`, the backtrace of transaction has already been recorded and logged.
+    /// See [`MetricsHandler::log_backtrace_on_long_read_transaction`].
+    backtrace_recorded: AtomicBool,
+    pub(super) env_metrics: Arc<DatabaseEnvMetrics>,
+    _tx_kind: PhantomData<K>,
```
hrxi:
```suggestion
    transaction_mode: TransactionMode,
```
would probably be better, then there aren't two instantiations of this struct.

database/src/mdbx/metrics_handler.rs
```diff
@@ -0,0 +1,132 @@
...
+    /// If `true`, the metric about transaction closing has already been recorded and we don't need
+    /// to do anything on [`Drop::drop`].
+    close_recorded: bool,
+    /// If `true`, the backtrace of transaction will be recorded and logged.
+    /// See [`MetricsHandler::log_backtrace_on_long_read_transaction`].
+    record_backtrace: bool,
+    /// If `true`, the backtrace of transaction has already been recorded and logged.
+    /// See [`MetricsHandler::log_backtrace_on_long_read_transaction`].
+    backtrace_recorded: AtomicBool,
+    pub(super) env_metrics: Arc<DatabaseEnvMetrics>,
```
hrxi, line 36: I think you should be able to get away with a reference here.

hrxi (commented):
Looks good to me. Most of the comments are minor or questions. The license/version this was taken from should probably be stated.

database/src/metrics.rs
paberr: Speed, I took it over from the reth repo

metrics-server/src/metrics.rs
paberr: Oh, that means the `Unit` is wrong?

database/src/mdbx/metrics_handler.rs
paberr: Not sure what you mean with that. Replacing the `Arc` with a `&`?

paberr: I don't think there's much value in that as the `Database` struct needs to have an `Arc` anyways.

database/src/metrics.rs
paberr: Changed, but now we need to manually implement `Debug` for `Database` and `Transaction`.

metrics-server/src/metrics.rs
hrxi: Yes: https://en.wikipedia.org/w/index.php?title=Byte&oldid=1227965783#Multiple-byte_units.

paberr:
Incompatible with new DB refactor

PR #2642, opened 2024-06-18
Handle unsupported light history store functionality

### LINKED ISSUE #2639: Panic `history_store.prove_chunk` for full node is unimplemented
### General information

- Library version(s): v0.22.2
- Node type: full node
- Validator: yes
- Caller: https://github.com/nimiq/core-rs-albatross/blob/albatross/consensus/src/messages/handlers.rs#L160

[panic-light-history-store.log](https://github.com/user-attachments/files/15886094/panic-light-history-store.log)

viquezclaudio:
Dont panic if a request that is not supported by the light history store is performed Fixes #2639

...
## Pull request checklist

### DIFF
--- a/blockchain/src/history/light_history_store.rs
+++ b/blockchain/src/history/light_history_store.rs
@@ -276,7 +276,7 @@ impl HistoryInterface for LightHistoryStore {
         _epoch_number: u32,
         _num_hist_txs: usize,
     ) -> Option<(Blake2bHash, u64)> {
-        None
+        unimplemented!()
     }
 
     fn tx_in_validity_window(
@@ -294,7 +294,7 @@ impl HistoryInterface for LightHistoryStore {
         _tx_hash: &Blake2bHash,
         _txn_option: Option<&TransactionProxy>,
     ) -> Vec<HistoricTransaction> {
-        todo!()
+        unimplemented!()
     }
 
     fn get_block_transactions(
@@ -302,7 +302,7 @@ impl HistoryInterface for LightHistoryStore {
         _block_number: u32,
         _txn_option: Option<&TransactionProxy>,
     ) -> Vec<HistoricTransaction> {
-        todo!()
+        unimplemented!()
     }
 
     fn get_epoch_transactions(
@@ -424,6 +424,10 @@ impl HistoryInterface for LightHistoryStore {
         nimiq_mmr::error::Error,
     > {
         unimplemented!()
+    }
+
+    fn is_light(&self) -> bool {
+        true
     }
 }
 
--- a/consensus/src/messages/handlers.rs
+++ b/consensus/src/messages/handlers.rs
@@ -90,10 +90,15 @@ impl<N: Network> Handle<N, BlockchainProxy> for RequestMacroChain {
 impl<N: Network> Handle<N, Arc<RwLock<Blockchain>>> for RequestBatchSet {
     fn handle(
         &self,
-        _peer_id: N::PeerId,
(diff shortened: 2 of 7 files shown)


blockchain/src/history/light_history_store.rs
```diff
@@ -294,23 +294,26 @@ impl HistoryInterface for LightHistoryStore {
         _tx_hash: &Blake2bHash,
         _txn_option: Option<&TransactionProxy>,
     ) -> Vec<HistoricTransaction> {
-        todo!()
+        log::warn!("Get history tx by hash is unsupported in the light history store");
+        vec![]
```
sisou: Hmm. Not sure just retuning an empty list is the way to go.
If I make this request to a remote full node via RPC, I don't see this log message.

How much work is it to change to a `Result` return type?

Eligioo (approved):

nibhar (commented):
I would prefer this to still panic if something is called which does not make any sense like `get_epoch_transactions`. That way we get a stacktrace of the offending code path. I would much rather, as you have done for `RequestBatchSet::handle`, make sure these are in fact `unreachable!()`.

Obviously there may also be malicious actors who can send any and all requests to any and all nodes and they must safeguard for it, but they should do it earlier than when they try to retrieve data.

viquezclaudio:
According to what we discussed, requests to unsupported functionality will not panic the node but will log who made those requests. (Should we disconnect from those peers and classify them as malicious? as this should not happen)
Deliberately calling functions from the light history store that are not supported will panic with unimplemented.

rpc-server/src/dispatchers/blockchain.rs
```diff
@@ -311,8 +311,16 @@ impl BlockchainInterface for BlockchainDispatcher {
         max: Option<u16>,
     ) -> RPCResult<Vec<Blake2bHash>, (), Self::Error> {
         if let BlockchainProxy::Full(blockchain) = &self.blockchain {
+            let blockchain = blockchain.read();
+
+            if blockchain.history_store.is_light() {
+                log::warn!(
+                    "Transaction hashes by address is not supported by the light history store"
```
sisou, line 351: @viquezclaudio I think this also applies to `get_transaction_by_hash`.

sisou (commented):
Light history stores can also not provide transaction proofs to light clients, I don't know if this is the PR for that.

consensus/src/messages/handlers.rs
```diff
@@ -154,10 +159,17 @@ impl<N: Network> Handle<N, Arc<RwLock<Blockchain>>> for RequestBatchSet {
...
         blockchain: &Arc<RwLock<Blockchain>>,
     ) -> Result<HistoryChunk, HistoryChunkError> {
-        if let Some(chunk) = blockchain.read().history_store.prove_chunk(
+        let blockchain = blockchain.read();
+
+        if blockchain.history_store.is_light() {
+            log::warn!(peer=%peer_id,"Request history chunk is not supported by the light history store");
+            return Err(HistoryChunkError::Other);
+        }
+
```
hrxi, line 171: I think it would be better if we didn't even listen for these requests in the light client. Instead of erroring out once we get a request.

viquezclaudio:
This PR might no longer be necessary in favor of re-using the history store with no indexing for full nodes as per 
#2692

paberr:
> This PR might no longer be necessary in favor of re-using the history store with no indexing for full nodes as per #2692

Indeed, PR #2692 removes the light history store completely.
I'm going to close this PR.

PR #2644, opened 2024-06-18
Tiny fixes to Macro blocks

nibhar:
While looking for performance improvements I came across two smaller problems.

* The Validator should keep on polling the macro producer even if it encounters an outdated state. Without it no repolling will be done until a subsequent wake. 
* Tendermint should make a better choice of the next future contribution to verify as some of them have a subset (or equal) contributors than those which have already been verified, making the work done to verify them again pointless. This is not causing any actual wrong behaviour, just does some unnecessary work. 

The selection criteria could be more sophisticated than that and if it is considered necessary to improve it, I will create an issue for it, as I don't think it is at this point necessary to do so.


hrxi (approved):
LGTM, thanks!

PR #2658, opened 2024-06-20
Limiting transactions sender and recipient data.

### LINKED ISSUE #2604: Recipient and sender data lengths are not checked / limited for regular transactions
In Nimiq PoW, extra data was limited to 64 bytes for regular (non-contract-creation) transactions.
At first glance, such limit does not seem to exist in core-rs-albatross currently.
Specifically, `verify_incoming_transaction` and `verify_outgoing_transaction` of `BasicAccountVerifier` do not perform such checks.

Similarly, I was wondering, that `verify_outgoing_transaction` of `StakingContractVerifier` is not verifying the sender data.
This might be covered in later stages of staking transaction processing though.

ii-cruz:
Sets a maximum for the sender and recipient data. The check is added on deserialization and the transaction verification steps as discussed on the issue.

- The sender data can only have 1 byte for the staking contract outgoing transactions
- The recipient data can have all the validator fields in the case of the Create Validator and Update Validator which are the biggest ones

The Incoming Staking Transaction Data has for these cases 464 bytes (SchnorrPublicKey :32, BlsPublicKey : 285, reward_address :20, signal_data: 32, BlsSignature: 95) + the SignatureProof.

The Signature Proof is technically unbounded due to the web auth fields and Merkle proof. So this PR adds limits to those as well. 

The proposal is to set these Limits
Recipient data total size 2112 bytes (464 +6 + Signature Proof = 1641)
Web auth total size ~~500 bytes as suggested in #2686 by @sisou~~ changed to 512 by feedback.
Merkle proof max 32 items. 1029 bytes

## Pull request checklist

### DIFF
--- a/primitives/src/policy.rs
+++ b/primitives/src/policy.rs
@@ -58,6 +58,22 @@ impl Policy {
     /// The maximum allowed size, in bytes, for a micro block body.
     pub const MAX_SIZE_MICRO_BODY: usize = 100_000;
 
+    // Maximum lengths for transaction sender_data and recipient_data fields used during deserialization.
+    /// Maximum size for the sender data field of the transactions.
+    /// The only transactions using this field are RemoveStake and Delete Validator.
+    pub const MAX_TX_SENDER_DATA_SIZE: usize = 1;
+    /// Maximum size for the recipient data field of the transactions.
+    /// This is used by the HTLC, Vesting and staking contract transactions. Basic transactions are allowed to use it as well.
+    /// The biggest transactions are the create validator and update validator with a web auth signature.
+    pub const MAX_TX_RECIPIENT_DATA_SIZE: usize = 2112;
+    /// Maximum length for basic transaction recipient_data used during intrinsic transaction verification.
+    pub const MAX_BASIC_TX_RECIPIENT_DATA_SIZE: usize = 64;
+    /// Maximum size for transaction's merkle path proofs.
+    /// 32 bytes for Merkle Path Node * 32 max length + 4 bytes to encode the bools for left/right + 1 byte for the serialize vec count.
+    pub const MAX_MERKLE_PATH_SIZE: usize = 1029;
+    /// Maximum size for the total web auth fields.
+    pub const MAX_SUPPORTED_WEB_AUTH_SIZE: usize = 512;
+
     /// The current version number of the protocol. Changing this always results in a hard fork.
     pub const VERSION: u16 = 1;
 
--- a/primitives/transaction/src/account/basic_account.rs
+++ b/primitives/transaction/src/account/basic_account.rs
@@ -1,5 +1,4 @@
-use log::error;
-use nimiq_primitives::account::AccountType;
+use nimiq_primitives::{account::AccountType, policy::Policy};
 use nimiq_serde::Deserialize;
 
 use crate::{
@@ -15,11 +14,19 @@ impl AccountTransactionVerification for BasicAccountVerifier {
         assert_eq!(transaction.recipient_type, AccountType::Basic);
 
         if transaction.value.is_zero() {
-            error!(
+            warn!(
                 "The following transaction can't have a zero value:\n{:?}",
                 transaction
             );
             return Err(TransactionError::ZeroValue);
+        }
+
+        if transaction.recipient_data.len() > Policy::MAX_BASIC_TX_RECIPIENT_DATA_SIZE {
+            warn!(
+                "The following transaction's recipient data exceeds the maximum size:\n{:?}",
+                transaction
+            );
+            return Err(TransactionError::Overflow);
         }
 
         if transaction
@@ -27,7 +34,7 @@ impl AccountTransactionVerification for BasicAccountVerifier {
             .contains(TransactionFlags::CONTRACT_CREATION)
             || transaction.flags.contains(TransactionFlags::SIGNALING)
         {
-            error!(
+            warn!(
                 "Contract creation and signaling not allowed for this transaction:\n{:?}",
                 transaction
             );
@@ -40,13 +47,21 @@ impl AccountTransactionVerification for BasicAccountVerifier {
     fn verify_outgoing_transaction(transaction: &Transaction) -> Result<(), TransactionError> {
         assert_eq!(transaction.sender_type, AccountType::Basic);
 
+        if !transaction.sender_data.is_empty() {
+            warn!(
+                "The following transaction can't have sender data:\n{:?}",
+                transaction
+            );
+            return Err(TransactionError::Overflow);
+        }
+
         // Verify signer & signature.
         let signature_proof = SignatureProof::deserialize_all(&transaction.proof)?;
 
         if !signature_proof.is_signed_by(&transaction.sender)
             || !signature_proof.verify(&transaction.serialize_content())
         {
-            error!(
+            warn!(
                 "The following transaction has an invalid proof:\n{:?}",
                 transaction
             );
--- a/primitives/transaction/src/account/htlc_contract.rs
+++ b/primitives/transaction/src/account/htlc_contract.rs
@@ -1,6 +1,5 @@
 use std::{borrow::Cow, str::FromStr};
 
-use log::error;
 use nimiq_hash::{
     sha512::{Sha512Hash, Sha512Hasher},
     Blake2bHash, Blake2bHasher, Hasher, Sha256Hash, Sha256Hasher,
@@ -26,7 +25,7 @@ impl AccountTransactionVerification for HashedTimeLockedContractVerifier {
             .flags
             .contains(TransactionFlags::CONTRACT_CREATION)
         {
-            error!(
+            warn!(
                 "Only contract creation is allowed for the following transaction:\n{:?}",
                 transaction
             );
@@ -34,7 +33,7 @@ impl AccountTransactionVerification for HashedTimeLockedContractVerifier {
         }
 
         if transaction.flags.contains(TransactionFlags::SIGNALING) {
-            error!(
+            warn!(
                 "Signaling not allowed for the following transaction:\n{:?}",
                 transaction
             );
@@ -63,6 +62,14 @@ impl AccountTransactionVerification for HashedTimeLockedContractVerifier {
 
     fn verify_outgoing_transaction(transaction: &Transaction) -> Result<(), TransactionError> {
         assert_eq!(transaction.sender_type, AccountType::HTLC);
+
+        if !transaction.sender_data.is_empty() {
+            warn!(
+                "The following transaction can't have sender data:\n{:?}",
+                transaction
+            );
+            return Err(TransactionError::Overflow);
+        }
 
         let proof = OutgoingHTLCTransactionProof::parse(transaction)?;
         proof.verify(transaction)?;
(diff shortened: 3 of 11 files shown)


jsdanielh (commented):
Could you add a test?

sisou:
Do we not want to keep the 64 byte limit for regular basic-to-basic transactions?

hrxi (approved):
I was a bit worried that these constants are unchecked, meaning that if we make any of the sender/recipient data longer, we won't immediately notice that we might go over these sizes. I talked with @ii-cruz about it, and we could add tests that assert this, so I'm fine with this going in as-is.

primitives/src/policy.rs
```diff
@@ -58,6 +58,12 @@ impl Policy {
     /// The maximum allowed size, in bytes, for a micro block body.
     pub const MAX_SIZE_MICRO_BODY: usize = 100_000;
 
+    /// Maximum lengths for transaction sender_data and recipient_data fields used during deserialization.
+    pub const MAX_TX_SENDER_DATA_SIZE: usize = 2;
```
sisou: Where are 2 bytes used? I think only txs from the staking contract use this and they use 1 byte only. Or does this include the 1 byte "prefix" for the data length?

primitives/src/policy.rs
```diff
@@ -58,6 +58,12 @@ impl Policy {
...
+    pub const MAX_TX_RECIPIENT_DATA_SIZE: usize = 569;
```
sisou: Can you add a comment where this arbitrary looking value comes from?

primitives/transaction/src/account/basic_account.rs
```diff
@@ -22,6 +22,14 @@ impl AccountTransactionVerification for BasicAccountVerifier {
             return Err(TransactionError::ZeroValue);
         }
 
+        if transaction.recipient_data.len() > Policy::MAX_BASIC_TX_RECIPIENT_DATA_SIZE {
+            error!(
+                "The following transaction recipient's data exceeds the maximum size:\n{:?}",
+                transaction
+            );
```
sisou: Since you are logging the whole transaction, and not just the data, I'd reword to:
```suggestion
            error!(
                "The following transaction's recipient data exceeds the maximum size:\n{:?}",
                transaction
            );
```

sisou: On further thought: should this even be logged in the node that processes the transaction? The transaction will be correctly rejected, so it's not an error for the node (operator), i.e. it's not unexpected behavior.

primitives/transaction/src/account/basic_account.rs
```diff
@@ -40,6 +48,14 @@ impl AccountTransactionVerification for BasicAccountVerifier {
     fn verify_outgoing_transaction(transaction: &Transaction) -> Result<(), TransactionError> {
         assert_eq!(transaction.sender_type, AccountType::Basic);
 
+        if !transaction.sender_data.is_empty() {
+            error!(
+                "The following transaction can't have sender data:\n{:?}",
+                transaction
+            );
```
sisou: Same comment here about the need for an error log. It's not faulty behavior of the node.

primitives/transaction/src/account/htlc_contract.rs
```diff
@@ -64,6 +64,14 @@ impl AccountTransactionVerification for HashedTimeLockedContractVerifier {
     fn verify_outgoing_transaction(transaction: &Transaction) -> Result<(), TransactionError> {
         assert_eq!(transaction.sender_type, AccountType::HTLC);
 
+        if !transaction.sender_data.is_empty() {
+            error!(
+                "The following transaction can't have sender data:\n{:?}",
+                transaction
+            );
```
sisou: And here.

primitives/transaction/src/account/vesting_contract.rs
```diff
@@ -56,6 +56,14 @@ impl AccountTransactionVerification for VestingContractVerifier {
     fn verify_outgoing_transaction(transaction: &Transaction) -> Result<(), TransactionError> {
         assert_eq!(transaction.sender_type, AccountType::Vesting);
 
+        if !transaction.sender_data.is_empty() {
+            error!(
+                "The following transaction can't have sender data:\n{:?}",
+                transaction
+            );
```
sisou: And here.

primitives/transaction/src/lib.rs
```diff
@@ -912,18 +905,37 @@ mod serde_derive {
...
             let sender_data: Vec<u8> = seq
                 .next_element()?
                 .ok_or_else(|| Error::invalid_length(2, &self))?;
+            let sender_data_len = sender_data.len();
+            if sender_data_len > Policy::MAX_TX_SENDER_DATA_SIZE {
+                return Err(Error::custom(format!(
+                    "sender and recipient data length exceeds the maximum allowed length of {}",
+                    Policy::MAX_TX_SENDER_DATA_SIZE
+                )));
+            }
```
sisou:
```suggestion
            if sender_data.len() > Policy::MAX_TX_SENDER_DATA_SIZE {
                return Err(Error::custom(format!(
                    "sender data length exceeds the maximum allowed length of {}",
                    Policy::MAX_TX_SENDER_DATA_SIZE
                )));
            }
```

primitives/transaction/src/lib.rs
```diff
@@ -912,18 +905,37 @@ mod serde_derive {
...
+            // Enforce maximum recipient data length.
             let recipient_data: Vec<u8> = seq
                 .next_element()?
                 .ok_or_else(|| Error::invalid_length(5, &self))?;
+            if recipient_data.len() > Policy::MAX_TX_RECIPIENT_DATA_SIZE {
+                return Err(Error::custom(format!(
+                    "sender and recipient data length exceeds the maximum allowed length of {}",
+                    Policy::MAX_TX_RECIPIENT_DATA_SIZE
+                )));
+            }
```
sisou:
```suggestion
            if recipient_data.len() > Policy::MAX_TX_RECIPIENT_DATA_SIZE {
                return Err(Error::custom(format!(
                    "recipient data length exceeds the maximum allowed length of {}",
                    Policy::MAX_TX_RECIPIENT_DATA_SIZE
                )));
            }
```

sisou:
Does this need the `testnet-restart` label? Looks to me like a breaking change, as updated nodes won't accept previously valid blocks anymore?

hrxi:
> Does this need the `testnet-restart` label? Looks to me like a breaking change, as updated nodes won't accept previously valid blocks anymore?

No, since people (likely?) didn't publish such blocks.

primitives/src/policy.rs
ii-cruz: Nah, it is just one byte you are correct.

primitives/transaction/src/account/basic_account.rs
ii-cruz: i turned them all into warnings

ii-cruz:
Soeren actually exploited it on the testnet, so it is a breaking change

primitives/transaction/src/signature_proof.rs
```diff
@@ -16,6 +17,16 @@ pub struct SignatureProof {
...
 }
 
+impl SerializedMaxSize for SignatureProof {
+    #[allow(clippy::identity_op)]
+    const MAX_SIZE: usize = 0
+        + PublicKey::MAX_SIZE
+        + Policy::MAX_MERKLE_PATH_SIZE
+        + Signature::MAX_SIZE
+        + /* option */ 1
+        + WebauthnExtraFields::MAX_SIZE;
```
hrxi:
```suggestion
        + nimiq_serde::option_max_size(WebauthnExtraFields::MAX_SIZE);
```

primitives/transaction/tests/serialization.rs
```diff
@@ -105,3 +108,45 @@ fn it_can_serialize_and_deserialize_signature_proofs() {
...
+
+    assert!(serialized.len() < SignatureProof::MAX_SIZE);
+    assert!(proof.webauthn_fields.serialized_size() > WebauthnExtraFields::MAX_SIZE);
+    error!("{:?}", proof.webauthn_fields.serialized_size());
+
+    let deserialized: Result<SignatureProof, DeserializeError> =
+        Deserialize::deserialize_from_vec(&serialized[..]);
+
+    assert!(deserialized.is_err());
+}
```
hrxi, line 133: Could you put this test into primitives/transaction/tests/serialization.rs instead? Then `WebauthnExtraFields` (and its members) wouldn't need to be `pub`.

primitives/src/policy.rs
```diff
@@ -674,7 +690,7 @@ impl Default for Policy {
             batches_per_epoch: 720,
             tendermint_timeout_init: 1000,
             tendermint_timeout_delta: 1000,
-            state_chunks_max_size: 1000,
+            state_chunks_max_size: 200, // #Nodes/accounts 200, TODO: Simulate with different sizes
```
hrxi: Was this change intentional?

primitives/src/policy.rs
```diff
@@ -58,6 +58,22 @@ impl Policy {
     /// The maximum allowed size, in bytes, for a micro block body.
     pub const MAX_SIZE_MICRO_BODY: usize = 100_000;
 
+    // Maximum lengths for transaction sender_data and recipient_data fields used during deserialization.
+    /// Maximum size for the sender data field of the transactions.
+    /// The only transactions using this field are RemoveStake and Delete Validator.
+    pub const MAX_TX_SENDER_DATA_SIZE: usize = 1;
```
hrxi, line 64: Could you add a test that checks that `MAX_TX_SENDER_DATA_SIZE` is larger or equal to `OutgoingStakingTransactionData::MAX_SIZE`?

primitives/src/policy.rs
```diff
@@ -58,6 +58,22 @@ impl Policy {
...
+    /// This is used by the HTLC, Vesting and staking contract transactions. Basic transactions are allowed to use it as well.
+    /// The biggest transactions are the create validator and update validator with a web auth signature.
+    pub const MAX_TX_RECIPIENT_DATA_SIZE: usize = 1024;
+    /// Maximum length for basic transaction recipient_data used during intrinsic transaction verification.
+    pub const MAX_BASIC_TX_RECIPIENT_DATA_SIZE: usize = 64;
+    /// Maximum size for transaction's merkle path proofs.
+    /// 32 bytes for Merkle Path Node * 32 max length + 4 bytes to encode the bools for left/right + 1 byte for the serialize vec count.
+    pub const MAX_MERKLE_PATH_SIZE: usize = 1029;
+    /// Maximum size for the total web auth fields.
+    pub const MAX_SUPPORTED_WEB_AUTH_SIZE: usize = 500;
```
hrxi:
```suggestion
    pub const MAX_SUPPORTED_WEB_AUTH_SIZE: usize = 512;
```

Completely optional, but I like powers of 2.

hrxi (commented):
Looks good to me, I left some comments.

primitives/src/policy.rs
ii-cruz: not at all, thank you!

primitives/src/policy.rs
ii-cruz: Added tests to both. Thank you, caught another miscalculation, the max recipient data was too small because the Merkle proof size was not being well accounted for.

primitives/transaction/tests/serialization.rs
ii-cruz: Talked with @hrxi, and it will stay where it is

PR #2667, opened 2024-06-21
Log reason of peer banning

### LINKED ISSUE #2661: Logging the ban reason

Eligioo:
All paths to a `CloseReason::MaliciousPeer` as reason for closing the connection now contain a WARN log message to annotate the reason for banning a peer.

I chose for WARN as log level for now but can easily changed if one believes there is a reason for it.

## Pull request checklist

### DIFF
--- a/validator/src/proposal_buffer.rs
+++ b/validator/src/proposal_buffer.rs
@@ -159,8 +159,14 @@ where
             .validate_message::<ProposalTopic<TValidatorNetwork>>(pubsub_id, MsgAcceptance::Reject);
 
         let network = Arc::clone(&self.network);
+
         // disconnect the peer
         let future = async move {
+            log::warn!(
+                peer_id = %source,
+                "Banning peer because an invalid proposal was received"
+            );
+
             network
                 .disconnect_peer(source, CloseReason::MaliciousPeer)
                 .await;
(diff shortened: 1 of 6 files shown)


validator/src/proposal_buffer.rs
```diff
@@ -159,6 +159,11 @@ where
             .validate_message::<ProposalTopic<TValidatorNetwork>>(pubsub_id, MsgAcceptance::Reject);
 
         let network = Arc::clone(&self.network);
+
+        log::warn!(
+            peer_id = %source,
+            "Banning peer because an invalid proposal was received"
+        );
```
jsdanielh: I'd move this into the future such that we have better correlation of events

validator/src/proposal_buffer.rs
```diff
@@ -455,6 +467,12 @@ where
 
             // Punish the propagation source of the proposal for propagating invalid proposal messages, i.e ban the peer.
             let nw = Arc::clone(&self.network);
+
+            log::warn!(
+                peer_id = %source,
+                "Disconnecting peer because proposal signature verfication failed"
```
jsdanielh:
```suggestion
                "Disconnecting peer because proposal signature verification failed."
```

jsdanielh (commented):
I'd have changed the `CloseReason::MaliciousPeer` variant to be `CloseReason::MaliciousPeer(String)` such that we enforce always the reason and then we log it in the network

Eligioo:
@jsdanielh 
> I'd have changed the `CloseReason::MaliciousPeer` variant to be `CloseReason::MaliciousPeer(String)` such that we enforce always the reason and then we log it in the network

I thought about this but this will cause the log messages to double:
- the new log in the network layer as you suggested
- the existing tracing logs creating context at the caller location that initiates the closing connection e.g.:
```
log::warn!(
    given_checkpoint_epoch,
    expected_checkpoint_epoch,
    %peer_id,
    "Banning peer because requesting macro chain failed: invalid checkpoint",
);
```

Basically to remove the existing tracing logs but don't lose context, you would need to format the extra values into the `String` of `CloseReason::MaliciousPeer(String)`. I'm not a fan of that.

hrxi (approved):
LGTM.

@Eligioo remarked that the peer banning in the testnet should already have a log for the ban unless it came from validator/src/proposal_buffer.rs.

hrxi:
Rebased for merge.

PR #2668, opened 2024-06-21
validator-network: Improve logging

jsdanielh:
Improve `validator-network` logging by moving most of the recurring errors to `trace` and only log to `debug` when there is a change in the state of the cache.

## Pull request checklist

### DIFF
--- a/validator-network/src/network_impl.rs
+++ b/validator-network/src/network_impl.rs
@@ -128,11 +128,11 @@ where
                 Ok(peer_id)
             }
             Ok(None) => {
-                log::trace!(validator_id, %public_key, "Unable to resolve validator peer ID: Entry not found in DHT");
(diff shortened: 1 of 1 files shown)


validator-network/src/network_impl.rs
```diff
@@ -128,11 +128,11 @@ where
                 Ok(peer_id)
             }
             Ok(None) => {
-                log::trace!(validator_id, %public_key, "Unable to resolve validator peer ID: Entry not found in DHT");
+                log::debug!(validator_id, %public_key, "Unable to resolve validator peer ID: Entry not found in DHT");
```
hrxi, line 131: I think it'd be better to still log these messages the first time they appear, but stop logging them afterwards.

hrxi (approved):
After talking to @jsdanielh, this looks good to me.

PR #2674, opened 2024-06-24
Unban banned peers after a duration

### LINKED ISSUE #2660: We ban important peers like other elected validators
Temporary ban peers instead. Ban expiration time of 1 min
Specifically for elected validators, probably validators too

Eligioo:
Until now we don't unban a peer in any way after we ban it. This PR now will unban a banned peer after 10 minutes allowing the banned peer to dial us again.

## Pull request checklist

### DIFF
--- a/network-libp2p/src/connection_pool/behaviour.rs
+++ b/network-libp2p/src/connection_pool/behaviour.rs
@@ -1,12 +1,13 @@
 use std::{
     collections::{BTreeMap, BTreeSet, HashMap, HashSet, VecDeque},
     net::IpAddr,
+    pin::Pin,
     sync::Arc,
     task::{Context, Poll, Waker},
     time::Duration,
 };
 
-use futures::StreamExt;
+use futures::{future::BoxFuture, Future, FutureExt, StreamExt};
 use instant::Instant;
 use ip_network::IpNetwork;
 use libp2p::{
@@ -20,7 +21,7 @@ use libp2p::{
     Multiaddr, PeerId, TransportError,
 };
 use nimiq_network_interface::{network::CloseReason, peer_info::Services};
-use nimiq_time::{interval, Interval};
+use nimiq_time::{interval, sleep_until, Interval};
 use nimiq_utils::WakerExt as _;
 use parking_lot::RwLock;
 use rand::{seq::IteratorRandom, thread_rng};
@@ -98,6 +99,12 @@ struct ConnectionState<T> {
     connected: BTreeMap<T, Option<Services>>,
     /// Set of connection IDs marked as banned.
     banned: BTreeSet<T>,
+    /// List of subsequent banned peers with their unban deadlines in ascending order.
+    unban_deadlines: VecDeque<(T, Instant)>,
+    /// Deadline for first peer that can be unbanned.
+    unban_timeout: Option<(T, BoxFuture<'static, ()>)>,
+    /// The time that needs to pass to unban a banned peer.
+    ban_time: Duration,
     /// Set of connection IDs mark as failed.
     failed: BTreeMap<T, usize>,
     /// Set of connection IDs mark as down.
@@ -113,25 +120,65 @@ struct ConnectionState<T> {
     desired_connections: usize,
     /// The set of services that this peer requires.
     required_services: Services,
+    /// Waker used for the next poll
+    waker: Option<Waker>,
 }
 
-impl<T: Ord> ConnectionState<T> {
+impl<T: Clone + Ord + Unpin> Future for ConnectionState<T> {
+    type Output = ();
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        while let Some((id, sleep)) = self.unban_timeout.as_mut() {
+            if sleep.poll_unpin(cx).is_ready() {
+                let id = id.clone();
+                self.banned.remove(&id);
+                self.unban_timeout = None;
+
+                // While we are unbanning, try to unban more peers if their deadline also have been reached.
+                // Otherwise schedule a sleep until for the first peer that can be unbanned.
+                while let Some((next_id, deadline)) = self.unban_deadlines.pop_front() {
+                    if Instant::now() >= deadline {
+                        self.banned.remove(&next_id);
+                    } else {
+                        self.unban_timeout = Some((next_id, Box::pin(sleep_until(deadline))));
+                        break;
+                    }
+                }
+            } else {
+                break;
+            }
+        }
+
+        if self.unban_timeout.is_none() {
+            self.waker.store_waker(cx);
+        }
+
+        Poll::Pending
+    }
+}
+
+impl<T: Clone + Ord> ConnectionState<T> {
     fn new(
         max_failures: usize,
         retry_down_after: Duration,
         desired_connections: usize,
         required_services: Services,
+        ban_time: Duration,
     ) -> Self {
         Self {
             dialing: BTreeSet::new(),
             connected: BTreeMap::new(),
             banned: BTreeSet::new(),
+            ban_time,
+            unban_deadlines: VecDeque::new(),
+            unban_timeout: None,
             failed: BTreeMap::new(),
             down: BTreeMap::new(),
             max_failures,
             retry_down_after,
             desired_connections,
             required_services,
+            waker: None,
         }
     }
 
@@ -161,12 +208,17 @@ impl<T: Ord> ConnectionState<T> {
     fn mark_banned(&mut self, id: T) {
         self.failed.remove(&id);
         self.down.remove(&id);
-        self.banned.insert(id);
-    }
+        self.banned.insert(id.clone());
 
-    /// Removes a connection ID from the banned set
-    fn unmark_banned(&mut self, id: T) {
-        self.banned.remove(&id);
+        let unban_deadline = Instant::now() + self.ban_time;
+        // If no peer is scheduled to be unbanned, put is as the first-to-be unbanned peer.
+        // Otherwise queue the peer such that it will be scheduled later when the peer is the first in line.
+        if self.unban_timeout.is_none() {
+            self.unban_timeout = Some((id, Box::pin(sleep_until(unban_deadline))));
+            self.waker.wake();
+        } else {
+            self.unban_deadlines.push_back((id.clone(), unban_deadline));
+        }
     }
 
     /// Returns whether a connection ID is banned
@@ -352,12 +404,14 @@ impl Behaviour {
                 config.retry_down_after,
                 desired_peer_count,
                 required_services,
+                Duration::from_secs(60 * 10), // 10 minutes
             ),
             addresses: ConnectionState::new(
                 4,
                 config.retry_down_after,
                 desired_peer_count,
                 required_services,
+                Duration::from_secs(60 * 10), // 10 minutes
             ),
             actions: VecDeque::new(),
             active: false,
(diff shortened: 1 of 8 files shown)


sisou:
The linked issue suggested a ban duration of only 1 minute. What made you decide to set it to 10 minutes?

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -115,6 +118,32 @@ struct ConnectionState<T> {
...
+            }
+
+            false
+        }) {
+            return Poll::Ready(Some(ConnectionStateAction::UnbanPeer(*peer_id)));
+        }
+
+        Poll::Pending
+    }
+}
```
hrxi, line 158: (Already talked to @Eligioo about this:) The `poll_next` function doesn't register with a waker, so in isolation, it wouldn't be called again.

hrxi (changes requested):

Eligioo:
> The linked issue suggested a ban duration of only 1 minute. What made you decide to set it to 10 minutes?

@sisou it was discussed later on

time/src/gloo.rs
```diff
@@ -41,3 +42,7 @@ pub fn sleep(duration: Duration) -> impl Future<Output = ()> {
         .expect("Duration as millis must fit in u32");
     SendWrapper::new(TimeoutFuture::new(millis))
 }
+
+pub fn sleep_until(deadline: Instant) -> impl Future<Output = ()> {
+    sleep(deadline - Instant::now())
```
hrxi:
```suggestion
    sleep(deadline.saturating_duration_since(Instant::now()))
```
Otherwise, this could panic if `deadline` is in the past.

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -98,6 +99,12 @@ struct ConnectionState<T> {
     connected: BTreeMap<T, Option<Services>>,
     /// Set of connection IDs marked as banned.
     banned: BTreeSet<T>,
+    /// The first peer that can be unbanned when its ban expiration is reached.
+    next_unban_peer: Option<(T, Instant)>,
+    /// List of subsequent peer unban deadlines in ascending order.
+    subsequent_unban_deadlines: VecDeque<(T, Instant)>,
```
hrxi:
```suggestion
    /// List of peer unban deadlines in ascending order.
    unban_deadlines: VecDeque<(T, Instant)>,
```
The extra `Option` seems superfluous to me, it's acting as the first element of the `subsequent_unban_deadlines`.

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -115,6 +122,29 @@ struct ConnectionState<T> {
...
+    UnbanPeer(PeerId),
+}
+
+impl Stream for ConnectionState<PeerId> {
+    type Item = ConnectionStateAction;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        // Handle the first banned peer that can be unbanned
+        if let Some((peer_id, due_to)) = self.next_unban_peer {
+            let mut sleep_until_pin = pin!(sleep_until(due_to));
```
hrxi: Since we drop the `sleep_until_pin` immediately, it will deregister itself and the runtime will not call our `poll_next` function again. We need to store it inside of our struct, probably in a `Option<Pin<Box<Sleep>>>`.

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -115,6 +122,29 @@ struct ConnectionState<T> {
...
+
+            // Wait until the ban period has expired
+            if sleep_until_pin.poll_unpin(cx).is_ready() {
+                self.next_unban_peer = self.subsequent_unban_deadlines.pop_front();
+                return Poll::Ready(Some(ConnectionStateAction::UnbanPeer(peer_id)));
```
hrxi: I think it'd be better if we'd call `unmark_banned` directly **instead** of implementing a `Stream` here and instead of calling `unban_connection` on the parent struct, because
a) it's keeping the struct itself consistent (things in `banned` have corresponding deadlines.
b) we'll consistently unban the correct IP address after banning it, even if it might have changed in the meantime.

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -161,7 +198,16 @@ impl<T: Ord> ConnectionState<T> {
...
         self.failed.remove(&id);
         self.down.remove(&id);
-        self.banned.insert(id);
+        self.banned.insert(id.clone());
+
+        let unban_deadline = Instant::now() + self.ban_time;
+        // If no peer is scheduled to be unbanned, put is as the first-to-be unbanned peer.
+        // Otherwise queue the peer such that it will be scheduled later when the peer is the first in line.
+        if self.unban_timeout.is_none() {
+            self.unban_timeout = Some((id, Box::pin(sleep_until(unban_deadline))));
```
hrxi, line 217: You'll need to wake yourself here, so that the `sleep_until` future gets polled at some point.

```suggestion
            self.unban_timeout = Some((id, Box::pin(sleep_until(unban_deadline))));
            if let Some(waker) = self.waker.take() {
                waker.wake();
            }
```

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -115,7 +122,34 @@ struct ConnectionState<T> {
...
-impl<T: Ord> ConnectionState<T> {
+impl<T: Clone + Ord + Unpin> Future for ConnectionState<T> {
+    type Output = ();
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let mut unban_id = None;
+
+        if let Some((id, sleep)) = self.unban_timeout.as_mut() {
+            if sleep.poll_unpin(cx).is_ready() {
+                unban_id = Some(id.clone());
```
hrxi:
```suggestion
                let id = id.clone();
                self.unmark_banned(id);
```

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -115,7 +122,34 @@ struct ConnectionState<T> {
...
 }
 
-impl<T: Ord> ConnectionState<T> {
+impl<T: Clone + Ord + Unpin> Future for ConnectionState<T> {
+    type Output = ();
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let mut unban_id = None;
+
+        if let Some((id, sleep)) = self.unban_timeout.as_mut() {
```
hrxi:
```suggestion
        while let Some((id, sleep)) = self.unban_timeout.as_mut() {
```

Better make this a `while` loop so that every created sleep in there is actually polled until it returns `Poll::Pending`. Needs to have a `break;` in the `else` of `if sleep.poll_unpin(cx).is_ready()` though.

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -115,7 +122,34 @@ struct ConnectionState<T> {
...
+            if sleep.poll_unpin(cx).is_ready() {
+                unban_id = Some(id.clone());
+
+                if let Some((id, deadline)) = self.unban_deadlines.pop_front() {
+                    self.unban_timeout = Some((id, Box::pin(sleep_until(deadline))));
+                    cx.waker().wake_by_ref();
```
hrxi: This waking becomes unnecessary if you do the suggested `while` change above.

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -115,7 +122,34 @@ struct ConnectionState<T> {
...
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let mut unban_id = None;
+
+        if let Some((id, sleep)) = self.unban_timeout.as_mut() {
+            if sleep.poll_unpin(cx).is_ready() {
+                unban_id = Some(id.clone());
+
+                if let Some((id, deadline)) = self.unban_deadlines.pop_front() {
+                    self.unban_timeout = Some((id, Box::pin(sleep_until(deadline))));
```
hrxi: Optimally, this would first check whether the deadline has already expired. In this case, we could avoid creating a `sleep_until`.

hrxi (commented):
Looks quite good already.

Once the code is finished, it'd also be nice to have a test that verifies that peers are indeed unbanned after some time. Can be a local unit test of `ConnectionState`. Ping me if you have questions about this.

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -113,25 +120,63 @@ struct ConnectionState<T> {
...
+        while let Some((id, sleep)) = self.unban_timeout.as_mut() {
+            if sleep.poll_unpin(cx).is_ready() {
+                let id = id.clone();
+                self.unmark_banned(id);
+                self.unban_timeout = None;
+
+                // While we are unbanning, try to unban more peers if their deadline also have been reached.
+                // Otherwise schedule a sleep until for the first peer that can be unbanned.
+                while let Some((next_id, deadline)) = self.unban_deadlines.pop_front() {
+                    if deadline.elapsed() >= self.ban_time {
```
hrxi:
```suggestion
                    if Instant::now() >= deadline {
```

`deadline` is the instant where the ban is supposed to expire. The current code checks that the expiry is at least `self.ban_time` away, i.e. that the ban has essentially expired twice.

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -113,25 +120,63 @@ struct ConnectionState<T> {
...
+    /// Waker used for the next poll
+    waker: Option<Waker>,
+}
+
+impl<T: Clone + Ord + Unpin> Future for ConnectionState<T> {
+    type Output = ();
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        self.waker.store_waker(cx);
+
```
hrxi:
```suggestion
```

See below.

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -113,25 +120,63 @@ struct ConnectionState<T> {
...
+                    } else {
+                        self.unban_timeout = Some((next_id, Box::pin(sleep_until(deadline))));
+                        break;
+                    }
+                }
+            } else {
+                break;
+            }
+        }
+
```
hrxi, line 151:
```suggestion

        if self.unban_timeout.is_none() {
            self.waker.store_waker(cx);
        }
```
Only when no `Sleep` has registered to a waker we need to be able to wake up ourselves.

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -161,7 +206,17 @@ impl<T: Ord> ConnectionState<T> {
...
+        // Otherwise queue the peer such that it will be scheduled later when the peer is the first in line.
+        if self.unban_timeout.is_none() {
+            self.unban_timeout = Some((id, Box::pin(sleep_until(unban_deadline))));
+            self.waker.wake();
+        } else {
+            self.unban_deadlines.push_back((id.clone(), unban_deadline));
+        }
     }
 
     /// Removes a connection ID from the banned set
```
hrxi: The `unmark_banned` function below doesn't clean up the `self.unban_deadlines`. I'd suggest to just remove the function since it's currently unused.

hrxi (commented):
Found a small bug, two nits and a buggy unused function. Perhaps we can just delete the unused function?

network-libp2p/src/connection_pool/behaviour.rs
Eligioo: Removed `unmark_banned` and `unban_connection` including its unused caller and `NetworkAction`.

hrxi (approved):
Looks good to me, thanks for your patience with my reviews!

PR #2676, opened 2024-06-24
Always make sure to read the `Transaction::proof` field completely

### LINKED ISSUE #2672: Transaction proof data is not checked for unused bytes at the end

hrxi:
Fixes #2672.

### DIFF
--- a/blockchain/tests/block_production.rs
+++ b/blockchain/tests/block_production.rs
@@ -1377,7 +1377,7 @@ fn it_can_revert_basic_and_create_contracts_txns() {
 }
 
 fn ed25519_key_pair(secret_key: &str) -> SchnorrKeyPair {
-    let priv_key: SchnorrPrivateKey =
-        Deserialize::deserialize_from_vec(&hex::decode(secret_key).unwrap()[..]).unwrap();
+    let priv_key =
+        SchnorrPrivateKey::deserialize_from_vec(&hex::decode(secret_key).unwrap()).unwrap();
     priv_key.into()
 }
--- a/bls/tests/tests.rs
+++ b/bls/tests/tests.rs
@@ -109,8 +109,7 @@ fn serialize_deserialize() {
 fn uncompress_compress() {
     let hex_public_key = "ae4cc2e31e04add9a6d379b4379b02f302971503cbac8d02fdc5d2dc8204d24ec8d095627d037de747f1a8ea7bf3c1693262d947f78e0cc73c18ecc2f2ec5b2249d551e1680fe0c973a7951bd78d4fbe0326be71286ed34004d2443eb3b00167a02edffcfd2b8539448fa116c5454da2d181dc03ea8cfe3fedb58b9b945d5e506c794deb3ba73983005b3ff799212bf59030a8dd17ff48fd5d015695195a022fed8ba4fab28a4c3e2d6f41be0e6315e41824df161219c02be5a281c215011c13131184187e9100d2d6a5321fd9b154806ecc78e93b91331a5334b8876fd1b8ea62b17ce6045fc9e1af60b7705b0cf86dba79f5bcb8320c99a45f3b7c7178f8f87ba953de2755c61af882059c1de1d7a35357f06cd4a7d954e4bb211900";
     let raw_public_key: Vec<u8> = hex::decode(hex_public_key).unwrap();
-    let compressed_public_key: CompressedPublicKey =
-        Deserialize::deserialize_from_vec(&raw_public_key).unwrap();
+    let compressed_public_key = CompressedPublicKey::deserialize_from_vec(&raw_public_key).unwrap();
 
     println!(
         "{:?}",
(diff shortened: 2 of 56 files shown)


blockchain/tests/block_production.rs
```diff
@@ -1377,7 +1377,7 @@ fn it_can_revert_basic_and_create_contracts_txns() {
 }
 
 fn ed25519_key_pair(secret_key: &str) -> SchnorrKeyPair {
-    let priv_key: SchnorrPrivateKey =
-        Deserialize::deserialize_from_vec(&hex::decode(secret_key).unwrap()[..]).unwrap();
+    let priv_key =
+        SchnorrPrivateKey::deserialize_from_vec(&hex::decode(secret_key).unwrap()).unwrap();
```
sisou, line 1381: I see that you now changed all deserializations to use the struct, which I like! But is there anything safeguarding against writing it the old way in the future? Can we add a custom clippy rule for this? Otherwise this won't necessarily stick.

hrxi: I haven't replaced all of the `Deserialize::deserialize` calls, only those where it wasn't difficult to specify.

I don't know about custom clippy lints, but it seems like a good idea in general. ^^

sisou: For future reference: https://github.com/trailofbits/dylint

serde/src/lib.rs
```diff
@@ -10,36 +10,64 @@ pub use serde_derive::{Deserialize, Serialize};
 
 /// Deserialization Error. This error is a wrapper over `postcard::Error`.
 #[derive(Eq, PartialEq)]
-pub struct DeserializeError(postcard::Error);
+pub struct DeserializeError(DeserializeErrorInner);
+
+#[derive(Eq, PartialEq)]
+enum DeserializeErrorInner {
```
jsdanielh, line 19: Isn't it better to derive the error using `thiserror`?

serde/src/lib.rs
```diff
@@ -10,36 +10,64 @@ pub use serde_derive::{Deserialize, Serialize};
 
 /// Deserialization Error. This error is a wrapper over `postcard::Error`.
```
jsdanielh: The error is strictly speaking no longer a wrapper over `postcard::Error`

primitives/transaction/src/lib.rs
```diff
@@ -713,9 +713,8 @@ mod serde_derive {
                         VARIANTS[0],
                         BASIC_FIELDS.len(),
                     )?;
-                    let signature_proof: SignatureProof =
-                        Deserialize::deserialize_from_vec(&self.proof)
-                            .map_err(|_| S::Error::custom("Could not serialize signature proof"))?;
+                    let signature_proof = SignatureProof::deserialize_all(&self.proof)
+                        .map_err(|_| S::Error::custom("Could not serialize signature proof"))?;
```
jsdanielh: Maybe deserialize?

bls/tests/tests.rs
```diff
@@ -109,8 +109,7 @@ fn serialize_deserialize() {
 fn uncompress_compress() {
     let hex_public_key = "ae4cc2e31e04add9a6d379b4379b02f302971503cbac8d02fdc5d2dc8204d24ec8d095627d037de747f1a8ea7bf3c1693262d947f78e0cc73c18ecc2f2ec5b2249d551e1680fe0c973a7951bd78d4fbe0326be71286ed34004d2443eb3b00167a02edffcfd2b8539448fa116c5454da2d181dc03ea8cfe3fedb58b9b945d5e506c794deb3ba73983005b3ff799212bf59030a8dd17ff48fd5d015695195a022fed8ba4fab28a4c3e2d6f41be0e6315e41824df161219c02be5a281c215011c13131184187e9100d2d6a5321fd9b154806ecc78e93b91331a5334b8876fd1b8ea62b17ce6045fc9e1af60b7705b0cf86dba79f5bcb8320c99a45f3b7c7178f8f87ba953de2755c61af882059c1de1d7a35357f06cd4a7d954e4bb211900";
     let raw_public_key: Vec<u8> = hex::decode(hex_public_key).unwrap();
-    let compressed_public_key: CompressedPublicKey =
-        Deserialize::deserialize_from_vec(&raw_public_key).unwrap();
+    let compressed_public_key = CompressedPublicKey::deserialize_from_vec(&raw_public_key).unwrap();
```
jsdanielh, line 112: Maybe this should be a `deserialize_all`?

hrxi: I want to address this in a follow-up PR since it'd be a larger change.

primitives/transaction/src/lib.rs
hrxi: Done.

serde/src/lib.rs
hrxi: Fixed.

serde/src/lib.rs
hrxi: I think I prefer the current way because it minimizes the length of the `Debug` implementation, since it directly reduces to the error it's wrapping.

jsdanielh (approved):

PR #2677, opened 2024-06-24
Fix eternally paused state sync and chunk ordering issues

paberr:
In combination with validity sync issues, we saw state sync getting stuck.
This could happen if we synced all chunks (`end_key = None`) and subsequently paused state sync but then failed to apply the last chunk.

This PR addresses the issue in two ways:
- Store the block number at which the last chunk should be applied. If we pass it and are not complete, unpause the state sync.
- If a chunk was still buffered from before a reset, a new duplicate chunk with the same start key would be discarded – resulting in an unordered list of chunks that eventually will fail to apply. We thus replace the duplicate check by a check that the buffer is ordered. It it isn't, we discard old entries in favour of new ones.

## Pull request checklist

### DIFF
--- a/consensus/src/sync/live/state_queue/mod.rs
+++ b/consensus/src/sync/live/state_queue/mod.rs
@@ -102,14 +102,20 @@ pub enum QueuedStateChunks<N: Network> {
 }
 
 /// This represents the behavior for the next chunk request. When the accounts trie is:
-/// Complete: there are no further requests to make. (Can only set after a macro block).
-/// Reset: the next request should start based on the blockchain accounts trie state.
-/// Continue: the next request should start on the specified key, which is the end of the previous request.
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum ChunkRequestState {
+    /// The accounts trie is complete and all chunks were applied.
+    /// There are no further requests to make. (Can only set after a macro block).
     Complete,
-    Paused,
+    /// We successfully requested all chunks but they are not applied yet.
+    /// Also records the block number for which we received the last chunk.
+    /// If we pass this block and are not complete, we reset.
+    Paused(u32),
+    /// Restart the state sync.
+    /// The next request should start based on the blockchain accounts trie state.
     Reset,
+    /// Continue the state sync from the given key.
+    /// The next request should start on the specified key, which is the end of the previous request.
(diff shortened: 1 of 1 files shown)


viquezclaudio (approved):

viquezclaudio:
I tested this fixes the validity sync stuck issue

consensus/src/sync/live/state_queue/mod.rs
```diff
@@ -102,14 +102,20 @@ pub enum QueuedStateChunks<N: Network> {
...
-    Paused,
+    /// We successfully requested all chunks but they are not applied yet.
+    /// Also records the block number for which we received the last chunk.
+    /// If we pass this block and are not complete, we reset.
+    Paused(u32),
+    /// Restart the state sync.
+    /// The next request should start based on the blockchain accounts trie state.
     Reset,
+    /// Continue the state sync from the given key.
+    ///Tthe next request should start on the specified key, which is the end of the previous request.
```
jsdanielh:
```suggestion
    /// The next request should start on the specified key, which is the end of the previous request.
```

consensus/src/sync/live/state_queue/mod.rs
```diff
@@ -275,20 +281,23 @@ impl<N: Network> StateQueue<N> {
...
-                response.chunk,
-                response.block_number,
-                response.block_hash
-            );
-            return;
+        // We enforce the invariant that chunks are ordered here.
+        // If the new chunk is for an earlier key, we clear the vector
+        // and discard old chunks.
+        if let Some(chunk) = chunks.last() {
+            // TODO: Potentially we can enforce that only one chain of chunks exist by also
```
jsdanielh:
```suggestion
            // TODO: Potentially we can enforce that only one chain of chunks exists by also
```

consensus/src/sync/live/state_queue/mod.rs
```diff
@@ -275,20 +281,23 @@ impl<N: Network> StateQueue<N> {
...
+            // checking that the previous end key matches the new start key.
+            // But that's only an optimisation.
```
jsdanielh:
```suggestion
            // But that's only an optimization.
```

jsdanielh (approved):

PR #2679, opened 2024-06-24
Dont move incompatible peers to live sync too early.

### LINKED ISSUE #2675: Missing blocks are tried to be pushed when the state root can not be verfied yet.
In several instances of testnet runs,  when full nodes are syncing the validity window, the following kind of log messages can be  observed:

`
Pushing block #121:MI:f023dca1a4 from missing blocks response
Rejecting block block=121:MI:f023dca1a4 block_root=d14ec6d3cac4714955df120c178242a95cdbfa54d0211d959f4c871d3d0feff4 history_root=13349e54e46d47f9497705d158c328d5b1ec37dd447971ec10aa68b49672e948 reason="History root doesn't match real history root"

 Rejecting block block=#121:MI:f023dca1a4 reason="bad state" error=Invalid block: Invalid history root error=[Invalid history root]

`

This is because for this particular case,, we started the validity sync to sync up to macro block #120:
`Starting validity window synchronization process target_macro=120`

But when we see a block higher than macro block #120 being announced and try to apply those, it fails because the history root cannot be verified yet.

Once the validity sync is complete, and the history root can be verified, everything works as expected.

However, this is misleading and can create potential errors during the state sync process since a failed push block error is returned from the blockchain.

A better strategy is to not push blocks until we can verify the history root.

viquezclaudio:
If peers are classified as incompatible and moved to live sync before the validity sync is complete, this creates issues where missing blocks can be pushed from the state sync, even though, we cannot verify the history root yet.
With this change, we first complete the validity sync and after that, we move the incompatible peers to live sync.

Fixes #2675

...
## Pull request checklist

### DIFF
--- a/blockchain/src/blockchain/abstract_blockchain.rs
+++ b/blockchain/src/blockchain/abstract_blockchain.rs
@@ -5,7 +5,10 @@ use nimiq_blockchain_interface::{
 };
 use nimiq_genesis::NetworkId;
 use nimiq_hash::Blake2bHash;
-use nimiq_primitives::slots_allocation::{Slot, Validators};
+use nimiq_primitives::{
+    policy::Policy,
+    slots_allocation::{Slot, Validators},
+};
 use tokio_stream::wrappers::BroadcastStream;
 
 use crate::Blockchain;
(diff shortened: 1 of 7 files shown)


blockchain/src/blockchain/abstract_blockchain.rs
```diff
@@ -43,6 +46,20 @@ impl AbstractBlockchain for Blockchain {
         self.state.main_chain.head.epoch_number()
     }
 
+    fn can_enforce_validity_window(&self) -> bool {
```
Eligioo, line 49: Would it you be able to remove the method from the blockchain wrapper(blockchain/src/blockchain/wrappers.rs:238)?

If you can enforce the validity window is now defined in two places.

viquezclaudio: Yup, good catch!
I intended to remove this from the blockchain wrapper.

Eligioo (approved):

jsdanielh (approved):

Issue #2625, opened 2024-06-17
Add a mechanism to sync the mempool


styppo:

sisou:
Is the mempool not synced between validators?

hrxi:
No, they're just accidentally synced if they see the same transactions.

sisou:
As mentioned in https://github.com/nimiq/core-rs-albatross/pull/2218#discussion_r1637657311, should full nodes and light nodes also be able to see the mempool?

PR #2681, opened 2024-06-24
Extend transaction topic mesh to non-validator nodes

### LINKED ISSUE #2615: Extend Gossipsub mesh of the transaction topic to nodes other than validators.

paberr:

This PR extends the nodes that listen to and validate transactions on the network.
This is done by:
- Separating the mempool from the validator.
- Moving the "can validate" logic from the validator/consensus/blockchain into the consensus.
- Initialising the mempool also for non-validator history and full clients.

## Pull request checklist

### DIFF
--- a/consensus/src/consensus/mod.rs
+++ b/consensus/src/consensus/mod.rs
@@ -13,7 +13,7 @@ use futures::{FutureExt, StreamExt};
 use instant::Instant;
 use nimiq_block::Block;
 use nimiq_blockchain_interface::AbstractBlockchain;
-use nimiq_blockchain_proxy::BlockchainProxy;
+use nimiq_blockchain_proxy::{BlockchainProxy, BlockchainReadProxy};
 use nimiq_hash::Blake2bHash;
 use nimiq_network_interface::{network::Network, request::request_handler};
 use nimiq_utils::spawn::spawn;
@@ -55,7 +55,9 @@ mod remote_event_dispatcher;
 #[derive(Clone)]
 pub enum ConsensusEvent {
     /// Consensus is established
-    Established,
+    /// Also includes a flag that indicates if we are ready for transaction verification.
+    /// The established event can be triggered multiple times with different values for the flag.
+    Established { synced_validity_window: bool },
     /// Consensus was lost
     Lost,
 }
@@ -128,6 +130,8 @@ pub struct Consensus<N: Network> {
 
     events: BroadcastSender<ConsensusEvent>,
     established_flag: Arc<AtomicBool>,
+    last_batch_number: u32,
+    synced_validity_window_flag: Arc<AtomicBool>,
     head_requests: Option<HeadRequests<N>>,
     head_requests_time: Option<Instant>,
 
@@ -192,6 +196,14 @@ impl<N: Network> Consensus<N> {
         Self::init_remote_event_dispatcher(&network, &blockchain);
 
         let established_flag = Arc::new(AtomicBool::new(false));
+        let mut synced_validity_window_flag = true;
+        #[cfg(feature = "full")]
+        {
+            if let BlockchainReadProxy::Full(blockchain) = blockchain.read() {
+                synced_validity_window_flag = blockchain.can_enforce_validity_window();
+            }
+        }
+        let synced_validity_window_flag = Arc::new(AtomicBool::new(synced_validity_window_flag));
 
         Consensus {
             blockchain,
@@ -199,6 +211,8 @@ impl<N: Network> Consensus<N> {
             sync: syncer,
             events: tx,
             established_flag,
+            last_batch_number: 0,
+            synced_validity_window_flag,
             head_requests: None,
             head_requests_time: None,
             min_peers,
@@ -286,6 +300,7 @@ impl<N: Network> Consensus<N> {
             blockchain: self.blockchain.clone(),
             network: Arc::clone(&self.network),
             established_flag: Arc::clone(&self.established_flag),
+            synced_validity_window_flag: Arc::clone(&self.synced_validity_window_flag),
             events: self.events.clone(),
             request: self.requests.0.clone(),
         }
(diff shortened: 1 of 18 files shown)


sisou:
Does this address #2625? Or is that another mechanism besides the transaction topic?

hrxi:
That's a different thing, the mempool is still not synced (it's only synced accidentally).

sisou:
What is the transaction topic then, if not a way of exchanging transactions between nodes?

I would think that's only relevant for mempool tx, as after inclusion the included txs are shared via their blocks on the block topic.

jsdanielh:
> What is the transaction topic then, if not a way of exchanging transactions between nodes?

As you say it is for transmission of transactions between nodes but this transmission doesn't warrant any synchronicity between them.

mempool/mempool-task/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+                        .expect("Head block not found");
+
+                    self.mempool
+                        .update(&vec![(hash.clone(), block)], [].as_ref());
+                }
+            }
+            BlockchainEvent::Rebranched(old_chain, new_chain) => {
+                // Mempool updates are only done once we are synced.
+                if self.consensus.is_ready_for_validation() {
+                    self.mempool.update(&new_chain, &old_chain);
```
jsdanielh: Clippy warning:
```suggestion
                    self.mempool.update(new_chain, old_chain);
```

mempool/mempool-task/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+                }
+            }
+            BlockchainEvent::Extended(hash) => {
+                // Mempool updates are only done once we are synced.
+                if self.consensus.is_ready_for_validation() {
+                    let block = self
+                        .consensus
+                        .blockchain
+                        .read()
+                        .get_block(&hash, true)
```
jsdanielh: Clippy warning
```suggestion
                        .get_block(hash, true)
```

mempool/mempool-task/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+
+[lints]
+workspace = true
+
+[dependencies]
+futures = { workspace = true }
+log = { workspace = true }
+parking_lot = "0.12"
+tokio-stream = { version = "0.1", features = ["sync"] }
+tokio-metrics = { version = "0.3", optional = true }
```
jsdanielh:
```suggestion
tokio-metrics = { version = "0.3", optional = true }
tokio-stream = { version = "0.1", features = ["sync"] }
```

mempool/mempool-task/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+
+nimiq-blockchain = { workspace = true }
+nimiq-blockchain-interface = { workspace = true }
+nimiq-consensus = { workspace = true }
+nimiq-mempool = { workspace = true }
+nimiq-utils = { workspace = true, features = ["time"] }
+nimiq-network-interface = { workspace = true }
```
jsdanielh:
```suggestion
nimiq-network-interface = { workspace = true }
nimiq-utils = { workspace = true, features = ["time"] }
```

consensus/src/consensus/mod.rs
```diff
@@ -301,7 +316,48 @@ impl<N: Network> Consensus<N> {
...
+        let (synced_validity_window, _) = self.check_validity_window();
+        self.events
+            .send(ConsensusEvent::Established {
+                synced_validity_window,
+            })
+            .ok();
+    }
+
+    /// Checks if the validity window is available.
+    /// This function contains optimisations to only run the check when necessary.
```
jsdanielh:
```suggestion
    /// This function contains optimizations to only run the check when necessary.
```

lib/Cargo.toml
```diff
@@ -83,18 +91,54 @@ nimiq-test-log = { workspace = true }
...
 tokio-console = ["console-subscriber", "logging", "tokio/tracing"]
 tokio-websocket = ["nimiq-network-libp2p/tokio-websocket"]
-validator = ["database-storage", "nimiq-mempool", "nimiq-validator", "nimiq-validator-network", "nimiq-rpc-server"]
+validator = [
+    "database-storage",
+    "nimiq-mempool",
+    "nimiq-mempool-task",
+    "nimiq-validator",
+    "nimiq-validator-network",
+    "nimiq-rpc-server",
```
jsdanielh, line 129:
```suggestion
    "database-storage",
    "nimiq-mempool",
    "nimiq-mempool-task",
    "nimiq-rpc-server",
    "nimiq-validator",
    "nimiq-validator-network",
```

lib/src/client.rs
```diff
@@ -16,6 +16,8 @@ use nimiq_genesis::NetworkInfo;
 use nimiq_light_blockchain::LightBlockchain;
 #[cfg(feature = "validator")]
 use nimiq_mempool::mempool::Mempool;
+#[cfg(feature = "validator")]
```
jsdanielh, line 19: Maybe this shouldn't be part of this feature

mempool/mempool-task/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct MempoolEvent(BlockchainEvent);
+
+impl From<MempoolEvent> for BlockchainEvent {
+    fn from(event: MempoolEvent) -> Self {
+        event.0
+    }
+}
+
+pub struct MempoolTask<N: Network> {
```
jsdanielh, line 32: Could you add some `rustdoc`?

consensus/src/consensus/mod.rs
```diff
@@ -301,7 +316,48 @@ impl<N: Network> Consensus<N> {
...
+                let old_value = self
+                    .synced_validity_window_flag
+                    .swap(can_enforce_validity_window, Ordering::Release);
+                // If the value changed, send an Established event.
+                let mut event = None;
+                if old_value != can_enforce_validity_window {
+                    event = Some(ConsensusEvent::Established {
+                        synced_validity_window: can_enforce_validity_window,
+                    });
+                }
```
hrxi, line 356:
```suggestion
                let event = (old_value != can_enforce_validity_window)
                    .then_some(ConsensusEvent::Established {
                        synced_validity_window: can_enforce_validity_window,
                    });
```
Just an idea, not sure if more readable.

lib/src/client.rs
```diff
@@ -68,6 +70,15 @@ pub type ValidatorProxy = AbstractValidatorProxy;
 pub type ZKPComponent = AbstractZKPComponent<Network>;
 pub type ZKPComponentProxy = AbstractZKPComponentProxy<Network>;
 
+#[cfg(feature = "validator")]
+pub type MempoolTask = AbstractMempoolTask<Network>;
+
+#[cfg(feature = "validator")]
+pub enum ValidatorOrMempool {
+    Validator(Validator),
```
hrxi, line 78: I think only the `Validator` variant should be behind the feature flag.

lib/Cargo.toml
```diff
@@ -61,14 +66,17 @@ nimiq-keys = { workspace = true }
...
 nimiq-network-libp2p = { workspace = true }
 nimiq-network-interface = { workspace = true }
 nimiq-primitives = { workspace = true, features = ["networks"] }
 nimiq-rpc-server = { workspace = true, optional = true }
 nimiq-serde = { workspace = true }
 nimiq-utils = { workspace = true, features = ["time", "key-store"] }
-nimiq-validator = { workspace = true, optional = true, features = ["trusted_push"] }
+nimiq-validator = { workspace = true, optional = true, features = [
+    "trusted_push",
+] }
```
hrxi, line 78: I can try to undo the autoformatter, it doesn't really make things better here.

mempool/mempool-task/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+            return Poll::Ready(Some(MempoolEvent(event)));
+        }
+
+        Poll::Pending
+    }
+}
+
+impl<TValidatorNetwork: ValidatorNetwork> Future for MempoolTask<TValidatorNetwork>
+where
+    PubsubId<TValidatorNetwork>: std::fmt::Debug + Unpin,
```
hrxi: These should be put onto the `ValidatorNetwork` trait instead.

mempool/mempool-task/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+        while let Poll::Ready(Some(event)) = self.consensus_event_rx.poll_next_unpin(cx) {
+            match event {
+                Ok(ConsensusEvent::Established {
+                    synced_validity_window: true,
+                }) => self.init_mempool(),
+                Ok(ConsensusEvent::Lost)
+                | Ok(ConsensusEvent::Established {
+                    synced_validity_window: false,
+                }) => self.pause(),
+                Err(_) => return Poll::Ready(None),
```
hrxi: In what case does this error?

If this can happen during normal operation, then this violates the `Stream` protocol because the other stream isn't polled.

hrxi (commented):
Haven't finished my review completely, but will move onto something different for now.

lib/Cargo.toml
paberr: I'll revert the changes to the formatting.

lib/src/client.rs
paberr: We don't have a better feature flag right now, this should be addresses as part of #2719

consensus/src/consensus/mod.rs
paberr: I prefer the if 😄

lib/src/client.rs
paberr: This would make the web client not compile anymore.
We don't have a better feature flag right now, this should be addresses as part of #2719

lib/Cargo.toml
paberr: I'll do that.

mempool/mempool-task/src/lib.rs
paberr: Outdated.

paberr:
Addressed all comments so far.

jsdanielh (approved):
LGTM

PR #2682, opened 2024-06-24
primitives: Update supply curve

### LINKED ISSUE #2380: Update new coin emission based on the Supply Curve Vote result
As per the outcome of the recent Supply Curve Vote, we should update the policy parameters to reflect the new emission accordingly.

jsdanielh:
Update supply curve to reflect the community voted curve which is the same used in the PoW chain (ideal curve).

This closes #2380.

## Pull request checklist

### DIFF
--- a/blockchain/tests/inherents.rs
+++ b/blockchain/tests/inherents.rs
@@ -47,7 +47,7 @@ fn it_can_create_batch_finalization_inherents() {
         version: 1,
         block_number: Policy::macro_block_of(2).unwrap(),
         round: 0,
-        timestamp: blockchain.state.election_head.header.timestamp + 1,
+        timestamp: blockchain.state.election_head.header.timestamp + 20000,
         parent_hash: Blake2bHash::default(),
         parent_election_hash: Blake2bHash::default(),
         interlink: None,
@@ -89,7 +89,7 @@ fn it_can_create_batch_finalization_inherents() {
     for inherent in &inherents {
         match inherent {
             Inherent::Reward { value, .. } => {
-                assert_eq!(*value, Coin::from_u64_unchecked(875));
+                assert_eq!(*value, Coin::from_u64_unchecked(942));
                 got_reward = true;
             }
             Inherent::FinalizeBatch => {
@@ -145,7 +145,7 @@ fn it_can_create_batch_finalization_inherents() {
 
     let inherents = blockchain.finalize_previous_batch(&macro_block);
     assert_eq!(inherents.len(), 3);
-    let one_slot_reward = 875 / Policy::SLOTS as u64;
+    let one_slot_reward = 942 / Policy::SLOTS as u64;
     let mut got_reward = false;
     let mut got_penalize = false;
     let mut got_finalize_batch = false;
@@ -163,7 +163,7 @@ fn it_can_create_batch_finalization_inherents() {
                     got_penalize = true;
                 } else {
                     assert_eq!(*actual_validator_address, *validator_address);
-                    assert_eq!(*value, Coin::from_u64_unchecked(875 - one_slot_reward));
+                    assert_eq!(*value, Coin::from_u64_unchecked(942 - one_slot_reward));
                     got_reward = true;
                 }
             }
--- a/blockchain/tests/inherents.rs
+++ b/blockchain/tests/inherents.rs
@@ -204,14 +204,9 @@ fn it_can_penalize_delayed_batch() {
     let (genesis_supply, genesis_timestamp) = blockchain.get_genesis_parameters();
 
     // Total reward for the previous batch
-    let prev_supply = Policy::supply_at(
-        u64::from(genesis_supply),
-        genesis_timestamp,
-        genesis_timestamp,
-    );
+    let prev_supply = Policy::supply_at(u64::from(genesis_supply), genesis_timestamp);
 
-    let current_supply =
-        Policy::supply_at(u64::from(genesis_supply), genesis_timestamp, next_timestamp);
+    let current_supply = Policy::supply_at(u64::from(genesis_supply), next_timestamp);
 
     let max_reward = current_supply - prev_supply;
 
--- a/primitives/src/policy.rs
+++ b/primitives/src/policy.rs
@@ -138,7 +138,7 @@ impl Policy {
     /// This is the number of Lunas (1 NIM = 100,000 Lunas) created by millisecond at the genesis of the
     /// Nimiq 2.0 chain. The velocity then decreases following the formula:
     /// Supply_velocity (t) = Initial_supply_velocity * 2^(- Supply_decay * t)
-    /// Where t is the time in milliseconds since the genesis block.
+    /// Where t is the time in milliseconds since the PoW genesis block.
     pub const INITIAL_SUPPLY_VELOCITY: f64 = 0.10594132556065439;
 
     /// The supply decay is a constant that is calculated so that the supply velocity decreases at a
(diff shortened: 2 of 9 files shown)


primitives/src/policy.rs
```diff
@@ -121,13 +121,13 @@ impl Policy {
 
     /// This is the number of Lunas (1 NIM = 100,000 Lunas) created by millisecond at the genesis of the
     /// Nimiq 2.0 chain. The velocity then decreases following the formula:
-    /// Supply_velocity (t) = Initial_supply_velocity * e^(- Supply_decay * t)
-    /// Where e is the exponential function and t is the time in milliseconds since the genesis block.
-    pub const INITIAL_SUPPLY_VELOCITY: f64 = 875.0;
+    /// Supply_velocity (t) = Initial_supply_velocity * 2^(- Supply_decay * t)
+    /// Where t is the time in milliseconds since the genesis block.
+    pub const INITIAL_SUPPLY_VELOCITY: f64 = 0.10594132556065439;
```
sisou, line 142: Is this a number that is dependent on the PoS genesis block timestamp? Or can we fix it to a stable timestamp, e.g. PoW genesis timestamp?

primitives/src/policy.rs
```diff
@@ -502,7 +502,8 @@ impl Policy {
         let exponent = -Policy::SUPPLY_DECAY * t;
 
         let supply = genesis_supply
-            + (Self::INITIAL_SUPPLY_VELOCITY / Self::SUPPLY_DECAY * (1.0 - exp(exponent))) as u64;
+            + (Self::INITIAL_SUPPLY_VELOCITY / Self::SUPPLY_DECAY
+                * (1.0 - f64::powf(2.0, exponent))) as u64;
```
hrxi: We can't use the `f64::powf` function here, since this is consensus-relevant and https://doc.rust-lang.org/std/primitive.f64.html#method.powf says:
> The precision of this function varies by platform and Rust version.

hrxi (changes requested):

jsdanielh:
Numbers for the regressed curved were calculated using: https://github.com/jsdanielh/supply-curve

The result is the following:
image

primitives/src/policy.rs
jsdanielh: Yes, if we want to 'continue' the curve, we would need to use the PoW timestamp. Will make a change to use it.

jsdanielh: Done

sisou:
@jsdanielh but careful now with the mix of parameters. You are using the PoW timestamp, but the PoS supply. Should it use the PoW supply, to start the curve at the right point?

Also, the PoW supply was based on _block number_. The block numbers have diverged slightly from the ideal timeline, I don't remember if we are _early_ or _late_. So using the "idealized" curve from the PoS genesis will either skip a few block rewards, or pay them out double. Just something to think about.

primitives/src/policy.rs
```diff
@@ -136,6 +136,9 @@ impl Policy {
     /// 25 MB.
     pub const HISTORY_CHUNKS_MAX_SIZE: u64 = 25 * 1024 * 1024;
 
+    /// The PoW genesis timestamp.
+    pub const POW_GENESIS_TIMESTAMP: u64 = 1523727000;
```
hrxi: What's the unit here? It looks like a timestamp using seconds as resolution, but I think we need milliseconds.

jsdanielh: Yeah, you're right

primitives/src/policy.rs
jsdanielh: Ok, I added `pow` to our math utils and changed the function to use it.

jsdanielh:
> @jsdanielh but careful now with the mix of parameters. You are using the PoW timestamp, but the PoS supply. Should it use the PoW supply, to start the curve at the right point?

I don't think so. The genesis supply is the initial factor in the supply curve and just sets an offset on which supply we start with. The rate of decay in the other hand needs to be preserved since it is exponentially decreasing in time and needs to consider that the supply rate already decayed (6+ years of PoW life).

sisou:
You are right that the genesis supply has no influence on the batch rewards (as for that, the increase of `supply_at` between two macro block timestamps is compared, so the initial supply added to both values cancels out).

But if you wanted to compute the actual supply at a certain point in time, using the PoS genesis supply generates a way-too-high number, as the graph is moved up on the Y-axis by the supply difference between PoW genesis and PoS genesis.

blockchain/tests/inherents.rs
```diff
@@ -47,7 +47,7 @@ fn it_can_create_batch_finalization_inherents() {
         version: 1,
         block_number: Policy::macro_block_of(2).unwrap(),
         round: 0,
-        timestamp: blockchain.state.election_head.header.timestamp + 1,
+        timestamp: blockchain.state.election_head.header.timestamp + 20000,
```
sisou, line 50: Any reason for this 20s?

jsdanielh: The test has assumptions in terms of the number of reward being paid which with the new supply curve don't hold for such a short timestamp difference.

utils/src/math/mod.rs
```diff
@@ -1,3 +1,5 @@
 mod exp;
+mod pow;
 
 pub use exp::exp;
+pub use pow::pow;
```
hrxi, line 5: You should be able to remove the `exp` implementation.

hrxi (commented):
I think there will be a discontinuity if we simply base it on the time since the PoW genesis block as the PoW rewards are not based on time AFAIK. Is that a problem?

utils/src/math/mod.rs
jsdanielh: Unfortunately it is still used to calculate batch delay penalties

jsdanielh:
> I think there will be a discontinuity if we simply base it on the time since the PoW genesis block as the PoW rewards are not based on time AFAIK. Is that a problem?

We're basing it in the _ideal_ PoW emission curve based on the ideal expected number of blocks per year. Having said that, indeed this is a subject of discussion.

hrxi (approved):

primitives/src/policy.rs
```diff
@@ -490,19 +493,20 @@ impl Policy {
...
 
-        let t = (current_time - genesis_time) as f64;
+        let t = (current_time - Self::POW_GENESIS_TIMESTAMP) as f64;
 
         let exponent = -Policy::SUPPLY_DECAY * t;
 
         let supply = genesis_supply
-            + (Self::INITIAL_SUPPLY_VELOCITY / Self::SUPPLY_DECAY * (1.0 - exp(exponent))) as u64;
+            + (Self::INITIAL_SUPPLY_VELOCITY / Self::SUPPLY_DECAY * (1.0 - pow(2.0, exponent)))
+                as u64;
```
hrxi, line 503: Looking at this again, this looks wrong.

This function returns more than the PoS `genesis_supply` at the `current_time` of the PoS genesis block.

hrxi (changes requested):

primitives/src/policy.rs
hrxi: Since I'm looking at simplifying the formula anyway, I'll probably also fix that part.

sisou: That is what [I commented above](https://github.com/nimiq/core-rs-albatross/pull/2682#issuecomment-2188872842):
> But if you wanted to compute the actual supply at a certain point in time, using the PoS genesis supply generates a way-too-high number, as the graph is moved up on the Y-axis by the supply difference between PoW genesis and PoS genesis.

sisou: If you move the start of the curve to the PoW genesis, you also need to move the genesis supply there. 

_Or_, if you want to use the PoS genesis supply, you need to start the curve with the initial velocity it will have at the PoS genesis time, which is not yet fixed, so the velocity cannot be fixed yet.

PR #2687, opened 2024-06-25
web-client: Include a various core-js utility classes in the package

sisou:
For example the coveted `BufferUtils` and `SerialBuffer`, which didn't have a replacement in the WASM client until now, and are used _everywhere_ in our apps.


web-client/extras/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "target": "es2022",
+    "allowJs": true,
+    "resolveJsonModule": true,
+    "moduleDetection": "force",
+    "isolatedModules": true,
+    "verbatimModuleSyntax": true,
+
+    /* Strictness */
+    "strict": true,
+    // "noUncheckedIndexedAccess": true,
```
jsdanielh:
```suggestion
```

web-client/extras/lib/BufferUtils.ts
```diff
@@ -0,0 +1,352 @@
...
+    static isHex(str: string, length?: number): boolean {
+        if (!/^[0-9A-Fa-f]*$/.test(str)) return false;
+        if (str.length % 2 !== 0) return false;
+        if (typeof length === 'number' && str.length / 2 !== length) return false;
+        return true;
+    }
+
+    static fromHex(hex: string, length?: number): Uint8Array {
+        hex = hex.trim();
+        if (!BufferUtils.isHex(hex, length)) throw new Error('String is not an hex string (of matching length)');
```
jsdanielh:
```suggestion
        if (!BufferUtils.isHex(hex, length)) throw new Error('String is not a hex string (of matching length)');
```

web-client/extras/tsconfig.json
jsdanielh:
```suggestion
```

jsdanielh (approved):

hrxi:
It seems to me that this adds a lot of unused code, and even for the used code it's not immediately clear to me that there aren't Web API replacements. Could we look into that?

web-client/extras/lib/ArrayUtils.ts
```diff
@@ -0,0 +1,53 @@
...
+            if (!found) {
+                return;
+            }
+            indices[i] += 1;
+            for (const j of Array.from(new Array(k - i - 1), (x, k) => i + k + 1)) {
+                indices[j] = indices[j - 1] + 1;
+            }
+            yield indices.map(i => list[i]);
+        }
+    }
```
hrxi, line 52: `randomElement` and `k_combinations` (different naming style?) seem unused. Since they're also not documented, maybe we can remove them?

web-client/extras/lib/BufferUtils.ts
```diff
@@ -0,0 +1,363 @@
...
+    static toAscii(buffer: Uint8Array): string {
+        const chunkSize = 0x2000;
+        const buf = BufferUtils._toUint8View(buffer);
+
+        let ascii = '';
+        for (let i = 0; i < buf.length; i += chunkSize) {
+            ascii += String.fromCharCode.apply(null, [...buf.subarray(i, i + chunkSize)]);
+        }
+        return ascii;
+    }
```
hrxi, line 37: This isn't `toAscii`, it's more like `toIso8559_1`. It doesn't throw on bytes >= 128.

web-client/extras/lib/BufferUtils.ts
```diff
@@ -0,0 +1,363 @@
...
+
+    static fromAscii(string: string): Uint8Array {
+        const buf = new Uint8Array(string.length);
+        for (let i = 0; i < string.length; ++i) {
+            buf[i] = string.charCodeAt(i);
+        }
+        return buf;
+    }
```
hrxi, line 45: Same problem as `fromAscii`.

web-client/extras/lib/BufferUtils.ts
```diff
@@ -0,0 +1,363 @@
...
+        const uint8View = BufferUtils._toUint8View(buffer);
+        return BufferUtils._ISO_8859_15_DECODER.decode(uint8View)
+            .replace(/\u20ac/g, '\u00a4') // € => ¤
+            .replace(/\u0160/g, '\u00a6') // Š => ¦
+            .replace(/\u0161/g, '\u00a8') // š => ¨
+            .replace(/\u017d/g, '\u00b4') // Ž => ´
+            .replace(/\u017e/g, '\u00b8') // ž => ¸
+            .replace(/\u0152/g, '\u00bc') // Œ => ¼
+            .replace(/\u0153/g, '\u00bd') // œ => ½
+            .replace(/\u0178/g, '\u00be'); // Ÿ => ¾
```
hrxi, line 67: What do these replacements mean?

web-client/extras/lib/BufferUtils.ts
```diff
@@ -0,0 +1,363 @@
...
+                out[p++] = ((c >> 6) & 63) | 128;
+                out[p++] = (c & 63) | 128;
+            } else {
+                out[p++] = (c >> 12) | 224;
+                out[p++] = ((c >> 6) & 63) | 128;
+                out[p++] = (c & 63) | 128;
+            }
+        }
+        return new Uint8Array(out);
+    }
```
hrxi, line 263: JS probably has a UTF-8 encoder?

web-client/extras/lib/BufferUtils.ts
```diff
@@ -0,0 +1,363 @@
...
+        const viewB = BufferUtils._toUint8View(b);
+        if (viewA.length !== viewB.length) return false;
+        for (let i = 0; i < viewA.length; i++) {
+            if (viewA[i] !== viewB[i]) return false;
+        }
+        return true;
+    }
+
+    /**
+     * Returns -1 if a is smaller than b, 1 if a is larger than b, 0 if a equals b.
```
hrxi, line 326: Should probably add: Shorter arrays are always considered smaller than longer ones.

web-client/extras/lib/NumberUtils.ts
```diff
@@ -0,0 +1,42 @@
...
+            && val >= 0 && val <= NumberUtils.UINT64_MAX;
+    }
+
+    static randomUint32(): number {
+        return Math.floor(Math.random() * (NumberUtils.UINT32_MAX + 1));
+    }
+
+    static randomUint64(): number {
+        return Math.floor(Math.random() * (NumberUtils.UINT64_MAX + 1));
+    }
```
hrxi, line 37: This doesn't reach every 64-bit integer due to the float precision. Probably unused and can be removed anyway.

web-client/extras/lib/SerialBuffer.ts
```diff
@@ -0,0 +1,205 @@
...
+            this.writeUint8(0xFD);
+            this.writeUint16(value);
+        } else if (value <= 0xFFFFFFFF) {
+            this.writeUint8(0xFE);
+            this.writeUint32(value);
+        } else {
+            this.writeUint8(0xFF);
+            this.writeUint64(value);
+        }
+    }
```
hrxi, line 134: I don't think we have these kinds of varuints in the code base anymore.

web-client/extras/lib/SerialBuffer.ts
```diff
@@ -0,0 +1,205 @@
...
+        const bytes = this.read(length);
+        return BufferUtils.toAscii(bytes);
+    }
+
+    writeVarLengthString(value: string): void {
+        if (StringUtils.isMultibyte(value) || !NumberUtils.isUint8(value.length)) throw new Error('Malformed value');
+        const bytes = BufferUtils.fromAscii(value);
+        this.writeUint8(bytes.byteLength);
+        this.write(bytes);
+    }
```
hrxi, line 199: That's not how we do variable-length strings in the codebase.

web-client/extras/lib/StringUtils.ts
```diff
@@ -0,0 +1,29 @@
+export class StringUtils {
+    static isMultibyte(str: string): boolean {
```
hrxi, line 2: `isMultibyte` seems like the wrong name. It also seems to be used as a `isAscii` check, which this function does not implement.

web-client/extras/lib/SerialBuffer.ts
sisou: Varuints definitely exist in core-js-albatross, for example I think the sender-data length is encoded as that? Maybe even the proof length?

sisou: Or is that a different _kind_ of varint?

web-client/extras/lib/SerialBuffer.ts
sisou: How do we do it then? With a varint length?

web-client/extras/lib/ArrayUtils.ts
sisou: Addressed in #2975

web-client/extras/lib/BufferUtils.ts
sisou: Addressed in #2975

web-client/extras/lib/BufferUtils.ts
sisou: Addressed in #2975

web-client/extras/lib/BufferUtils.ts
sisou: Addressed in #2975

web-client/extras/lib/BufferUtils.ts
sisou: Addressed in #2975

web-client/extras/lib/NumberUtils.ts
sisou: Addressed in #2975

web-client/extras/lib/StringUtils.ts
sisou: It checks if all characters in the string are in the ascii range. So it's definitely not multibyte when this returns true. Or what am I missing?

hrxi: As far as I can tell, it checks whether the string has any [surrogates](https://en.wikipedia.org/w/index.php?title=UTF-16&oldid=1245839466#U+D800_to_U+DFFF_(surrogates)) in UTF-16 encoding. ASCII would be checking that all characters are between U+0 and U+7F.

web-client/extras/lib/SerialBuffer.ts
hrxi: Yes, with varint length: https://postcard.jamesmunns.com/wire-format#15---string.

web-client/extras/lib/SerialBuffer.ts
hrxi: Different kind of varint: https://postcard.jamesmunns.com/wire-format#varint-encoded-integers.

web-client/extras/lib/StringUtils.ts
sisou: Hmm. So what should this method do instead? Or should I research?

hrxi: I don't really know myself. Looking at the usage, it looks like [isWellFormed](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/isWellFormed) would be a good replacement. It checks whether the JS string is actually a well-formed Unicode string. This guarantees that it can be encoded as UTF-8. Looking at the MDN page, it might be a new API that we can't rely on yet, so maybe there's a polyfill version?

sisou: Done in https://github.com/nimiq/core-rs-albatross/pull/2715/commits/c805864e562615813f27b57ee43d895f867bdb93

web-client/extras/lib/SerialBuffer.ts
sisou: Done in https://github.com/nimiq/core-rs-albatross/pull/2715/commits/3103f08eaf93752c0376f93e25e2b079f81aaae1

web-client/extras/lib/SerialBuffer.ts
sisou: Done in https://github.com/nimiq/core-rs-albatross/pull/2715/commits/3103f08eaf93752c0376f93e25e2b079f81aaae1

PR #2691, opened 2024-06-25
Make vesting contract creation parsing more resilient

### LINKED ISSUE #2690: Related addresses on vesting contract creation are wrong
https://github.com/nimiq/core-rs-albatross/blob/5cfa101819663f7ff4a2151346dfc75ea83fe2e2/primitives/transaction/src/lib.rs#L477-L478

`Deserialize::deserialize_from_vec` of this struct is wrong and can't be used here.

hrxi:
Use `serde` to do it, but remove the `Deserialize` impl because it was error-prone.

Based on top of #2676, so this should wait until #2676 is approved.

Fixes #2690.

### DIFF
--- a/primitives/transaction/src/account/vesting_contract.rs
+++ b/primitives/transaction/src/account/vesting_contract.rs
@@ -1,7 +1,7 @@
 use log::error;
 use nimiq_keys::Address;
 use nimiq_primitives::{account::AccountType, coin::Coin};
-use nimiq_serde::{Deserialize, Serialize};
+use nimiq_serde::{Deserialize, Serialize, SerializedSize};
 
 use crate::{
     account::AccountTransactionVerification, SignatureProof, Transaction, TransactionError,
@@ -40,16 +40,6 @@ impl AccountTransactionVerification for VestingContractVerifier {
             return Err(TransactionError::InvalidForRecipient);
         }
 
-        let allowed_sizes = [Address::SIZE + 8, Address::SIZE + 24, Address::SIZE + 32];
-        if !allowed_sizes.contains(&transaction.recipient_data.len()) {
-            warn!(
-                len = transaction.recipient_data.len(),
-                ?transaction,
-                "Invalid data length for this transaction",
-            );
-            return Err(TransactionError::InvalidData);
-        }
-
         CreationTransactionData::parse(transaction).map(|_| ())
     }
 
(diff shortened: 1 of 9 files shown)


primitives/transaction/src/account/vesting_contract.rs
```diff
@@ -68,79 +68,132 @@ impl AccountTransactionVerification for VestingContractVerifier {
     }
 }
 
-#[derive(Default, Clone, Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Default)]
 pub struct CreationTransactionData {
```
jsdanielh, line 65: Could you add some rustdoc?

primitives/transaction/src/account/vesting_contract.rs
```diff
@@ -68,79 +68,132 @@ impl AccountTransactionVerification for VestingContractVerifier {
...
-        if transaction.recipient_data.len() == Address::SIZE + 8 {
+    fn parse_impl(data: &[u8], tx_value: Coin) -> Result<Self, TransactionError> {
+        // TODO: Change this into `const` block once Rust 1.79 is sufficiently old.
+        #[allow(clippy::assertions_on_constants)]
+        {
+            assert!(CreationTransactionData8::SIZE != CreationTransactionData24::SIZE);
+            assert!(CreationTransactionData8::SIZE != CreationTransactionData32::SIZE);
+            assert!(CreationTransactionData24::SIZE != CreationTransactionData32::SIZE);
+        }
+        Ok(if data.len() == CreationTransactionData8::SIZE {
```
jsdanielh: Why are we not using here a `match` statement?

serde/src/lib.rs
```diff
@@ -91,14 +91,22 @@ pub trait SerializedMaxSize {
     const MAX_SIZE: usize;
 }
 
+/// Types implementing this trait have an alternative fixed-size serialization length.
+pub trait SerializedFixedSize {
```
jsdanielh, line 95: I think a better naming could be `FixedSerializedSize`. wdyt?

primitives/transaction/src/account/vesting_contract.rs
hrxi: Fixed.

primitives/transaction/src/account/vesting_contract.rs
hrxi: Done.

serde/src/lib.rs
hrxi: I think I prefer the current name.

jsdanielh (approved):

Issue #2623, opened 2024-06-17
Investigate macro block production being slow / not terminating


styppo:
We can see this happening in the ValidatorStress scenario

jsdanielh:
Closed since handel was refactored.

PR #2693, opened 2024-06-26
Create poll function for handel aggregations

nibhar:
This PR implements a poll function for handel.

Other smaller additions have been made like removing the tokio dependence and some changes to lock scoping as well as wakers. 

Lastly the test has been updated to be more reliable.

While this PR may affect #2623 I do not expect it to have a big impact.

### DIFF
--- a/handel/Cargo.toml
+++ b/handel/Cargo.toml
@@ -38,4 +38,4 @@ nimiq-network-mock = { workspace = true }
 nimiq-test-log = { workspace = true }
 
 tokio = { version = "1.38", features = ["rt", "time", "macros"] }
-nimiq-utils = { workspace = true }
+nimiq-utils = { workspace = true, features = ["spawn"] }
--- a/handel/Cargo.toml
+++ b/handel/Cargo.toml
@@ -22,8 +22,6 @@ parking_lot = "0.12"
 rand = "0.8"
 serde = "1.0"
 thiserror = "1.0"
-tokio = { version = "1.38", features = ["macros", "rt", "time"] }
-tokio-stream = "0.1"
 
 nimiq-bls = { workspace = true }
 nimiq-collections = { workspace = true }
--- a/handel/src/aggregation.rs
+++ b/handel/src/aggregation.rs
@@ -117,7 +117,7 @@ where
     }
 
     /// Starts level `level`
-    fn start_level(&mut self, level: usize) {
+    fn start_level(&mut self, level: usize, store: &<P as Protocol<TId>>::Store) {
         let level = self
             .levels
             .get(level)
@@ -135,11 +135,7 @@ where
             // Don't do anything for level 0 as it only contains this node
             if level.id > 0 {
                 // Get the current best for the level. Freeing the lock as soon as possible to continue working on todos
-                let best = {
-                    let store = self.protocol.store();
-                    let store = store.read();
-                    store.combined(level.id - 1)
-                };
+                let best = store.combined(level.id - 1);
 
                 if let Some(best) = best {
                     self.send_update(
@@ -165,7 +161,7 @@ where
     }
 
     /// Check if a level was completed
-    fn check_completed_level(&mut self, level_id: usize) {
+    fn check_completed_level(&mut self, level_id: usize, store: &<P as Protocol<TId>>::Store) {
         let num_peers = {
             let level = self
                 .levels
@@ -188,14 +184,10 @@ where
 
         // first get the current contributor count for this level. Release the lock as soon as possible
         // to continue working on todos.
-        let num_contributors = {
-            let store = self.protocol.store();
-            let store = store.read();
-            let best = store
-                .best(level_id)
-                .unwrap_or_else(|| panic!("Expected a best signature for level {}", level_id));
-            self.num_contributors(best)
-        };
+        let best = store
+            .best(level_id)
+            .unwrap_or_else(|| panic!("Expected a best signature for level {}", level_id));
+        let num_contributors = self.num_contributors(best);
 
         // If the number of contributors on this level is equal to the number of peers on this level it is completed.
         if num_contributors == num_peers {
@@ -216,19 +208,14 @@ where
             // if there is a level with a higher id than the completed one it needs to be activated.
             if level_id + 1 < self.levels.len() {
                 // activate next level
-                self.start_level(level_id + 1);
+                self.start_level(level_id + 1, store);
             }
         }
 
         // In order to send updated messages iterate all levels higher than the given level.
         let level_count = self.levels.len();
         for i in level_id + 1..level_count {
-            let combined = {
-                // Acquire read lock to retrieve the current combined contribution for the next lower level
-                let store = self.protocol.store();
-                let store = store.read();
-                store.combined(i - 1)
-            };
+            let combined = store.combined(i - 1);
 
             // if there is an aggregate contribution for given level i send it out to the peers of that level.
             if let Some(multisig) = combined {
@@ -327,8 +314,11 @@ where
             // next time the timeout triggers the next level needs activating
             self.next_level_timeout += 1;
 
+            let store_rw = self.protocol.store();
+            let store = store_rw.read();
+
             // finally start the level.
-            self.start_level(level);
+            self.start_level(level, &*store);
         }
     }
 
@@ -346,30 +336,26 @@ where
             return todo.contribution;
         }
 
+        let store_rw = self.protocol.store();
+        let mut store = store_rw.write();
+
         // if the contribution is valid push it to the store, creating a new aggregate
-        {
-            let store = self.protocol.store();
-            let mut store = store.write();
-            store.put(
-                todo.contribution.clone(),
-                todo.level,
-                self.protocol.registry(),
-                self.protocol.identify(),
-            );
-        }
+        store.put(
+            todo.contribution.clone(),
+            todo.level,
+            self.protocol.registry(),
+            self.protocol.identify(),
+        );
 
         // in case the level of this todo has not started, start it now as we have already contributions on it.
-        self.start_level(todo.level);
+        self.start_level(todo.level, &*store);
         // check if a level was completed by the addition of the contribution
-        self.check_completed_level(todo.level);
+        self.check_completed_level(todo.level, &*store);
 
         // get the best aggregate
         let last_level = self.levels.last().expect("No levels");
-        let best = {
-            let store = self.protocol.store();
-            let store = store.read();
-            store.combined(last_level.id)
-        };
+
+        let best = store.combined(last_level.id);
 
         best.expect("A best signature must exist after applying a contribution.")
     }
--- a/handel/src/aggregation.rs
+++ b/handel/src/aggregation.rs
@@ -1,5 +1,4 @@
 use std::{
-    fmt::Debug,
     mem,
     pin::Pin,
     task::{Context, Poll},
@@ -23,6 +22,7 @@ use crate::{
     todo::{TodoItem, TodoList},
     update::LevelUpdate,
     verifier::{VerificationResult, Verifier},
+    Identifier,
 };
 
 // TODOS:
@@ -35,7 +35,7 @@ type LevelUpdateStream<P, T> = BoxStream<'static, LevelUpdate<<P as Protocol<T>>
 /// Future implementation for the next aggregation event
 pub struct OngoingAggregation<TId, P, N>
 where
-    TId: Debug + Clone + Unpin + Send + 'static,
+    TId: Identifier,
     P: Protocol<TId>,
     N: Network<Contribution = P::Contribution>,
 {
@@ -73,7 +73,7 @@ where
 
 impl<TId, P, N> OngoingAggregation<TId, P, N>
 where
-    TId: Debug + Clone + Unpin + Send + 'static,
+    TId: Identifier,
     P: Protocol<TId>,
     N: Network<Contribution = P::Contribution>,
 {
@@ -394,7 +394,7 @@ where
 
 impl<TId, P, N> Stream for OngoingAggregation<TId, P, N>
 where
-    TId: Debug + Clone + Unpin + Send + 'static,
+    TId: Identifier,
     P: Protocol<TId>,
     N: Network<Contribution = P::Contribution>,
 {
@@ -477,7 +477,7 @@ where
 
 pub struct FinishedAggregation<TId, P, N>
 where
-    TId: Debug + Clone + Unpin + Send + 'static,
+    TId: Identifier,
     P: Protocol<TId>,
     N: Network<Contribution = P::Contribution>,
 {
@@ -488,7 +488,7 @@ where
 
 impl<TId, P, N> FinishedAggregation<TId, P, N>
 where
-    TId: Debug + Clone + Unpin + Send + 'static,
+    TId: Identifier,
     P: Protocol<TId>,
     N: Network<Contribution = P::Contribution>,
 {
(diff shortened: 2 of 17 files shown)


handel/src/aggregation.rs
```diff
@@ -321,105 +314,194 @@ impl<
...
+        // if the contribution is valid push it to the store, creating a new aggregate
+        store.put(
+            todo.contribution.clone(),
+            todo.level,
+            self.protocol.registry(),
+            self.protocol.identify(),
+        );
+
+        // in case the level of this todo has not started, start it now as we have already contributions on it.
+        self.start_level(todo.level, &*store);
```
ii-cruz: We do not check if the level has been started so we might be repeatedly starting the level.

handel/src/aggregation.rs
```diff
@@ -321,105 +314,194 @@ impl<
...
+where
+    TId: Debug + Clone + Unpin + Send + 'static,
+    P: Protocol<TId>,
+    N: Network<Contribution = P::Contribution>,
+{
+    type Item = P::Contribution;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        // Check if the automatic update interval triggers, if so perform the update.
+        if let Poll::Ready(_instant) = self.periodic_update_interval.as_mut().poll_tick(cx) {
```
ii-cruz: By increasing this timeout, we might be falling into these periodic updates more often and thus increase the number of messages sent.

handel/src/aggregation.rs
```diff
@@ -321,105 +314,194 @@ impl<
...
+        }
+
+        if let Poll::Ready(_instant) = self.start_level_interval.as_mut().poll_tick(cx) {
+            // Activates the next level if there is a next level.
+            // This potentially creates new messages in the sender.
+            self.activate_next_level();
+        }
+
+        // Poll the verification future if there is one.
+        let mut best_aggregate = if let Some(verification_future) = &mut self.current_verification {
```
ii-cruz, line 405: why is the verification after the intervals and timeouts? Shouldn't it be the other way around?

handel/src/aggregation.rs
ii-cruz: Talked with @nibhar and both timeouts were increased so this is not relevant

handel/src/aggregation.rs
ii-cruz: my bad this is a none issue

handel/src/aggregation.rs
nibhar: I changed the ordering such that the timers are polled last, meaning that any updates created in the potential subsequent calls will already be able to use any new aggregate produced by a previous signature being added.

handel/src/todo.rs
```diff
@@ -102,22 +99,15 @@ where
             list: HashSet::new(),
             evaluator,
             input_stream,
-            waker: None,
         }
     }
 
+    /// Safe without a wake as it is only called in the constructor of Aggregation.
```
hrxi, line 107: Could you add a member variable that tracks whether it has ever been polled, and add an assert here that this hasn't happened? This would ensure that this is robust in case of future code modification.

handel/src/protocol.rs
```diff
@@ -9,11 +9,15 @@ use crate::{
     identity::{IdentityRegistry, WeightRegistry},
     partitioner::Partitioner,
     store::ContributionStore,
-    verifier::{VerificationResult, Verifier},
+    verifier::Verifier,
 };
 
 #[async_trait]
```
hrxi: The `#[async_trait]` can go, this trait no longer contains any `async` functions.

handel/src/evaluator.rs
```diff
@@ -256,13 +247,18 @@ where
...
+        if msg.level as usize == self.partitioner.levels() {
+            let weight = self
+                .weights
+                .signers_identity(&msg.aggregate.contributors())
+                .len();
+            // Only available to full aggregations
+            return weight == self.partitioner.size();
         }
-        let range = self.partitioner.range(level);
+        let range = self.partitioner.range(msg.level as usize);
```
hrxi, line 261: The level should probably not be passed as an `usize` (it has nothing to do with the host architecture's pointer bit width), but I guess it doesn't have to be in this pull request.

handel/src/aggregation.rs
```diff
@@ -326,100 +332,190 @@ impl<
...
+where
+    TId: Debug + Clone + Unpin + Send + 'static,
+    P: Protocol<TId>,
+    N: Network<Contribution = P::Contribution>,
+{
+    type Item = P::Contribution;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        // Check if the automatic update interval triggers, if so perform the update.
+        if let Poll::Ready(_instant) = self.periodic_update_interval.as_mut().poll_tick(cx) {
```
hrxi:
```suggestion
        while let Poll::Ready(_instant) = self.periodic_update_interval.as_mut().poll_tick(cx) {
```
This needs to be a `while let`, otherwise the interval doesn't register itself with the waker if `Poll::Ready` is returned: https://docs.rs/tokio/1.38.0/tokio/time/struct.Interval.html#method.poll_tick.

handel/src/aggregation.rs
```diff
@@ -326,100 +332,190 @@ impl<
...
+            // This creates new messages in the sender.
+            self.automatic_update();
+        }
+
+        if let Poll::Ready(_instant) = self.start_level_interval.as_mut().poll_tick(cx) {
```
hrxi:
```suggestion
        while let Poll::Ready(_instant) = self.start_level_interval.as_mut().poll_tick(cx) {
```

handel/src/aggregation.rs
```diff
@@ -326,100 +332,190 @@ impl<
...
+                } else {
+                    break;
+                }
             }
         }
-    }
 
-    fn into_inner(self) -> (LevelUpdateStream<P, TId>, LevelUpdateSender<N>) {
-        (self.todos.into_stream(), self.sender)
+        // Poll sender to until it is pending sending as many messages as possible.
```
hrxi:
```suggestion
        // Poll sender until it is pending, sending as many messages as possible.
```

handel/src/aggregation.rs
```diff
@@ -326,100 +332,190 @@ impl<
...
+        while let Poll::Ready(t) = self.sender.poll_next_unpin(cx) {
+            t.expect("Sender should never return None");
+        }
```
hrxi: It seems that the `LevelUpdateSender` can never return `Poll::Ready`

```suggestion
        assert!(self.sender.poll_next_unpin(cx).is_pending());
```

Since that is the case, maybe it shouldn't implement `Stream`, but rather `Future`.

handel/src/aggregation.rs
```diff
@@ -430,98 +526,132 @@ impl<
...
+                log::error!("Failed to poll input stream; None returned")
             }
-
-            let response = (self.level_update.clone(), msg.origin());
-            self.sender.send(response);
         }
 
-        Poll::Ready((self.level_update.aggregate.clone(), None))
+        // Send as many messages as possible.
+        let _ = self.sender.poll_next_unpin(cx);
```
hrxi:
```suggestion
        assert!(self.sender.poll_next_unpin(cx).is_pending());
```

handel/src/aggregation.rs
```diff
@@ -430,98 +526,132 @@ impl<
...
     P: Protocol<TId>,
     N: Network<Contribution = P::Contribution>,
-> {
-    next_aggregation:
-        Option<BoxFuture<'static, (P::Contribution, Option<NextAggregation<TId, P, N>>)>>,
+{
+    /// The aggregation is currently ongoing and it is going to produce stream items as well as network traffic.
+    Ongoing(OngoingAggregation<TId, P, N>),
+    /// The aggregation is finished and it will no longer produce stream items and it will reduce network traffic
+    /// to answering messages with an unfinished aggregate within them
```
hrxi:
```suggestion
    /// to answering messages with an unfinished aggregate within them.
```

handel/src/aggregation.rs
```diff
@@ -430,98 +526,132 @@ impl<
...
-            Some(next_aggregation) => next_aggregation,
-            // If there is None that means all signatories have signed the payload so there is nothing more to aggregate after the last returned value.
-            // Thus return Poll::Ready(None) as the previous value can no longer be improved.
-            None => return Poll::Ready(None),
+        // Poll either the ongoing or the finished aggregation
+        let result = match &mut *self {
+            Self::Transitioning => {
+                unreachable!("Aggregation should never be transitioning when polled.")
+            }
+            Self::Finished(ref mut finished_aggregation) => {
```
hrxi:
```suggestion
            Self::Finished(finished_aggregation) => {
```
Here and below, I thought the `ref mut` is unnecessary, but maybe I'm wrong.

handel/src/aggregation.rs
```diff
@@ -430,98 +526,132 @@ impl<
...
+                // Finished aggregations are simply polled as the future will never produce a value.
+                return finished_aggregation.poll_unpin(cx);
```
hrxi:
```suggestion
                assert!(finished_aggregation.poll_unpin(cx).is_pending());
                return Poll::Pending;
```

handel/src/aggregation.rs
```diff
@@ -430,98 +526,132 @@ impl<
...
+            }
+            Self::Ongoing(ref mut ongoing_aggregation) => {
```
hrxi:
```suggestion
            Self::Ongoing(ongoing_aggregation) => {
```
(maybe incorrect.)

handel/Cargo.toml
```diff
@@ -22,8 +22,6 @@ parking_lot = "0.12"
 rand = "0.8"
 serde = "1.0"
 thiserror = "1.0"
-tokio = { version = "1.38", features = ["macros", "rt", "time"] }
-tokio-stream = "0.1"
```
hrxi, line 26: 🎉

network-mock/src/lib.rs
```diff
@@ -31,7 +31,7 @@ pub struct MockAddress(u64);
     From,
     Into,
 )]
-pub struct MockPeerId(u64);
+pub struct MockPeerId(pub u64);
```
hrxi, line 34: The `pub` seems unnecessary in this commit.
```suggestion
pub struct MockPeerId(u64);
```

handel/tests/mod.rs
```diff
@@ -330,7 +338,14 @@ async fn it_can_aggregate() {
     net.disconnect();
 
     // give the other aggregations time to complete themselves
-    nimiq_time::sleep(Duration::from_millis(100)).await;
```
hrxi, line 328: 🎉

handel/src/aggregation.rs
```diff
@@ -327,8 +314,11 @@ where
             // next time the timeout triggers the next level needs activating
             self.next_level_timeout += 1;
 
+            let store_rw = self.protocol.store();
+            let store = store_rw.read();
+
             // finally start the level.
-            self.start_level(level);
+            self.start_level(level, &*store);
```
hrxi:
```suggestion
            self.start_level(level, &store);
```
The dereference is likely unnecessary.

handel/src/aggregation.rs
```diff
@@ -346,30 +336,26 @@ where
...
+        store.put(
+            todo.contribution.clone(),
+            todo.level,
+            self.protocol.registry(),
+            self.protocol.identify(),
+        );
 
         // in case the level of this todo has not started, start it now as we have already contributions on it.
-        self.start_level(todo.level);
+        self.start_level(todo.level, &*store);
```
hrxi:
```suggestion
        self.start_level(todo.level, &store);
```

handel/src/aggregation.rs
```diff
@@ -346,30 +336,26 @@ where
...
         // check if a level was completed by the addition of the contribution
-        self.check_completed_level(todo.level);
+        self.check_completed_level(todo.level, &*store);
```
hrxi:
```suggestion
        self.check_completed_level(todo.level, &store);
```

primitives/block/Cargo.toml
```diff
@@ -43,5 +42,6 @@ nimiq-utils = { workspace = true, features = ["merkle"] }
 nimiq-vrf = { workspace = true, features = ["serde-derive"] }
 
 [dev-dependencies]
+nimiq-handel = { workspace = true }
```
hrxi, line 45: 🎉

handel/src/aggregation.rs
```diff
@@ -435,6 +423,18 @@ where
             None
         };
 
+        // Check if the automatic update interval triggers, if so perform the update.
+        if let Poll::Ready(_instant) = self.periodic_update_interval.as_mut().poll_tick(cx) {
```
hrxi:
```suggestion
        while let Poll::Ready(_instant) = self.periodic_update_interval.as_mut().poll_tick(cx) {
```

handel/src/aggregation.rs
```diff
@@ -435,6 +423,18 @@ where
...
+            // This creates new messages in the sender.
+            self.automatic_update();
+        }
+
+        if let Poll::Ready(_instant) = self.start_level_interval.as_mut().poll_tick(cx) {
```
hrxi:
```suggestion
        while let Poll::Ready(_instant) = self.start_level_interval.as_mut().poll_tick(cx) {
```

hrxi (commented):
The general changes look very good to me. :)

I found a couple of minor issues.

handel/src/evaluator.rs
nibhar: I believe that is going to be the case in a couple more places. Would you say this would be more of a cosmetic change or do you see a more functional benefit?

I would do it later if it is considered beneficial.

handel/src/todo.rs
nibhar: Good call! I add it.

handel/src/aggregation.rs
nibhar: Good find! I made it a Future instead. I also added the asserts.

network-mock/src/lib.rs
nibhar: The test uses it to create a specific Peer to send a message to. I think without it being pub it cannot be used like so:
https://github.com/nimiq/core-rs-albatross/pull/2693/files#diff-026455fd64af083e01fb8f83c0d5430cd7e37ba6096fa32800f3a08a022f2deeR192

handel/src/aggregation.rs
nibhar: Works 👍!

handel/src/aggregation.rs
nibhar: I think these are duplicates with an above comment from a prior commit?

handel/src/aggregation.rs
nibhar: I think these are duplicates with an above comment from a prior commit?

handel/src/evaluator.rs
hrxi: Cosmetic. Our levels don't get large enough to get in the range of the platform differences.

network-mock/src/lib.rs
hrxi: Hmmm. That test kinda confuses `MockPeerId`s and handel's IDs. There's probably a mapping between them somewhere, no?

hrxi: Ah, it manually sets up the peer IDs for that to work.

You can go from `u64` to `MockPeerId` using `MockPeerId::from`.

hrxi (approved):
Nice changes! They look good to me now. :)

PR #2699, opened 2024-06-26
Update state sync chunk max size

### LINKED ISSUE #2542: Set `state_chunks_max_size`
https://github.com/nimiq/core-rs-albatross/blob/3e08cee265122eb58eebfe0f84090a272f674355/primitives/src/policy.rs#L677

viquezclaudio:
Fixes #2542

### DIFF
--- a/primitives/src/policy.rs
+++ b/primitives/src/policy.rs
@@ -674,7 +674,7 @@ impl Default for Policy {
             batches_per_epoch: 720,
             tendermint_timeout_init: 1000,
             tendermint_timeout_delta: 1000,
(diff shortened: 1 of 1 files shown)


viquezclaudio:
This number was updated according to testing using the devnet scenarios, measuring the time it takes to finish state sync for various state chunk max sizes

viquezclaudio:
200 (as it was) is considered the baseline and in some cases it took more than 100 seconds to sync two batches full of transactions.
With 1000 and 2000, it takes around ~15s, with 2000 taking a bit longer in some cases
With 10000, on multiple occasions, it took more than 100s to sync two batches.

jsdanielh (approved):

PR #2702, opened 2024-06-26
LiveSync: move `BlockQueue` to own task

### LINKED ISSUE #2624: Refactor Consensus task structure

styppo:
Move `BlockQueue` to its own task such that it is polled independently of the `Consensus` task. This ensures that network events are processed immediately and not blocked by push operations.

Closes #2624


consensus/tests/state_live_sync.rs
```diff
@@ -939,6 +956,7 @@ async fn clears_buffer_after_macro_block() {
         ),
         "Should accept chunks"
     );
+    yield_now().await;
```
hrxi, line 959: I don't think it's a good idea to rely on this function for our tests: https://docs.rs/tokio/1.38.0/tokio/task/fn.yield_now.html says that it's not guaranteed that the runtime will schedule other tasks on a `yield_now`. Could we maybe instead move this test to a multithreaded runtime?

PR #2704, opened 2024-06-27
Smaller Updates to Handel

nibhar:
While doing some code review with @hrxi we found a couple of minor changes we wanted to make. None of these should affect how the code executes with the exception of the early return which should just make code which executed unnecessarily no longer execute.

### DIFF
--- a/handel/src/store.rs
+++ b/handel/src/store.rs
@@ -17,6 +17,7 @@ where
     Self: Send + Sync,
 {
     /// Put `signature` into the store for level `level`.
+    /// In the case of individual signatures they must not already exist.
     fn put(
         &mut self,
         contribution: TProtocol::Contribution,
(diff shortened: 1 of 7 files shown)


handel/src/store.rs
```diff
@@ -230,20 +230,21 @@ where
                 .get_mut(level)
                 .unwrap_or_else(|| panic!("Missing Level {level}"))
                 .combine(&identity, false);
+            // ggf expect() here
```
hrxi: Leftover comment.

handel/src/store.rs
```diff
@@ -230,20 +230,21 @@ where
...
 
             self.individual_contributions
                 .get_mut(level)
                 .unwrap_or_else(|| panic!("Missing Level {level}"))
                 .insert(identity, contribution.clone());
+            // ggf expect ggf check for change
```
hrxi: Leftover comment.

Eligioo (approved):

handel/src/store.rs
```diff
@@ -159,7 +165,10 @@ impl<TId: Clone + std::fmt::Debug + 'static, TProtocol: Protocol<TId>>
                 contributors
             }
         } else {
-            registry.signers_identity(&contribution.contributors())
+            // Combining worked meaning the contributions did not overlap. Since individuals should already
+            // have been present in the previous best this the best signature available.
```
jsdanielh:
```suggestion
            // have been present in the previous best this is the best signature available.
```

hrxi: One more word to the left.

ii-cruz (approved):
LGTM

PR #2653, opened 2024-06-20
Set argon2id params

hrxi:
The values are taken from
https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#argon2id.

Fixes #2472.

Thanks to @Eligioo for finding these values.

PR #2377, opened 2024-04-17
Use argon2id by default

paberr:
## What's in this pull request?
Replaces `argon2d` with `argon2id` while remaining backwards compatible with existing wallets.
`argon2id` has additional side-channel protection.

## Pull request checklist

- [x] All tests pass. The project builds and runs.
- [x] I have resolved any merge conflicts.
- [x] I have resolved all `clippy` and `rustfmt` warnings.

PR #2705, opened 2024-06-27
Use PoW memory cost for Argon2d KDF

sisou:
But keep the updated cost for Argon2id.

This makes the function backward-compatible with the PoW JS library, to support decoding existing LoginFiles.

Partly reverses 9f8db328a368c689bb5a1366f92d39da52e5d2f4 (#2653) and changes it to be in the spirit of #2377, which meant to preserve backward-compatibility.

### DIFF
--- a/hash/src/argon2kdf.rs
+++ b/hash/src/argon2kdf.rs
@@ -1,8 +1,10 @@
 use argon2::Config;
 pub use argon2::{Error as Argon2Error, Variant as Argon2Variant};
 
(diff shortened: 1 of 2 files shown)


hash/src/argon2kdf.rs
```diff
@@ -14,7 +16,10 @@ pub fn compute_argon2_kdf(
     let config = Config {
         time_cost: iterations,
         hash_length: derived_key_length as u32,
-        mem_cost: MEMORY_COST,
+        mem_cost: match variant {
+            Argon2Variant::Argon2d => MEMORY_COST_ARGON2D,
+            _ => MEMORY_COST_ARGON2ID,
```
Eligioo:
```suggestion
            Argon2Variant::Argon2i | Argon2Variant::Argon2id => MEMORY_COST_ARGON2ID,
```

hrxi (approved):

PR #2715, opened 2024-06-27
web-client: Enable HD wallet key derivation

sisou:
Add classes/structs and methods to support HD wallet key derivation as was possible with the core-js library.

- [ ] Review of encryption/decryption methods by @paberr and/or @hrxi for best practices
- [x] Move of CryptoUtils/hashing methods into their own worker so they don't block the main thread


key-derivation/src/lib.rs
```diff
@@ -5,8 +5,9 @@ use nimiq_hash::{hmac::*, sha512::Sha512Hash};
 use nimiq_keys::{Address, Ed25519PublicKey, PrivateKey};
 use nimiq_serde::Serialize;
 use regex::Regex;
+use serde::Deserialize;
```
Eligioo: Instead of using `serde::Deserialize`, shouldn't we use `nimiq_serde::Deserialize`?

mnemonic/src/lib.rs
```diff
@@ -58,6 +63,106 @@ impl Entropy {
...
+    pub fn from_encrypted(buf: &[u8], key: &[u8]) -> Result<Entropy, String> {
+        let version = buf[0];
+        let rounds_log = buf[1];
+        if rounds_log > 32 {
+            return Err(format!("Rounds out-of-bounds: 2^{}", rounds_log));
+        }
+        let rounds = 2u32.pow(rounds_log as u32);
+
+        match version {
+            1 => unimplemented!(),
```
Eligioo, line 127: Are we going to fully discard older versions?

web-client/example/README.md
```diff
@@ -24,7 +24,7 @@ to build the WASM files. Then go into `example/node` and run
 yarn
 ```
 
-to install the dependencies. Other package managers will also work, but the `preinstall` script uses `yarn`.
+to install the dependencies. Other package managers will also work.
```
Eligioo, line 27: Yarn isn't mandatory all together anymore?

Eligioo (commented):
Halfway through the review, already some comments.

web-client/example/README.md
sisou: Not here, as I removed the preinstall script (below). So packages can be installed with any manager for the example.

web-client/src/primitives/es256_signature.rs
```diff
@@ -17,24 +17,22 @@ impl ES256Signature {
     /// Deserializes an ES256 signature from a byte array.
     ///
     /// Throws when the byte array contains less than 64 bytes.
-    #[wasm_bindgen(js_name = fromBytes)]
-    pub fn deserialize(bytes: &[u8]) -> Result<ES256Signature, JsError> {
+    pub fn unserialize(bytes: &[u8]) -> Result<ES256Signature, JsError> {
```
Eligioo: I assume you switched from `deserialize` to `unserialize` to stay closer to the core-js PoW naming convention? 

Though throughout the rest of the Rust codebase, `deserialize` is the naming convention for deserializing bytes. And as the `web-client` is breaking anyway compared to the PoW core-js, imo this is the right time to use a different naming convention. Then I would be in favor of using `deserialize` over `unserialize`.

web-client/extras/launcher/cryptoutils-worker-proxy.ts
```diff
@@ -0,0 +1,60 @@
+import type { Remote } from 'comlink';
+
+interface CryptoUtilsWorker {
+    otpKdf(message: Uint8Array, key: Uint8Array, salt: Uint8Array, iterations: number): Promise<Uint8Array>;
+}
+
+let remote: Remote<CryptoUtilsWorker> | undefined;
+
+// class CryptoUtilsProxy {}
```
Eligioo: Forgot to remove?

web-client/extras/lib/Secret.ts
```diff
@@ -0,0 +1,154 @@
+import { CryptoUtils, Hash, PrivateKey } from '@nimiq/core';
+import { BufferUtils } from './BufferUtils';
+import { Entropy } from './Entropy';
+import { SerialBuffer } from './SerialBuffer';
+import { Serializable } from './Serializable';
+
+abstract class Secret extends Serializable {
```
Eligioo, line 7: Would it be an idea to also port over the tests? For the `ExtendedPrivateKey`, `Secret`, etc.?

mnemonic/src/lib.rs
sisou: The older versions use a KDF OTP function that core-rs-albatross no longer implements.

web-client/extras/lib/Secret.ts
sisou: I can try, but then we need a JS test-runner, too :/

sisou: Added the relevant tests with `vitest` :muscle:

sisou:
@jsdanielh Rebased and squashed a few earlier commit together.

Eligioo:
Has this been further reviewed?

PR #2720, opened 2024-06-28
Use only primitive IEEE 754 features in supply curve calculations

hrxi:
Additionally use a relatively simple square-and-multiply exponentiation.

The key realization is that `x ** (y * z)` is equal to `(x ** y) ** z`. We can use that to only require calculating integer powers of floating point numbers, but that can be done with the square-and-multiply algorithm.

We find that `exp(-1.1e-9) ≈ 0.9999999989` and `2 ** -5.7327557121556496e-12 ≈ 0.9999999999960264`, both as exact as `f64` (confirmed with WolframAlpha).

The results of the supply curve differ only by less than 0.001%

https://en.wikipedia.org/w/index.php?title=Exponentiation_by_squaring&oldid=1229001691#With_constant_auxiliary_memory


primitives/src/policy.rs
```diff
@@ -128,6 +127,7 @@ impl Policy {
     /// The supply decay is a constant that is calculated so that the supply velocity decreases at a
     /// the velocity of the PoW chain supply curve.
     pub const SUPPLY_DECAY: f64 = 5.7327557121556496e-12;
+    pub const SUPPLY_DECAY_BASE: f64 = 0.9999999999960264;
```
paberr: I would add a comment on what these variables mean/how they are related.

primitives/src/policy.rs
```diff
@@ -94,8 +94,7 @@ impl Policy {
     /// system time. We only care about drifting to the future.
     pub const TIMESTAMP_MAX_DRIFT: u64 = 600000;
 
-    /// The slope of the exponential decay used to punish validators for not producing block in time
-    pub const BLOCKS_DELAY_DECAY: f64 = 1.1e-9;
+    pub const BLOCKS_DELAY_DECAY_BASE: f64 = 0.9999999989;
```
paberr: I would add a comment on what these variables mean/how they are related.

paberr (commented):
Generally LGTM

PR #2722, opened 2024-07-01
Propagate block headers and bodies through separate gossipsub topics

### LINKED ISSUE #2499: Light clients get send broadcasted BlockHeaders multiple times
After the merge of #2343 light clients now receive the block headers from basically every single full node they are connected with. That is unnecessary and leads to unwanted network traffic.

### LINKED ISSUE #2560: Don't publish on header topic until consensus has been established

styppo:
The current gossipsub topic structure for block propagation suffers from poor mesh quality on the `block-header` topic, duplicated messages due to re-publishing as well as data duplication across topics.

This PR replaces the `block`/`block-header` gossipsub topics by a `block-header` and a `block-body` topic, eliminating cross topic data duplication. All nodes subscribe to the `block-header` topic, improving mesh quality, while only full/history nodes subscribe to the `block-body` topic.

Additionally, `validators` and `next_batch_initial_punished_set` are moved from the macro body to the macro header while preserving ZKP circuit compatibility. These properties are always required by light nodes, so they need to be part of the header.

Finally, block source tracking was refactored to account for the fact that a single block can now have two pubsub ids associated with it.

Closes #2499, closes #2560

### DIFF
--- a/blockchain/src/block_production/mod.rs
+++ b/blockchain/src/block_production/mod.rs
@@ -266,6 +266,22 @@ impl BlockProducer {
             .seed()
             .sign_next_with_rng(&self.signing_key, rng);
 
+        // If this is an election block, calculate the validator set for the next epoch.
+        let validators = match Policy::is_election_block_at(block_number) {
+            true => Some(blockchain.next_validators(&seed)),
+            false => None,
+        };
+
+        // Get the staking contract PRIOR to any state changes.
+        let staking_contract = blockchain
+            .get_staking_contract_if_complete(None)
+            .expect("Staking Contract must be complete to create a macro body");
+
+        // Calculate the punished set for the next batch.
+        let next_batch_initial_punished_set = staking_contract
+            .punished_slots
+            .next_batch_initial_punished_set(block_number, &staking_contract.active_validators);
+
         // Create the header for the macro block without the state root and the transactions root.
         // We need several fields of this header in order to calculate the transactions and the
         // state.
@@ -284,6 +300,8 @@ impl BlockProducer {
             body_root: Blake2sHash::default(),
             diff_root: Blake2bHash::default(),
             history_root: Blake2bHash::default(),
+            validators,
+            next_batch_initial_punished_set,
         };
 
         // Create the body for the macro block.
@@ -345,28 +363,12 @@ impl BlockProducer {
             .get_staking_contract_if_complete(txn_option)
             .expect("Staking Contract must be complete to create a macro body");
 
-        // Calculate the punished set for the next batch.
-        let next_batch_initial_punished_set = staking_contract
-            .punished_slots
-            .next_batch_initial_punished_set(
-                macro_header.block_number,
-                &staking_contract.active_validators,
-            );
-
         // Calculate the reward transactions.
         let reward_transactions =
             blockchain.create_reward_transactions(macro_header, &staking_contract);
 
-        // If this is an election block, calculate the validator set for the next epoch.
-        let validators = match Policy::is_election_block_at(macro_header.block_number) {
-            true => Some(blockchain.next_validators(&macro_header.seed)),
-            false => None,
-        };
-
         // Create the body for the macro block.
         MacroBody {
-            validators,
-            next_batch_initial_punished_set,
             transactions: reward_transactions,
         }
     }
--- a/consensus/src/consensus/consensus_proxy.rs
+++ b/consensus/src/consensus/consensus_proxy.rs
@@ -603,7 +603,7 @@ impl<N: Network> ConsensusProxy<N> {
         self,
         block_number: u32,
         block_hash: Blake2bHash,
-        pubsub_id: N::PubsubId,
+        first_peer_id: N::PeerId,
     ) -> Result<Block, ResolveBlockError<N>> {
         // Create the oneshot sender whose receiver this fn will await and whose
         // sender will be given to the consensus proper to resolve the call.
@@ -613,7 +613,7 @@ impl<N: Network> ConsensusProxy<N> {
         let request = ResolveBlockRequest {
             block_number,
             block_hash,
-            pubsub_id,
+            first_peer_id,
             response_sender,
         };
 
--- a/consensus/src/consensus/consensus_proxy.rs
+++ b/consensus/src/consensus/consensus_proxy.rs
@@ -592,10 +592,7 @@ impl<N: Network> ConsensusProxy<N> {
     }
 
     /// Attempts to resolve a block with `block_hash` header hash at the given `block_height`.
-    /// The `pubsub_id` must contain the pubsup id containing the information which referenced
-    /// the block being resolved here. One example is the predecessor, other could be possible.
-    /// The peers present in the id will be asked for the information as they should be able to
-    /// produce it.
+    /// The first resolution attempt is performed with the peer specified by `first_peer_id`.
     ///
     /// This function fails, if the consensus cannot accept more requests or if the consensus drops
     /// the request on its side, generally indicating that it is no longer of use.
--- a/consensus/src/messages/mod.rs
+++ b/consensus/src/messages/mod.rs
@@ -148,7 +148,7 @@ impl Topic for BlockBodyTopic {
 
     const BUFFER_SIZE: usize = 16;
     const NAME: &'static str = "block-body";
-    const VALIDATE: bool = false;
+    const VALIDATE: bool = true;
 }
 
 /*
@@ -361,11 +361,8 @@ pub enum HistoryChunkError {
 pub struct RequestBlock {
     /// The hash of the block that is requested.
     pub hash: Blake2bHash,
-    /// Whether to include the body if the requested block is a micro block.
-    ///
-    /// If the requested block is a macro block, this flag has no effect and
-    /// the response will always contain the body.
-    pub include_micro_bodies: bool,
+    /// Whether to include the body.
+    pub include_body: bool,
 }
 
 /// Error response for [`RequestBlock`].
@@ -442,10 +439,8 @@ impl Debug for ResponseBlocks {
 pub struct RequestMissingBlocks {
     /// Target block hash.
     pub target_hash: Blake2bHash,
-    /// Whether to include bodies for micro blocks.
-    ///
-    /// Bodies of macro blocks are always included.
-    pub include_micro_bodies: bool,
+    /// Whether to include block bodies.
+    pub include_body: bool,
     /// Set of hashes of blocks known to the requester, possible starts of the
     /// answer chain.
     ///
--- a/consensus/src/messages/mod.rs
+++ b/consensus/src/messages/mod.rs
@@ -1,10 +1,16 @@
-use std::fmt::{Debug, Formatter};
+use std::{
+    fmt::{Debug, Display, Formatter},
+    io::Write,
+};
 
-use nimiq_block::{Block, BlockInclusionProof, MacroBlock};
+use nimiq_block::{
+    Block, BlockBody, BlockInclusionProof, BlockType, MacroBlock, MacroHeader, MicroBlock,
+    MicroHeader, MicroJustification, TendermintProof,
+};
 #[cfg(feature = "full")]
 use nimiq_blockchain::HistoryTreeChunk;
 use nimiq_blockchain_interface::Direction;
-use nimiq_hash::Blake2bHash;
+use nimiq_hash::{Blake2bHash, Blake2sHash, Hash, HashOutput, SerializeContent};
 use nimiq_keys::Address;
 use nimiq_macros::test_max_req_size;
 use nimiq_mmr::mmr::proof::SizeProof;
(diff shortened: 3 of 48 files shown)


blockchain/src/block_production/mod.rs
```diff
@@ -284,6 +300,8 @@ impl BlockProducer {
             body_root: Blake2sHash::default(),
             diff_root: Blake2bHash::default(),
             history_root: Blake2bHash::default(),
+            validators,
+            next_batch_initial_punished_set,
         };
 
         // Create the body for the macro block.
```
nibhar, line 307: `staking_contract` could be reused in the call below this line, to not have to get it twice, as `next_macro_body(...)` also needs it.

consensus/src/messages/mod.rs
```diff
@@ -23,6 +29,128 @@ use crate::error::SubscribeToAddressesError;
...
+    }
+
+    pub fn body_root(&self) -> &Blake2sHash {
+        match self {
+            BlockHeaderMessage::Macro { header, .. } => &header.body_root,
+            BlockHeaderMessage::Micro { header, .. } => &header.body_root,
+        }
+    }
+
+    pub fn split_block(block: Block) -> (BlockHeaderMessage, BlockBodyMessage) {
```
nibhar, line 59: To me this seems more suited for `Block::split_block(self) -> (BlockHeaderMessage, BlockBodyMessage)`, than here.

consensus/src/messages/mod.rs
```diff
@@ -23,6 +29,128 @@ use crate::error::SubscribeToAddressesError;
...
+
+#[derive(Clone, Debug, Default)]
+pub struct BlockBodyTopic;
+
+impl Topic for BlockBodyTopic {
+    type Item = BlockBodyMessage;
+
+    const BUFFER_SIZE: usize = 16;
+    const NAME: &'static str = "block-body";
+    const VALIDATE: bool = false;
```
nibhar: This is weird to me, why would we not validate Block bodies?

consensus/src/sync/live/block_queue/assembler.rs
```diff
@@ -0,0 +1,329 @@
...
+
+        Poll::Pending
+    }
+}
+
+struct TimeLimitedCache<K, V> {
+    map: HashMap<K, (V, Instant)>,
+    ttl: Duration,
+    interval: Interval,
+}
```
nibhar, line 196: I would prefer this moved to e.g `nimiq-collections` as it could be used anywhere.

consensus/src/sync/live/block_queue/assembler.rs
```diff
@@ -0,0 +1,329 @@
...
+
+            if let Some(body) = self.cached_bodies.remove(&body_key) {
+                // Check that header and body type match.
+                if header.0.ty() != body.ty() {
+                    debug!(
+                        block = %header.0,
+                        peer_id = %header.1.propagation_source(),
+                        "Discarding block - header and body types don't match"
+                    );
+                    // TODO ban peer
```
nibhar: Todo.

consensus/src/sync/live/block_queue/assembler.rs
```diff
@@ -0,0 +1,329 @@
...
+use nimiq_block::{Block, BlockBody, MacroBlock, MicroBlock};
+use nimiq_hash::{Blake2bHash, Blake2sHash, Hash};
+use nimiq_network_interface::network::{Network, PubsubId};
+use nimiq_time::{interval, Interval};
+
+use crate::messages::{BlockBodyMessage, BlockHeaderMessage};
+
+type PubsubHeader<N> = (BlockHeaderMessage, <N as Network>::PubsubId);
+type PubsubBody<N> = (BlockBodyMessage, <N as Network>::PubsubId);
+type BodyKey = (Blake2sHash, Blake2bHash);
```
nibhar:
```suggestion
/// First is the body root and second is the header hash.
type BodyKey = (Blake2sHash, Blake2bHash);
```
or similar.

consensus/src/sync/live/block_queue/assembler.rs
```diff
@@ -0,0 +1,329 @@
...
+    type Item = (Block, N::PubsubId);
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        // The cache stream never returns None, so it's ok to ignore that case here.
+        while let Poll::Ready(Some(num_evicted)) = self.cached_headers.poll_next_unpin(cx) {
+            debug!(num_evicted, "Evicted {} headers from cache", num_evicted);
+        }
+        while let Poll::Ready(Some(num_evicted)) = self.cached_bodies.poll_next_unpin(cx) {
+            debug!(num_evicted, "Evicted {} bodies from cache", num_evicted);
+        }
```
nibhar: To me it seems to make more sense to evict last. Even after the early returns the `BlockAssembler` should be polled again and return `Poll::Pending` eventually and before that it could evict.

consensus/src/sync/live/block_queue/assembler.rs
```diff
@@ -0,0 +1,329 @@
...
+    }
+}
+
+impl<N: Network> Stream for BlockAssembler<N> {
+    type Item = (Block, N::PubsubId);
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        // The cache stream never returns None, so it's ok to ignore that case here.
+        while let Poll::Ready(Some(num_evicted)) = self.cached_headers.poll_next_unpin(cx) {
```
nibhar: This does drop the header and thus the network should validate the message as Ignored. Same for the bodies if they are validated.

consensus/src/sync/live/block_queue/assembler.rs
```diff
@@ -0,0 +1,329 @@
...
+                    let header = self.cached_headers.remove(&body.0.header_hash).unwrap();
+
+                    // Check that header and body type match.
+                    if header.0.ty() != body.0.body.ty() {
+                        debug!(
+                            block = %header.0,
+                            peer_id = %header.1.propagation_source(),
+                            "Discarding block - header and body types don't match"
+                        );
+                        // TODO ban peer
```
nibhar: Todo

consensus/src/sync/live/block_queue/assembler.rs
```diff
@@ -0,0 +1,329 @@
...
+
+                    return Poll::Ready(Some((
+                        Self::assemble_block(header.0, body.0.body),
+                        header.1,
+                    )));
+                }
+            }
+
+            let body_key = (hash, body.0.header_hash);
+            self.cached_bodies.insert(body_key, body.0.body);
```
nibhar: I think this is an unbounded buffer on a non validated topic. Obviously it has a time bounded component to it, but still this is not ideal I would say.

consensus/src/sync/live/queue.rs
```diff
@@ -472,7 +474,8 @@ fn validate_message<N: Network>(
     network: Arc<N>,
     pubsub_id: Option<N::PubsubId>,
     block_push_result: &Option<Result<PushResult, PushError>>,
-    include_body: bool,
+    // TODO Remove this
```
nibhar: Todo

primitives/block/src/macro_block.rs
```diff
@@ -197,12 +197,61 @@ pub struct MacroHeader {
...
+        let punished_set_hash = self
+            .next_batch_initial_punished_set
+            .serialize_to_vec()
+            .hash::<Blake2sHash>();
+        punished_set_hash.serialize_to_writer(writer)?;
+
+        self.body_root.serialize_to_writer(writer)?;
+
+        Ok(())
+    }
```
nibhar: How long will these two be necessary? I suspect forever because of the zkp setup and then I would maybe consider naming them as such, to not make it seem like they can be removed at some later time.

blockchain/src/block_production/mod.rs
styppo: True, but also minor optimization not really related to this PR

consensus/src/messages/mod.rs
styppo: I agree that it would be nicer, but it would introduce a dependency from `Block` to `BlockHeaderMessage` and `BlockBodyMessage` which are defined in the `consensus` crate.

consensus/src/messages/mod.rs
styppo: They are now validated as well with the new block source tracking.

consensus/src/sync/live/block_queue/assembler.rs
styppo: This would add a dependency to `nimiq-collections`, so we can move it once we need it.

consensus/src/sync/live/block_queue/assembler.rs
styppo: Block body topic is now validated.

consensus/src/consensus/consensus_proxy.rs
```diff
@@ -603,7 +603,7 @@ impl<N: Network> ConsensusProxy<N> {
         self,
         block_number: u32,
         block_hash: Blake2bHash,
-        pubsub_id: N::PubsubId,
+        first_peer_id: N::PeerId,
```
Eligioo, line 603: Rustdoc still references `pubsub_id` instead of `first_peer_id`

consensus/src/messages/mod.rs
```diff
@@ -23,6 +29,128 @@ use crate::error::SubscribeToAddressesError;
...
+            BlockHeaderMessage::Micro { header, .. } => &header.body_root,
+        }
+    }
+
+    pub fn split_block(block: Block) -> (BlockHeaderMessage, BlockBodyMessage) {
+        match block {
+            Block::Macro(block) => {
+                let header = BlockHeaderMessage::Macro {
+                    header: block.header,
+                    justification: block.justification.unwrap(),
```
Eligioo: `expects` instead of `unwrap`s?

consensus/src/sync/live/block_queue/assembler.rs
```diff
@@ -0,0 +1,392 @@
...
+        }
+    }
+
+    fn reject_messages(&self, pubsub_id_header: N::PubsubId, pubsub_id_body: N::PubsubId) {
+        let network = Arc::clone(&self.network);
+        let peer_id_header = pubsub_id_header.propagation_source();
+        let peer_id_body = pubsub_id_body.propagation_source();
+        spawn(async move {
+            network
+                .disconnect_peer(peer_id_header, CloseReason::MaliciousPeer)
```
Eligioo, line 80: Add a log here that we are banning this peer including the reason

consensus/src/sync/live/block_queue/assembler.rs
```diff
@@ -0,0 +1,392 @@
...
+                .await;
+            if peer_id_header != peer_id_body {
+                network
+                    .disconnect_peer(peer_id_body, CloseReason::MaliciousPeer)
```
Eligioo, line 84: Same as above

primitives/block/src/block.rs
```diff
@@ -363,14 +339,15 @@ impl Block {
     }
 
     /// Updates validator keys from a public key cache.
+    // TODO remove this function
```
Eligioo, line 342: TODO

consensus/src/consensus/consensus_proxy.rs
styppo: Fixed

consensus/src/messages/mod.rs
styppo: Fixed

consensus/src/sync/live/block_queue/assembler.rs
styppo: A log already exists at the call site

consensus/src/sync/live/block_queue/assembler.rs
styppo: A log already exists at the call site

primitives/block/src/block.rs
styppo: Removing this function is out of scope for this PR

consensus/src/sync/live/block_queue/assembler.rs
Eligioo: At the call site we only log `debug!(%peer_id, "Banned peer")` but not where it actually is coming from or why we are banning this peer.

We do something similar at every other place where we disconnect from a peer with a `CloseReason::MaliciousPeer` reason:
https://github.com/nimiq/core-rs-albatross/blob/a287b24fb772756a595b89b77045125b743c2e06/consensus/src/sync/light/sync_requests.rs#L98-L105

https://github.com/nimiq/core-rs-albatross/blob/a287b24fb772756a595b89b77045125b743c2e06/consensus/src/sync/light/validity_window.rs#L375-L386

styppo: I was referring to the call site of the current function `BlockAssembler::reject_messages()`, which logs a debug message with the concrete reason [here](https://github.com/nimiq/core-rs-albatross/blob/styppo/block-gossip/consensus/src/sync/live/block_queue/assembler.rs#L112) and [here](https://github.com/nimiq/core-rs-albatross/blob/styppo/block-gossip/consensus/src/sync/live/block_queue/assembler.rs#L146).

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -1,35 +1,103 @@
 use futures::stream::BoxStream;
 use nimiq_block::Block;
-use nimiq_network_interface::network::Network;
+use nimiq_network_interface::network::{MsgAcceptance, Network, PubsubId};
+pub use proxy::BlockQueueProxy as BlockQueue;
```
nibhar, line 4: Seems out of place here. I liked it better where it was.

consensus/src/sync/live/block_queue/mod.rs
```diff
@@ -1,35 +1,103 @@
...
-    TooFarBehind(Block, N::PeerId),
+    Head(BlockAndSource<N>),
+    Buffered(Vec<BlockAndSource<N>>),
+    Missing(Vec<BlockAndSource<N>>),
+    TooFarAhead(N::PeerId),
+    TooFarBehind(N::PeerId),
+}
+
+#[derive(Debug)]
+pub enum BlockSource<N: Network> {
```
nibhar, line 34: I would have this in its own file alongside `BlockAndSource` for less clutter. Personal taste though.

consensus/tests/history.rs
```diff
@@ -637,7 +637,7 @@ async fn request_missing_blocks_across_macro_block() {
     // Run the block_queue one iteration, i.e. until it processed one batch.
     // Needs to be pulled again to send the request missing block over the network.
     syncer.next().await;
-    // syncer.next().await;
```
nibhar, line 640: Probably an oversight, but why is this no longer needed?

nibhar (approved):
LGTM

Just some very minor structural comments. I think this should be tested however with a prover node to make sure the zkp works.

consensus/tests/history.rs
sisou: This just removes a commented dublicate line. The actual active line stays.

nibhar: Clearly read that too fast... Thanks!

PR #2726, opened 2024-07-03
Make sure we don't overflow in reward calculation

hrxi:
This makes sure we don't hand out ~2**64 Luna when our reward curve isn't strictly monotone.

Our rewards curve should still be monotone as otherwise more than the total supply could be handed out.


sisou:
This seems like an important piece of code to have a test for :wink:

hrxi:
> This seems like an important piece of code to have a test for 😉

True…

I don't see an easy way to add a test there, because the reward curve is currently hardcoded (and it shouldn't be possible to find a point in time where my code matter). Perhaps we could pass in a custom reward curve to this function that exhibits this problematic behavior and then check that it correctly does it.

jsdanielh (approved):

PR #2734, opened 2024-07-10
Use a patched libp2p-websocket-websys crate to avoid use-after-free errors

sisou:
Use a patched `libp2p-websocket-websys` crate to avoid https://github.com/libp2p/rust-libp2p/issues/5490

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3206,8 +3206,7 @@ dependencies = [
 [[package]]
 name = "libp2p-websocket-websys"
 version = "0.3.2"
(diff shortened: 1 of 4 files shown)


Eligioo (approved):

hrxi:
Is there an upstream PR so that we can eventually go back to using an unpatched version?

sisou:
Not yet, I want to wait for the reaction to the issue, if my primitive fix is ok, or if we should do it proper.

jsdanielh (approved):

PR #2735, opened 2024-07-10
Simplify supply curve formula and continue smoothly after the upgrade

hrxi:


primitives/src/policy.rs
```diff
@@ -493,31 +486,31 @@ impl Policy {
...
+    /// ```text
+    /// supply(t) = total_supply - (total_supply - genesis_supply) * supply_decay^t
+    /// ```
+    /// Where t is the time in milliseconds since the PoS genesis block and `genesis_supply` is the supply at
     /// the genesis of the Nimiq 2.0 chain.
     #[cfg_attr(feature = "ts-types", wasm_bindgen(js_name = supplyAt))]
-    pub fn supply_at(genesis_supply: u64, current_time: u64) -> u64 {
-        assert!(current_time >= Self::POW_GENESIS_TIMESTAMP);
-
-        let t = current_time - Self::POW_GENESIS_TIMESTAMP;
```
sisou: You can remove the `POW_GENESIS_TIMESTAMP` constant now.

hrxi: Done.

sisou: It's used in other places in the code though, so this will now lead to a compile error.

rpc-server/src/dispatchers/policy.rs
```diff
@@ -166,6 +166,7 @@ impl PolicyInterface for PolicyDispatcher {
     async fn get_supply_at(
         &mut self,
         genesis_supply: u64,
+        genesis_time: u64,
         current_time: u64,
     ) -> RPCResult<u64, (), Self::Error> {
         if Policy::POW_GENESIS_TIMESTAMP > current_time {
```
sisou: E.g. here @hrxi

primitives/src/policy.rs
```diff
@@ -493,31 +483,31 @@ impl Policy {
...
+            .expect("current_time must be greater or equal to genesis_time");
 
-        let supply = genesis_supply
-            + (Self::INITIAL_SUPPLY_VELOCITY / Self::SUPPLY_DECAY
-                * (1.0 - powi(Self::SUPPLY_DECAY_BASE, t))) as u64;
-
-        cmp::min(supply, Policy::TOTAL_SUPPLY)
+        Policy::TOTAL_SUPPLY
+            - ((Policy::TOTAL_SUPPLY - genesis_supply) as f64 * powi(Policy::SUPPLY_DECAY, t))
+                as u64
```
sisou: Can you explain how the new formula works? It seems you are changing the calculation from a forward calculation (genesis supply + diff) to a backwards calculation (total supply - diff). How does that work, and what is `powi`?

hrxi: Only addressing a tiny part of your comment: `powi` is "to the power of", but implemented in a way that is not different between machines.

blockchain/tests/inherents.rs
```diff
@@ -84,12 +84,14 @@ fn it_can_create_batch_finalization_inherents() {
 
     let (validator_address, _) = active_validators.iter().next().unwrap();
 
+    const EXPECTED_REWARD: u64 = 166_810_895;
```
hrxi: This high number is expected, since in the unit test chain, we have an initial supply of 1.01e12 but a total supply of 2100e12. The remaining supply is 2098.99e+12, which is 500_438_213 shifted 22 to the right. Since the block time is 20 seconds, so we expect the reward to be roughly(!) one third of it, so 166_812_737. Hence, the value seems to be correct.

paberr (approved):

PR #2736, opened 2024-07-10
Tendermint impl use policy consts

Eligioo:
Currently the `TENDERMINT_TIMEOUT_DELTA` and `TENDERMINT_TIMEOUT_INIT` constants are defined at two places. This PR fixes this by letting the Tendermint impl use the constants that are defined in the Policy.

## Pull request checklist

### DIFF
--- a/primitives/src/policy.rs
+++ b/primitives/src/policy.rs
@@ -18,16 +18,6 @@ pub struct Policy {
     /// How many batches constitute an epoch
     #[cfg_attr(feature = "ts-types", wasm_bindgen(skip))]
     pub batches_per_epoch: u16,
-    /// Tendermint's initial timeout, in milliseconds.
-    ///
-    /// See <https://arxiv.org/abs/1807.04938v3> for more information.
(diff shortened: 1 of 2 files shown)


primitives/src/policy.rs
```diff
@@ -688,8 +666,6 @@ impl Default for Policy {
         Policy {
             blocks_per_batch: 60,
             batches_per_epoch: 720,
-            tendermint_timeout_init: 1000,
-            tendermint_timeout_delta: 1000,
```
sisou, line 692: They might have been here on purpose to make them more easily configurable for tests, etc.?

Eligioo: With these changes it will still be flexible as the actual values are set on the implementors of the `tendermint::Protocol` trait.

Here is an example: https://github.com/nimiq/core-rs-albatross/blob/albatross/tendermint/tests/common/mod.rs#L137-L138

jsdanielh (approved):

PR #2737, opened 2024-07-11
Get rid of unnecessary `Waker::wake_by_ref` calls

hrxi:
Add `WakerExt::wake` to conveniently call `wake` on `Option<Waker>`. Remove self-wake from `read_to_buf`.

### DIFF
--- a/consensus/src/sync/history/sync.rs
+++ b/consensus/src/sync/history/sync.rs
@@ -8,6 +8,7 @@ use futures::{future::BoxFuture, stream::FuturesUnordered, FutureExt};
 use nimiq_blockchain::Blockchain;
 use nimiq_hash::Blake2bHash;
 use nimiq_network_interface::network::{Network, SubscribeEvents};
+use nimiq_utils::WakerExt as _;
 use parking_lot::RwLock;
 
 use crate::{
@@ -119,8 +120,6 @@ impl<TNetwork: Network> MacroSync<TNetwork::PeerId> for HistoryMacroSync<TNetwor
 
         // Pushing the future to FuturesUnordered above does not wake the task that
         // polls `epoch_ids_stream`. Therefore, we need to wake the task manually.
-        if let Some(waker) = &self.waker {
-            waker.wake_by_ref();
-        }
+        self.waker.wake();
     }
 }
--- a/consensus/src/sync/sync_queue.rs
+++ b/consensus/src/sync/sync_queue.rs
@@ -260,9 +260,7 @@ where
                 self.peers.read().len(),
             );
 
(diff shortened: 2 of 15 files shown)


network-libp2p/src/discovery/message_codec/reader.rs
```diff
@@ -14,55 +14,46 @@ use super::header::Header;
 /// Try to read, such that at most `n` bytes are in the buffer. This will return `Poll::Pending` until the buffer
 /// has `n` bytes in it. This returns `Poll::Ready(Ok(false))` in case of EOF.
```
jsdanielh:
```suggestion
/// has `wanted_len` bytes in it. This returns `Poll::Ready(Ok(false))` in case of EOF.
```

consensus/src/sync/history/sync.rs
```diff
@@ -8,6 +8,7 @@ use futures::{future::BoxFuture, stream::FuturesUnordered, FutureExt};
 use nimiq_blockchain::Blockchain;
 use nimiq_hash::Blake2bHash;
 use nimiq_network_interface::network::{Network, SubscribeEvents};
+use nimiq_utils::WakerExt as _;
```
Eligioo, line 11:
```suggestion
use nimiq_utils::WakerExt;
```

I don't think the import has to be marked as `_`.

consensus/src/sync/sync_queue.rs
```diff
@@ -260,9 +260,7 @@ where
                 self.peers.read().len(),
             );
 
-            if let Some(waker) = self.waker.take() {
-                waker.wake();
-            }
+            self.waker.wake();
         }
```
Eligioo, line 264: Around line 330, there is another `self.waker.take()`

network-libp2p/src/discovery/message_codec/reader.rs
hrxi: Fixed.

consensus/src/sync/sync_queue.rs
hrxi: Fixed.

consensus/src/sync/history/sync.rs
hrxi: Importing it as `_` is indeed not necessary.

By importing it as `_` (i.e. not giving this import any name, `_` is special), I make sure that the `WakerExt` symbol does not unnecessarily get imported into the module when it is actually unused. I just want to import the trait to be able to call its methods.

jsdanielh (approved):

Eligioo (approved):

PR #2748, opened 2024-07-15
Verify configured voting and signing keys when checking validator status

sisou:
To prevent validator nodes running undetected with wrong configured voting and signing keys, I added checks to the inactive check to compare the configured keys to the on-chain keys to alert node operators about a potential missmatch.

### DIFF
--- a/validator/src/validator.rs
+++ b/validator/src/validator.rs
@@ -54,8 +54,7 @@ use crate::{
 enum ValidatorStakingState {
     Active,
     Inactive(Option<u32>),
-    NoStake,
-    Unknown,
+    UnknownOrNoStake,
 }
 
 pub struct ConsensusState {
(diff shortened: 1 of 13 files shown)


viquezclaudio (approved):

sisou:
@viquezclaudio Testing this, and looking at the code, it looks like `get_staking_state()` is called for every block (when the Validator is polled). Is that fine? Should I make the config check happen less often? In the usual case, once per node startup is enough (although the keys registered in the staking contract can be changed at any time, of course).

viquezclaudio:
Yes, I think it will be better to do it just once, at init

sisou:
@viquezclaudio I have moved the check to the `init_epoch()` method which gets called both on init and for each new epoch. That way the check is only done at init and every 12 hours, so it still re-checks if the validator was updated in the staking contract.

validator/src/validator.rs
```diff
@@ -299,6 +300,20 @@ where
         spawn(async move {
             network.set_validators(voting_keys).await;
         });
+
+        // Check validator configuration
+        let _ = self.get_validator(&blockchain)
```
Eligioo:
```suggestion
        if let Ok(Some(validator)) = self.get_validator(&blockchain) {
            ...
        }
```

As you are only interested in the `if let Ok(Some(validator))` case, this is basically the same without the nested `map`s.

validator/src/validator.rs
```diff
@@ -630,25 +645,31 @@ where
         self.consensus.is_ready_for_validation()
     }
 
-    fn get_staking_state(&self, blockchain: &Blockchain) -> ValidatorStakingState {
+    fn get_validator(&self, blockchain: &Blockchain) -> Result<Option<ValidatorAccount>, ()> {
```
Eligioo: Why is the return type `Result<Option<ValidatorAccount>, ()>`? Returning `Option<ValidatorAccount>` already is enough and lets you get rid of the `()` error case.

Then `get_staking_state` could return `ValidatorStakingState::Unknown` in the case of `get_validator` returning `None`.

sisou: Nope. Unknown is when we fail to read the staking contract. When it's none, it returns Inactive.

Eligioo: Now if `get_validator` returns `Err(())`, it means that the `staking_contract` can't be found. That feels weird. You can only know this by reading the function. Result should then return a proper Error here instead of `()`.

Looking at it again you can even get rid of `ValidatorStakingState::NoStake` as it's a dead variant. Then it can be `get_validator -> Option<ValidatorAccount>`

Eligioo (approved):
LGTM

PR #2758, opened 2024-07-19
primitives: Improve `KeyNibbles` serialization

jsdanielh:
- Improve `KeyNibbles` serialization performance by using`serde_byte`s `ByteBuf`.
- Commit e983d05 introduced a `TrieNode` serialization performance improvement by using `serde_byte`'s `Bytes` but forgot to also use it for deserizalization. But the problem is that we're deserializing into a `Vec<u8>` and we can't use `Bytes` since the size is unknown. So this commit changes (de)serialization to use `ByteBuf` instead.

## Pull request checklist

### DIFF
--- a/primitives/src/key_nibbles.rs
+++ b/primitives/src/key_nibbles.rs
@@ -363,6 +363,7 @@ mod serde_derive {
         de::{Deserialize, Deserializer, Error, SeqAccess, Unexpected, Visitor},
         ser::{Serialize, SerializeStruct, Serializer},
     };
+    use serde_bytes::ByteBuf;
 
     use super::KeyNibbles;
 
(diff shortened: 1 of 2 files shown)


jsdanielh:
@hrxi Per the profiling I did, this improves a lot the serialization of `KeyNibbles` but now we're potentially storing more data than needed. Any idea on how to improve it?

hrxi (changes requested):
You could probably gain the same performance by using [`ByteBuf`](https://docs.rs/serde_bytes/0.11.15/serde_bytes/struct.ByteBuf.html) instead of `Vec` for de-/serialization, without changing the on-disk format.

hrxi:
Added a9a1d370d098c1e8f91bd0c5f54da4ce6687fa9b to #2766.

PR #2759, opened 2024-07-19
Dynamic validator block producer timeout

### LINKED ISSUE #2664: Change skip block timeout to a dynamic one
Also for tendermint

Eligioo:
Introduce a dynamic micro block producer timeout based on the average block time of the node's head and its 30 immediate predecessor blocks.

Parameters:
Timeout: 4 x the average block time
Timeout lower bound: `Policy::MINIMUM_PRODUCER_TIMEOUT` (4000 ms)
Timeout upper bound: `Validator::MAXIMUM_PRODUCER_TIMEOUT` (30_000 ms)
Final, canonical skip block timestamp: `head.timestamp() + Policy::MINIMUM_PRODUCER_TIMEOUT`

## Pull request checklist

### DIFF
--- a/validator/src/micro.rs
+++ b/validator/src/micro.rs
@@ -11,6 +11,7 @@ use nimiq_block::{Block, EquivocationProof, MicroBlock, SkipBlockInfo};
 use nimiq_blockchain::{BlockProducer, BlockProducerError, Blockchain};
 use nimiq_blockchain_interface::{AbstractBlockchain, PushResult};
 use nimiq_mempool::mempool::Mempool;
+use nimiq_primitives::policy::Policy;
 use nimiq_time::sleep;
 use nimiq_utils::time::systemtime_to_timestamp;
 use nimiq_validator_network::ValidatorNetwork;
@@ -201,12 +202,18 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
         let now = systemtime_to_timestamp(SystemTime::now());
         let wait_until_min = now + self.producer_timeout.as_millis() as u64;
         let wait_until_expected = expected_next_ts
-            + (self.producer_timeout - self.block_separation_time).as_millis() as u64;
+            + (self
+                .producer_timeout
+                .saturating_sub(self.block_separation_time))
+            .as_millis() as u64;
+
         let next_block_timeout = cmp::max(wait_until_min, wait_until_expected) - now;
-        sleep(Duration::from_millis(next_block_timeout)).await;
+        let timeout = Duration::from_millis(next_block_timeout);
+        sleep(timeout).await;
 
         info!(
             block_number = self.block_number,
+            ?timeout,
             "No micro block received within timeout, producing skip block"
         );
 
@@ -246,7 +253,7 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
             if !in_current_state(head) {
                 None
             } else {
-                let timestamp = head.timestamp() + self.producer_timeout.as_millis() as u64;
+                let timestamp = head.timestamp() + Policy::MINIMUM_PRODUCER_TIMEOUT;
 
                 let skip_block = self.block_producer.next_micro_block(
                     &blockchain,
--- a/validator/src/micro.rs
+++ b/validator/src/micro.rs
@@ -253,7 +253,7 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
             if !in_current_state(head) {
                 None
             } else {
-                let timestamp = head.timestamp() + Policy::MINIMUM_PRODUCER_TIMEOUT;
+                let timestamp = head.timestamp() + Policy::MIN_PRODUCER_TIMEOUT;
 
                 let skip_block = self.block_producer.next_micro_block(
                     &blockchain,
--- a/validator/src/validator.rs
+++ b/validator/src/validator.rs
@@ -132,7 +132,10 @@ impl<TValidatorNetwork: ValidatorNetwork> Validator<TValidatorNetwork>
 where
     PubsubId<TValidatorNetwork>: std::fmt::Debug + Unpin,
 {
-    const PRODUCER_TIMEOUT: Duration = Duration::from_millis(Policy::BLOCK_PRODUCER_TIMEOUT);
+    const MAXIMUM_PRODUCER_TIMEOUT: Duration = Duration::from_millis(16_000); // 16 seconds
+    const MINIMUM_PRODUCER_TIMEOUT: Duration =
+        Duration::from_millis(Policy::MINIMUM_PRODUCER_TIMEOUT); // 4 seconds
+    const NUM_BLOCKS_FOR_TIMEOUT_CALCULATION: u32 = 60;
     const BLOCK_SEPARATION_TIME: Duration = Duration::from_millis(Policy::BLOCK_SEPARATION_TIME);
     const EQUIVOCATION_PROOFS_MAX_SIZE: usize = 1_000; // bytes
 
@@ -234,6 +237,58 @@ where
     fn init(&mut self, head_hash: Option<&Blake2bHash>) {
         self.init_epoch();
         self.init_block_producer(head_hash);
+    }
+
+    /// Computes the micro block producer timeout based on historical block data.
+    ///
+    /// This method calculates the timeout duration for a micro block producer by computing the average block
+    /// time over a defined number of previous blocks (`Self::NUM_BLOCKS_FOR_TIMEOUT_CALCULATION`).
+    /// If sufficient block history is not available, it uses the genesis block as a fallback for the calculation.
+    ///
+    /// If the block window is equal or smaller than 1, `Self::MINIMUM_PRODUCER_TIMEOUT` is returned.
+    ///
+    /// The timeout is determined by multiplying the average block time by four. For example
+    /// an average block time of 2 seconds results in a 8 second timeout.
+    ///
+    /// Lastly, the timeout is clamped between `Self::MINIMUM_PRODUCER_TIMEOUT` and `Self::MAXIMUM_PRODUCER_TIMEOUT`.
+    fn compute_micro_block_producer_timeout(
+        head_block: &Block,
+        blockchain: &Blockchain,
+    ) -> Duration {
+        let lower_timestamp;
+        let lower_block_number;
+
+        if let Ok(window_block) = blockchain.get_block_at(
+            head_block
+                .block_number()
+                .saturating_sub(Self::NUM_BLOCKS_FOR_TIMEOUT_CALCULATION),
+            false,
+            None,
+        ) {
+            lower_timestamp = window_block.timestamp();
+            lower_block_number = window_block.block_number();
+        } else {
+            let genesis_block = blockchain
+                .get_block_at(Policy::genesis_block_number(), false, None)
+                .expect("genesis block must be present in the ChainStore");
+
+            lower_timestamp = genesis_block.timestamp();
+            lower_block_number = genesis_block.block_number();
+        }
+
+        let block_window = head_block.block_number().saturating_sub(lower_block_number) as u64;
+        if block_window <= 1 {
+            return Self::MINIMUM_PRODUCER_TIMEOUT;
+        }
+
+        let avg_block_time =
+            (head_block.timestamp().saturating_sub(lower_timestamp)) / (block_window - 1);
+        let timeout = Duration::from_millis(avg_block_time * 4);
+
+        timeout.clamp(
+            Self::MINIMUM_PRODUCER_TIMEOUT,
+            Self::MAXIMUM_PRODUCER_TIMEOUT,
+        )
     }
 
     /// Resets the reactivation state if we sent a reactivate transaction that expired.
@@ -374,8 +429,6 @@ where
                     .get_equivocation_proofs_for_block(Self::EQUIVOCATION_PROOFS_MAX_SIZE);
                 let prev_seed = head.seed().clone();
 
-                drop(blockchain);
-
                 self.micro_producer = Some(ProduceMicroBlock::new(
                     Arc::clone(&self.blockchain),
                     Arc::clone(&self.mempool_task.mempool),
@@ -385,7 +438,7 @@ where
                     equivocation_proofs,
                     prev_seed,
                     next_block_number,
-                    Self::PRODUCER_TIMEOUT,
+                    Self::compute_micro_block_producer_timeout(head, &blockchain),
                     Self::BLOCK_SEPARATION_TIME,
                 ));
             }
--- a/validator/src/validator.rs
+++ b/validator/src/validator.rs
@@ -134,10 +134,12 @@ where
 {
     /// The minimum time to wait for a micro block before starting a skip block.
     const MIN_PRODUCER_TIMEOUT: Duration = Duration::from_millis(Policy::MIN_PRODUCER_TIMEOUT);
-    /// The maximum time to wait for a micro block before starting to skip block.
+    /// The maximum time to wait for a micro block before starting a skip block.
     const MAX_PRODUCER_TIMEOUT: Duration = Duration::from_secs(16);
     /// The number of blocks to consider when calculating the micro block producer timeout.
     const PRODUCER_TIMEOUT_WINDOW_SIZE: u32 = 120;
+    /// Multiplier applied to the average block time when calculating the micro block producer timeout.
+    const PRODUCER_TIMEOUT_MULTIPLIER: u64 = 4;
     /// The targeted block time.
     const BLOCK_SEPARATION_TIME: Duration = Duration::from_millis(Policy::BLOCK_SEPARATION_TIME);
     /// The maximum number of bytes that equivocation proofs are allowed to take up in a block.
@@ -254,11 +256,12 @@ where
         let end_block_number = head_block.block_number();
         let end_timestamp = head_block.timestamp();
 
-        // Calculate the block number at the start of the window.
-        // Make sure this never points before the genesis block.
+        // Calculate the block number at the start of the window. Make sure this points to at least
+        // one block after the genesis block, as the genesis timestamp can be unreliable.
         let start_block_number =
             end_block_number.saturating_sub(Self::PRODUCER_TIMEOUT_WINDOW_SIZE);
-        let mut start_block_number = u32::max(start_block_number, Policy::genesis_block_number());
+        let mut start_block_number =
+            u32::max(start_block_number, Policy::genesis_block_number() + 1);
 
         // We might not have the block at `start_block_number` in our database. If it's missing,
         // move up the chain to the next succeeding macro block(s), adjusting `start_block_number`
@@ -287,7 +290,7 @@ where
 
         // Calculate the timeout and clamp it to the allowed range.
         let avg_block_time = end_timestamp.saturating_sub(start_timestamp) / effective_window_size;
-        let timeout = Duration::from_millis(avg_block_time * 4);
+        let timeout = Duration::from_millis(avg_block_time * Self::PRODUCER_TIMEOUT_MULTIPLIER);
         timeout.clamp(Self::MIN_PRODUCER_TIMEOUT, Self::MAX_PRODUCER_TIMEOUT)
     }
 
--- a/validator/src/validator.rs
+++ b/validator/src/validator.rs
@@ -132,11 +132,15 @@ impl<TValidatorNetwork: ValidatorNetwork> Validator<TValidatorNetwork>
 where
     PubsubId<TValidatorNetwork>: std::fmt::Debug + Unpin,
 {
-    const MAXIMUM_PRODUCER_TIMEOUT: Duration = Duration::from_millis(16_000); // 16 seconds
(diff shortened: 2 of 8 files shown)


validator/src/validator.rs
```diff
@@ -232,6 +237,50 @@ where
...
+        let amount_of_timestamps = self.recent_block_timestamps.len();
+
+        // If 0 or 1 timestamp have been stored, simply return the Policy::BLOCK_PRODUCER_TIMEOUT
+        if amount_of_timestamps <= 1 {
+            return Policy::BLOCK_PRODUCER_TIMEOUT;
+        }
+
+        let mut summed_durations = 0;
+        for i in 1..amount_of_timestamps {
+            // Compute the difference between the timestamp (i) and it's immediate predecessor (i - 1)
```
hrxi:
```suggestion
            // Compute the difference between the timestamp (i) and its immediate predecessor (i - 1)
```

validator/src/validator.rs
```diff
@@ -232,6 +237,50 @@ where
...
+        let average_duration: u64 = summed_durations / ((amount_of_timestamps as u64) - 1);
+        let timeout = average_duration * 4;
+
+        if timeout < Policy::BLOCK_PRODUCER_TIMEOUT {
+            return Policy::BLOCK_PRODUCER_TIMEOUT;
+        } else if timeout > Self::MAXIMUM_PRODUCER_TIMEOUT {
+            return Self::MAXIMUM_PRODUCER_TIMEOUT;
+        }
+
+        timeout
```
hrxi:
```suggestion
        timeout.clamp(Policy::BLOCK_PRODUCER_TIMEOUT, Self::MAXIMUM_PRODUCER_TIMEOUT)
```

validator/src/validator.rs
```diff
@@ -232,6 +237,50 @@ where
...
+        }
+
+        let mut summed_durations = 0;
+        for i in 1..amount_of_timestamps {
+            // Compute the difference between the timestamp (i) and it's immediate predecessor (i - 1)
+            summed_durations +=
+                self.recent_block_timestamps[i] - self.recent_block_timestamps[i - 1];
+        }
+
+        let average_duration: u64 = summed_durations / ((amount_of_timestamps as u64) - 1);
```
hrxi:
```suggestion
        let complete_duration = self.recent_block_timestamps[amount_of_timestamps - 1] - self.recent_block_timestamps[0];
        let average_duration: u64 = complete_duration / ((amount_of_timestamps as u64) - 1);
```

validator/src/validator.rs
```diff
@@ -415,6 +464,7 @@ where
 
         self.check_reactivate(block.block_number());
         self.init_block_producer(Some(hash));
+        self.push_block_timestamp(block.timestamp());
     }
 
     fn on_blockchain_rebranched(
```
hrxi: I have a feeling this function needs adjustment of `self.recent_block_timestamps` as well.

validator/src/validator.rs
```diff
@@ -743,7 +793,10 @@ where
                 Ok(ConsensusEvent::Lost)
                 | Ok(ConsensusEvent::Established {
                     synced_validity_window: false,
-                }) => self.pause(),
+                }) => {
+                    self.pause();
+                    self.recent_block_timestamps.clear();
```
hrxi: I *think* this shouldn't be necessary, as the `self.recent_block_timestamps` should be updated correctly even in case of lost consensus.

validator/src/validator.rs
```diff
@@ -232,6 +237,50 @@ where
         self.init_block_producer(head_hash);
     }
 
+    fn push_block_timestamp(&mut self, timestamp: u64) {
+        // Only push the timestamps once caught up with the chain
```
hrxi: Why? This is consensus relevant, so everyone should have the same view.

hrxi (commented):
Perhaps this entire thing should be read out of the blockchain (no way to get out of sync), instead of being tracked by the validator.

What about the canonical block timestamp of the skip block, what does it get set to? Is it verified somewhere? Perhaps the whole checking could go into the blockchain itself?

Eligioo:
@hrxi To clarify this change a bit more, this PR doesn't change the canonical timestamp of a skip block (the timestamp of the skip block that ends up in the blockchain), it only changes the time validators will wait before they start the aggregations for the skip block (and eventually produce it).

This is also why I don't think it's a protocol breaking change as the timestamp of the skip block will become `head.timestamp() + Policy::BLOCK_PRODUCER_TIMEOUT`, regardless how long the produce timeout was. Which can be found in this PR. This should also answer your question about what the timestamp of the skip block is going to be.

To summarise:
- previously the `self.producer_timeout` would be used for actually waiting to start aggregations but could also be used to compute the  timestamps of the skip blocks. This now has been split up as `self.producer_timeout` will become dynamic.
- The skip block timestamp with this PR thus becomes `head.timestamp() + Policy::BLOCK_PRODUCER_TIMEOUT` instead of `head.timestamp() + self.producer_timeout`.
- Even if the validators will wait 10/20/30 secs to start the skip block aggregations, the skip block timestamp will still be deterministic.

validator/src/validator.rs
```diff
@@ -210,6 +213,8 @@ where
             validator_state: None,
             automatic_reactivate,
 
+            recent_block_timestamps: VecDeque::with_capacity(30),
```
nibhar: Magic Number. This one should probably be configurable. If it is important for every network participant to have the same constant this should go to policy I guess.

validator/src/validator.rs
```diff
@@ -232,6 +237,50 @@ where
         self.init_block_producer(head_hash);
     }
 
+    fn push_block_timestamp(&mut self, timestamp: u64) {
```
nibhar: This does not seem very robust to me. What happens in case of a rebranch?

validator/src/validator.rs
nibhar: Pretty sure it does.

validator/src/validator.rs
```diff
@@ -232,6 +237,50 @@ where
...
+            assert!(amount_of_timestamps < 31);
+
+            self.recent_block_timestamps.pop_front();
+        }
+
+        self.recent_block_timestamps.push_back(timestamp);
+    }
+
+    fn compute_producer_timeout(&self) -> u64 {
+        let amount_of_timestamps = self.recent_block_timestamps.len();
```
nibhar: This having variance is not ideal. I remember talking about it and I see the problem of having to have a set amount of blocks before the current one in order to have the full 30 (or whatever the number) deltas, but validators already need those.
Since the canonical skip block will just have a fixed delta I think this should always be a fixed number.

validator/src/validator.rs
nibhar: This is only true if the blocks are in order. The presence of skip blocks makes this a non guaranteed property.

nibhar (changes requested):
I think rebranching needs to be accounted for here.

I also think that it would be very beneficial to have everyone have the same block timestamps, and the same number of them. Otherwise everyone may start the skip blocks at different times, which in itself is not a problem, but also not exactly helpful.

validator/src/validator.rs
hrxi: > This is only true if the blocks are in order.

This isn't true. The change I proposed doesn't need to assign any semantic value to the elements in the list.

You're computing `(a[1] - a[0]) + (a[2] - a[1]) + … + (a[n] - a[n-1])`, reordering, we get `(a[n] - a[n-1]) + (a[n-1] - a[n-2]) + … + (a[2] - a[1]) + (a[1] - a[0])`, associativity `a[n] + (-a[n-1] + a[n-1]) + … + (-a[1] + a[1]) - a[0]`, so `a[n] + 0 + 0 + … + 0 + a[0] = a[n] - a[0]`.

So if I got everything correctly, and if you think my formula is wrong, then the current formula is wrong, too.

nibhar: Apologies, yes i meant the marked section to already be wrong, as blocks are not necessarily pushed in ascending timestamp order. 

After a little more thought it should not be a problem though as long as rebranches get worked into this, as that should restore that order. Depending on what guarantees are given for what blocks are at least in there that is not a problem, but if there are no guarantees `summed_duration` might even get negative because of it. 

Importantly I don't see that the first block is necessarily on the same fork as the last block.

Your change obviously has nothing t do with that, I guess I should have created a new comment here instead of adding to yours.

Eligioo:
@nibhar @hrxi based on the feedback I have taken a different, more simple approach that shouldn't be susceptible to rebranching. The `VecDeque` keeping track of the recent pushed blocks is now out of the way. This has been replaced with a single method that you feed the an older timestamp, a more recent timestamp and the amount of blocks that are produced within those timestamps. The information is gathered directly from the `blockchain`.

validator/src/validator.rs
```diff
@@ -233,6 +234,35 @@ where
...
+        if amount_of_blocks <= 1 {
+            return Policy::BLOCK_PRODUCER_TIMEOUT;
+        }
+
+        let total_duration = max_block_timestamp.saturating_sub(min_block_timestamp);
+        let average_block_time: u64 = total_duration / (amount_of_blocks as u64 - 1);
+        let timeout = average_block_time * 4;
+
+        timeout.clamp(
+            Policy::BLOCK_PRODUCER_TIMEOUT,
```
hrxi: This constant should probably be called `MINIMUM_PRODUCER_TIMEOUT`.

validator/src/validator.rs
```diff
@@ -355,6 +385,17 @@ where
                     .get_equivocation_proofs_for_block(Self::EQUIVOCATION_PROOFS_MAX_SIZE);
                 let prev_seed = head.seed().clone();
 
+                // Get the lower timestamp necessary to compute the producer time-out.
+                let timestamp = blockchain
+                    .get_block_at(
+                        head.block_number()
+                            .saturating_sub(Self::BLOCK_DISTANCE_FROM_HEAD - 1),
```
hrxi:
```suggestion
                            .saturating_sub(Self::BLOCK_DISTANCE_FROM_HEAD),
```
Why `- 1`? Or maybe rename `BLOCK_DISTANCE_FROM_HEAD` to `NUM_BLOCKS_FOR_TIMEOUT_CALCULATION`.

validator/src/validator.rs
```diff
@@ -355,6 +385,17 @@ where
...
+                        false,
+                        None,
+                    )
+                    .map(|block| block.timestamp())
+                    .unwrap_or_default();
```
hrxi: A timestamp of `0` looks wrong here. Perhaps get the earliest block that's later than the PoS genesis block?

validator/src/validator.rs
```diff
@@ -366,7 +407,11 @@ where
                     equivocation_proofs,
                     prev_seed,
                     next_block_number,
-                    Self::PRODUCER_TIMEOUT,
+                    Duration::from_millis(self.compute_micro_block_producer_timeout(
+                        timestamp,
+                        head.timestamp(),
+                        Self::BLOCK_DISTANCE_FROM_HEAD,
```
hrxi: This might need adjustment, because there aren't always `Self::BLOCK_DISTANCE_FROM_HEAD` blocks into the past (near the PoS genesis block, or do we look into PoW history)?

validator/src/micro.rs
```diff
@@ -216,7 +223,7 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
             if !in_current_state(&head) {
                 None
             } else {
-                let timestamp = head.timestamp() + self.producer_timeout.as_millis() as u64;
+                let timestamp = head.timestamp() + Policy::BLOCK_PRODUCER_TIMEOUT;
```
hrxi: I think this is problematic: Since this always takes the minimum timestamp, it'll factor into the average block production time later. This means that having skip blocks will decrease the average block time, even if the skip blocks get spaced apart further, since the skip block timestamps are incorrect. Perhaps we can use the `self.producer_timeout.as_millis() as u64` here?

hrxi (commented):
Found some smaller things, but the one of skip blocks reducing the skip block timeouts sounds bad.

validator/src/validator.rs
nibhar: Yeah this should probably not fall back to 0 but to whatever is the earliest timestamp available combined with the proper amount of blocks it is away from the head.

validator/src/validator.rs
nibhar: This ties into the previous comment, but I think it should be adjusted.

validator/src/micro.rs
nibhar: What happens if a skip block takes very long to produce? What happens if the stated timestamp is more off the actual timestamp it got produced than `MAXIMUM_PRODUCER_TIMOUT`?

hrxi: > I think this is problematic: Since this always takes the minimum timestamp, it'll factor into the average block production time later. This means that having skip blocks will decrease the average block time, even if the skip blocks get spaced apart further, since the skip block timestamps are incorrect.

We talked about this and would fix it in another place: When doing the averaging, don't take skip blocks into account.

> What happens if a skip block takes very long to produce? What happens if the stated timestamp is more off the actual timestamp it got produced than `MAXIMUM_PRODUCER_TIMOUT`?

Producer timeouts only start when a peer receives a block. The actual block timestamp is irrelevant for this, so there's no problem with receiving blocks whose timestamps are far in the past.

validator/src/validator.rs
Eligioo: Fixed

validator/src/validator.rs
Eligioo: Fixed

validator/src/micro.rs
Eligioo: Fixed

validator/src/validator.rs
```diff
@@ -382,13 +415,52 @@ where
...
                     Self::BLOCK_SEPARATION_TIME,
                 ));
             }
         }
     }
 
+    /// Returns a timestamp and the number of blocks produced between this timestamp and the provided head block (both inclusive).
+    /// The number of skip blocks are filtered out in this time range as their timestamp is canonical thus not representative.
+    fn producer_timeout_block_range(
+        blockchain: &RwLockReadGuard<Blockchain>,
```
styppo: This can be simplified to `&Blockchain`.

validator/src/micro.rs
```diff
@@ -201,12 +202,18 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
...
+            .as_millis() as u64;
+
         let next_block_timeout = cmp::max(wait_until_min, wait_until_expected) - now;
-        sleep(Duration::from_millis(next_block_timeout)).await;
+        let duration = Duration::from_millis(next_block_timeout);
+        sleep(duration).await;
 
         info!(
             block_number = self.block_number,
+            timeout_ms = duration.as_millis(),
```
styppo:
```suggestion
            timeout = ?duration,
```

validator/src/validator.rs
```diff
@@ -132,7 +134,8 @@ impl<TValidatorNetwork: ValidatorNetwork> Validator<TValidatorNetwork>
 where
     PubsubId<TValidatorNetwork>: std::fmt::Debug + Unpin,
 {
-    const PRODUCER_TIMEOUT: Duration = Duration::from_millis(Policy::BLOCK_PRODUCER_TIMEOUT);
+    const MAXIMUM_PRODUCER_TIMEOUT: u64 = 30_000; // 30 seconds
```
styppo: 30 seconds maximum timeout seems excessive. I think this should be between 12 and 16 seconds, i.e. 3-4 times the minimum timeout to not allow the chain to slow down too much.

Also, typing this as `Duration` instead of `u64` seems more expressive.
```suggestion
    const MAXIMUM_PRODUCER_TIMEOUT: Duration = Duration::from_secs(16);
```

validator/src/validator.rs
```diff
@@ -132,7 +134,8 @@ impl<TValidatorNetwork: ValidatorNetwork> Validator<TValidatorNetwork>
...
+    const NUM_BLOCKS_FOR_TIMEOUT_CALCULATION: u32 = 30;
```
styppo: I think this window needs to be bigger to make the average block time computation more robust against one-off skip blocks. It's also good if it includes at least one macro block to work around the edge case where we have only skip blocks in the window. Two batches worth of blocks seems like a good ball park.

validator/src/validator.rs
```diff
@@ -382,13 +415,52 @@ where
...
+        let genesis_block = blockchain
+            .get_block_at(Policy::genesis_block_number(), false, None)
+            .expect("genesis block must be present in the ChainStore");
+
+        (
+            genesis_block.timestamp(),
+            head.block_number() - genesis_block.block_number() + 1,
+        )
+    }
+
```
styppo: The filtering of the skip blocks here doesn't really make sense. Consider the case where we are adding only skip blocks to the end of the calculation window, then the timeout doesn't change even though it should be increasing.

A simple solution could be to compute the average block time as

```
avg_block_time = (current_system_time - block_at(current_block_number - window_size).timestamp) / window_size
```
By using the current system time instead of the head block timestamp, we approximate the block arrival time which is more useful in the face of canonical skip block timestamps. This would give a slightly longer average block time than desired when the block at the start of the window is a skip block. Alternatively, we could change the canonical skip block timestamp to `parent_block.timestamp + MINIMUM_PRODUCER_TIMEOUT` instead of `parent_block.timestamp + BLOCK_SEPARATION_TIME`, which better approximates the actual production time of the skip block.

styppo (changes requested):

validator/src/validator.rs
Eligioo: This was the formula I was using before I started filtering skip blocks except I used the head timestamp instead of the current system time 😅

You have to keep the fallback on the genesis block. There are situations (e.g. a very short chain) where the `block_at(current_block_number - window_size)` can't return a block.

validator/src/validator.rs
Eligioo: It can be done but it always needs to get converted to a `u64` at runtime which kinda defeats the purpose of it being a `const` since it's only being used for clamping. Not a fan of it.

styppo: We need a `Duration` in the end to pass into `sleep()`. So you can just convert the `u64` timeout into a `Duration` and then use `Duration::clamp` to restrict the range.

styppo (approved):

validator/src/validator.rs
```diff
@@ -236,6 +243,55 @@ where
...
+        // Calculate the effective window size.
+        // If there are no blocks in the window, we return the default value.
+        let effective_window_size = end_block_number.saturating_sub(start_block_number) as u64;
+        if effective_window_size < 1 {
+            return Self::MIN_PRODUCER_TIMEOUT;
+        }
+
+        // Calculate the timeout and clamp it to the allowed range.
+        let avg_block_time = end_timestamp.saturating_sub(start_timestamp) / effective_window_size;
+        let timeout = Duration::from_millis(avg_block_time * 4);
```
nibhar: Magic Number.
I suggest putting it  as a const like the rest of them.

nibhar (approved):

validator/src/validator.rs
Eligioo: Done

jsdanielh:
Rebasing for merge

validator/src/validator.rs
```diff
@@ -132,8 +132,17 @@ impl<TValidatorNetwork: ValidatorNetwork> Validator<TValidatorNetwork>
...
     PubsubId<TValidatorNetwork>: std::fmt::Debug + Unpin,
 {
-    const PRODUCER_TIMEOUT: Duration = Duration::from_millis(Policy::BLOCK_PRODUCER_TIMEOUT);
+    /// The minimum time to wait for a micro block before starting a skip block.
+    const MIN_PRODUCER_TIMEOUT: Duration = Duration::from_millis(Policy::MIN_PRODUCER_TIMEOUT);
+    /// The maximum time to wait for a micro block before starting to skip block.
+    const MAX_PRODUCER_TIMEOUT: Duration = Duration::from_secs(16);
+    /// The number of blocks to consider when calculating the micro block producer timeout.
+    const PRODUCER_TIMEOUT_WINDOW_SIZE: u32 = 120;
+    /// Multiplier applied to the average block time twhen calculating the micro block producer timeout.
```
jsdanielh:
```suggestion
    /// Multiplier applied to the average block time when calculating the micro block producer timeout.
```

validator/src/validator.rs
```diff
@@ -132,8 +132,17 @@ impl<TValidatorNetwork: ValidatorNetwork> Validator<TValidatorNetwork>
 where
     PubsubId<TValidatorNetwork>: std::fmt::Debug + Unpin,
 {
-    const PRODUCER_TIMEOUT: Duration = Duration::from_millis(Policy::BLOCK_PRODUCER_TIMEOUT);
+    /// The minimum time to wait for a micro block before starting a skip block.
+    const MIN_PRODUCER_TIMEOUT: Duration = Duration::from_millis(Policy::MIN_PRODUCER_TIMEOUT);
+    /// The maximum time to wait for a micro block before starting to skip block.
```
jsdanielh:
```suggestion
    /// The maximum time to wait for a micro block before starting a skip block.
```

jsdanielh (approved):

PR #2763, opened 2024-07-22
Tendermint: Start skipped aggregations

nibhar:
In case of a skip ahead to a future round prior to this PR Tendermint will just ignore and never start any aggregations in between.

Starting these aggregations is useful as it leads to a more "complete" state. In between aggregations have an easier time completing with the additional contribution of the node skipping ahead.

### DIFF
--- a/tendermint/src/tendermint.rs
+++ b/tendermint/src/tendermint.rs
@@ -213,6 +213,39 @@ impl<TProtocol: Protocol> Tendermint<TProtocol> {
             .is_some()
     }
 
+    /// Creates an aggregation for a given `id`. If that aggregation does already exist,
+    /// as indicated by the presence of `id` in `self.aggregation_senders` it has no effect.
+    ///
+    /// The vote for this aggregation must exist in `self.state.votes` otherwise this function panics.
+    pub(crate) fn create_aggregation(&mut self, id: (u32, Step)) {
+        // create the aggregation if it does not exist yet
+        if let Entry::Vacant(entry) = self.aggregation_senders.entry(id) {
+            log::debug!(?id, "Creating Aggregation");
+            // Retrieve the vote this node is going to take.
+            let vote = self
+                .state
+                .votes
+                .get(&id)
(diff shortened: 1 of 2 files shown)


tendermint/src/tendermint.rs
```diff
@@ -213,6 +219,34 @@ impl<TProtocol: Protocol> Tendermint<TProtocol> {
...
+    /// Creates an aggregation for a given `id`. If that aggregation does already exist,
+    /// as indicated by the presence of `id` in `self.aggregation_senders` it has no effect.
+    ///
+    /// The vote for this aggregation must exist in `self.state.votes` otherwise this function panics.
+    pub(crate) fn create_aggregation(&mut self, id: (u32, Step)) {
+        // create the aggregation if it does not exist yet
+        if let Entry::Vacant(entry) = self.aggregation_senders.entry(id) {
+            log::debug!(?id, "Creating Aggregation");
+            // Retrieve the vote this node is going to take.
+            let vote = self.state.votes.get(&id).expect("").clone();
```
hrxi:
```suggestion
            let vote = self.state.votes.get(&id).unwrap().clone();
```
or add a reason.

Ah, you just moved this code around. I guess you don't have to fix this.

hrxi (approved):
Looks good to me.

Eligioo (approved):

tendermint/src/tendermint.rs
nibhar: Fixed it still. Also the semantics here have changed, as I did not only copy, but introduced `id` as a parameter here. This can actually be encountered if one does not properly set up a vote for any aggregation that needs starting. Specifically (as always in Tendermint) the vote must be created in a poll call prior to the poll call creating the aggregation in order for the state to have been properly exported. 
Actually at seconds thought this seems a bit fragile, but I am not sure how to best address this.

hrxi (approved):

hrxi:
Rebased for signing.

PR #2778, opened 2024-07-25
Poll again after adding new block stream

### LINKED ISSUE #2745: Failure to catch up through head requests
In [this CI run](https://github.com/nimiq/core-rs-albatross/actions/runs/9929244435/job/27426574778), it was observed that validators 2 and 4 fail to catch up after being restarted. It appears that unknown blocks received from head requests are not being pushed onto the chain, but only after the restart has occurred.

Eligioo:
With the PR #2702 the `BlockQueue` was moved into its own task so that it can be polled independently from the `Consensus`. This turned out to reveal a race against the `LiveSyncQueue` when adding a second stream to the `BlockQueue::block_stream`. This second stream is for example used when we want to push blocks received from `HeadRequests` into the `BlockQueue`. However since the `BlockQueue` now only is polled before the second stream got added, the second stream didn't get a chance to store a waker.

Before, when the BlockQueue was still part of the Consensus task, the BlockQueue would getting polled when the Consensus gets polled.

This PR fixes this by waking up the `BlockQueue` stream after a new stream to the `BlockQueue::block_stream` has been added.

## Pull request checklist

### DIFF
--- a/consensus/src/sync/live/block_queue/queue.rs
+++ b/consensus/src/sync/live/block_queue/queue.rs
@@ -7,7 +7,7 @@ use std::{
     error::Error,
     pin::Pin,
     sync::Arc,
-    task::{Context, Poll},
+    task::{Context, Poll, Waker},
 };
 
 use futures::{stream::BoxStream, Stream, StreamExt};
@@ -17,7 +17,7 @@ use nimiq_blockchain_proxy::BlockchainProxy;
 use nimiq_hash::Blake2bHash;
 use nimiq_network_interface::network::{MsgAcceptance, Network, PubsubId};
 use nimiq_primitives::{policy::Policy, slots_allocation::Validators};
-use nimiq_utils::spawn::spawn;
+use nimiq_utils::{spawn::spawn, WakerExt};
 use parking_lot::RwLock;
 use tokio::sync::oneshot::Sender as OneshotSender;
 
@@ -79,6 +79,9 @@ pub struct BlockQueue<N: Network> {
     ///
     pending_requests:
         BTreeMap<u32, HashMap<Blake2bHash, OneshotSender<Result<Block, ResolveBlockError<N>>>>>,
+
+    /// Waker used for the poll function
+    pub(crate) waker: Option<Waker>,
 }
 
 impl<N: Network> BlockQueue<N> {
@@ -134,6 +137,7 @@ impl<N: Network> BlockQueue<N> {
             blocks_pending_push: BTreeSet::new(),
             current_macro_height,
             pending_requests: BTreeMap::default(),
+            waker: None,
         }
     }
 
@@ -933,6 +937,7 @@ impl<N: Network> Stream for BlockQueue<N> {
             }
         }
 
+        self.waker.store_waker(cx);
         Poll::Pending
     }
 }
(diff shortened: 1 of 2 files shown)


nibhar (approved):
While I think this is a good change, as there could be a new stream being pushed into this select, the only place where I see it actually happening is in `LiveSyncer::with_queue`. That is a constructor, so I am not sure if the analysis of the race actually holds. I might be missing something here though.

Eligioo:
@nibhar When the `BlockQueueProxy` is constructed, it also constructs a `BlockQueue` which is then immediately polled:
```
impl<N: Network> BlockQueueProxy<N> {
    pub async fn new(network: Arc<N>, blockchain: BlockchainProxy, config: QueueConfig) -> Self {
        let queue = BlockQueue::new(network, blockchain, config).await;
        Self::with_queue(queue)
    }
}
```

As part of the `BlockQueue` constructor, the `gossipsub` stream is added directly to the `block_stream`:
```
impl<N: Network> BlockQueue<N> {
    const MAX_BUFFERED_PER_PEER_PER_HEIGHT: usize = 5;

    pub async fn new(network: Arc<N>, blockchain: BlockchainProxy, config: QueueConfig) -> Self {
        let block_stream = if config.include_micro_bodies {
            network.subscribe::<BlockTopic>().await.unwrap().boxed()
        } else {
            network
                .subscribe::<BlockHeaderTopic>()
                .await
                .unwrap()
                .boxed()
        };

        Self::with_gossipsub_block_stream(blockchain, network, block_stream, config)
    }
}
```

Then when both are constructed, the `LiveSyncer::with_queue` is called which adds the second stream, that never gets polled, I was talking about.

styppo (approved):
LGTM 👍 

For completeness, let me mention that the root cause of the race is [the newly added `spawn()`](https://github.com/nimiq/core-rs-albatross/blob/albatross/consensus/src/sync/live/block_queue/proxy.rs#L75) in the `BlockQueueProxy` constructor, which causes the `BlockQueue`'s first `poll()` to race against the remaining `Consensus` initialization (in particular the `LiveSyncer` constructor which [adds the second block stream](https://github.com/nimiq/core-rs-albatross/blob/bc44e73833d59ba2049efd193fee1091574f672a/consensus/src/sync/live/mod.rs#L64)) that is executing on the original `Consensus` task.

consensus/src/sync/live/block_queue/queue.rs
```diff
@@ -865,6 +869,8 @@ impl<N: Network> Stream for BlockQueue<N> {
     type Item = QueuedBlock<N>;
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<Option<Self::Item>> {
+        self.waker.store_waker(cx);
```
hrxi: We only have to store the waker in the very bottom of this function, if we return `Poll::Pending`.

consensus/src/sync/live/block_queue/queue.rs
```diff
@@ -79,6 +79,9 @@ pub struct BlockQueue<N: Network> {
     ///
     pending_requests:
         BTreeMap<u32, HashMap<Blake2bHash, OneshotSender<Result<Block, ResolveBlockError<N>>>>>,
+
+    /// Waker used for the poll function
+    pub(crate) waker: Option<Waker>,
```
ii-cruz, line 84: Perhaps add why we need the waker otherwise some other day we may think we no longer need it and fall into the same issue.

ii-cruz (approved):

consensus/src/sync/live/block_queue/queue.rs
Eligioo: As the `waker` is now referenced in other parts of the codebase, I think it's enough that you shouldn't suddenly remove it without consequences.

PR #2781, opened 2024-07-26
Various Clippy fixes

Eligioo:
Various Clippy fixes with Rust 1.80.

Noticeable changes causing most new Clippy warnings:
- Amongst some others, `size_of` is now part of the prelude https://github.com/rust-lang/rust/pull/123168
- `-Zcheck-cfg` is now enabled by default https://github.com/rust-lang/cargo/pull/13571/

The unexpected `cfg` condition value `parallel` will be addressed in a separate PR.

### DIFF
--- a/bls/src/types/compressed_signature.rs
+++ b/bls/src/types/compressed_signature.rs
@@ -93,7 +93,7 @@ impl fmt::Display for CompressedSignature {
 impl From<G1Projective> for CompressedSignature {
     fn from(signature: G1Projective) -> Self {
         let mut buffer = [0u8; SIZE];
-        CanonicalSerialize::serialize_compressed(&signature.into_affine(), &mut &mut buffer[..])
+        CanonicalSerialize::serialize_compressed(&signature.into_affine(), &mut buffer[..])
             .unwrap();
         CompressedSignature { signature: buffer }
     }
--- a/primitives/transaction/src/account/staking_contract/structs.rs
+++ b/primitives/transaction/src/account/staking_contract/structs.rs
@@ -7,7 +7,8 @@ use nimiq_serde::{Deserialize, DeserializeError, Serialize, SerializedMaxSize};
 use crate::{SignatureProof, Transaction, TransactionError};
 
 /// We need to distinguish two types of transactions:
-/// 1. Incoming transactions, which include:
+/// 1. Incoming transactions. The type of transaction, parameters and proof are given in the `data` field of this transaction.
+///    Supported incoming transactions are:
 ///     - Validator
 ///         * Create
 ///         * Update
@@ -18,13 +19,13 @@ use crate::{SignatureProof, Transaction, TransactionError};
 ///         * Create
 ///         * Update
 ///         * AddStake
-///     The type of transaction, parameters and proof are given in the data field of the transaction.
-/// 2. Outgoing transactions, which include:
+///
+/// 2. Outgoing transactions. The type of transaction, parameters and proof are given in the `proof` field of this transaction.
+///    Supported outgoing transactions are:
 ///     - Validator
 ///         * Delete
 ///     - Staker
 ///         * RemoveStake
-///     The type of transaction, parameters and proof are given in the proof field of the transaction.
 ///
 /// It is important to note that all `signature` fields contain the signature
 /// over the complete transaction with the `signature` field set to `Default::default()`.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -170,6 +170,7 @@ clippy.empty_docs = "allow" # false positives: https://github.com/rust-lang/rust
 clippy.large_enum_variant = "allow"
 clippy.too_many_arguments = "allow"
 clippy.type_complexity = "allow"
+rust.unexpected_cfgs = { level = "warn", check-cfg = ['cfg(tokio_unstable)'] }
 rust.unused_qualifications = "warn"
 
 [workspace.dependencies]
--- a/bls/src/types/public_key.rs
+++ b/bls/src/types/public_key.rs
@@ -60,11 +60,8 @@ impl PublicKey {
     /// and one bit indicating if it is the "point-at-infinity".
     pub fn compress(&self) -> CompressedPublicKey {
         let mut buffer = [0u8; CompressedPublicKey::SIZE];
-        CanonicalSerialize::serialize_compressed(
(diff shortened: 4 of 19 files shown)


database/database-value/src/lib.rs
```diff
@@ -98,7 +98,7 @@ macro_rules! as_db_bytes {
                     #[allow(clippy::size_of_in_element_count)]
                     Cow::Borrowed(slice::from_raw_parts(
                         self as *const $typ as *const u8,
-                        mem::size_of::<$typ>(),
+                        size_of::<$typ>(),
```
hrxi: What's the minimum required Rust version for this?

lib/src/extras/logging.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(unexpected_cfgs)]
```
hrxi: Could you fix this warning instead? I think the warning can be quite useful in detecting typos.

primitives/transaction/src/account/staking_contract/structs.rs
```diff
@@ -18,13 +18,16 @@ use crate::{SignatureProof, Transaction, TransactionError};
 ///         * Create
 ///         * Update
 ///         * AddStake
-///     The type of transaction, parameters and proof are given in the data field of the transaction.
+///
+/// The type of transaction, parameters and proof are given in the data field of the transaction.
```
hrxi: This fix seems incorrect(?). It looks like it should belong inside the `1.`.

primitives/transaction/src/account/staking_contract/structs.rs
```diff
@@ -18,13 +18,16 @@ use crate::{SignatureProof, Transaction, TransactionError};
...
+///
 /// 2. Outgoing transactions, which include:
 ///     - Validator
 ///         * Delete
 ///     - Staker
 ///         * RemoveStake
-///     The type of transaction, parameters and proof are given in the proof field of the transaction.
+///
+/// The type of transaction, parameters and proof are given in the proof field of the transaction.
```
hrxi: Same as above.

bls/src/types/compressed_signature.rs
```diff
@@ -93,7 +93,7 @@ impl fmt::Display for CompressedSignature {
 impl From<G1Projective> for CompressedSignature {
     fn from(signature: G1Projective) -> Self {
         let mut buffer = [0u8; SIZE];
-        CanonicalSerialize::serialize_compressed(&signature.into_affine(), &mut &mut buffer[..])
+        CanonicalSerialize::serialize_compressed(&signature.into_affine(), &mut buffer[..])
```
hrxi, line 96: What warning does this fix?

hrxi (commented):
Can you add the names of the clippy warnings that you fix to the commit description?

bls/src/types/compressed_signature.rs
Eligioo:
```
warning: the borrowed expression implements the required traits
  --> bls/src/types/compressed_signature.rs:96:76
   |
96 |         CanonicalSerialize::serialize_compressed(&signature.into_affine(), &mut &mut buffer[..])
   |                                                                            ^^^^^^^^^^^^^^^^^^^^ help: change this to: `&mut buffer[..]`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrows_for_generic_args
   = note: `#[warn(clippy::needless_borrows_for_generic_args)]` on by default
```

primitives/transaction/src/account/staking_contract/structs.rs
Eligioo: It's part of `1.` but being on the next line directly after `* AddStake` it thinks its part of the list thus needs to be indented up to the same level.

primitives/transaction/src/account/staking_contract/structs.rs
Eligioo: Commented above

database/database-value/src/lib.rs
Eligioo: Rust 1.80: https://github.com/rust-lang/rust/pull/123168

lib/src/extras/logging.rs
Eligioo: It is regarding the `tokio_unstable` feature that isn't defined in the `Cargo.toml` but can be passed as a `RUSTFLAGS` during compilation.

Another way I could possibly fix this is by adding a workspace lint that was introduced on purpose for the `unexpected_cfgs` https://github.com/rust-lang/cargo/pull/13913/. Do you prefer that?

hrxi: Yes, I prefer that. `#[allow(unexpected_cfgs)]` does not fix the warning but just hides it. You can use the lint configuration you linked to declare what cfgs we expect in our code. This will allow us to catch typos, because it's declaring a whitelist of lints we expect.

Eligioo: Fixed

primitives/transaction/src/account/staking_contract/structs.rs
hrxi: Before: image.
After: image.

Both are incorrect, as far as I can tell.

Before, it was part of `* AddStake` which is clearly incorrect. Afterwards, it's now a free-standing text, but it should be part of the first enumeration item "1.".

I don't know if that can be fixed properly.

If not, perhaps we could reword the text so that

> The type of transaction, parameters and proof are given in the data field of the transaction.

can right after the "1."?

database/database-value/src/lib.rs
hrxi: Hm. Sounds a bit annoying to bump up our MSRV just to silence a warning. Perhaps we can simply disable the warning for now?

primitives/transaction/src/account/staking_contract/structs.rs
Eligioo: Addressed

primitives/transaction/src/account/staking_contract/structs.rs
Eligioo: Addressed

database/database-value/src/lib.rs
Eligioo: Fixed by allowing this Clippy lint

hrxi (approved):
Thanks!

PR #2785, opened 2024-07-26
Fixing parallel feature for the primitives crate.

ii-cruz:
The parallel feature for the primitives crate was not functional. This PR fixes that. Additionally, the prover feature could use some refinement that is addressed in this PR.

## Pull request checklist


Eligioo (commented):
LGTM.

rust-analyzer always does weird things when it comes to editing `Cargo.toml` files as we usually keep everything on the same line. You may can fix that.

primitives/src/slots_allocation.rs
```diff
@@ -227,7 +228,7 @@ impl Hash for Validators {
         #[cfg(not(feature = "parallel"))]
         let iter = public_keys.iter();
         #[cfg(feature = "parallel")]
-        let iter = self.public_keys.par_iter();
+        let iter = public_keys.par_iter();
```
paberr, line 231: good catch

paberr (approved):

PR #2787, opened 2024-07-29
Test that we don't require the latest stable Rust version

hrxi:
I chose 1.78.0 because it's two versions behind the latest Rust version and I've seen other Rust projects choose that.

We can bump that any time, I'd like to ensure that bumps in our MSRV are visible in PRs.

### DIFF
--- a/.github/workflows/build+test.yml
+++ b/.github/workflows/build+test.yml
@@ -26,7 +26,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
-    - uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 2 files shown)


Eligioo:
Maybe add to the README a paragraph how we currently have setup our MSRV with a disclaimer this policy may change in the future?

jsdanielh (approved):
LGTM. I just add a comment that we fixed the version here to test the MSRV

hrxi:
Addressed the comments.

jsdanielh (approved):

Eligioo (approved):

PR #2791, opened 2024-07-30
Change starting of rounds from the wrong current id to the correct given id.

nibhar:
A bug I introduced in #2763.

After changing `start_aggregation` to take the `id` as an argument, it would still invoke `Protocol::start_aggregation` with the current step and current round. This would not only not start the round which needed starting, but also panics if the `current_step` is `Step::Propose`. This fixes it.


jsdanielh (approved):
LGTM

sisou:
Would it be easy to add a test for this bug?

hrxi (approved):
LGTM, thanks!

hrxi:
Rebased for merging.

Issue #2550, opened 2024-05-31
Audit early returns in `Future::poll` functions in the codebase


hrxi:
This can easily starve sub-futures if not handled carefully.

hrxi:
- [ ] consensus/src/consensus/head_requests.rs
- [ ] consensus/src/consensus/mod.rs
- [ ] consensus/src/consensus/remote_event_dispatcher.rs
- [ ] consensus/src/sync/history/cluster.rs
- [ ] consensus/src/sync/history/sync_stream.rs
- [ ] consensus/src/sync/light/sync.rs
- [ ] consensus/src/sync/light/sync_stream.rs
- [ ] consensus/src/sync/live/block_queue/block_request_component.rs
- [ ] consensus/src/sync/live/block_queue/mod.rs
- [ ] consensus/src/sync/live/diff_queue/mod.rs
- [ ] consensus/src/sync/live/mod.rs
- [ ] consensus/src/sync/live/state_queue/chunk_request_component.rs
- [ ] consensus/src/sync/live/state_queue/mod.rs
- [ ] consensus/src/sync/sync_queue.rs
- [ ] consensus/src/sync/syncer.rs
- [ ] consensus/src/sync/syncer_proxy.rs
- [ ] consensus/tests/history.rs
- [ ] handel/src/aggregation.rs
- [ ] handel/src/network.rs
- [ ] handel/src/todo.rs
- [ ] mempool/src/executor.rs
- [ ] network-libp2p/src/connection_pool/bans.rs
- [ ] network-libp2p/src/discovery/message_codec/reader.rs
- [ ] network-libp2p/src/network.rs
- [ ] tendermint/src/tendermint.rs
- [x] test-utils/src/mock_node.rs
- [x] validator/src/aggregation/skip_block.rs
- [x] validator/src/macro.rs
- [x] validator/src/proposal_buffer.rs
- [x] validator/src/validator.rs
- [x] zkp-component/src/zkp_component.rs
- [x] zkp-component/src/zkp_prover.rs
- [x] zkp-component/src/zkp_requests.rs

PR #2798, opened 2024-08-02
Add our own versions of `FuturesUnordered` and `FuturesOrdered`

hrxi:
Unlike their `futures_util::stream::*` counterparts, they wake themselves when some future is pushed.

This gets rid of a couple of manual waker instances.

It also fixes some waking-related bugs (CC #2550).

Use clippy to warn for usages of the old types.

### DIFF
--- a/consensus/src/sync/live/state_queue/chunk_request_component.rs
+++ b/consensus/src/sync/live/state_queue/chunk_request_component.rs
@@ -4,7 +4,7 @@ use std::{
     task::{Context, Poll},
 };
 
-use futures::{FutureExt, Stream, StreamExt};
+use futures::{future::BoxFuture, FutureExt, Stream, StreamExt};
 use nimiq_network_interface::{network::Network, request::RequestError};
 use nimiq_primitives::key_nibbles::KeyNibbles;
 use parking_lot::RwLock;
@@ -71,6 +71,14 @@ impl<N: Network> ChunkRequestComponent<N> {
         request: RequestChunk,
     ) -> Result<ResponseChunk, RequestError> {
         network.request::<RequestChunk>(request, peer_id).await
+    }
+
+    /// Returns a future that resolves when the peer list of the chunk request
+    /// component becomes nonempty.
+    ///
+    /// Returns `None` is the chunk request component already has peers.
+    pub fn wait_for_peers(&self) -> Option<BoxFuture<'static, ()>> {
+        self.peers.read().wait_for_peers()
     }
 }
 
--- a/consensus/src/sync/live/state_queue/mod.rs
+++ b/consensus/src/sync/live/state_queue/mod.rs
@@ -9,7 +9,7 @@ use std::{
     task::{Context, Poll},
 };
 
-use futures::{stream::BoxStream, Stream, StreamExt};
+use futures::{future::BoxFuture, stream::BoxStream, Stream, StreamExt};
 use nimiq_block::Block;
 use nimiq_blockchain::Blockchain;
 use nimiq_blockchain_interface::{AbstractBlockchain, BlockchainEvent};
@@ -29,7 +29,7 @@ use serde::{Deserialize, Serialize};
 use self::chunk_request_component::ChunkRequestComponent;
 use super::{
     block_queue::BlockAndId,
-    queue::{ChunkAndId, LiveSyncQueue, QueueConfig},
+    queue::{ChunkAndId, QueueConfig},
 };
 use crate::sync::live::diff_queue::{DiffQueue, QueuedDiff};
 
@@ -173,6 +173,14 @@ pub struct StateQueue<N: Network> {
 
     /// The blockchain event stream.
     blockchain_rx: BoxStream<'static, BlockchainEvent>,
+
+    /// Waiter for the peer list to become nonempty.
+    ///
+    /// Since we only want to dispatch requests from the
+    /// `ChunkRequestComponent` when its peer list is nonempty, we need some
+    /// notification mechanism to wake us up once the list becomes nonempty if
+    /// we find it empty.
+    peers_became_nonempty: Option<BoxFuture<'static, ()>>,
 }
 
 impl<N: Network> StateQueue<N> {
@@ -211,6 +219,7 @@ impl<N: Network> StateQueue<N> {
             current_macro_height,
             start_key,
             blockchain_rx,
+            peers_became_nonempty: None,
         }
     }
 
@@ -574,8 +583,22 @@ impl<N: Network> Stream for StateQueue<N> {
             }
         }
 
+        // Check if we have peers.
+        if self.peers_became_nonempty.is_none() {
+            self.peers_became_nonempty = self.chunk_request_component.wait_for_peers();
+        }
+        if let Some(peers_became_nonempty) = &mut self.peers_became_nonempty {
+            if peers_became_nonempty.as_mut().poll(cx).is_ready() {
+                self.peers_became_nonempty = None;
+            }
+        }
+        // Obvious TOCTOU, but it would otherwise need to lock the
+        // `chunk_request_component`'s peer list.
+        //
         // Request chunks via ChunkRequestComponent.
-        if !self.chunk_request_component.has_pending_requests() && self.num_peers() > 0 {
+        if self.peers_became_nonempty.is_none()
+            && !self.chunk_request_component.has_pending_requests()
+        {
             self.request_chunk();
         }
 
--- a/consensus/src/sync/peer_list.rs
+++ b/consensus/src/sync/peer_list.rs
@@ -1,6 +1,8 @@
-use std::{collections::HashSet, fmt, ops::Index, slice::SliceIndex};
+use std::{collections::HashSet, fmt, ops::Index, slice::SliceIndex, sync::Arc};
 
+use futures::future::BoxFuture;
 use nimiq_network_interface::network::Network;
+use tokio::sync::Notify;
 
 /// A list of peers to be used while syncing.
 /// This contains an ordered list of peers as well as a hashmap.
@@ -9,6 +11,8 @@ use nimiq_network_interface::network::Network;
 pub struct PeerList<N: Network> {
     peers_set: HashSet<N::PeerId>,
     peers: Vec<N::PeerId>,
+    /// Used to notify listeners when the peer list becomes nonempty.
+    notify_nonempty: Arc<Notify>,
 }
 
 /// Stores an index into a [`PeerList`].
@@ -58,6 +62,7 @@ impl<N: Network> Default for PeerList<N> {
         Self {
             peers_set: Default::default(),
             peers: Default::default(),
+            notify_nonempty: Default::default(),
         }
     }
 }
@@ -67,6 +72,7 @@ impl<N: Network> Clone for PeerList<N> {
         Self {
             peers_set: self.peers_set.clone(),
             peers: self.peers.clone(),
+            notify_nonempty: Default::default(),
         }
     }
 }
@@ -74,6 +80,9 @@ impl<N: Network> Clone for PeerList<N> {
 impl<N: Network> PeerList<N> {
     pub fn add_peer(&mut self, peer_id: N::PeerId) -> bool {
         if self.peers_set.insert(peer_id) {
+            if self.peers.is_empty() {
+                self.notify_nonempty.notify_waiters();
+            }
             self.peers.push(peer_id);
             return true;
         }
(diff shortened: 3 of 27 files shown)


jsdanielh (commented):
Looks like we're still using the `futures`'s version in `consensus/src/sync/syncer.rs`.

consensus/src/sync/live/state_queue/mod.rs
```diff
@@ -173,6 +173,8 @@ pub struct StateQueue<N: Network> {
 
     /// The blockchain event stream.
     blockchain_rx: BoxStream<'static, BlockchainEvent>,
+
+    nonempty_peers: Option<BoxFuture<'static, ()>>,
```
Eligioo: Add rustdoc

consensus/src/sync/live/state_queue/mod.rs
```diff
@@ -568,9 +571,24 @@ impl<N: Network> Stream for StateQueue<N> {
...
+            }
+        }
+        // Obvious TOCTOU, but it would otherwise need to lock the
+        // `chunk_request_component`'s peer list.
+        //
         // Request chunks via ChunkRequestComponent.
-        if !self.chunk_request_component.has_pending_requests() && self.num_peers() > 0 {
-            self.request_chunk();
+        if self.nonempty_peers.is_none() && !self.chunk_request_component.has_pending_requests() {
+            println!("{}", self.request_chunk());
```
Eligioo: Forgot to remove the `println!`?

consensus/src/sync/peer_list.rs
```diff
@@ -58,22 +60,27 @@ impl<N: Network> Default for PeerList<N> {
...
             peers: Default::default(),
+            notify_nonempty: Default::default(),
         }
     }
 }
 
 impl<N: Network> Clone for PeerList<N> {
-    fn clone(&self) -> Self {
-        Self {
+    fn clone(&self) -> PeerList<N> {
```
Eligioo: Why can't both lines be `Self`?

Eligioo (commented):
LGTM, some minor feedback points.

CC #2832 to use this new abstraction once this PR is in.

consensus/src/sync/live/state_queue/mod.rs
hrxi: Thanks, fixed.

consensus/src/sync/peer_list.rs
hrxi: Can both be `Self`. I deleted the implementation and recreated it during the creation of the PR, so I used the more explicit style. Interesting that it still ended up having such a small diff. I can change it to `Self` if you want.

consensus/src/sync/live/state_queue/mod.rs
hrxi: Thanks, fixed.

consensus/src/sync/peer_list.rs
Eligioo: It's nitpicking but by having `Self` it can stay `Self`, even if the name of the struct changes :)

hrxi: Changed it.

Eligioo (approved):

utils/src/stream.rs
```diff
@@ -0,0 +1,114 @@
...
+    task::{Context, Poll, Waker},
+};
+
+use futures::{stream as inner, Stream};
+use pin_project::pin_project;
+
+use crate::WakerExt as _;
+
+#[pin_project]
+pub struct FuturesOrdered<F: Future> {
```
jsdanielh, line 22: I'd add some `rustdoc`. Something along these lines:
```
/// Wrapper over a `FuturesOredered` that auto-wakes on push
```

utils/src/stream.rs
```diff
@@ -0,0 +1,114 @@
...
+    type Item = F::Output;
+    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<F::Output>> {
+        let this = self.project();
+        this.waker.store_waker(cx);
+        this.inner.poll_next(cx)
+    }
+}
+
+#[pin_project]
+pub struct FuturesUnordered<F: Future> {
```
jsdanielh, line 89: Same as above. I'd add some `rustdoc`

utils/src/lib.rs
```diff
@@ -18,6 +18,8 @@ pub mod tagged_signing;
 #[cfg(feature = "time")]
 pub mod time;
 
+pub mod stream;
```
jsdanielh, line 22: Shouldn't we add a feature such that we don't pull unnecessary dependencies when we depend on this crate and we don't need anything on the stream module?

consensus/src/sync/live/state_queue/chunk_request_component.rs
```diff
@@ -72,6 +73,11 @@ impl<N: Network> ChunkRequestComponent<N> {
     ) -> Result<ResponseChunk, RequestError> {
         network.request::<RequestChunk>(request, peer_id).await
     }
+
+    pub fn notify_nonempty_peers(&self) -> Option<Arc<Notify>> {
```
jsdanielh: Could you add some `rustdoc`?

consensus/src/sync/live/state_queue/mod.rs
```diff
@@ -568,8 +577,23 @@ impl<N: Network> Stream for StateQueue<N> {
             }
         }
 
+        // Check if we have peers.
+        if self.nonempty_peers.is_none() {
```
jsdanielh: Isn't it better to use `get_or_insert`?

consensus/src/sync/live/state_queue/mod.rs
```diff
@@ -568,8 +577,23 @@ impl<N: Network> Stream for StateQueue<N> {
...
+            self.nonempty_peers = self
+                .chunk_request_component
+                .notify_nonempty_peers()
+                .map(|notify| Box::pin(async move { notify.notified().await }) as BoxFuture<()>);
+        }
+        if let Some(nonempty_peers) = &mut self.nonempty_peers {
+            if nonempty_peers.as_mut().poll(cx).is_ready() {
```
jsdanielh: Can't we just `take` it?

consensus/src/sync/peer_list.rs
```diff
@@ -130,6 +137,10 @@ impl<N: Network> PeerList<N> {
         peer_index.index = peer_index.index.wrapping_add(1) % self.peers.len();
         Some(self.peers[peer_index.index])
     }
+
+    pub fn notify_nonempty(&self) -> Arc<Notify> {
```
jsdanielh: Could you add some `rustdoc`?

consensus/src/sync/live/state_queue/mod.rs
hrxi: We only want to remove the future from the member variable when it resolves. `Option::take` can't do that.

consensus/src/sync/live/state_queue/mod.rs
hrxi: This only inserts something if `self.chunk_request_component.notify_nonempty_peers()` returns a non-`None` value. This doesn't fit `Option::get_or_insert`'s function signature.

consensus/src/sync/peer_list.rs
hrxi: Done.

utils/src/stream.rs
hrxi: Done.

utils/src/stream.rs
hrxi: Done.

consensus/src/sync/live/state_queue/chunk_request_component.rs
hrxi: Done.

consensus/src/sync/live/state_queue/mod.rs
```diff
@@ -173,6 +173,14 @@ pub struct StateQueue<N: Network> {
...
     /// The blockchain event stream.
     blockchain_rx: BoxStream<'static, BlockchainEvent>,
+
+    /// Waiter for the peer list to become nonempty.
+    ///
+    /// Since we only want to dispatch requests from the
+    /// `ChunkRequestComponent` when its peer list is nonempty, we need some
+    /// notification mechanism to wake us up once the list becomes nonempty if
+    /// we find it empty.
+    nonempty_peers: Option<BoxFuture<'static, ()>>,
```
nibhar: I would prefer this to be named `peers_became_nonempty` or similar. This field does not hold any peers itself, but the name implies it to some extend.

consensus/src/sync/peer_list.rs
```diff
@@ -67,13 +70,17 @@ impl<N: Network> Clone for PeerList<N> {
...
             peers: self.peers.clone(),
+            notify_nonempty: Default::default(),
         }
     }
 }
 
 impl<N: Network> PeerList<N> {
     pub fn add_peer(&mut self, peer_id: N::PeerId) -> bool {
         if self.peers_set.insert(peer_id) {
+            if self.peers.is_empty() {
```
nibhar, line 83: This seems weird, why does the set need to be empty to notify nonempty? Is there a guarantee that you cannot register on this Notify when the peer list is already non empty? 

Would there be any harm in just moving this down one line and inverting the condition?

That way whenever the set gets added to it will trigger a notify_waiters, which there should not be any if it was previously not empty, but if there are they now know it is (still) not empty. Otherwise they will be waiting forever as long as the list never gets empty again.

consensus/src/sync/peer_list.rs
```diff
@@ -9,6 +10,7 @@ use nimiq_network_interface::network::Network;
 pub struct PeerList<N: Network> {
     peers_set: HashSet<N::PeerId>,
     peers: Vec<N::PeerId>,
+    notify_nonempty: Arc<Notify>,
```
nibhar, line 15: This could probably do with some comment. I am also not too happy about the name, but right now have no better suggestion.

consensus/src/sync/live/state_queue/mod.rs
nibhar: `notify_nonempty_peers()` can return `None` thus the following assignment can be to `None` again, which `get_or_insert` does not allow for.

consensus/src/sync/sync_queue.rs
```diff
@@ -22,7 +20,7 @@ use crate::sync::peer_list::PeerListIndex;
 
 #[pin_project]
 #[derive(Debug)]
-struct OrderWrapper<TId, TOutput> {
+struct OrderWrapper<TId: Clone, TOutput> {
```
nibhar: Why does this need to be `Clone`? I cannot find it as new requirement.

utils/src/lib.rs
hrxi: Done, although I doubt its usefulness. All of our crates depend on `futures` directly or indirectly.

consensus/src/sync/sync_queue.rs
```diff
@@ -452,14 +450,24 @@ where
                         }
                     }
                 }
-                None => {
+                Poll::Ready(None) => {
                     return if self.ids_to_request.is_empty() || self.peers.read().is_empty() {
-                        Poll::Ready(None)
+                        self.waker.store_waker(cx);
+                        Poll::Pending
```
nibhar: The return could be moved out of the conditional as they are now the same. As half the condition is repeated in the else case this can maybe also be combined.

consensus/src/sync/live/state_queue/mod.rs
hrxi: Done.

consensus/src/sync/peer_list.rs
hrxi: Added a comment.

consensus/src/sync/peer_list.rs
hrxi: > This seems weird, why does the set need to be empty to notify nonempty?

The set only becomes nonempty when it was previously empty.

> Is there a guarantee that you cannot register on this Notify when the peer list is already non empty?

Not directly here, although I guess I could add it. The `notify_nonempty_peers` ensures it by returning an `Option` that only contains a notifier when the set is empty. I guess I should move that guarantee to this function.

> Would there be any harm in just moving this down one line and inverting the condition?
> 
> That way whenever the set gets added to it will trigger a notify_waiters, which there should not be any if it was previously not empty, but if there are they now know it is (still) not empty. Otherwise they will be waiting forever as long as the list never gets empty again.

The harm would mostly be hiding bugs and unnecessary wakeups. We could also wake this from many other places, so it's not clear that "adding a peer to an otherwise nonempty list" should trigger the "peer list became nonempty" any more than, say, "removing a peer from a list, and it's still nonempty".

consensus/src/sync/sync_queue.rs
hrxi: True, removed it again.

consensus/src/sync/sync_queue.rs
hrxi: Did the former. Not sure how to do the latter.

consensus/src/sync/peer_list.rs
hrxi: > Is there a guarantee that you cannot register on this Notify when the peer list is already non empty?

I added that guarantee to the API now. Thanks for bringing it up.

jsdanielh (approved):

nibhar (approved):

jsdanielh:
Rebasing branch to merge it

PR #2801, opened 2024-08-02
Don't swallow return value in `spawn` and use our `spawn`

hrxi:
Add `nimiq_utils::spawn_local` for spawns on the local task.

Also rename `nimiq_utils::spawn::spawn` to `nimiq_utils::spawn` for less stuttering.


jsdanielh (approved):
LGTM

PR #2803, opened 2024-08-05
Go back to upstream libp2p-websocket-websys

### LINKED ISSUE #5490: websocket-websys: Use-after-free of onclose and onerror event handlers
### Summary

Already mentioned in #5229, but overlooked (my fault), `onclose` and sometimes also `onerror` event handlers are called from JS after they have been dropped in Rust.

### Expected behavior

No errors thrown.

### Actual behavior

The Websocket connection gets closed inside the `drop` implementation for the socket, which then still triggers a `close` event on the Websocket in JS, for which the handler in Rust has however already been dropped:

```
Error: closure invoked recursively or after being dropped
```

This is an uncaught exception, which only produces console output in browsers, but makes NodeJS programs exit.

### Possible Solution

A simple solution would be to unregister the relevant event handlers in the Drop implementation, like so:

```rust
self.inner.socket.set_onclose(None);
self.inner.socket.set_onerror(None);
```

I have done this in my fork for now: https://github.com/sisou/rust-libp2p/commit/1b14ee1a5ae5375a11731328afce361fafa8ad6a#diff-c6771ff4ec042ffe20b4e526aa5be035c9e3d532f3edbd965e47541ab56fdffd

According to @nibhar, there must be a better way, though. He suggested storing the socket's event handlers outside the connection, having them clean-up themselves once the `onclose` handler is called.

### Version

- libp2p-core: 0.41.2
- libp2p-websocket-websys: 0.3.2

### Would you like to work on fixing this bug ?

Yes

sisou:
As they fixed the issue in v0.3.3: https://github.com/libp2p/rust-libp2p/pull/5521#issuecomment-2265755680

## Pull request checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3177,8 +3177,9 @@ dependencies = [
 
 [[package]]
 name = "libp2p-websocket-websys"
-version = "0.3.2"
-source = "git+https://github.com/sisou/rust-libp2p?branch=sisou/websocket-websys-drop#1b14ee1a5ae5375a11731328afce361fafa8ad6a"
+version = "0.3.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f95cd8a32fcf94ad1e5c2de37c2a05a5a4188d8358b005859a0fc9e63b6953bc"
 dependencies = [
  "bytes",
  "futures",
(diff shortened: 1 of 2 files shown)


nibhar:
What about #2301? Is this going to be closed by the upstream as well?

sisou:
@nibhar No. Unfortunately, in https://github.com/libp2p/rust-libp2p/pull/4889 they made us change the implementation from being runtime agnostic to only accepting environments with a `Window` or `WorkerGlobalScope` (scopes supported by the `web-sys` crate). And Node doesn't have those. So that workaround is kept.

Eligioo (approved):

hrxi:
Thanks!

PR #2804, opened 2024-08-05
Node 22 built-in Websocket support

sisou:
Only set `global.WebSocket` when it doesn't yet exist. [Node v22 added a global `WebSocket` client](https://nodejs.org/en/blog/announcements/v22-release-announce#websocket), which is browser-compatible, so can be used instead. I tested that it indeed works with Node v22.

Additionally adds peer count and block height stats to the Node example applications, to see that the apps are working properly.


hrxi (approved):
Looks good to me.

jsdanielh (approved):
LGTM

PR #2805, opened 2024-08-05
RPC: Always include all fields for staker/validator

sisou:
- **Do not omit unset fields in Staker and Validator structs when returned from the RPC API**
  This is so users of these structs learn about these potentially set fields even while their staker/validator is running smoothly, so that they won't be surprised by unknown fields once they become set.
- **Return an error when requesting data from a block body, but the node doesn't have the block body**
  Previously this silently omitted the body data, which made the `include_body` flag seem broken.

## Pull request checklist

### DIFF
--- a/rpc-interface/src/types.rs
+++ b/rpc-interface/src/types.rs
@@ -901,7 +901,6 @@ impl Account {
 pub struct Staker {
     pub address: Address,
     pub balance: Coin,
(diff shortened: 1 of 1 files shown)


hrxi (approved):
LGTM.

jsdanielh (approved):

PR #2808, opened 2024-08-05
Adding the ZKP Ceremony verifying key

### LINKED ISSUE #1247: Securely generate ZKP keys
For the devnet, we simply create the ZKP keys from a fixed, public seed.
**It should be noted that this is insecure (but ok for the devnet)!**

For the mainnet, we should use a different procedure to generate the keys.
Also, we need a mechanism to distribute the proving keys (which are very large) to prover nodes.

ii-cruz:
Adapts the prove, verify, and setup binaries to use a default folder '.zkp_example' and adds a couple of miscellaneous fixes on these example binaries.
Adds the zkp ceremony verifying key to both testnet and mainnet folders.
When we restart we must update the prover node keys.

This closes #1247.

## Pull request checklist


hrxi (approved):
🎉

jsdanielh (approved):

Issue #2790, opened 2024-07-29
CI run - Panic Trying to revert macro block


ii-cruz:
Panic validator 2

CI run https://github.com/nimiq/core-rs-albatross/actions/runs/9922584262

Eligioo:
Closed with #2752.

In the following scenario:
- The first batch in the first epoch of the chain has some transactions,
- A macro block gets produced and we proceed to the next batch,
- This new batch doesn't contain any transactions and we try to push with history sync (`Blockchain::push_history_sync`)

The known history returned by getting the `get_nonfinal_epoch_transactions` goes over a macro block. Causing to create a `diverging_block` before the macro block we already have, when reverting to a common state, resulting into reverting beyond this macro block. Which shouldn't be possible.

A test should catch this, which I will create.

PR #2752, opened 2024-07-17
HistoryStore: fix `get_nonfinal_epoch_transactions()` returning incorrect results for batch 0

styppo:
Refactor related functions and remove dead/duplicate code.

Closes #2740

PR #2819, opened 2024-08-07
Add History Sync test to verify that it does not revert beyond a macro block

Eligioo:
Adds a test that should catch the panic `Panic Trying to revert macro block` observed in #2790. The issue is already solved by a previous PR.

## Pull request checklist

### DIFF
--- a/blockchain/tests/history_sync.rs
+++ b/blockchain/tests/history_sync.rs
@@ -13,6 +13,90 @@ use nimiq_test_utils::blockchain::{
 use nimiq_utils::time::OffsetTime;
 use parking_lot::RwLock;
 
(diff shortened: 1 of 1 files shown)


hrxi (commented):
I don't understand what the test tries to test. I don't see a revert in there. Can you add some comments, e.g. stating on top of the test what it tries to test. You can check the next function in the file for inspiration.

Eligioo:
@hrxi This test is about that it shouldn't revert the history in the case when you want to push an empty batch through history sync. Searching for a common state based on the history we already have and the new history we receive, is part of the `Blockchain::push_history_sync`. Before #2752 this reverting mechanism was broken and resulted into reverting beyond a macro block.

I clarified the test with documentation.

hrxi (approved):
Thanks for the clear explanation! :)

PR #2823, opened 2024-08-08
Optimize the chain store block table

jsdanielh:
Optimize the chain store block table by differentiating pushed vs stored blocks in the chain store.
Pushed blocks are partially restored and are rebuilt by using the chain info DB and the history store. Stored blocks are fully stored since there is no associated history.
The `chain_info.on_main_chain` flag is used to differentiate where to store the block and how to rebuild it (if needed).

## Pull request checklist

### DIFF
--- a/blockchain/src/blockchain/history_sync.rs
+++ b/blockchain/src/blockchain/history_sync.rs
@@ -378,7 +378,7 @@ impl Blockchain {
         );
 
         this.chain_store
-            .put_chain_info(&mut txn, &block_hash, &chain_info, true);
+            .put_chain_info(&mut txn, &block_hash, &chain_info, false);
 
         let wanted_history_root = this
             .history_store
--- a/blockchain/src/blockchain/rebranch_utils.rs
+++ b/blockchain/src/blockchain/rebranch_utils.rs
@@ -166,6 +166,13 @@ impl Blockchain {
                 );
             }
 
+            // Remove block from chain store (we will add it fully later on with its full body)
+            self.chain_store.remove_chain_info(
+                write_txn,
+                &current.0,
+                current.1.head.block_number(),
+            );
+
             // Block was reverted, add it to the reverted chain collection.
             revert_chain.push(current);
 
@@ -225,8 +232,10 @@ impl Blockchain {
             reverted_block.1.on_main_chain = false;
             reverted_block.1.main_chain_successor = None;
 
+            // Block isn't pushed since it was reverted (no history associated) and we need to store fully the block again
+            // since it was partially stored before and removed from the `pushed` set of blocks above.
             self.chain_store
-                .put_chain_info(write_txn, &reverted_block.0, &reverted_block.1, false);
+                .put_chain_info(write_txn, &reverted_block.0, &reverted_block.1, true);
         }
 
         // Update the main_chain_successor of the common ancestor block.
@@ -246,9 +255,11 @@ impl Blockchain {
             fork_block.1.on_main_chain = true;
             fork_block.1.main_chain_successor = main_chain_successor;
 
-            // Include the body of the new block (at position 0).
+            // Block is pushed and we have history for it, so we make sure that block ends in the `pushed` block table (by setting `chain_info.on_main_chain`).
+            // The block was probably also previously stored in the `stored` block table (inferior chain path) but it is not pruned since worst case
+            // scenario it will be removed on a batch finalization.
             self.chain_store
-                .put_chain_info(write_txn, &fork_block.0, &fork_block.1, i == 0);
+                .put_chain_info(write_txn, &fork_block.0, &fork_block.1, true);
         }
 
         // Update the head.
--- a/blockchain/src/chain_store.rs
+++ b/blockchain/src/chain_store.rs
@@ -1,21 +1,119 @@
+use std::sync::Arc;
+
 use nimiq_account::RevertInfo;
-use nimiq_block::Block;
+use nimiq_block::{Block, BlockType, EquivocationProof, MacroBody, MicroBody};
 use nimiq_blockchain_interface::{BlockchainError, ChainInfo, Direction};
 use nimiq_database::{
     declare_table,
     mdbx::{MdbxDatabase, MdbxReadTransaction, MdbxWriteTransaction, OptionalTransaction},
     traits::{Database, DupReadCursor, ReadCursor, ReadTransaction, WriteCursor, WriteTransaction},
 };
+use nimiq_database_value_derive::DbSerializable;
 use nimiq_hash::Blake2bHash;
 use nimiq_primitives::{policy::Policy, trie::trie_diff::TrieDiff};
+use nimiq_serde::{Deserialize, Serialize};
+use nimiq_transaction::{historic_transaction::HistoricTransactionData, reward::RewardTransaction};
+
+use crate::{history::interface::HistoryInterface, history_store_proxy::HistoryStoreProxy};
 
 declare_table!(HeadTable, "Head", () => Blake2bHash);
 declare_table!(ChainTable, "ChainData", Blake2bHash => ChainInfo);
-declare_table!(BlockTable, "Block", Blake2bHash => Block);
+declare_table!(PushedBlockTable, "PushedBlockTable", Blake2bHash => PushedBlock);
+declare_table!(StoredBlockTable, "StoredBlockTable", Blake2bHash => Block);
 declare_table!(HeightIndex, "HeightIndex", u32 => dup(Blake2bHash));
 declare_table!(RevertTable, "Receipts", u32 => RevertInfo);
 declare_table!(AccountsDiffTable, "AccountsDiff", Blake2bHash => TrieDiff);
 
+/// The non-header content of a block except that transactions are not stored to
+/// optimize blocks storage. This assumes that a block has been pushed and that there
+/// is history associated with this block.
+#[derive(Debug, Serialize, Deserialize, DbSerializable)]
+pub enum PushedBlock {
+    Macro,
+    Micro {
+        equivocation_proofs: Vec<EquivocationProof>,
+    },
+}
+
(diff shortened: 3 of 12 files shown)


sisou:
So blocks will be spread across three tables. Do all of them get pruned the same way? So if I request a past block in a full node that pruned that epoch, will I get the correct error? Or half a block?

styppo:
These changes would benefit from rebasing them onto #2722. Said PR moves `validators` and `next_batch_initial_punished_set` from the `MacroBody` to the `MacroHeader`, reducing the `MacroBody` to only the reward transactions. Thus, it should be possible to simplify this code by never storing block bodies (including macro blocks) and always recovering them from the `HistoryStore`.

jsdanielh:
> So blocks will be spread across three tables.

If you add the history store's history tree table, yes.

> Do all of them get pruned the same way?

No, they get pruned in a different way. For instance, the `stored_block_table` gets pruned on a batch finalization.

> So if I request a past block in a full node that pruned that epoch, will I get the correct error? Or half a block?

The correct error: No block found. The mechanism is the following:
1. Block is looked up in the `store_block_table`. If so, we return it.
2. Block is looked up in the `pushed_block_table`.
  2.1 If so and if it's a micro block, transactions are retrieved from the HS.
3. If both 1 and 2 are a miss an error is returned.

As mentioned before `stored_block_table` is pruned at the end of a batch and `pushed_block_table` might get pruned after an epoch finalization depending on the value of `config.max_epochs_stored` (which is also the case for the HS on full nodes).

jsdanielh:
> These changes would benefit from rebasing them onto #2722. Said PR moves `validators` and `next_batch_initial_punished_set` from the `MacroBody` to the `MacroHeader`, reducing the `MacroBody` to only the reward transactions. Thus, it should be possible to simplify this code by never storing block bodies (including macro blocks) and always recovering them from the `HistoryStore`.

Rebased and implemented @styppo

blockchain/src/blockchain/rebranch_utils.rs
```diff
@@ -248,7 +262,7 @@ impl Blockchain {
 
             // Include the body of the new block (at position 0).
             self.chain_store
-                .put_chain_info(write_txn, &fork_block.0, &fork_block.1, i == 0);
+                .put_chain_info(write_txn, &fork_block.0, &fork_block.1, i == 0, true);
```
styppo: This line looks a bit fishy. We're iterating over `target_chain` here, which are the blocks on the fork we're rebranching to, so these blocks are `stored` in the chain store with full body. When `i == 0` and thus `include_body = true`, the respective block will now also be stored as `pushed`, while all other blocks remain `stored`, even though they should now be considered `pushed`. This doesn't really cause any problems as the blocks can still be discovered, but it is inconsistent and leads to data duplication. However, it might actually be faster to duplicate the transactions instead of deleting them.

blockchain/src/chain_store.rs
```diff
@@ -1,21 +1,136 @@
...
 declare_table!(HeightIndex, "HeightIndex", u32 => dup(Blake2bHash));
 declare_table!(RevertTable, "Receipts", u32 => RevertInfo);
 declare_table!(AccountsDiffTable, "AccountsDiff", Blake2bHash => TrieDiff);
 
+/// The non-header content of a block except that transactions are not stored to
+/// optimize blocks storage. This assumes that a block has been pushed and that there
+/// is history associated with this block.
+#[derive(Debug, Serialize, Deserialize, DbSerializable)]
+pub struct PushedBlock {
+    justification: Option<BlockJustification>,
```
styppo: The `justification` is already stored as part of `ChainInfo`, so we actually don't need it here.

blockchain/src/chain_store.rs
```diff
@@ -1,21 +1,136 @@
...
+            justification: block.justification(),
+            body: block.body(),
+        }
+    }
+}
+
+impl PushedBlock {
+    fn block_from_header(
+        &self,
+        block: &Block,
```
styppo: We could save a bunch of `clone()`s by taking ownership of `self` and `block` here.

```suggestion
    fn into_block(
        self,
        block: Block,
```

blockchain/src/chain_store.rs
```diff
@@ -1,21 +1,136 @@
...
+        history_store: &Arc<HistoryStoreProxy>,
+        txn: &MdbxReadTransaction,
+    ) -> Block {
+        if let Some(justification) = &self.justification {
+            assert!(block.ty() == justification.ty());
```
styppo:
```suggestion
            assert_eq!(block.ty(), justification.ty());
```

blockchain/src/chain_store.rs
```diff
@@ -1,21 +1,136 @@
...
+        }
+        if let Some(body) = &self.body {
+            assert!(block.ty() == body.ty());
```
styppo:
```suggestion
            assert_eq!(block.ty(), body.ty());
```

blockchain/src/chain_store.rs
```diff
@@ -1,21 +1,136 @@
...
+        }
+        match block {
+            Block::Macro(header_only_block) => Block::Macro(MacroBlock {
+                header: header_only_block.header.clone(),
+                justification: self
+                    .justification
+                    .as_ref()
+                    .map(|justification| justification.clone().unwrap_macro()),
+                body: self.body.as_ref().map(|_| {
```
styppo: This is a bit strange. We could just completely ignore `self.body` in this case (or even better, get rid of it entirely for macro blocks)?

blockchain/src/chain_store.rs
```diff
@@ -1,21 +1,136 @@
...
 
+/// The non-header content of a block except that transactions are not stored to
+/// optimize blocks storage. This assumes that a block has been pushed and that there
+/// is history associated with this block.
+#[derive(Debug, Serialize, Deserialize, DbSerializable)]
+pub struct PushedBlock {
+    justification: Option<BlockJustification>,
+    #[serde(serialize_with = "PushedBlock::serialize_body")]
+    body: Option<BlockBody>,
+}
```
styppo:
```suggestion
pub enum PushedBlock {
    Macro,
    Micro {
        equivocation_proofs: Vec<EquivocationProof>,
    }
}
```
This could be an alternative way to model this. Technically, we wouldn't need to store anything for macro blocks at all when they are `pushed` (which they should always be).

styppo (approved):

blockchain/src/chain_store.rs
```diff
@@ -95,8 +205,12 @@ impl ChainStore {
         };
 
         if include_body {
-            if let Some(block) = txn.get(&self.block_table, hash) {
-                chain_info.head = block;
+            // Check the stored block table, then the pushed block table
```
styppo: We could eliminate one database fetch operation here by storing a `pushed` flag in `ChainInfo`, which tells us which block table to use.

jsdanielh: Just implemented this but used the already present `chain_info.on_main_chain` flag.

blockchain/src/blockchain/history_sync.rs
```diff
@@ -378,7 +378,7 @@ impl Blockchain {
         );
 
         this.chain_store
-            .put_chain_info(&mut txn, &block_hash, &chain_info, true);
+            .put_chain_info(&mut txn, &block_hash, &chain_info, false);
```
jsdanielh, line 381: @styppo I believe this can be `false` since we are already storing the `chain_info` with `chain_info.on_main_chain` and we're setting `include_body=true` at [line 172](https://github.com/nimiq/core-rs-albatross/pull/2823/files#diff-347ff6e4fe84ae58c6cf527d349b6acc2f21444c88e11efc6deb7028915daec4R172).

blockchain/src/chain_store.rs
```diff
@@ -1,21 +1,121 @@
...
+    }
+}
+
+impl PushedBlock {
+    fn ty(&self) -> BlockType {
+        match self {
+            PushedBlock::Macro => BlockType::Macro,
+            PushedBlock::Micro {
+                equivocation_proofs: _,
+            } => BlockType::Micro,
```
styppo:
```suggestion
            PushedBlock::Micro { .. } => BlockType::Micro,
```

blockchain/src/chain_store.rs
```diff
@@ -1,21 +1,121 @@
...
+        }
+    }
+    fn into_block(
+        self,
+        block: &Block,
+        history_store: &Arc<HistoryStoreProxy>,
+        txn: &MdbxReadTransaction,
+    ) -> Block {
+        assert_eq!(self.ty(), block.ty());
+        let mut block_to_return = block.clone();
```
styppo:
```suggestion
    fn populate_body(
        self,
        block: &mut Block,
        history_store: &Arc<HistoryStoreProxy>,
        txn: &MdbxReadTransaction,
    ) {
        assert_eq!(self.ty(), block.ty());
```
We can eliminate the `block.clone()` here by passing a mutable reference to the partial block.

blockchain/src/chain_store.rs
```diff
@@ -161,7 +287,18 @@ impl ChainStore {
...
         if include_body {
-            txn.put_reserve(&self.block_table, hash, &chain_info.head);
+            // If block is on main chain we store it in the `pushed` block table: we should be able to retrieve
+            // the block from the `chain_info` table and the history store.
+            // Otherwise it goes completely to the `stored` block table.
+            if chain_info.on_main_chain {
+                txn.put_reserve(
+                    &self.pushed_block_table,
+                    hash,
+                    &PushedBlock::from(chain_info.head.clone()),
```
styppo: The `clone` here is unnecessarily expensive since it clones the full block, but `PushedBlock::from` only needs the `equivocation_proofs` for micro blocks. It would be more efficient to implement `From<&Block>` and clone only the `equivocation_proofs` within the trait implementation.

styppo:
LGTM

PR #2832, opened 2024-08-15
Implementation of Mempool Syncer

### LINKED ISSUE #2625: Add a mechanism to sync the mempool

Eligioo:

The Mempool Syncer is responsible for acquiring the transactions that are currently in the mempool of other nodes directly after reaching consensus.

The mempool syncers get initiated once the Mempool Executors are spawned. One Syncer per Executor is started in order to sync the regular and control transactions separately. First the Syncer starts to discover which transaction hashes other peers with a mempool have and compare those with what we have locally. Then we distribute those unknown hashes among the peers we know have those transactions and retrieve the actual transaction. Lastly for every received transaction we do a full verification which should add it to our local mempool if everything checks out.

## Open design questions
- [x] Define the parameters for the Mempool Syncer: max transactions to sync, transactions per request, etc.
        **Max transactions we acquire per peer: 25.000
        Amount of transaction per request: 500**
- [x] Should the Mempool Syncer end its stream after x amount of time
        **After 10 minutes**

## Pull request checklist


mempool/src/executor.rs
```diff
@@ -114,16 +114,18 @@ impl<N: Network, T: Topic + Unpin + Sync> Future for MempoolExecutor<N, T> {
...
-                    Ok(_) => MsgAcceptance::Accept,
-                    // Reject the message if signature verification fails or transaction is invalid
-                    // for current validation window
-                    Err(VerifyErr::InvalidTransaction(_)) => MsgAcceptance::Reject,
-                    Err(VerifyErr::AlreadyIncluded) => MsgAcceptance::Reject,
-                    Err(_) => MsgAcceptance::Ignore,
-                };
-
-                network.validate_message::<T>(pubsub_id, acceptance);
+                if let Some(pubsub_id) = pubsub_id {
```
hrxi, line 157: We should still somehow punish the sender of an invalid transaction, even if it's not sent via gossipsub.

Eligioo: Added this. If some of the verification variants are encountered, the peer is banned.

sisou:
Why is this a breaking change (`testnet-restart` label)?

Eligioo:
> Why is this a breaking change (`testnet-restart` label)?

Because it introduces some new network request messages. Technically it is behind a feature flag because the syncer utilizes an yet unused service flag so these new request messages would only go to peers that also have this new component. So even though it's breaking, it shouldn't be too difficult to get in but for bookkeeping I kept the label around.

consensus/src/sync/syncer.rs
```diff
@@ -16,6 +16,7 @@ use nimiq_network_interface::network::{CloseReason, Network, NetworkEvent, Subsc
 use nimiq_primitives::policy::Policy;
 use nimiq_time::{interval, Interval};
 use nimiq_utils::stream::FuturesUnordered;
+use tokio::sync::broadcast::{channel as broadcast, Sender as BroadcastSender};
```
hrxi:
```suggestion
use tokio::sync::broadcast;
```
And then use `broadcast::channel()` and `broadcast::Sender`. We use it this way in the rest of the code, but I don't know how to enforce that with a clippy lint…

consensus/src/sync/syncer.rs
```diff
@@ -159,12 +168,15 @@ impl<N: Network, M: MacroSync<N::PeerId>, L: LiveSync<N>> Syncer<N, M, L> {
...
         let network_events = network.subscribe_events();
+        let (tx, _rx) = broadcast(256);
+
         Syncer {
             live_sync,
             macro_sync,
             blockchain,
             network,
             network_events,
+            events: tx,
```
hrxi:
```suggestion
            events: broadcast::Sender::new(256),
```

consensus/src/sync/syncer.rs
```diff
@@ -185,12 +197,17 @@ impl<N: Network, M: MacroSync<N::PeerId>, L: LiveSync<N>> Syncer<N, M, L> {
     pub fn move_peer_into_live_sync(&mut self, peer_id: N::PeerId) {
         debug!(%peer_id, "Adding peer to live sync");
         self.live_sync.add_peer(peer_id);
+        self.events.send(SyncEvent::AddLiveSync(peer_id)).ok();
```
hrxi:
```suggestion
        let _ = self.events.send(SyncEvent::AddLiveSync(peer_id));
```

Using `.ok()` doesn't communicate that you want to ignore the error. `.ok()` is used to convert the `Ok` case of a `Result<T, E>` into an `Option<T>`.

consensus/src/sync/syncer.rs
```diff
@@ -185,12 +197,17 @@ impl<N: Network, M: MacroSync<N::PeerId>, L: LiveSync<N>> Syncer<N, M, L> {
...
     }
 
     pub fn num_peers(&self) -> usize {
         self.live_sync.num_peers()
     }
 
+    pub fn broadcast_sender(&self) -> BroadcastSender<SyncEvent<<N as Network>::PeerId>> {
+        self.events.clone()
+    }
+
```
hrxi:
```suggestion
```

Seems unused. Should probably get a better name if it was used, maybe `events_sender`.

consensus/src/sync/syncer_proxy.rs
```diff
@@ -227,6 +228,11 @@ impl<N: Network> SyncerProxy<N> {
         gen_syncer_match!(self, accepted_block_announcements)
     }
 
+    /// Returns a broadcast sender for sync events
+    pub fn broadcast_sender(&self) -> BroadcastSender<SyncEvent<<N as Network>::PeerId>> {
+        gen_syncer_match!(self, broadcast_sender)
+    }
+
```
hrxi: Unused, too.

mempool/mempool-task/src/lib.rs
```diff
@@ -167,12 +195,39 @@ impl<N: Network> MempoolTask<N> {
...
 }
 
 impl<N: Network> Stream for MempoolTask<N> {
     type Item = MempoolEvent;
 
     fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        // Process consensus sync updates.
+        while let Poll::Ready(Some(Ok(event))) = self.sync_event_rx.poll_next_unpin(cx) {
+            self.on_consensus_sync_event(event)
+        }
```
hrxi, line 227: This should handle errors, but currently ignores them entirely.

mempool/mempool-task/src/lib.rs
```diff
@@ -167,12 +195,39 @@ impl<N: Network> MempoolTask<N> {
...
+
+        // Process network updates.
+        while let Poll::Ready(Some(Ok(event))) = self.network_event_rx.poll_next_unpin(cx) {
+            self.on_network_event(event)
+        }
```
hrxi, line 237: This should handle errors as well.

mempool/src/executor.rs
```diff
@@ -113,16 +123,37 @@ impl<N: Network, T: Topic + Unpin + Sync> Future for MempoolExecutor<N, T> {
...
-                    Err(_) => MsgAcceptance::Ignore,
-                };
-
-                network.validate_message::<T>(pubsub_id, acceptance);
+                match pubsub_or_peer_id {
+                    PubsubIdOrPeerId::PeerId(peer_id) => match verify_tx_ret {
+                        Ok(_) => (),
+                        Err(VerifyErr::InvalidAccount(_))
+                        | Err(VerifyErr::InvalidBlockNumber)
+                        | Err(VerifyErr::InvalidTransaction(_)) => {
```
hrxi: What's the reasoning for including specifically these errors for banning the client and not the others?

Can a transaction become invalid with one of these errors while we're requesting them because we apply a new block?

It'd be nice to have a comment above to explain which error types were chosen.

mempool/src/sync/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+    Pending,
+    Requested,
+    Received,
+}
+
+/// Struct for keeping track which peers have a specific hash in their mempool and
+/// store if we have a requested the corresponding transaction already to one of those peers.
+struct HashRequestStatus<N: Network> {
+    status: RequestStatus,
+    peer_ids: Vec<N::PeerId>,
```
hrxi:
```suggestion
    peer_ids: HashSet<N::PeerId>,
```
Looks like this wants fast lookup, so a `HashSet` would be more appropriate.

mempool/src/sync/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+const MAX_HASHES_PER_REQUEST: usize = 500;
+const MAX_TOTAL_HASHES: usize = 25_000;
+const SHUTDOWN_TIMEOUT_DURATION: Duration = Duration::from_secs(10 * 60); // 10 minutes
+
+/// Struct responsible for discovering hashes and retrieving transactions from the mempool of other nodes that have a mempool
+pub(crate) struct MempoolSyncer<N: Network> {
+    /// Timeout to gracefully shutdown the mempool syncer entirely
+    shutdown_timer: Pin<Box<Sleep>>,
+
+    /// Consensus sync event receiver
```
hrxi:
```suggestion
```
The comment doesn't tell me what it does, can be omitted.

mempool/src/sync/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+    consensus_sync_event_rx: BroadcastStream<SyncEvent<<N>::PeerId>>,
+
+    /// Blockchain reference
```
hrxi:
```suggestion
```

mempool/src/sync/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+    /// Retrieved transactions that are ready to get verified and pushed into our local mempool
+    transactions: VecDeque<(Transaction, N::PeerId)>,
+
+    /// Requests to other peers for fetching transactions by their hashes
+    transactions_requests: FuturesUnordered<
+        BoxFuture<'static, (N::PeerId, Result<ResponseMempoolTransactions, RequestError>)>,
+    >,
+
+    /// Collection of transaction hashes not present in the local mempool
+    unknown_hashes: HashMap<Blake2bHash, HashRequestStatus<N>>,
```
hrxi: I don't see anything removing transactions from this hashmap.

mempool/src/sync/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+        peers: Vec<N::PeerId>,
+        transaction_type: MempoolTransactionType,
+        blockchain: Arc<RwLock<Blockchain>>,
+        consensus: ConsensusProxy<N>,
+        mempool_state: Arc<RwLock<MempoolState>>,
+    ) -> Self {
+        let hashes_requests = peers
+            .iter()
+            .map(|peer_id| {
+                let peer_id = *peer_id;
```
hrxi:
```suggestion
            .map(|&peer_id| {
```

mempool/src/sync/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+
+    /// Push newly discovered hashes into the `unknown_hashes` and keep track which peers have those hashes
+    fn push_unknown_hashes(&mut self, hashes: Vec<Blake2bHash>, peer_id: N::PeerId) -> bool {
+        let blockchain = self.blockchain.read();
+        let state = self.mempool_state.read();
+
+        debug!(peer_id = %peer_id, num = %hashes.len(), "Received unknown mempool hashes");
+        let mut new_hashes_discovered = false;
+
+        hashes.into_iter().take(MAX_TOTAL_HASHES).for_each(|hash| {
```
hrxi: What is the `.take(MAX_TOTAL_HASHES)` supposed to do here?

hrxi: `for` loop would probably be more readable here, rather than `for_each(…)`

mempool/src/sync/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+
+            match self.unknown_hashes.entry(hash) {
+                Occupied(mut entry) => {
+                    entry.get_mut().add_peer(peer_id);
+                }
+                Vacant(entry) => {
+                    entry.insert(HashRequestStatus::new(vec![peer_id]));
+                    new_hashes_discovered = true;
+                }
+            };
```
hrxi:
```suggestion
            match self.unknown_hashes.entry(hash).or_insert_with(|| {
                new_hashes_discovered = true;
                HashRequestStatus::new(vec![])
            }).add_peer(peer_id);
```

mempool/src/sync/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+
+        new_hashes_discovered
+    }
+
+    /// Add peer to discover its mempool
+    fn add_peer(&mut self, peer_id: N::PeerId) {
+        if self.peers.contains(&peer_id)
+            || !self
+                .network
+                .peer_provides_services(peer_id, Services::MEMPOOL)
```
hrxi, line 134: Since the mempool syncer is doing the filtering here, it should probably also do that in the constructor, rather than the caller doing that.

mempool/src/sync/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+    }
+
+    pub fn mark_as_requested(&mut self) {
+        self.status = RequestStatus::Requested;
+    }
+
+    pub fn mark_as_received(&mut self) {
+        self.status = RequestStatus::Received
+    }
+}
```
hrxi: Since the members are also accessed directly, maybe we can skip all these member functions and operate directly on the members.

mempool/src/sync/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+        {
+            for i in 0..self.peers.len() {
+                let peer = self.peers[i];
+                if let Some(request) = self.batch_hashes_by_peer_id(peer) {
+                    prepared_requests.push(request);
+                }
+            }
+        }
+
+        let requests = prepared_requests.into_iter().map(|request| {
```
hrxi:
```suggestion
        let requests = prepared_requests.into_iter().map(|(request, peer_id)| {
```

mempool/src/sync/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+        let mut prepared_requests = vec![];
+
+        while self
+            .unknown_hashes
+            .iter()
+            .any(|(_, request_status)| !request_status.is_requested())
+        {
+            for i in 0..self.peers.len() {
+                let peer = self.peers[i];
+                if let Some(request) = self.batch_hashes_by_peer_id(peer) {
```
hrxi: This looks like O(n²) over the `self.unknown_hashes`. Maybe it would be possible to keep unrequested hashes in a different collection than requested hashes so that we don't have to filter all the time?

I guess one could also invert this loop — for each unknown hash, pick one of the possible peers and add it to that peer's list.

mempool/src/sync/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+        // Then we check if we should shutdown ourself
+        if self.shutdown_timer.poll_unpin(cx).is_ready() {
+            info!(
+                syncer_type = ?self.mempool_transaction_type,
+                "Shutdown mempool syncer"
+            );
+            return Poll::Ready(None);
+        }
+
+        // Then we check if peers got added to live sync in the mean time
```
hrxi:
```suggestion
        // Then we check if peers got added to live sync in the meantime
```
"meantime" has fewer meanings than "mean time".

mempool/src/sync/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+                    if transactions.is_empty() {
+                        continue;
+                    }
+
+                    info!(num = %transactions.len(), "Synced mempool transactions");
+                    self.transactions.extend(transactions);
+                }
+                Err(err) => {
+                    error!(%err, %peer_id, "Failed to fetch mempool transactions");
+                }
```
hrxi, line 363: We should probably re-request these transactions from another peer.

mempool/src/sync/mod.rs
```diff
@@ -0,0 +1,421 @@
...
+                            }
+                        })
+                        .collect();
+
+                    if transactions.is_empty() {
+                        continue;
+                    }
+
+                    info!(num = %transactions.len(), "Synced mempool transactions");
+                    self.transactions.extend(transactions);
```
hrxi, line 343: What about the response missing transactions that we requested?

mempool/src/executor.rs
```diff
@@ -113,16 +123,37 @@ impl<N: Network, T: Topic + Unpin + Sync> Future for MempoolExecutor<N, T> {
...
+                            Ok(_) => MsgAcceptance::Accept,
+                            // Reject the message if signature verification fails or transaction is invalid
+                            // for current validation window
+                            Err(VerifyErr::InvalidTransaction(_)) => MsgAcceptance::Reject,
+                            Err(VerifyErr::AlreadyIncluded) => MsgAcceptance::Reject,
+                            Err(_) => MsgAcceptance::Ignore,
+                        };
+
+                        network.validate_message::<T>(pubsub_id, acceptance);
+                    }
```
hrxi, line 154: Can we somehow unify these branches so that we don't have duplication on what errors we consider bad?

hrxi:
> > Why is this a breaking change (`testnet-restart` label)?
> 
> Because it introduces some new network request messages. Technically it is behind a feature flag because the syncer utilizes an yet unused service flag so these new request messages would only go to peers that also have this new component. So even though it's breaking, it shouldn't be too difficult to get in but for bookkeeping I kept the label around.

I think I agree with @sisou that it's not breaking, precisely because it uses a previously unused feature flag.

mempool/src/sync/mod.rs
Eligioo: To bound the amount of hashes we receive and process from the peers we are requesting them from. On the server side it's also already capped by the `MAX_TOTAL_HASHES`. Those hashes can even be fake so I wanted it to limit it some way.

hrxi: So should it be `MAX_HASHES_PER_REQUEST`?

mempool/src/sync/mod.rs
Eligioo: I my opinion that's OK when you don't receive (all) transactions you are asking for. By the time the peer received your request, the particular transaction(s) could already be adopted in a block.

mempool/src/sync/mod.rs
Eligioo: No, since we want to receive at most `MAX_TOTAL_HASHES` (25.000) hashes from a peer when requesting it mempool hashes.

`MAX_HASHES_PER_REQUEST` is the constant to limit the amount of the actual transactions we ask for per request. So for example we want to request 2000 transactions in total,  `MAX_HASHES_PER_REQUEST` makes sure those transactions are batched into 4 separate requests each containing `MAX_HASHES_PER_REQUEST` transactions.

mempool/src/executor.rs
Eligioo: If you look at all the variants of [`VerifyErr`](https://github.com/nimiq/core-rs-albatross/blob/8280e813459a1ee51c1a53d3363396aff9a18765/mempool/src/verify.rs#L16-L31), not all of them should result into banning the peer. So I picked the ones that are considered malicious and will cause into a ban.

I also unified the `pubsub_or_peer_id` and made the match on the `VerifyErr` exhaustive in the case.

Added comments to each variant

Eligioo:
@hrxi either your comments are resolved or has received a reply

PR #2835, opened 2024-08-16
Update all tests to use an Albatross network ID

sisou:
Most tests still used the `NetworkId::Dummy`, which is not an Albatross ID, so it doesn't include `sender_data` into a transaction's `serialized_content`, thus not testing the whole Albatross range.

## Pull request checklist


hrxi (approved):
LGTM, I verified that these were all the instances.

jsdanielh (approved):

PR #2842, opened 2024-08-20
Limit genesis-builder volatile database to 100GB

### LINKED ISSUE #2840: Volatile mdbx opening fails on RISC-V
It is not clear if this is a RISC-V limitation or a limitation specific to the server, but on Scaleway's RV1 cloud servers, compilation fails because of a `Other(12)` MDBX error when generating the genesis blocks. Code 12 stands for ENOMEM, which means "not enough memory".

Through trial and error @jsdanielh found that the virtual mapped database size of 1TB, with which a volatile DB is opened for the build, is the problem. The server only seems to support 171GB, although on another server that was still too much.

For databases during runtime, the client respects a database `size` client config option. But such a config cannot be used for the compilation step. 

I think the volatile database for genesis generation can be reasonably limited to 100GB, which could be passed as `max_size` config option into the database opening method here: https://github.com/nimiq/core-rs-albatross/blob/fab6cbcf6c1b5d80bc35e0b61321b896bea971e2/genesis-builder/src/main.rs#L20

sisou:
This is to support platforms that don't support 1TB mapped databases, such as RISC-V in testing. At runtime, the database size can be limited via a config option, but for compilation, no such configuration is possible.

100GB should be more than enough for genesis generation.

## Pull request checklist

### DIFF
--- a/database/src/mdbx/database.rs
+++ b/database/src/mdbx/database.rs
@@ -10,8 +10,8 @@ use crate::{
     Error,
 };
 
-const GIGABYTE: usize = 1024 * 1024 * 1024;
-const TERABYTE: usize = GIGABYTE * 1024;
+pub const GIGABYTE: usize = 1024 * 1024 * 1024;
+pub const TERABYTE: usize = GIGABYTE * 1024;
 
 /// Database config options.
 pub struct DatabaseConfig {
--- a/genesis-builder/src/main.rs
+++ b/genesis-builder/src/main.rs
@@ -1,6 +1,6 @@
 use std::{env, process::exit};
 
-use nimiq_database::mdbx::MdbxDatabase;
+use nimiq_database::mdbx::{DatabaseConfig, MdbxDatabase, GIGABYTE};
 use nimiq_genesis_builder::{GenesisBuilder, GenesisInfo};
 
 fn usage(args: Vec<String>) -> ! {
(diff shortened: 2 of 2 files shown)


jsdanielh:
Mmm, to me this is a shot in the dark but could also be a temporal WA. We need to figure out the correct way to map the maximum we can per architecture and set the default.

sisou:
A dev in Telegram had this issue with a Pi4B board with an ARM processor and Debian. He said with Ubuntu it worked fine.

In my case I ran Ubuntu.

It might just be a problem with these small boards and processors.

sisou:
But I don't think we need the maximum for the genesis block generation, that's the point of this PR.

The rest can stay, and the maximum can stay as is, as it's configurable during runtime.

jsdanielh:
> But I don't think we need the maximum for the genesis block generation, that's the point of this PR.
> 
> The rest can stay, and the maximum can stay as is, as it's configurable during runtime.

We also use it for some testing utilities and also it is awful UI if you as a user, have to guess a value where the default doesn't work.

sisou:
Of course it's not _great_, but also for majority of users it will just work.

What else is the `size` config option for, if not to change the size when the default doesn't work for you.

jsdanielh:
> Of course it's not great, but also for majority of users it will just work.

I'm not arguing that. I agree that this is a working workaround for most (but also potentially not all) users.

> What else is the size config option for, if not to change the size when the default doesn't work for you.

What is not great is the guessing part of it. What is a good value, what is the recommended one?

sisou:
Failed now to build with this branch :( Even reducing it to ~~10GB~~ 1GB still fails. Something relevant changed between v0.23.0 and v0.24.0.

sisou:
This change seems to no longer be enough to fix the build of the `genesis-builder` crate on RISC-V.

sisou:
Turns out I changed the wrong volatile database invocation in this PR. #2906 has the right file changed.

PR #2844, opened 2024-08-21
Update README.md

danithedevops:
Fixing a minor typo in the README :)

...
## Pull request checklist


ines-toupeira:
"within" alludes to a range of time, which is not the case. "in" is clearer and aligns more closely with the idea that validators must hit specific timestamps exactly

PR #2845, opened 2024-08-22
Allow chunks request retries

ii-cruz:
Enhances the chunk requests component by having retries in case of failure. This is done by moving the chunks response verification to the sync queue.

## Pull request checklist


jsdanielh (approved):

nibhar (approved):
Maybe lets think about some other potential checks, which would warrant a retry and open an issue for those.

PR #2847, opened 2024-08-22
Overhaul the example config file and related docs

sisou:
Add missing settings, normalize formatting, add missing descriptions.

### TODOs:
- [x] Mempool filter documentation


lib/src/config/config_file/client.example.toml
```diff
@@ -1,61 +1,73 @@
...
 ]
 
-# Optionally specify address(es) that will be advertised to peers instead of the ones in `listen_addresses`
+# Address(es) that will be advertised to peers instead of the ones in `listen_addresses`.
 #
 # This can be used to advertise the public URL and port that peers should connect to, while
 # `listen_addresses` contains the loopback IP and port that this nodes listens on, which may
 # not be publicly reachable.
-# For validators it is strongly recommended to list a public reachable IPv4 IP.
+# For validators it is strongly recommended to list a public reachable IPv4 IP or a domain that resolves to one.
```
jsdanielh: Mmm, if you add a domain, it is strongly recommended to add a public reachable IP for autonat to work.

lib/src/config/config_file/client.example.toml
```diff
@@ -66,229 +78,268 @@ seed_nodes = [
 #
 ##############################################################################
 #[network.tls]
-#private_key = "./my_private_key.pem"
-#certificates = "./my_certificate.pem"
+#private_key = "./path/to/private_key.pem"
+#certificates = "./path/to/fullchain.pem"
```
jsdanielh: `fullchain` here could be misunderstood. Maybe `certificate-chain`?

lib/src/config/config_file/client.example.toml
```diff
@@ -66,229 +78,268 @@ seed_nodes = [
...
+# The network to connect to.
+# Possible values: "main-albatross", "test-albatross", "dev-albatross"
+# Default: "test-albatross"
 network = "test-albatross"
 
-# Specify the sync menchanism according to the client type
-# Possible values: history, full or light
+# The sync mode, which determines the client type.
+# Validator nodes require "full" or "history" sync mode.
+# Possible values: history, full, light
```
jsdanielh:
```suggestion
# Possible values: "history", "full", "light".
```

lib/src/config/config_file/client.example.toml
```diff
@@ -66,229 +78,268 @@ seed_nodes = [
...
+# Default: history
```
jsdanielh:
```suggestion
# Default: "history"
```

lib/src/config/config_file/client.example.toml
```diff
@@ -66,229 +78,268 @@ seed_nodes = [
...
 # Default: 25000
 #blacklist_limit = 25000
 
-# Rules to filter certain transaction
+# Rules to filter mempool transaction by.
 #[mempool.filter]
+
+# Minimum fee for all transactions.
 #tx_fee = 0
+# Minimum tfee per byte for all ransactions.
```
jsdanielh:
```suggestion
# Minimum fee per byte for all transactions.
```

lib/src/config/config_file/mod.rs
```diff
@@ -350,36 +350,48 @@ pub struct MempoolSettings {
 #[derive(Clone, Debug, Deserialize)]
 #[serde(deny_unknown_fields)]
 pub struct MempoolFilterSettings {
+    /// Minimum fee for all transactions
     #[serde(deserialize_with = "deserialize_coin")]
     #[serde(default)]
     pub tx_fee: Coin,
+    /// Minimum tfee per byte for all ransactions
```
```suggestion
    /// Minimum fee per byte for all transactions
```

mempool/src/filter.rs
```diff
@@ -110,29 +110,29 @@ impl Default for MempoolFilter {
 /// Struct defining a Mempool rule
 #[derive(Debug, Clone)]
 pub struct MempoolRules {
-    /// Transaction fee
+    /// Minimum fee for all transactions
     pub tx_fee: Coin,
-    /// Transaction fee per byte
+    /// Minimum tfee per byte for all ransactions
```
jsdanielh:
```suggestion
    /// Minimum fee per byte for all transactions
```

lib/src/config/config_file/client.example.toml
sisou: I cannot find that specific recommendation in the documentation. Can you point to it?

jsdanielh: It's an autonat limitation: https://github.com/libp2p/specs/blob/master/autonat/README.md#autonat-protocol

sisou: I read `based on the IP address` that DNS is also allowed.

The more important limitation (that we do not document here at all yet) is that the IP must be the one of the server itself, not another proxy.

jsdanielh: Complete text is: `implementations MUST NOT dial any multiaddress unless it is based on the IP address the requesting node is observed as`, so there must be an multiaddress based IP since autonat doesn't resolve DNS addresses.

sisou: I understand 'based on the IP address' as anything that resolves to an IP address. If you say autonat doesn't resolve DNS (which I did not read anywhere), then yes, it must be an IP-multiaddress, ok.

jsdanielh: In v1, the rust implementation checks by default if the provided address is a global IP: https://github.com/libp2p/rust-libp2p/blob/64c6eb299ce905b2f81fd1317ec616f56dd37986/protocols/autonat/src/v1/behaviour.rs#L331 and since it doesn't resolve DNS, it won't work.

sisou: Perfekt :+1:

sisou: Reworded like this: https://github.com/nimiq/core-rs-albatross/compare/8f125169001439461d6f4693af6cc2daf30f4a95..0008b8240d5929558d900633a016911b5305c6af

jsdanielh (approved):

Issue #2751, opened 2024-07-15
Panic during block production: InsufficientFunds


viquezclaudio:
A performance degradation was observed in the testnet after the spammer was started.
Block production /push times were high:

![image](https://github.com/user-attachments/assets/dda22112-bfa7-4ae3-9fe2-f5d021713ccd)

Also, validator 2 exhibited a behavior where from a certain point in time, it started accumulating all transactions in the mempool without removing any, which caused the mempool to get to its maximum capacity. 

![image](https://github.com/user-attachments/assets/5d175063-e570-44ca-a309-0335f5517196)

After dumping logs, the following panic was observed in validator 2:

`
ERROR panic  | thread 'tokio-runtime-worker' panicked at 'Failed to compute accounts hash during block production: InsufficientFunds { needed: Coin(2), balance: Coin(0) }': blockchain/src/block_production/mod.rs:133
    0: <backtrace::capture::Backtrace as core::default::Default>::default
    1: log_panics::Config::install_panic_hook::{{closure}}
    2: std::panicking::rust_panic_with_hook
    3: std::panicking::begin_panic_handler::{{closure}}
    4: std::sys_common::backtrace::__rust_end_short_backtrace
    5: rust_begin_unwind
    6: core::panicking::panic_fmt
    7: core::result::unwrap_failed
    8: nimiq_blockchain::block_production::BlockProducer::next_micro_block
    9: nimiq_validator::micro::NextProduceMicroBlockEvent<TValidatorNetwork>::next::{{closure}}
   10: <nimiq_validator::validator::Validator<TValidatorNetwork> as core::future::future::Future>::poll
   11: <tokio_metrics::task::Instrumented<T> as core::future::future::Future>::poll
   12: tokio::runtime::task::core::Core<T,S>::poll
   13: tokio::runtime::task::harness::Harness<T,S>::poll
   14: tokio::runtime::scheduler::multi_thread::worker::Context::run_task
   15: tokio::runtime::scheduler::multi_thread::worker::Context::run
   16: tokio::runtime::context::set_scheduler
   17: tokio::runtime::context::runtime::enter_runtime
   18: tokio::runtime::scheduler::multi_thread::worker::run
   19: <tokio::runtime::blocking::task::BlockingTask<T> as core::future::future::Future>::poll
   20: tokio::runtime::task::core::Core<T,S>::poll
   21: tokio::runtime::task::harness::Harness<T,S>::poll
[...]

PR #2848, opened 2024-08-23
Increase validator robustness against mempool errors

styppo:
Allow validators to continue block production, even after failing to construct a block from the current mempool.

Log a detailed error message when block production fails.

fb224ca33f5819380247c8ed23139e69960dd500 adds an artificial panic to stop block production to help debug issue #2751. This should be reverted once that issue is fixed.


blockchain/src/block_production/mod.rs
```diff
@@ -13,9 +13,43 @@ use nimiq_transaction::{
...
+    #[error("Failed to add to history")]
+    HistoryError,
+    #[error("Accounts are incomplete")]
+    AccountsIncomplete,
+}
+
+impl BlockProducerError {
+    fn accounts_error(
+        blockchain: &Blockchain,
+        error: AccountsError,
```
nibhar, line 33:
```suggestion
    fn from_accounts_error(
        error: AccountsError,
        blockchain: &Blockchain,
```
I like this better, it is not a deal breaker though.

validator/src/micro.rs
```diff
@@ -84,75 +84,103 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
         let mut expected_next_ts;
 
         let return_value = loop {
+            // Wait for the expected timestamp to arrive before producing the block.
+            // We sleep at the beginning of the loop such that we can `continue` into the sleep.
+            // We can't sleep while holding the blockchain lock.
+            if delay > Duration::default() {
```
nibhar:
```suggestion
            if delay.is_zero() {
```
I like `delay.is_zero()` better here, as it is in my opinion a more meaningful expression.

If yielding is not a problem the conditional could also be removed, sleeping unconditionally albeit for a zero duration. I did not check if that is problematic or not and I just assume that `sleep(Duration::ZERO).await;` will in fact yield.

validator/src/micro.rs
```diff
@@ -84,75 +84,103 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
...
+                        "Failed to construct micro block"
+                    );
+
+                    // Add an artificial panic here so that it's easier to catch any remaining
+                    // mempool bugs.
+                    // TODO Remove this.
+                    panic!("Failed to construct micro block");
+
+                    // Sleep a bit to allow the task to be cancelled externally if needed.
+                    delay = Duration::from_millis(50);
```
nibhar, line 143: Mark this to allow unreachable code, for the time that the above panic is in place.

validator/src/micro.rs
```diff
@@ -84,75 +84,103 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
...
+                Err(error) => {
+                    error!(
+                        block_number = self.block_number,
+                        %error,
+                        "Failed to construct micro block"
+                    );
+
+                    // Add an artificial panic here so that it's easier to catch any remaining
+                    // mempool bugs.
+                    // TODO Remove this.
```
nibhar: Please create an issue for this once this gets merged.

nibhar (approved):
Some minor comments, but it looks good to me.

Eligioo (approved):

PR #2856, opened 2024-08-26
Publish/validate before db commit

paberr:
This PR adds a post-validation hook to the blockchain and uses it to publish/validate blocks before the database commit is happening. This way, the blocks are forwarded to the network without waiting for the database.

## Pull request checklist


blockchain/src/blockchain/mod.rs
```diff
@@ -10,3 +13,25 @@ pub mod slots;
 pub mod verify;
 pub mod wrappers;
 pub mod zkp_sync;
+
+/// A post-validation hook that can be registered with the blockchain to run after block validation.
+/// For `extend` and `rebranch` operations, this is called before a block is committed to the database.
+/// It thus provides a timing advantage.
+pub trait PostValidationHook {
+    /// Run the pre-commit hook.
```
nibhar: `post-validation` or `pre-commit`. I assume they are the same, but one name should be picked. It is mentioned only once, but there is other places where the names are used.

blockchain/src/blockchain/push.rs
```diff
@@ -40,12 +41,13 @@ impl Blockchain {
...
     /// Set the trusted flag to true to skip VRF and signature verifications: when the source of the
     /// block can be trusted.
-    fn do_push(
+    fn do_push<F: PostValidationHook>(
         this: RwLockUpgradableReadGuard<Self>,
         block: Block,
         trusted: bool,
         diff: Option<TrieDiff>,
         chunks: Vec<TrieChunkWithStart>,
+        post_validation_hook: &F,
```
nibhar, line 50: I guess this could have been a closure.

blockchain/src/blockchain/mod.rs
```diff
@@ -10,3 +13,25 @@ pub mod slots;
...
+impl PostValidationHook for () {
+    fn post_validation(&self, _block: Block, _push_result: Result<PushResult, PushError>) {}
+}
+
+impl<F: PostValidationHook> PostValidationHook for Option<F> {
+    fn post_validation(&self, block: Block, push_result: Result<PushResult, PushError>) {
+        match self {
+            Some(hook) => hook.post_validation(block, push_result),
+            None => {}
+        }
```
nibhar: This should work, I think.
```suggestion
        self.as_ref().inspect(|hook| hook.post_validation(block, push_result));
```

blockchain/src/blockchain/mod.rs
```diff
@@ -10,3 +13,25 @@ pub mod slots;
 pub mod verify;
 pub mod wrappers;
 pub mod zkp_sync;
+
+/// A post-validation hook that can be registered with the blockchain to run after block validation.
+/// For `extend` and `rebranch` operations, this is called before a block is committed to the database.
+/// It thus provides a timing advantage.
+pub trait PostValidationHook {
+    /// Run the pre-commit hook.
+    fn post_validation(&self, block: Block, push_result: Result<PushResult, PushError>);
```
nibhar: I think it would be nice to be able to do 
```suggestion
    fn post_validation(&self, block: &Block, push_result: Result<&PushResult, &PushError>);
```
here.

I realize that to publish this will have to be cloned within, but it will then not be cloned to do validate message, where it is unnecessary. Also the result types are not necessary to be consumed here either and can just be passed as reference.

nibhar (commented):
Lots of additional `Clone` derives, which indicate that we now clone more. Not entirely sure where this is and how big the impact is, pretty sure it could be avoided though from what I read.

nibhar:
In https://github.com/nimiq/core-rs-albatross/commit/bad136e8b649517dd3ae3cc3b67863b8c8c92ddc, which is on top of this PR, I removed a bunch of the clones again, and passed the arguments to `post_validation` by reference.

blockchain/src/blockchain/mod.rs
paberr: post validation because it's not always pre-commit. it's still a leftover from a previous implementation

blockchain/src/blockchain/push.rs
paberr: I opted against a closure mainly to avoid code duplication as there are only two types of hooks we currently use.
Obviously, something like `push(..., |block, result| hook.post_validation(block, result))` would have been an option, too.

blockchain/src/blockchain/mod.rs
nibhar: I will change it then to `post-validation` where applicable.

nibhar:
In https://github.com/nimiq/core-rs-albatross/compare/nibhar/publish-pre-commit?expand=1 I also added the changed documentation, as well as a `push_with_hook` function which can be used to include the hook, whereas push remains unchanged. That cuts down on all those `&()` additions.

blockchain/src/blockchain/push.rs
```diff
@@ -176,26 +180,36 @@ impl Blockchain {
...
         debug_assert!(
             this.get_missing_accounts_range(None).is_none(),
             "Should call push only for complete tries"
         );
-        Self::push_wrapperfn(this, block, false, None, vec![]).map(|res| res.0)
+        Self::push_wrapperfn(this, block, false, None, vec![], post_validation_hook)
+            .map(|res| res.0)
+    }
+
+    pub fn push(
```
Eligioo, line 199: Nice in order to not break all the tests :-)

validator/src/validator.rs
```diff
@@ -509,41 +513,25 @@ where
...
+                            &block_publisher,
+                        )
                     };
-
-                    if result == Some(PushResult::Extended)
-                        || result == Some(PushResult::Rebranched)
-                    {
-                        if block_copy.is_election() {
-                            info!(
-                                block_number = &block_copy.header.block_number,
```
Eligioo: You are losing the info logs I believe. Is that desired?

Eligioo (approved):
LGTM, nice way to decouple the publish and commit. 1 small log thing.

blockchain/src/blockchain/push.rs
```diff
@@ -261,34 +276,63 @@ impl Blockchain {
...
+                trust,
+                diff,
+                chunks,
+                post_validation_hook,
+            );
+            match res {
+                // Extended and Rebranched are already handled pre-commit.
+                Ok((PushResult::Extended | PushResult::Rebranched, _)) => {}
+                Ok((ref res, _)) => post_validation_hook.post_validation(&block, &Ok(res.clone())),
+                Err(ref res) => post_validation_hook.post_validation(&block, &Err(res.clone())),
```
styppo: Compilation error when the metrics feature is disabled.

```suggestion
                Ok((ref res, _)) => post_validation_hook.post_validation(&block, Ok(res)),
                Err(ref res) => post_validation_hook.post_validation(&block, Err(res)),
```

blockchain/src/blockchain/push.rs
```diff
@@ -261,34 +276,63 @@ impl Blockchain {
...
         diff: Option<TrieDiff>,
         chunks: Vec<TrieChunkWithStart>,
+        post_validation_hook: &F,
     ) -> Result<(PushResult, Result<ChunksPushResult, ChunksPushError>), PushError> {
         #[cfg(not(feature = "metrics"))]
         {
-            Self::do_push(this, block, trust, diff, chunks)
+            let res = Self::do_push(
+                this,
+                block.clone(),
```
styppo: We should eliminate the cloning of the block here by moving the call to `post_validation_hook.post_validation()` into `Blockchain::do_push()`, making it consistent with the `Extended`/`Rebranched` case.

blockchain/src/blockchain/push.rs
```diff
@@ -261,34 +276,63 @@ impl Blockchain {
...
+            }
+            res
         }
         #[cfg(feature = "metrics")]
         {
             let metrics = this.metrics.clone();
-            let res = Self::do_push(this, block, trust, diff, chunks);
+            let res = Self::do_push(
+                this,
+                block.clone(),
```
styppo: Same as above.

validator/src/micro.rs
```diff
@@ -9,21 +9,18 @@ use std::{
...
-use crate::aggregation::skip_block::SkipBlockAggregation;
+use crate::{aggregation::skip_block::SkipBlockAggregation, validator::BlockPublisher};
 
-// Ignoring this clippy warning since size difference is not that much (320
-// bytes) and we probably don't want the performance penalty of the allocation.
-#[allow(clippy::large_enum_variant)]
 pub(crate) enum ProduceMicroBlockEvent {
-    MicroBlock(MicroBlock, PushResult),
+    MicroBlock,
 }
```
styppo, line 25: Looks like this event is now unused and we could get rid of it. But it's also not breaking anything so I think it would be fine to address this in a separate PR.

styppo (changes requested):

validator/src/validator.rs
styppo: I think it's fine to get rid of those.

blockchain/src/blockchain/push.rs
nibhar: I could do that, but it is quite annoying as there is a bunch of early returns and `?` uses within `do_push`. All of those instances would need to be updated to call the `post_validation_hook`. The way it is done here, there is no need for that. 
Also if in the future there will be new early returns, they will still invoke the `post_validation_hook` the way it is currently done, whereas invoking it inside of `do_push` that call could be omitted by mistake fairly easily.

validator/src/micro.rs
nibhar: Yes it is unused and could be removed.

Removing it will likely also entail making the micro Producer no longer having to implement Stream. That probably will mean a bunch of additional changes (for the better).

I opted not to do it here as it is out of scope of this PR.

hrxi: Maybe the "fix clippy warning" commit could be dropped instead of half-removing the `MicroBlock` event. A `#[allow()]` could be used instead, leaving the (full) removal for a future PR.

blockchain/src/blockchain/push.rs
paberr: Somehow my previous comment is shown as "pending" but yes that's exactly the reason for the current design.

validator/src/micro.rs
nibhar: For the time being I don't plan on doing any additional work there. The Event is unused, but removing the stream makes us less flexible in the future. If the micro producer is ever going to return something meaningful again it would have to be re implemented. It does not break anything as it is and I think it is semantically not too bad either.

styppo (approved):
LGTM

jsdanielh:
Rebasing branch and signing commits for merge

PR #2858, opened 2024-08-27
Improve performance of `Mempool::update`

### LINKED ISSUE #2855: Mempool update can take several seconds
Executing the mempool update function, which is done after we adopt blocks, can take several seconds: it depends on the number of txns we have in the mempool and if we have affected senders by the block that was adopted.

This is extracted after running a private tesnet with a spammer, sending 1000tpb from the same sender:
 `Updated mempool, took 8.73577319s`

styppo:
Eliminate unnecessary cloning of the blockchain's head block by returning a reference from `AbstractBlockchain::head`.

Closes #2855

### DIFF
--- a/blockchain-interface/src/abstract_blockchain.rs
+++ b/blockchain-interface/src/abstract_blockchain.rs
@@ -23,7 +23,7 @@ pub trait AbstractBlockchain {
     fn now(&self) -> u64;
 
     /// Returns the head of the main chain.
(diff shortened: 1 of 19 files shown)


jsdanielh (commented):
Could you address the clippy warnings in `validator/src/micro.rs`?

nibhar (approved):

jsdanielh (approved):

PR #2859, opened 2024-08-27
Consensus: add interval to wake up consensus task for head requests

styppo:


consensus/src/consensus/mod.rs
```diff
@@ -134,8 +135,10 @@ pub struct Consensus<N: Network> {
     #[cfg(feature = "full")]
     last_batch_number: u32,
     synced_validity_window_flag: Arc<AtomicBool>,
+
     head_requests: Option<HeadRequests<N>>,
     head_requests_time: Option<Instant>,
+    head_requests_interval: Interval,
```
nibhar, line 141: As I understand it, every head request re initializes this interval and this interval not being the only trigger to it, it might make sense to change this to `BoxFuture<'static, ()>` or similar. This does not happen every `Self::HEAD_REQUEST_TIMEOUT`, but it triggers generally less often. 
Absolutely not required though.

consensus/src/consensus/mod.rs
```diff
@@ -442,29 +451,42 @@ impl<N: Network> Consensus<N> {
...
+        );
+
+        self.head_requests = Some(HeadRequests::new(
+            self.sync.peers(),
+            Arc::clone(&self.network),
+            self.blockchain.clone(),
+        ));
+
+        self.head_requests_time = Some(Instant::now());
+        self.head_requests_interval = interval(Self::HEAD_REQUESTS_TIMEOUT);
```
nibhar, line 489: The first tick always completes immediately. 
I am not sure if that is a problem, I believe most likely not as it only triggers a wake, but if it is, we should skip the first instance.

consensus/src/consensus/mod.rs
```diff
@@ -546,6 +569,9 @@ impl<N: Network> Future for Consensus<N> {
             }
         }
 
+        // Poll interval to wake up this task on a regular basis to perform head requests.
+        while self.head_requests_interval.poll_next_unpin(cx).is_ready() {}
```
Eligioo, line 573: Is `while` desired here if only intended to register the waker for the interval task? Could simply be `let _ =`.

Eligioo (approved):

PR #2860, opened 2024-08-27
Handel update

nibhar:
Allows Handel to to switch to response only mode before all signatures are accumulated. A Closure is used to determine if continuing the aggregation can lead to a different result and if it cannot, it switches to respond only. This produces less or even zero network activity for finished aggregations, while still allowing people to catch up as the aggregations do not fully terminate.

Might be related to #2623.

### DIFF
--- a/handel/src/evaluator.rs
+++ b/handel/src/evaluator.rs
@@ -33,7 +33,6 @@ where
 }
 
 /// A signature counts as it was signed N times, where N is the signers weight
-#[derive(Debug)]
 pub struct WeightedVote<TId, TProtocol>
 where
     TId: Identifier,
@@ -48,8 +47,8 @@ where
     /// Partitioner that registers the handel levels and its IDs.
     partitioner: Arc<TProtocol::Partitioner>,
 
-    /// Threshold after which a signature could be considered final according to the weights of the signers.
-    pub threshold: usize,
+    /// Closure used to determine the finality of a given contribution.
+    finality_check: Box<dyn Fn(&TProtocol::Contribution) -> bool + Send + Sync>,
 }
 
 impl<TId, TProtocol> WeightedVote<TId, TProtocol>
@@ -78,13 +77,13 @@ where
         store: Arc<RwLock<TProtocol::Store>>,
         weights: Arc<TProtocol::Registry>,
         partitioner: Arc<TProtocol::Partitioner>,
-        threshold: usize,
+        finality_check: impl Fn(&TProtocol::Contribution) -> bool + Send + Sync + 'static,
     ) -> Self {
         Self {
             store,
             weights,
             partitioner,
-            threshold,
+            finality_check: Box::new(finality_check),
         }
     }
 }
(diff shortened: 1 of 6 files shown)


handel/src/evaluator.rs
```diff
@@ -49,7 +48,7 @@ where
     partitioner: Arc<TProtocol::Partitioner>,
 
     /// Threshold after which a signature could be considered final according to the weights of the signers.
-    pub threshold: usize,
+    finality_check: Box<dyn Fn(&TProtocol::Contribution) -> bool + Send + Sync>,
```
jsdanielh, line 51: Is it still a threshold as the rustdoc says?

validator/src/aggregation/tendermint/protocol.rs
```diff
@@ -37,14 +35,60 @@ impl TendermintAggregationProtocol {
...
+                    // Any option which has at least 2f+1 votes make this contribution final.
+                    if weight >= policy::Policy::TWO_F_PLUS_ONE as usize {
+                        return true;
+                    }
+                    // If the proposal does not have 2f+1 check if it has the new biggest_weight
+                    if hash.is_some() && weight > biggest_weight {
+                        biggest_weight = weight;
+                    }
+                    // The cast votes need to be removed from the uncast votes
+                    uncast_votes -= weight;
```
jsdanielh: Couldn't we better use `saturating_sub`?

validator/src/aggregation/tendermint/protocol.rs
```diff
@@ -37,14 +35,60 @@ impl TendermintAggregationProtocol {
...
+                }
+
+                let cast_votes = policy::Policy::SLOTS as usize - uncast_votes;
```
jsdanielh: Couldn't we better use `saturating_sub`?

validator/src/aggregation/tendermint/protocol.rs
```diff
@@ -37,14 +35,60 @@ impl TendermintAggregationProtocol {
...
+                if cast_votes < policy::Policy::TWO_F_PLUS_ONE as usize {
+                    // This should not be here. The criteria that the best vote can be elevated above the threshold
+                    // using the uncast votes is enough. In our case that mans seeing f+1 different proposals wth 1
```
jsdanielh:
```suggestion
                    // using the uncast votes is enough. In our case that means that seeing f+1 different proposals with 1
```

handel/src/evaluator.rs
nibhar: No, I updated it to reflect the change. Thanks.

validator/src/aggregation/tendermint/protocol.rs
nibhar: I use `saturating_sub` here now. Note that if this would ever saturate the contribution would also be invalid as a whole as then there is more votes present as there is votes to begin with. A different place would have already filtered this out. 
It is still way better to use the `saturating_sub` here.

validator/src/aggregation/tendermint/protocol.rs
nibhar: Here it is not strictly necessary, as `uncast_votes` start as `policy::Policy::SLOTS` and then only ever decrease. With the `saturating_sub` from your previous comment it can no longer underflow and thus can only ever reach 0. I still added it, because it does no harm at all and is a better practice.
Thanks.

jsdanielh (approved):

jsdanielh:
PR reverted in 5de14219069b20789223dd9e1a0fe2974469b1da.

PR #2864, opened 2024-08-28
pow-migration: Do not use `Policy` functions to calculate epoch numbers

jsdanielh:
Change the blockchain history store interface to add a function (`add_to_history_for_epoch`) that receives the epoch number instead of relying on `Policy` to calculate it. This is useful in the history migration for `PoW` since the genesis block number wouldn't be known until the migration is over.

## Pull request checklist

### DIFF
--- a/blockchain/src/history/history_store.rs
+++ b/blockchain/src/history/history_store.rs
@@ -341,11 +341,10 @@ impl HistoryStore {
     pub(crate) fn put_historic_txns(
         &self,
         txn: &mut MdbxWriteTransaction,
+        epoch_number: u32,
         block_number: u32,
         hist_txs: &[HistoricTransaction],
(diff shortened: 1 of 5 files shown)


blockchain/src/history/history_store.rs
```diff
@@ -338,14 +338,13 @@ impl HistoryStore {
     }
 
     /// Internal function for `add_to_history`, which also returns leaf indices.
```
Eligioo: Rustdoc says it's an internal function though now it has public visibility. Is this intended?

Eligioo (approved):

viquezclaudio (approved):

blockchain/src/history/history_store.rs
jsdanielh: Changed the visibility to keep it internal within the crate.

PR #2869, opened 2024-08-29
Fix ordering of the diff queue

ii-cruz:
The introduction of the diffs needed flag made it so that the blocks are returned out of order. This does not seem to be a bug, however, it is a strange behavior.
This PR restores the original diff queue order by appending to the futures ordered diffs despite us not needing diffs. These futures will resolve immediately with no None diff attached to the block.

I refactored the poll function and the function arguments a bit as well.

## Pull request checklist

### DIFF
--- a/consensus/src/sync/live/diff_queue/mod.rs
+++ b/consensus/src/sync/live/diff_queue/mod.rs
@@ -1,6 +1,6 @@
 use std::{
     collections::HashSet,
-    future::Future,
+    future,
     pin::Pin,
     sync::Arc,
     task::{Context, Poll},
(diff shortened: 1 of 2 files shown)


consensus/src/sync/live/diff_queue/mod.rs
```diff
@@ -213,49 +224,28 @@ impl<N: Network> Stream for DiffQueue<N> {
...
-                        }));
-                    }
-
-                    let get_diff = self.diff_request_component.request_diff();
+        while let Poll::Ready(block) = self.block_queue.poll_next_unpin(cx) {
+            match block {
+                Some(block) => {
+                    let get_diff = self
+                        .diff_needed
+                        .then(|| self.diff_request_component.request_diff());
```
hrxi: I think the code would become clearer by outlining the `Option` out of the `augment_block` function. It's directly matched in that function.

E.g. like this:
```rs
if self.diff_needed {
    self.diffs.push_back(Box::pin(augment_block(block, get_diff)));
} else {
    self.diffs.push_back(Box::pin(future::ready(Ok(QueuedDiff::from_block_no_diff(block)))));
}
```

That would also allow to reduce the number of lines changed.

consensus/src/sync/live/diff_queue/mod.rs
```diff
@@ -213,49 +222,30 @@ impl<N: Network> Stream for DiffQueue<N> {
...
-                Poll::Ready(None) => {
-                    block_queue_done = true;
-                    break;
+                (Some(block), false) => {
+                    self.diffs.push_back(Box::pin(future::ready(Ok(
+                        QueuedDiff::from_block_no_diff(block),
+                    ))));
                 }
-                Poll::Pending => break,
+                (None, ..) => return Poll::Ready(None),
```
hrxi, line 240:
```suggestion
                // Immediately quit when the block queue stream has ended, not
                // waiting for any pending diff requests.
                (None, ..) => return Poll::Ready(None),
```
One last nit: Could you add a comment saying that we explicitly end the stream if the block queue stream has ended, not waiting for the diff requests (just to make sure that it's clear that this is intended)?

hrxi (approved):
Thanks!

hrxi:
Rebased for merging.

PR #2880, opened 2024-09-03
Handel Update

nibhar:
This PR allows Handel to terminate aggregates earlier than once every possible contributor is present. It accepts a closure to check for finality. 
This PR also updates Tendermint, such that it can accept a final contribution with less contributors than the previous, non-final contribution.


nibhar:
Closed as not planned as it currently stands. I keep the branch around but this will get replaces by a different mechanism to reduce spam.

PR #2882, opened 2024-09-03
Allow multiple rounds of validity sync

### LINKED ISSUE #2881: Second round of validity sync is problematic
When the validity sync is complete, it could be the case that another round of validity sync is needed because the blockchain state 
has advanced so much. This causes the issue where we start doing validity sync and state sync at the same time and try to push a block when we are not ready.

[invalid_history.log](https://github.com/user-attachments/files/16852418/invalid_history.log)

viquezclaudio:
When we are done with the validity sync, move the peers back to macro sync to re-check if there is a more recent state that we need to sync with.
The exit condition is:
 - No more new macro state
 - We can verify the history root

...
## Pull request checklist

### DIFF
--- a/consensus/src/sync/light/validity_window.rs
+++ b/consensus/src/sync/light/validity_window.rs
@@ -348,6 +348,7 @@ impl<TNetwork: Network> LightMacroSync<TNetwork> {
                                     "Validity window syncing is complete"
                                 );
 
+                                self.synced_validity_peers.push(peer_id);
                                 self.validity_queue.remove_peer(&peer_id);
                                 self.syncing_peers.remove(&peer_id);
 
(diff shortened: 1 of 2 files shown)


consensus/src/sync/light/validity_window.rs
```diff
@@ -360,8 +361,7 @@ impl<TNetwork: Network> LightMacroSync<TNetwork> {
                                 // We are complete so we emit the peer
```
Eligioo, line 361: This comment is now a bit misleading as you don't move the peer directly into live sync anymore.

ii-cruz (approved):
LGTM

I think the poll_validity_window_chunks can probably be simplified now with Result<(),Err> and fail only in the case of outdated peer. 
We can then change the err to Outdated peer on the poll function directly.

Eligioo (approved):
LGTM

PR #2886, opened 2024-09-04
consensus: Improve error handling for failed missing block requests

jsdanielh:
Improve error handling for failed missing block requests in the block queue by removing from the buffer the block successor chain if a request for a missing block failed.

## Pull request checklist

### DIFF
--- a/consensus/src/sync/live/block_queue/queue.rs
+++ b/consensus/src/sync/live/block_queue/queue.rs
@@ -826,7 +826,7 @@ impl<N: Network> Stream for BlockQueue<N> {
         loop {
             let poll_res = self.request_component.poll_next_unpin(cx);
             match poll_res {
-                Poll::Ready(Some(result)) => {
+                Poll::Ready(Some(Ok(result))) => {
                     if let Some(block) = self.handle_missing_blocks(
                         result.target_block_number,
                         result.target_block_hash,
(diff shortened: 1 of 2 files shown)


consensus/src/sync/live/block_queue/queue.rs
```diff
@@ -833,6 +833,27 @@ impl<N: Network> Stream for BlockQueue<N> {
...
+                        else {
+                            // No block found whose parent block is the one we're looking: we should have already removed
+                            // all blocks from the target block chain that errored.
+                            break;
+                        };
+                        block_number = block.block_number();
+                        block_hash = hash.clone();
+                        blocks.remove(&block_hash);
+                    }
+                    return Poll::Ready(None);
```
ii-cruz: I think this is incorrect. The block queue should not return None since that means it has terminated by rust stream rules. Here we should instead do nothing and let it loop again so that it registers the weaker or returns the next value.

ii-cruz (changes requested):

ii-cruz (commented):
This has a couple of clippy warnings. Besides that it's good to go

Eligioo (approved):

jsdanielh:
> This has a couple of clippy warnings. Besides that it's good to go

@viquezclaudio fixed them in `albatross` already. Rebased branch to reflect them here

ii-cruz (approved):

PR #2889, opened 2024-09-05
Consensus: return block number from head requests

styppo:
Detect if an incompatible peer is in sync using the returned block number.

### DIFF
--- a/consensus/src/sync/syncer.rs
+++ b/consensus/src/sync/syncer.rs
@@ -13,6 +13,7 @@ use nimiq_blockchain_interface::AbstractBlockchain;
 use nimiq_blockchain_proxy::BlockchainProxy;
 use nimiq_hash::Blake2bHash;
 use nimiq_network_interface::network::{CloseReason, Network, NetworkEvent, SubscribeEvents};
+use nimiq_primitives::policy::Policy;
 use nimiq_time::{interval, Interval};
 use nimiq_utils::stream::FuturesUnordered;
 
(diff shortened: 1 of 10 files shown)


jsdanielh (approved):
LGTM

consensus/src/sync/syncer.rs
```diff
@@ -248,47 +249,15 @@ impl<N: Network, M: MacroSync<N::PeerId>, L: LiveSync<N>> Syncer<N, M, L> {
...
-        // Fetch the reported checkpoint block from our chain.
-        // If we don't know the block, the peer is either ahead of us or on a different chain,
-        // We consider the peer as not in sync until we catch up and is sufficiently close (see below).
-        let checkpoint = match blockchain.get_block(&head.r#macro, false) {
-            Ok(block) => block,
-            Err(_) => return false,
-        };
-
-        // We consider the peer synced if it's at most one batch behind us.
-        if checkpoint.batch_number() < blockchain.macro_head().batch_number().saturating_sub(1) {
```
ii-cruz, line 287: Should we also check the block_hash? We are not doing anything with it. Tbh idk what we could do with just the hash of the head, not much.
This change loosens up a bit the checks of the head requests. Now the peer might be in a different chain and be considered synced. Is this ok?

ii-cruz (commented):
Should

Eligioo (commented):
LGTM but second the comment that @ii-cruz made earlier. Peers that just forked  shouldn't be considered as compatible.

consensus/src/sync/syncer.rs
```diff
@@ -248,47 +249,15 @@ impl<N: Network, M: MacroSync<N::PeerId>, L: LiveSync<N>> Syncer<N, M, L> {
...
+        // Check the peer's head block number. We consider the peer synced if the peer is
+        // at most one batch behind us.
+        let peers_batch = Policy::batch_at(head.block_number);
+        let our_batch = blockchain.read().batch_number();
+        if peers_batch < our_batch.saturating_sub(1) {
```
nibhar: This is not equivalent to the removed parts.

```suggestion
        if peers_batch > our_batch {
            debug!(
                %peer_id,
                peers_batch = checkpoint.batch_number(),
                our_batch = blockchain.macro_head().batch_number(),
                peers_batch,
                our_batch,
                "Incompatible peer is in a future batch"
            );
            return false;
        }
        if peers_batch < our_batch.saturating_sub(1) {
```
Maybe that is intended, but if so I cannot see it mentioned.

consensus/src/sync/syncer.rs
nibhar: This does consider the peer synced as soon as they are in the same batch or the one before (I made another comment about it). As a checkpoint is final by the security assumption this should not be a real problem.

The peers head state is however not in any way verifiable. Previously it could be sanity checked that at least the macro chain is the same, now they can just claim to be at whatever height, making up a hash. That may or may not be problematic, I am not too familiar with the caller side here.

nibhar (changes requested):
This changes how `is_peer_synced` works!

Previous to this change a peer was _not_ synced if the given election hash or the given checkpoint hash was not found. Meaning an ahead peer was not synced!
With this change an ahead peer is absolutely synced. Only thing they have to do is to send a sufficiently large block number with any random hash that we cannot verify. Not sure if that is intended.
Also the returned information becomes all but unverifiable. Also not sure as to the implications.

consensus/src/sync/syncer.rs
styppo: It's not a problem if the peer is on a different chain _within the same batch_. That other chain might become the canonical one after all, so we want to keep accepting blocks from that peer.

The function name `is_peer_synced` is probably a bit misleading. `is_peer_macro_synced` or `is_peer_eligible_for_live_sync` would better capture the semantics of this function.

The peer's claimed state is never verifiable, irrespective of whether it contains a hash of their state, so we're not losing anything in terms of verifiability here. However, we could consider reintroducing the checkpoint hash to check if the security assumption has been violated and ban peers that are on a different macro chain.

consensus/src/sync/syncer.rs
styppo: With 098f2819714a76dcc067fbb5e6407295262d4836, a peer will not be considered synced if it is too far ahead of our state. This avoids moving an incompatible peer back and forth between macro and live sync if it is further ahead than the BlockQueue's acceptance window.

consensus/src/sync/syncer.rs
nibhar: As it is just an additional hash which does provide some utility I would suggest to have it remain in the response. It is not the most solid "proof" of a peers local state if equal but if different it is an interesting observation to make.

ii-cruz: The renaming is simple and a good idea. I think it's ok to leave the rest as is

styppo:
Superseded by #2982

PR #2893, opened 2024-09-05
Fix production of invalid history chunks

### LINKED ISSUE #2890: Invalid history chunks
In the test net, it has been observed the following log messages:

`Banning peer because the validity history chunk didn't verify peer_id=12D3KooWKd6cY4saxrTradfE5ELkxDFNWTC3Vn79H8ckNQDREA75 chunk=138 verifier_block=3549090 epoch=12`

This is caused by a peer behind who generated a history chunk proof that doesn't verify because the verifier block number is ahead of his current state. This led to an improper banning of the peer.

For this case (server is behind), we should reply that we couldn't produce the proof such that the peer is emitted as outdated and not banned

viquezclaudio:
- Do not produce invalid history chunks if the verifier block number is ahead of our state
- Improve the validity window sync (log & process)

## Pull request checklist

### DIFF
--- a/blockchain/src/history/history_store.rs
+++ b/blockchain/src/history/history_store.rs
@@ -407,7 +407,12 @@ impl HistoryInterface for HistoryStore {
     /// Returns the length (i.e. the number of leaves) of the History Tree at a given block height.
     /// Note that this returns the number of leaves for only the epoch of the given block height,
     /// this is because we have separate History Trees for separate epochs.
-    fn length_at(&self, block_number: u32, txn_option: Option<&MdbxReadTransaction>) -> u32 {
+    /// If we dont have a block number at the given block height, we return None.
+    fn length_at(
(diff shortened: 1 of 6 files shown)


blockchain/src/history/history_store.rs
```diff
@@ -416,19 +421,19 @@ impl HistoryInterface for HistoryStore {
...
-                        i + 1
+                        Some(i + 1)
                     } else {
-                        0
+                        Some(0)
                     }
                 }
                 // If it doesn't exist, then the HistoryStore is empty at this block height.
-                None => 0,
+                None => None,
```
sisou: If you're just forwarding `None`, you should be able to use `Option::map` instead on `cursor.prev()`.

ii-cruz (approved):
LGTM

PR #2900, opened 2024-09-09
Network discovery: store the remote address of the peer after finishing a successful handshake

### LINKED ISSUE #2896: Network: banning incorrect IPs
When the node bans a peer, it currently bans all the IPs of the `PeerContact` that are associated to the peer. This can cause bans of undesired IPs like `127.0.0.1`. Instead we should only ban the connected IP and those confirmed by AutoNAT.

```
2024-09-04T18:28:58.253406282Z WARN  validity_window      | Banning peer because the validity history chunk didn't verify peer_id=12D3KooWGG7m9Lip5mXjhvDPW3KLGMWb755AjqRq73fuEXAoZjv1 chunk=2114 verifier_block=3565470 epoch=13
2024-09-04T18:28:58.253538715Z DEBUG behaviour            | Banned peer peer_id=12D3KooWGG7m9Lip5mXjhvDPW3KLGMWb755AjqRq73fuEXAoZjv1
2024-09-04T18:28:58.253552722Z DEBUG behaviour            | Banned address address=/ip4/127.0.0.1/tcp/8444/ws
2024-09-04T18:28:58.253577720Z DEBUG behaviour            | Banned address address=/ip4/172.17.0.2/tcp/8444/ws
```

Eligioo:
Previously we used the `remote_addr` directly from the `handle_pending_inbound_connection` to check if it's banned. However this will never work correctly because it also contains `Multiaddr` protocol parts like `/tcp/{random port}` and others. Since a new connection always gets opened on a new port you have a different `Multiaddr` on each connection thus  never find a banned `Multiaddr`.

This PR adds storing a remote address in the `PeerContactMeta` that only contains a `ip4`, `ip6`, `dns4` or `dns6` part of a `Multiaddr` that can be used to correctly verify if IP or hostname is banned.

Lastly this PR solves the problem of banning observed peer addresses, which could lead to banning `127.0.0.1`. Fixes #2896.

### DIFF
--- a/network-libp2p/src/connection_pool/behaviour.rs
+++ b/network-libp2p/src/connection_pool/behaviour.rs
@@ -28,7 +28,7 @@ use rand::{seq::IteratorRandom, thread_rng};
 use void::Void;
 
 use super::Error;
-use crate::discovery::peer_contacts::PeerContactBook;
+use crate::discovery::{handler, peer_contacts::PeerContactBook};
 
 /// Current state of connections and peers for connection limits
 #[derive(Clone, Debug)]
@@ -625,12 +625,11 @@ impl Behaviour {
         self.peer_ids.mark_banned(peer_id);
         debug!(%peer_id, "Banned peer");
 
-        // Mark its addresses as banned if we have them
+        // Mark its outer protocol address as banned if we have it
         if let Some(contact) = self.contacts.read().get(&peer_id) {
-            let addresses = contact.addresses();
-            for address in addresses {
-                self.addresses.mark_banned(address.clone());
-                debug!(%address, "Banned address");
+            if let Some(outer_protocol_address) = contact.get_outer_protocol_address() {
+                debug!(address = %outer_protocol_address, "Banned address");
+                self.addresses.mark_banned(outer_protocol_address);
             }
         }
     }
@@ -893,9 +892,11 @@ impl NetworkBehaviour for Behaviour {
         _local_addr: &Multiaddr,
         remote_addr: &Multiaddr,
     ) -> Result<(), ConnectionDenied> {
-        if self.addresses.is_banned(remote_addr.clone()) {
-            debug!(%remote_addr, "Address is banned");
-            return Err(ConnectionDenied::new(Error::BannedIp));
+        if let Some(outer_protocol_address) = handler::outer_protocol_address(remote_addr) {
+            if self.addresses.is_banned(outer_protocol_address) {
+                debug!(%remote_addr, "Address is banned");
+                return Err(ConnectionDenied::new(Error::BannedIp));
+            }
         }
 
         // Get IP from multiaddress if it exists.
--- a/network-libp2p/src/discovery/handler.rs
+++ b/network-libp2p/src/discovery/handler.rs
@@ -10,6 +10,7 @@ use futures_timer::Delay;
 use instant::Instant;
 use libp2p::{
     identity::Keypair,
+    multiaddr::Protocol,
     swarm::{
         handler::{
             ConnectionEvent, DialUpgradeError, FullyNegotiatedInbound, FullyNegotiatedOutbound,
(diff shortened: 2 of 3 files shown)


network-libp2p/src/discovery/handler.rs
```diff
@@ -244,6 +251,23 @@ impl Handler {
     }
 }
 
+/// Extract the `/ip4/`,`/ip6/`, `/dns4/` or `/dns6/` protocol part from a `Multiaddr`
+pub(crate) fn protocol_part(addr: &Multiaddr) -> Option<Multiaddr> {
```
jsdanielh: Mmm, maybe `outer_protocol_address`? Because there are other `protocols` on top of this outer protocols.

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -183,6 +183,7 @@ impl SignedPeerContact {
 /// Meta information attached to peer contact info objects. This is meant to be mutable and change over time.
 #[derive(Clone, Debug, Serialize, Deserialize)]
 struct PeerContactMeta {
+    protocol_addr: Option<Multiaddr>,
```
jsdanielh: Same as above: Maybe `outer_protocol_address`

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -271,6 +275,19 @@ impl PeerContactInfo {
     pub fn set_score(&self, score: f64) {
         self.meta.write().score = score;
     }
+
+    /// Gets the protocol address of the peer. For example `/ip4/x.x.x.x` or `/dns4/foo.bar`
+    pub fn get_protocol_addr(&self) -> Option<Multiaddr> {
```
jsdanielh: Same as above: Maybe `outer_protocol_address`

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -271,6 +275,19 @@ impl PeerContactInfo {
...
+        self.meta.read().protocol_addr.clone()
+    }
+
+    /// Sets the protocol address of the peer once
+    pub fn set_protocol_addr(&self, addr: Multiaddr) {
```
jsdanielh: Same as above: Maybe `outer_protocol_address`

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -271,6 +275,19 @@ impl PeerContactInfo {
...
+        self.meta.write().protocol_addr.get_or_insert_with(|| {
+            debug!(peer_id = %self.peer_id, %addr, "Set protocol address for peer");
```
```suggestion
            trace!(peer_id = %self.peer_id, %addr, "Set protocol address for peer");
```

network-libp2p/src/discovery/handler.rs
jsdanielh: We also need to be careful here when adding circuits, since we would be getting the relay address

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -625,12 +625,11 @@ impl Behaviour {
         self.peer_ids.mark_banned(peer_id);
         debug!(%peer_id, "Banned peer");
 
-        // Mark its addresses as banned if we have them
+        // Mark its protocol address as banned if we have it
```
jsdanielh:
```suggestion
        // Mark its outer protocol address as banned if we have it
```

network-libp2p/src/discovery/handler.rs
Eligioo: Correct, a valid CR address is `/ip4/198.51.100.0/tcp/55555/p2p/QmRelay/p2p-circuit/p2p/QmAlice`.
Everything prior to the `/p2p-circuit/` is the address of the relay which we don't want to ban of course. But that is a concern for a later PR.

jsdanielh (approved):
LGTM

PR #2906, opened 2024-09-10
Limit genesis building volatile database to 100GB

### LINKED ISSUE #2840: Volatile mdbx opening fails on RISC-V

sisou:
Replaces #2842. Please read the discussion there, too!

This is to support platforms that don't support 1TB mapped databases, such as RISC-V in testing. At runtime, the database size can be limited via a config option, but for compilation, no such configuration is possible.

100GB should be more than enough for genesis generation.

## Pull request checklist

### DIFF
--- a/genesis/build.rs
+++ b/genesis/build.rs
@@ -1,6 +1,6 @@
 use std::{env, fmt::Write, fs, path::Path};
 
-use nimiq_database::mdbx::MdbxDatabase;
+use nimiq_database::mdbx::{DatabaseConfig, MdbxDatabase};
 use nimiq_genesis_builder::GenesisBuilder;
 use nimiq_hash::Blake2bHash;
 
@@ -34,8 +34,12 @@ fn generate_albatross(name: &str, out_dir: &Path, src_dir: &Path) {
     let genesis_config = src_dir.join(format!("{name}.toml"));
     log::info!("genesis source file: {}", genesis_config.display());
 
-    let db =
-        MdbxDatabase::new_volatile(Default::default()).expect("Could not open a volatile database");
+    let db = MdbxDatabase::new_volatile(DatabaseConfig {
+        // Limit the database to 100GB to support platforms with a lower supported maximum
+        size: Some(0..(100 * 1024 * 1024 * 1024)),
+        ..Default::default()
+    })
+    .expect("Could not open a volatile database");
     let builder = GenesisBuilder::from_config_file(genesis_config).unwrap();
     let genesis_hash = builder.write_to_files(db, &directory).unwrap();
     write_genesis_rs(&directory, name, &genesis_hash);


genesis/build.rs
```diff
@@ -34,8 +34,12 @@ fn generate_albatross(name: &str, out_dir: &Path, src_dir: &Path) {
     let genesis_config = src_dir.join(format!("{name}.toml"));
     log::info!("genesis source file: {}", genesis_config.display());
 
-    let db =
-        MdbxDatabase::new_volatile(Default::default()).expect("Could not open a volatile database");
+    let db = MdbxDatabase::new_volatile(DatabaseConfig {
+        // Limit the database to 100GB to support platforms with a lower allowed maximum
+        size: Some(0..(100 * GIGABYTE as isize)),
```
hrxi:
```suggestion
        size: Some(0..(100 * 1024 * 1024 * 1024)),
```

hrxi (commented):
Looks good to me from the functionality side.

Could you maybe add "fixes #2840" to the commit message as well?

genesis/build.rs
sisou: Can you explain why you think this suggestion is better?

database/src/mdbx/database.rs
```diff
@@ -10,8 +10,8 @@ use crate::{
     Error,
 };
 
-const GIGABYTE: usize = 1024 * 1024 * 1024;
-const TERABYTE: usize = GIGABYTE * 1024;
+pub const GIGABYTE: usize = 1024 * 1024 * 1024;
+pub const TERABYTE: usize = GIGABYTE * 1024;
```
hrxi: Hmm, sorry, it seems like not all of my comments made it through. I wanted to comment here that I don't think it's good for the database crate to export `GIGABYTE`/`TERABYTE` constants as that seems pretty unrelated to databases.

```suggestion
const GIGABYTE: usize = 1024 * 1024 * 1024;
const TERABYTE: usize = GIGABYTE * 1024;
```

hrxi (approved):
Thanks!

hrxi:
Rebased for merging.

PR #2915, opened 2024-09-16
Add assertion for assumptions, replace `mul_f32` with integer operations

hrxi:
We know that the buffer is empty at that point because it was emptied by `ProposalBuffer::poll_proposal`.


jsdanielh (approved):
LGTM

PR #2922, opened 2024-09-19
Migration: introduce `async_retryer` helper function to allow multiple attempts for Futures that return a `Result`

### LINKED ISSUE #2921: Pow migration: we always assume a RPC call never fails
It has been observed that the `pow-migration` binary can panic because a call to the PoW node through RPC fails. Currently we always assume that such calls never fail but that's not the case.

Ideally we want to add some wrapper around the `pow_client` that retries the same request if it fails and tries again. And only if it fails successively, it actually exits the client with an error.

telegram-cloud-photo-size-4-6008322646678291101-y
telegram-cloud-photo-size-4-6008322646678291145-y

Eligioo:
- Introduce `async_retryer` helper function to allow multiple attempts for Futures that return a `Result`. This is helpful since RPC calls to the PoW node can fail at any point in time.
- Add log message to inform about the PoW history migration progress.
- Fix "off by one" to migrate the PoW history when the amount of block confirmations is reached instead of block confirmations + 1

## Pull request checklist

### DIFF
--- a/pow-migration/src/lib.rs
+++ b/pow-migration/src/lib.rs
@@ -6,6 +6,7 @@ pub mod types;
 
 use std::{
     fmt::Debug,
+    future::Future,
     path::PathBuf,
     process::{exit, Command, ExitStatus},
     time::Duration,
@@ -28,6 +29,8 @@ use crate::{
     state::{get_stakers, get_validators, setup_pow_rpc_server},
     types::{BlockWindows, Error, PoSRegisteredAgents},
 };
+
+const RETRYER_MAX_ATTEMPTS: u8 = 5;
 
 static TESTNET_BLOCK_WINDOWS: &BlockWindows = &BlockWindows {
     // The testnet blocks are produced ~every minute.
@@ -173,7 +176,7 @@ pub async fn migrate(
 
     // Wait for enough confirmations for the candidate block
     loop {
-        let current_height = pow_client.block_number().await.unwrap();
+        let current_height = async_retryer(|| pow_client.block_number()).await.unwrap();
 
         let next_candidate = candidate_block + block_windows.readiness_window;
 
@@ -250,7 +253,7 @@ pub async fn migrate(
     );
 
     loop {
-        let current_height = pow_client.block_number().await.unwrap();
+        let current_height = async_retryer(|| pow_client.block_number()).await.unwrap();
         log::info!(current_height);
 
         let next_candidate = candidate_block + block_windows.readiness_window;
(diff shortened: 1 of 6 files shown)


pow-migration/src/lib.rs
```diff
@@ -469,3 +468,32 @@ pub fn exit_with_error<E: Debug>(error: E, message: &'static str) -> ! {
     log::error!(?error, "{}", message);
     exit(1);
 }
+
+const MAX_ATTEMPTS: u8 = 5;
```
jsdanielh: I'd place the constant above in the beginning of the file

pow-migration/src/lib.rs
```diff
@@ -469,3 +468,32 @@ pub fn exit_with_error<E: Debug>(error: E, message: &'static str) -> ! {
...
+pub async fn async_retryer<F, Fut, T, E>(mut f: F) -> T
```
jsdanielh: Could you add some `rustdoc`?

pow-migration/src/lib.rs
```diff
@@ -469,3 +468,32 @@ pub fn exit_with_error<E: Debug>(error: E, message: &'static str) -> ! {
...
+                    exit_with_error(
+                        err,
+                        "Failed to retrieve data from PoW RPC server after multiple attempts",
+                    )
+                }
+                log::warn!(
+                    ?attempts,
+                    "Request to RPC server failed. Retrying request..."
+                );
+                sleep(Duration::from_millis(1000)).await;
```
jsdanielh: Isn't this too much? 1s seems too much to me

pow-migration/src/lib.rs
```diff
@@ -469,3 +468,32 @@ pub fn exit_with_error<E: Debug>(error: E, message: &'static str) -> ! {
...
+    loop {
+        attempts += 1;
+        match f().await {
+            Ok(val) => return val,
+            Err(err) => {
+                if attempts >= MAX_ATTEMPTS {
+                    exit_with_error(
+                        err,
+                        "Failed to retrieve data from PoW RPC server after multiple attempts",
+                    )
```
jsdanielh: I'd return an error instead of exiting such that we can preserve the call trace

pow-migration/src/lib.rs
Eligioo: Set it to 300ms to still give it some slack if it for example is under load

jsdanielh (approved):

viquezclaudio (approved):

jsdanielh:
Rebased branch for merging it

PR #2929, opened 2024-09-24
Disambiguate VRF seeds across non-competing branches

styppo:
Skip blocks do not change the VRF seed but simply pass it through to the next block, effectively making skip blocks a no-op in terms of VRF seed calculation. This means that before this PR, two non-competing branches (i.e. those that have different skip blocks) will (potentially) get the same VRF seeds for the same block numbers. This breaks some assumptions regarding fork detection and skip block applicability across branches. 

This PR adds the current block number to the VRF seed calculation for regular (non-skip) blocks, effectively encoding the position of skip blocks (because skipped block numbers are **not** part of the VRF), such that non-competing branches get different VRF seeds while competing ones still get the same.

### DIFF
--- a/vrf/src/vrf.rs
+++ b/vrf/src/vrf.rs
@@ -88,12 +88,13 @@ pub struct VrfSeed {
 impl VrfSeed {
     const SIZE: usize = 96;
 
-    /// Verifies the current VRF Seed given the previous VRF Seed (which is part of the message)
-    /// and the signer's public key.
+    /// Verifies the current VRF Seed given the previous VRF Seed (which is part of the message),
+    /// the signer's public key and the nonce.
     pub fn verify(
         &self,
         prev_seed: &VrfSeed,
         public_key: &Ed25519PublicKey,
+        nonce: u32,
     ) -> Result<(), VrfError> {
         // Deserialize signature.
         let V = CompressedEdwardsY::from_slice(&self.signature[..32])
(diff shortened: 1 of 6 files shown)


vrf/src/vrf.rs
```diff
@@ -94,6 +94,7 @@ impl VrfSeed {
         &self,
         prev_seed: &VrfSeed,
         public_key: &Ed25519PublicKey,
+        nonce: u32,
```
jsdanielh, line 97: Could you also update the `rustdoc` of the function?

vrf/src/vrf.rs
```diff
@@ -169,6 +171,7 @@ impl VrfSeed {
     pub fn sign_next_with_rng<R: RngCore + CryptoRng>(
         &self,
         keypair: &KeyPair,
+        nonce: u32,
```
jsdanielh, line 174: Could you also update the `rustdoc` of the function?

jsdanielh (commented):
Just some small comments

sisou:
I edited the PR description to make it clearer (for me and hopefully everyone else) what is happening.

nibhar (approved):
I second the comment updates, but other than that, looks good to me.

Eligioo (approved):

jsdanielh:
I'll update the comments and rebase this PR

jsdanielh (approved):

PR #2931, opened 2024-09-24
Re-request failed chunks during validity sync

### LINKED ISSUE #2930: If a peer does not provide history chunk because it is outdated, we do not re-request the chunk
Observed while trying to sync the seed node as full node to the pubtestnet.

While doing the validity sync process one peer is emitted as outdated:

`The peer could not provide the requested history chunk, we emit it as outdated err=couldn't produce proof peer_id=12D3KooWJuXdSaeeF81PiaLye9pkWytYNNoMkXVS49M6PP838bfc chunk_index=622 block_number=5027010`

However we do not try to re-request the history chunk and the validity sync doesn't complete.

viquezclaudio:
Fixes #2930
If a peer can not provide a history chunk to us, because it is outdated, we properly emit the peer as outdated, however, we also need to re-request the given chunk in order to continue syncing.

...
## Pull request checklist


Eligioo (approved):

Eligioo:
Signing commit

PR #2932, opened 2024-09-24
network-libp2p: Only use secure WS dial attempts when requested

jsdanielh:
Introduce a new transport wrapper (only_secure_ws_transport) that filters out any address that is not a web socket secure address. The transport is inspired on the [global_only transport from libp2p](https://github.com/libp2p/rust-libp2p/blob/master/core/src/transport/global_only.rs). The transport is only enabled when the configuration option `only_secure_ws_connections` is enabled.

## Pull request checklist

### DIFF
--- a/network-libp2p/src/discovery/peer_contacts.rs
+++ b/network-libp2p/src/discovery/peer_contacts.rs
@@ -17,6 +17,8 @@ use parking_lot::RwLock;
 use serde::{Deserialize, Serialize};
 use thiserror::Error;
 
+use crate::utils;
+
 #[derive(Debug, Error)]
 pub enum PeerContactError {
     #[error("Exceeded number of advertised addresses")]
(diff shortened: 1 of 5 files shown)


Eligioo (approved):

hrxi:
What's the motivation for this PR? Is it browser clients?

sisou:
Yes. There is a config setting used in the web-client 

https://github.com/nimiq/core-rs-albatross/blob/dba1ace89d77e53fd65c414d32e8a0fed03f9eac/web-client/src/client/lib.rs#L160

which didn't have the desired effect anymore. Browsers are still connecting to insecure websockets, triggering a UI warning to the user that the website tried to connect to insecure content (from the secure wallet origin). This warning is very visible in Firefox, less so in Chrome, but it scares users by showing them unknown URLs that we don't control and they don't recognize.

PR #2935, opened 2024-09-25
Create .nimiq dir when trying to create example config files

### LINKED ISSUE #2846: Cannot create example config when .nimiq dir doesn't exist
When starting a client without any command line arguments on a "clean" system, it looks for the config file at `$HOME/.nimiq/client.toml`. If that doesn't exist, it fails with a configuration error:

```
Configuration error: Config file not found. Please create one at /home/<user>/.nimiq/client.toml or specify a location using -c path/to/config.toml, see the example config file at /home/<user>/.nimiq/client.toml.example
```

But that error says the user should look for the example config file at `$HOME/.nimiq/client.toml.example`. But that also failed to create if the `.nimiq` directory does not already exist (curiously, there should be an `info` and a `warn` log about that, but those do not show?).

The relevant code is:

https://github.com/nimiq/core-rs-albatross/blob/04340beadf69a271ec5e4661c230236eb4d4600b/lib/src/config/config_file/mod.rs#L85-L107

Furthermore, because the example-config-creating code runs before the config file is read, a new example config file is created again after the user renamed the example file to `client.toml`.

### Issues
- [ ] `.nimiq` directory is not created, so example config file can not be written
- [ ] Failure to write example config file fails silently (logs are not printed)
- [ ] Example config file is written again, even if main config file exists

Related PRs:
- #2401 by @Eligioo
- #2414 by @hrxi

hrxi:
Also fixed a couple of TOCTOUs (probably unnecessary) and state the error if an example config file could not be created.

Fixes #2846.

### DIFF
--- a/lib/src/config/config_file/mod.rs
+++ b/lib/src/config/config_file/mod.rs
@@ -1,5 +1,6 @@
 use std::{
-    collections::HashMap, fmt::Debug, fs::read_to_string, num::NonZeroU8, path::Path, str::FromStr,
+    collections::HashMap, fmt::Debug, fs, io, io::Write as _, num::NonZeroU8, path::Path,
+    str::FromStr,
 };
 
 use log::level_filters::LevelFilter;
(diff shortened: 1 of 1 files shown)


jsdanielh (commented):
Could you address the clippy warnings?

jsdanielh (approved):

PR #2940, opened 2024-09-27
Implement pagination by tx hash for requesting of transaction history

### LINKED ISSUE #2247: [RPC] - U16 Limit for "getTransactionHashesByAddress" prevents all tx to be fetched
## New issue checklist

- [X] I have read all of the [`README`](https://github.com/nimiq-network/core/blob/master/README.md) 
- [X] I have searched [existing issues](https://github.com/nimiq-network/core/issues?q=is%3Aissue+sort%3Acreated-desc) and **this is not a duplicate**.

### General information

- Library version(s): 0.20.0 >
- Browser version(s): - 
- Devices/Simulators/Machine affected: -
- Reproducible in the testnet? (Yes/No):  Yes
- Related issues: -

#### Expected behavior

{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "getTransactionHashesByAddress",
    "params": [
        "NQ65 DHN8 4BSR 5YSX FC3V BB5J GKM2 GB2L H17C",
        90000
    ]
}

To fetch all 90k transactions 

#### Actual behavior

{
    "jsonrpc": "2.0",
    "error": {
        "code": -32602,
        "message": "Invalid params",
        "data": "Expected an object for the request parameters."
    },
    "id": 1
}

if the int is higher than 65535 there is invalid parameter error 

#### Steps to reproduce

--- CORRECT REQUEST
{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "getTransactionHashesByAddress",
    "params": [
        "NQ65 DHN8 4BSR 5YSX FC3V BB5J GKM2 GB2L H17C",
        65535
    ]
}

--- INCORRECT
{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "getTransactionHashesByAddress",
    "params": [
        "NQ65 DHN8 4BSR 5YSX FC3V BB5J GKM2 GB2L H17C",
        65536
    ]
}

NO

I only checked this specific RPC call but there could be more

### LINKED ISSUE #2697: Efficient transaction retrieval beyond limit
Libmdbx supports [set_lowerbound](https://docs.rs/libmdbx/0.5.0/libmdbx/struct.Cursor.html#method.set_lowerbound) on cursors, which allows to seek to a specific duplicate data item **efficiently**.
This is because duplicate items are represented as another B+Tree.

We can thus also retrieve older transaction history efficiently.

### LINKED ISSUE #2733: Allow paging for transactions-by-address queries
As @paberr recently discovered, the tree structure we use for the transaction store lends itself to efficiently request transactions with an offset.

This option should be added to the RPC API, and then to the relevant protocol message.

- [ ] offset in RPC `get_transactions_by_address`
- [ ] offset in protocol receipts message

The change to the message will be a protocol breaking change.

paberr:
This PR adds a `start_at` parameter to various functions used to request the transaction history.
This efficiently adds pagination of results from a known transaction hash (which will not be included in subsequent queries).

@sisou This changes the interface slightly, adding a new parameter. It would be good to have your input on this.

## Pull request checklist

### DIFF
--- a/blockchain/src/history/history_store_index.rs
+++ b/blockchain/src/history/history_store_index.rs
@@ -550,11 +550,13 @@ impl HistoryIndexInterface for HistoryStoreIndex {
 
     /// Returns a vector containing all transaction (and reward inherents) hashes corresponding to the given
     /// address. It fetches the transactions from most recent to least recent up to the maximum
-    /// number given.
+    /// number given. It allows to give a starting point to fetch the transactions from (exclusive). If this hash is given
+    /// but not found or does not belong to this address, the function will return an empty vector.
     fn get_tx_hashes_by_address(
         &self,
         address: &Address,
         max: u16,
+        start_at: Option<Blake2bHash>,
         txn_option: Option<&MdbxReadTransaction>,
     ) -> Vec<Blake2bHash> {
         if max == 0 {
@@ -568,13 +570,30 @@ impl HistoryIndexInterface for HistoryStoreIndex {
         // Seek to the first transaction hash at the given address. If there's none, stop here.
         let mut cursor = txn.dup_cursor(&self.address_table);
 
-        if cursor.set_key(address).is_none() {
-            return tx_hashes;
-        }
+        // Find start index.
+        if let Some(hash) = start_at {
+            let raw_tx_hash = RawTransactionHash::from(hash);
+            // A start hash is given, so we get the `EpochBasedIndex` first.
+            let Some(start_index) = txn.get(&self.tx_hash_table, &raw_tx_hash) else {
+                return tx_hashes;
+            };
 
-        // Then go to the last transaction hash at the given address and add it to the transaction
-        // hashes list.
-        tx_hashes.push(cursor.last_duplicate().expect("This shouldn't panic since we already verified before that there is at least one transactions at this address!").value);
+            // If no entry can be found, return an empty vector.
+            if cursor.set_subkey(address, &start_index).is_none() {
+                return tx_hashes;
+            }
+
+            // We don't add the current hash to the list.
+        } else {
+            // If no start hash is given, we start at the last transaction hash for this address.
+            if cursor.set_key(address).is_none() {
+                return tx_hashes;
+            }
+
+            // Then go to the last transaction hash at the given address and add it to the transaction
+            // hashes list.
+            tx_hashes.push(cursor.last_duplicate().expect("This shouldn't panic since we already verified before that there is at least one transactions at this address!").value);
+        }
 
         while tx_hashes.len() < max as usize {
             // Get previous transaction hash.
@@ -618,6 +637,8 @@ impl HistoryIndexInterface for HistoryStoreIndex {
 
 #[cfg(test)]
 mod tests {
+    use std::ops::Deref;
+
     use nimiq_database::mdbx::MdbxDatabase;
     use nimiq_primitives::{coin::Coin, networks::NetworkId};
     use nimiq_test_log::test;
@@ -1196,6 +1217,7 @@ mod tests {
             &Address::from_user_friendly_address("NQ09 VF5Y 1PKV MRM4 5LE1 55KV P6R2 GXYJ XYQF")
                 .unwrap(),
             99,
+            None,
             Some(&txn),
         );
 
@@ -1209,7 +1231,7 @@ mod tests {
         assert_eq!(query_1[4], *hashes[0]);
 
         let query_2 =
-            history_store.get_tx_hashes_by_address(&Address::burn_address(), 2, Some(&txn));
+            history_store.get_tx_hashes_by_address(&Address::burn_address(), 2, None, Some(&txn));
 
         assert_eq!(query_2.len(), 2);
         assert_eq!(query_2[0], *hashes[6]);
@@ -1219,6 +1241,7 @@ mod tests {
             &Address::from_user_friendly_address("NQ04 B79B R4FF 4NGU A9H0 2PT9 9ART 5A88 J73T")
                 .unwrap(),
             99,
+            None,
             Some(&txn),
         );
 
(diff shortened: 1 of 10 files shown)


consensus/src/consensus/consensus_proxy.rs
```diff
@@ -125,6 +126,7 @@ impl<N: Network> ConsensusProxy<N> {
                     RequestTransactionReceiptsByAddress {
                         address: address.clone(),
                         max,
+                        start_at: start_at.clone(),
```
nibhar, line 129: This `.clone()` is likely unnecessary.
```suggestion
                        start_at,
```

web-client/src/client/lib.rs
```diff
@@ -678,13 +681,20 @@ impl Client {
             }
         }
 
+        let start_at = if let Some(start_at) = start_at {
+            Some(Blake2bHash::from_str(&start_at)?)
+        } else {
+            None
+        };
+
```
nibhar, line 689: Possible, but not sure if better.
```suggestion
        let start_at = start_at.map(|s| Blake2bHash::from_str(&s).ok()).flatten();

```

web-client/src/client/lib.rs
```diff
@@ -718,12 +731,19 @@ impl Client {
...
     ) -> Result<PlainTransactionDetailsArrayType, JsError> {
         let since_block_height = since_block_height.unwrap_or(0);
         let min_peers = min_peers.unwrap_or(1);
 
+        let start_at = if let Some(start_at) = start_at {
+            Some(Blake2bHash::from_str(&start_at)?)
+        } else {
+            None
+        };
+
```
nibhar, line 746: Same as above.

nibhar (approved):
Just minor comments.

paberr:
Feedback from @sisou: for backwards-compatibility, it would be better to have new functions for the pagination and keep the old ones.

web-client/src/client/lib.rs
paberr: We probably want to fail if `Blake2bHash::from_str` doesn't work. That's why I went for the more complicated approach,

blockchain/src/history/interface.rs
```diff
@@ -195,11 +195,13 @@ pub trait HistoryIndexInterface {
 
     /// Returns a vector containing all transaction (and reward inherents) hashes corresponding to the given
     /// address. It fetches the transactions from most recent to least recent up to the maximum
-    /// number given.
+    /// number given. It allows to give a starting point to fetch the transactions from. If this hash is given
+    /// but not found, the function will return an empty vector.
```
Eligioo, line 199: Inform implementors whether this `start_at` hash is inclusive or exclusive.

blockchain/src/history/history_store_index.rs
```diff
@@ -550,11 +550,13 @@ impl HistoryIndexInterface for HistoryStoreIndex {
 
     /// Returns a vector containing all transaction (and reward inherents) hashes corresponding to the given
     /// address. It fetches the transactions from most recent to least recent up to the maximum
-    /// number given.
+    /// number given. It allows to give a starting point to fetch the transactions from. If this hash is given
+    /// but not found or does not belong to this address, the function will return an empty vector.
```
Eligioo, line 554: Inform callers whether this `start_at` hash is inclusive or exclusive.

rpc-interface/src/blockchain.rs
```diff
@@ -93,21 +93,27 @@ pub trait BlockchainInterface {
     /// Returns the hashes for the latest transactions for a given address. All the transactions
     /// where the given address is listed as a recipient or as a sender are considered. Reward
     /// transactions are also returned. It has an option to specify the maximum number of hashes to
-    /// fetch, it defaults to 500.
+    /// fetch, it defaults to 500. It has also an option to retrieve transactions before a given
+    /// transaction hash. If this hash is not found or does not belong to this address, it will return an empty list.
+    /// The transaction hashes are returned in descending order, meaning the latest transaction is the first.
```
Eligioo, line 98: Same as above

rpc-interface/src/blockchain.rs
```diff
@@ -93,21 +93,27 @@ pub trait BlockchainInterface {
...
+        start_at: Option<Blake2bHash>,
     ) -> RPCResult<Vec<Blake2bHash>, (), Self::Error>;
 
     /// Returns the latest transactions for a given address. All the transactions
     /// where the given address is listed as a recipient or as a sender are considered. Reward
     /// transactions are also returned. It has an option to specify the maximum number of transactions
-    /// to fetch, it defaults to 500.
+    /// to fetch, it defaults to 500. It has also an option to retrieve transactions before a given
+    /// transaction hash. If this hash is not found or does not belong to this address, it will return an empty list.
+    /// The transactions are returned in descending order, meaning the latest transaction is the first.
```
Eligioo, line 111: Same as above

Eligioo:
> Feedback from @sisou: for backwards-compatibility, it would be better to have new functions for the pagination and keep the old ones.

It feels a bit ambiguous to have a one method that supports and one which doesn't support the `start_at`. Also since we don't have a RC yet we could have a breaking RPC interface imo.

consensus/src/consensus/consensus_proxy.rs
paberr: No, because it's in a loop

paberr:
> > Feedback from @sisou: for backwards-compatibility, it would be better to have new functions for the pagination and keep the old ones.
> 
> It feels a bit ambiguous to have a one method that supports and one which doesn't support the `start_at`. Also since we don't have a RC yet we could have a breaking RPC interface imo.

This only applies to the web-client and his reasoning is that you always have to provide null for options if you don't want to pass them. But then, there are two `Option` already. Not sure if a third really makes a difference.

sisou:
The problem with having to send `null` applies only to the RPC methods. Javascript (web-client) works correctly with omitted properties (`undefined`).

jsdanielh:
Rebasing branch to merge it

Issue #2934, opened 2024-09-25
Verify that block requests return the correct block


styppo:
We need to check if the returned block has the correct hash (i.e. the one we requested).

PR #2943, opened 2024-09-30
Verifying response from request block and calling cached hash

### LINKED ISSUE #2933: Call `hash_cached()` for blocks received from block requests
This includes the `RequestMissingBlocks` and `RequestBlock`.

ii-cruz:
- Verifies the hash received is the one requested on the head requests.
- Call `hash_cached()` once we receive a block on the head requests and the missing block request of the block queue.

This fixes #2933 and #2934.

## Pull request checklist

### DIFF
--- a/consensus/src/consensus/head_requests.rs
+++ b/consensus/src/consensus/head_requests.rs
@@ -12,8 +12,11 @@ use nimiq_block::Block;
 use nimiq_blockchain_interface::AbstractBlockchain;
 use nimiq_blockchain_proxy::BlockchainProxy;
 use nimiq_hash::Blake2bHash;
-use nimiq_network_interface::{network::Network, request::RequestError};
-use nimiq_utils::stream::FuturesUnordered;
+use nimiq_network_interface::{
+    network::{CloseReason, Network},
+    request::RequestError,
+};
+use nimiq_utils::{spawn, stream::FuturesUnordered};
 
 use crate::messages::{BlockError, RequestBlock, RequestHead, ResponseHead};
 
@@ -101,8 +104,23 @@ impl<TNetwork: Network + 'static> HeadRequests<TNetwork> {
         include_body: bool,
     ) -> Result<Result<Block, BlockError>, RequestError> {
         network
-            .request::<RequestBlock>(RequestBlock { hash, include_body }, peer_id)
+            .request::<RequestBlock>(
+                RequestBlock {
+                    hash: hash.clone(),
+                    include_body,
+                },
+                peer_id,
+            )
             .await
+            .map(|block_res| {
+                // Modify response if the block received does not correspond to the one requested.
+                Ok(block_res.and_then(|mut block| {
+                    if block.hash_cached() != hash {
+                        return Err(BlockError::ResponseHashMismatch);
+                    }
+                    Ok(block)
+                }))
+            })?
     }
 }
 
@@ -151,9 +169,24 @@ impl<TNetwork: Network + 'static> Future for HeadRequests<TNetwork> {
                 (Ok(Ok(block)), peer_id) => {
                     self.unknown_blocks.push((block, peer_id));
                 }
+                // We disconnect from the malicious peer in case of deliberate lying
+                (Ok(Err(BlockError::ResponseHashMismatch)), peer_id) => {
+                    warn!(%peer_id,
+                        "Banning peer due to wrong reply to requested block",
+                    );
+
+                    let network = Arc::clone(&self.network);
+                    spawn(Box::pin({
+                        async move {
+                            network
+                                .disconnect_peer(peer_id, CloseReason::MaliciousPeer)
+                                .await;
+                        }
+                    }));
+                }
                 // We don't do anything with failed requests.
                 (Ok(Err(error)), peer_id) => {
-                    trace!(%error, %peer_id, "Block request failed on remote side");
+                    trace!(%error, %peer_id, "Block request failed on remote side")
                 }
                 (Err(error), peer_id) => {
                     trace!(%error, %peer_id, "Failed block request");
--- a/consensus/src/sync/live/block_queue/block_request_component.rs
+++ b/consensus/src/sync/live/block_queue/block_request_component.rs
@@ -170,7 +170,8 @@ impl<N: Network> BlockRequestComponent<N> {
                     // 3. Check that hash chain verifies
                     // We need to pass through the validators once we reach a new epoch
                     // to be able to verify macro blocks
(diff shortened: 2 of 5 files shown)


consensus/src/consensus/head_requests.rs
```diff
@@ -148,15 +158,30 @@ impl<TNetwork: Network + 'static> Future for HeadRequests<TNetwork> {
         // Then poll blocks.
         while let Poll::Ready(Some(result)) = self.head_blocks.poll_next_unpin(cx) {
             match result {
-                (Ok(Ok(block)), peer_id) => {
-                    self.unknown_blocks.push((block, peer_id));
+                (Ok(Ok(mut block)), peer_id, requested_hash) => {
+                    if requested_hash == block.hash_cached() {
```
nibhar: I would prefer this check to go into `Self::request_block(...)` The hash is known in there and you could just add some `.and_then` in there after the await to check for a header hash mismatch and then return a (new) error variant. That way you can remove the `requested_hash` tuple field again.
For banning you would then have to update the Ok(Err(error)) arm, or add your variant as an arm here.

consensus/src/sync/live/block_queue/block_request_component.rs
```diff
@@ -170,55 +170,72 @@ impl<N: Network> BlockRequestComponent<N> {
                     // 3. Check that hash chain verifies
                     // We need to pass through the validators once we reach a new epoch
                     // to be able to verify macro blocks
-                    let blocks = &response.blocks;
+                    {
```
nibhar: You could probably just get a `&mut` here already removing the need for the additional scope I believe.

consensus/src/sync/live/block_queue/block_request_component.rs
```diff
@@ -170,55 +170,72 @@ impl<N: Network> BlockRequestComponent<N> {
...
+                            if let Err(e) = last_block.verify_validators(&request.epoch_validators)
+                            {
+                                log::error!(
+                                    last_block = last_block.block_number(),
+                                    error = %e,
+                                    "Received invalid chain of missing blocks (macro block does not verify)"
+                                );
+                                return false;
+                            }
+                        }
```
nibhar: Is this cheaper to check than to `.hash_cached()` if that would actually hash + the associated checks there? If not then this check should remain where it was so we check cheap things which discard the block first.

consensus/src/sync/live/block_queue/block_request_component.rs
```diff
@@ -234,31 +251,19 @@ impl<N: Network> BlockRequestComponent<N> {
...
-                    for block in blocks.iter().skip(1) {
-                        if block.block_number() == previous.block_number() + 1
-                            && block.block_number() <= request.target_block_number
-                            && block.parent_hash() == &previous.hash()
+                    for i in 1..blocks.len() {
+                        let previous_block_hash = blocks[i - 1].hash_cached();
+
+                        if blocks[i].block_number() == blocks[i - 1].block_number() + 1
+                            && blocks[i].block_number() <= request.target_block_number
+                            && blocks[i].parent_hash() == &previous_block_hash
```
nibhar: Invert to remove the empty scope and the else.

nibhar (changes requested):
In my opinion a new `BlockError` variant should be added here and the logic to check the correct hash should move into the `request_block` function itself, like mentioned in one of my comments.

Other than that just minor style comments

consensus/src/sync/live/block_queue/block_request_component.rs
ii-cruz: Does not work since we are taking twice by reference the first and last blocks and storing them in temporary variables, we need to drop these before the `&mut` can be taken.

consensus/src/sync/live/block_queue/block_request_component.rs
ii-cruz: This was to keep things a bit more tidy since keeping the order means scoping the mut references too. 

Indeed the aggregation of the signatures is still more expense than the iteration over the `blocks`. I will fix it. Thanks!

consensus/src/consensus/head_requests.rs
```diff
@@ -101,8 +104,23 @@ impl<TNetwork: Network + 'static> HeadRequests<TNetwork> {
...
             .await
+            .and_then(|block_res| {
+                // Modify response if the block received does not correspond to the one requested.
+                Ok(block_res.and_then(|mut block| {
+                    if block.hash_cached() != hash {
+                        return Err(BlockError::ResponseHashMismatch);
+                    }
+                    Ok(block)
+                }))
+            })
```
nibhar: Clippy suggestion here
```suggestion
            .map(|block_res| {
                // Modify response if the block received does not correspond to the one requested.
                block_res.and_then(|mut block| {
                    if block.hash_cached() != hash {
                        return Err(BlockError::ResponseHashMismatch);
                    }
                    Ok(block)
                })
            })
```

consensus/src/consensus/head_requests.rs
```diff
@@ -151,9 +169,28 @@ impl<TNetwork: Network + 'static> Future for HeadRequests<TNetwork> {
...
+                            let network = Arc::clone(&self.network);
+                            spawn(Box::pin({
+                                async move {
+                                    network
+                                        .disconnect_peer(peer_id, CloseReason::MaliciousPeer)
+                                        .await;
+                                }
+                            }));
+                        }
+                    }
```
nibhar: I prefer this for it uses less intendation:
```rust
// We disconnect from the malicious peer in case of deliberate lying
(Ok(Err(BlockError::ResponseHashMismatch)), peer_id) => {
    warn!(%peer_id,
        "Banning peer due to wrong reply to requested block",
    );

    spawn(Box::pin({
        async move {
            network
                .disconnect_peer(peer_id, CloseReason::MaliciousPeer)
                .await;
        }
    }));
}
// We don't do anything with failed requests.
(Ok(Err(error)), peer_id) => {
  trace!(%error, %peer_id, "Block request failed on remote side")
}
```

consensus/src/sync/live/block_queue/block_request_component.rs
ii-cruz: For completion purposes, it is possible to remove the scopes because rust has Non-lexical lifetimes which allows the borrows to be dropped early if they are no longer used, thus allowing for other borrows to take place.

nibhar (approved):
Good to go from my end!

jsdanielh (approved):

jsdanielh:
Rebasing branch to merge it

PR #2945, opened 2024-10-01
Remove LevelUpdate origin from serialized representation

nibhar:
This PR removes the origin from Handel's `LevelUpdates` for the network traffic generated. `LevelUpdate` therefore no longer implements `Serialize` and `Deserialize` such that it cannot be accidentally send again in a future update.

A new utility structure is added, `SerializableLevelUpdate`, which implements `Serialize` and `Deserialize` again while removing the origin. Before sending a `LevelUpdate` is converted into a `SerializableLevelUpdate` and when receiving it uses the `ValidatorMessage`'s `validator_id` to recreate the `LevelUpdate`. This is necessary as in a different PR this origin will be used to ban misbehaving validators.

Minor cleanup to dependencies. I also moved some tests around as they make way more sense there than where they were previously.

### DIFF
--- a/validator/src/aggregation/update.rs
+++ b/validator/src/aggregation/update.rs
@@ -1,0 +1,43 @@
+use nimiq_handel::{contribution::AggregatableContribution, update::LevelUpdate};
+use serde::{Deserialize, Serialize};
+
+/// The serializable/deserializable representation of a LevelUpdate. It does omit the origin of the
+/// LevelUpdate itself, as the ValidatorNetwork's ValidatorMessage already includes it.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+#[serde(bound = "C: AggregatableContribution")]
+pub struct SerializableLevelUpdate<C>
+where
+    C: AggregatableContribution,
+{
+    pub aggregate: C,
+    pub individual: Option<C>,
+    pub level: u8,
+}
+
+impl<C> SerializableLevelUpdate<C>
+where
+    C: AggregatableContribution,
+{
+    /// Given an origin, transforms this SerializableLevelUpdate into a LevelUpdate.
+    pub fn into_level_update(self, origin: u16) -> LevelUpdate<C> {
+        LevelUpdate {
+            aggregate: self.aggregate,
+            individual: self.individual,
+            level: self.level,
+            origin,
+        }
+    }
+}
+
+impl<C> From<LevelUpdate<C>> for SerializableLevelUpdate<C>
+where
+    C: AggregatableContribution,
+{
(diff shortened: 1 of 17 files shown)


sisou:
_Why_ should the origin not be sent over the network?

nibhar:
> _Why_ should the origin not be sent over the network?

It is still send as part of `ValidatorMessage`. So before this PR the origin would be send twice. Here I remove one instance of it and use the other to populate the field in order to maintain the functionality exactly as is.

nibhar:
Note though that the origin which is send over the network as part of the `ValidatorMessage` which wraps the `SerializableLevelUpdate`'s is also not strictly necessary. We only send it as a convenience for easy lookup of the respective validator public key.
As this is a Request/Response we know the sender exactly and could do a lookup with their peer information.

sisou:
So the whole reason for this PR then is to remove a field from a message that is not critical to anything, but instead introduce a new struct and logic to merge the field back into the original struct?

nibhar:
The purpose is to remove unnecessary network traffic. These messages are send a lot so it does make sense to remove unnecessary bytes. This is also a breaking change as it changes a network messages serialized representation.

sisou:
Sent a lot, but not so much that the overhead of the different serialization and merging on deserialization becomes a problem for each node?

nibhar:
> Sent a lot, but not so much that the overhead of the different serialization and merging on deserialization becomes a problem for each node?

I don't understand what you mean? It is not doing a whole lot of computation. It just takes the identifier from one place and plugs it into the other after deserializing. This is not doing any heavy operations here. Where do you see the computational overhead you are talking about?

sisou:
Just checking. If you say it's negligible, then good.

validator/src/aggregation/update.rs
```diff
@@ -0,0 +1,43 @@
+use nimiq_handel::{contribution::AggregatableContribution, update::LevelUpdate};
+use serde::{Deserialize, Serialize};
+
+/// The serializable/deserializable representation of a LevelUpdate. It does away with the origin,
```
jsdanielh:
```suggestion
/// The serializable/deserializable representation of a LevelUpdate. It goes away with the origin,
```

sisou: `does` is correct. But if you want to reword it, we could use "It skips teg origin,"

nibhar: I have changed it for a more straight forward wording.

jsdanielh (approved):
LGTM

jsdanielh:
Rebasing and signing commits to merge it

Issue #2764, opened 2024-07-22
Potential DoS due to `KeyNibbles` deserialization


hrxi:
`\x01\x00\x00` causes the deserializer to panic.

PR #2948, opened 2024-10-02
Fuzz a couple of `Deserialize` implementations

### LINKED ISSUE #2765: Fuzz `Deserialize` implementations
Would have caught #2764 without me trying this by hand.

hrxi:
`Bitset`, `TrieNode` and `KeyNibbles`. I've also made sure that the respective implementations look sane.

I also checked that the fuzzing would have caught #2764, the fuzzer found these crashes immediately.

Some more implementations don't need fuzzing, I added some comments for that.

Fixes #2765.


fuzz/src/bin/00_testcases.rs
nibhar, line 1: Weird name for a file? What do the `00` stand for?

nibhar (approved):
Cool stuff!
Would these be supposed to be run continuously in our CI? Making sure nothing breaks?

jsdanielh (approved):

hrxi:
> Would these be supposed to be run continuously in our CI? Making sure nothing breaks?

That sounds like a good idea, but not implemented in this PR unfortunately.

fuzz/src/bin/00_testcases.rs
hrxi: It's a file different from the others. The 00 is so that it's sorted first. ^^

Issue #276, opened 2021-08-04
network-libp2p: improve reachability of nodes behind a NAT


vlvrd:
The goal of this issue is to improve the reachability of nodes that are behind a NAT (load balancer, reverse proxy, etc). Initially the idea to address this is by improving the reflected source IP of the node such that peers dial an address that is reachable from the public internet. The [libp2p feature request](https://github.com/libp2p/rust-libp2p/issues/2116) to implement the Proxy Protocol that could improve the reachability was marked as not planned. 

Now with the recent release of AutoNAT V2 (https://github.com/libp2p/rust-libp2p/pull/5526) combined with the [Circuit Relays protocol](https://docs.libp2p.io/concepts/nat/circuit-relay/), is meant to be the next proposal to improve this reachability.

Tasks:
- [x] Update the AutoNAT V1 in the root `NetworkBehaviour` to AutoNAT V2 (PR #2955)
- [ ] Add the Circuit Relay protocol to the root `NetworkBehaviour`

PR #2955, opened 2024-10-04
network-libp2p: Migrate AutoNAT v1 protocol to v2

Eligioo:
Transition from AutoNAT v1 protocol to v2
Main differences:
- `autonat v1` allowed testing reachability for the node. `autonat v2` allows testing reachability for an individual address.
- `autonat v2` provides a mechanism for nodes to verify whether the peer actually successfully dialled an address.
- `autonat v2` provides a mechanism for nodes to dial an IP address different from the requesting node's observed IP address without risking amplification attacks. `autonat v1` disallowed such dials to prevent amplification attacks.

This PR should contribute improving the issues stated by #276 

## Pull request checklist


network-libp2p/src/autonat.rs
```diff
@@ -0,0 +1,97 @@
+use std::collections::{HashMap, HashSet};
+
+use libp2p::Multiaddr;
+
+#[derive(Clone, Debug, Default, PartialEq)]
```
nibhar: This could probably get `Copy`, though it does not do much here.

network-libp2p/src/autonat.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    pub fn set_listen_address_nat_status(&mut self, address: Multiaddr, nat_status: NatStatus) {
+        if let Some(status) = self.listen_address_status.get_mut(&address) {
+            *status = nat_status.clone();
+
+            if nat_status == NatStatus::Public {
+                self.confirmed_addresses.insert(address);
+            } else {
+                self.confirmed_addresses.remove(&address);
+            }
```
nibhar: Swap around to avoid the `.clone()`.
```suggestion
            if nat_status == NatStatus::Public {
                self.confirmed_addresses.insert(address);
            } else {
                self.confirmed_addresses.remove(&address);
            }

            *status = nat_status;
```

network-libp2p/tests/network.rs
```diff
@@ -318,7 +317,7 @@ async fn create_network_with_n_peers(n_peers: usize) -> Vec<Network> {
 
 #[test(tokio::test)]
 async fn connections_stress_and_reconnect() {
-    let peers: usize = 10;
+    let peers: usize = 5;
```
nibhar, line 320: Why this change?

nibhar (approved):
LGTM, but i have limited knowledge of the inner workings of libp2p here. Probably a good idea to have another reviewer as well.

network-libp2p/src/autonat.rs
Eligioo: Is now a `Copy` 👍

network-libp2p/src/autonat.rs
Eligioo: Swap not necessary since it's now a `Copy` 👍

network-libp2p/tests/network.rs
Eligioo: This test is flaky and with the old number of peers, the network in the CI would progress too slow as there is more going on during network bootstrapping (namely a multi roundtrip process for the Autonat probes per listen adress).

Peers start to disconnect after 30 seconds thus assertions don't hold anymore. With 5 peers it seems the CI is able to keep up.

nibhar: Okay. We should maybe change the test setup then eventually.

network-libp2p/src/autonat.rs
```diff
@@ -0,0 +1,97 @@
...
+    Public,
+    Private,
+    #[default]
+    Unknown,
+}
+
+#[derive(Default)]
+pub(crate) struct NatState {
+    confirmed_addresses: HashSet<Multiaddr>,
+    listen_address_status: HashMap<Multiaddr, NatStatus>,
```
jsdanielh: Why listen? Maybe only: `addresses_status`?

network-libp2p/src/autonat.rs
```diff
@@ -0,0 +1,97 @@
...
+    status: NatStatus,
+}
+
+impl NatState {
+    pub fn add_listen_address(&mut self, address: Multiaddr) {
```
jsdanielh: Maybe only `add_address`?

network-libp2p/src/autonat.rs
```diff
@@ -0,0 +1,97 @@
+use std::collections::{HashMap, HashSet};
+
+use libp2p::Multiaddr;
+
+#[derive(Copy, Clone, Debug, Default, PartialEq)]
+pub(crate) enum NatStatus {
```
jsdanielh, line 7: Could you add some `rustdoc`?

network-libp2p/src/autonat.rs
jsdanielh: Could you add some `rustdoc`?

network-libp2p/src/autonat.rs
```diff
@@ -0,0 +1,97 @@
...
+    status: NatStatus,
+}
+
+impl NatState {
+    pub fn add_listen_address(&mut self, address: Multiaddr) {
+        self.listen_address_status
+            .insert(address, NatStatus::Unknown);
+    }
+
+    pub fn remove_listen_address(&mut self, address: &Multiaddr) {
```
jsdanielh: Could you add some `rustdoc`?

network-libp2p/src/autonat.rs
```diff
@@ -0,0 +1,97 @@
...
+        if let Some(status) = self.listen_address_status.get_mut(&address) {
+            *status = nat_status;
+
+            if nat_status == NatStatus::Public {
+                self.confirmed_addresses.insert(address);
+            } else {
+                self.confirmed_addresses.remove(&address);
+            }
+            self.update_state();
+        }
```
jsdanielh:
```suggestion
        let Some(status) = self.listen_address_status.get_mut(&address) else {return;};
        *status = nat_status;

        if nat_status == NatStatus::Public {
            self.confirmed_addresses.insert(address);
        } else {
            self.confirmed_addresses.remove(&address);
        }
        self.update_state();
}
```

network-libp2p/src/autonat.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    pub fn add_confirmed_address(&mut self, address: Multiaddr) {
+        if let Some(address_status) = self.listen_address_status.get_mut(&address) {
```
jsdanielh: To me, we should add it to `listen_address` if it doesn't exist and add them to `confirmed_addresses` regardless

network-libp2p/src/autonat.rs
```diff
@@ -0,0 +1,97 @@
...
+            *address_status = NatStatus::Public;
+
+            self.confirmed_addresses.insert(address);
+            self.update_state();
+        }
+    }
+
+    pub fn remove_confirmed_address(&mut self, address: &Multiaddr) {
+        if let Some(address_status) = self.listen_address_status.get_mut(address) {
```
jsdanielh: To me, we should add it to `listen_address` if it doesn't exist and remove it from `confirmed_addresses` regardless

network-libp2p/src/network_types.rs
```diff
@@ -238,6 +239,8 @@ pub(crate) struct TaskState {
     pub(crate) dht_bootstrap_state: DhtBootStrapState,
     /// DHT (kad) is in server mode
     pub(crate) dht_server_mode: bool,
+    /// The NAT status of the local peer
+    pub(crate) nat_status: NatState,
```
jsdanielh, line 243: Maybe I'm missing something here, but why do we want this `nat_status`? I don't see any reader of this

jsdanielh:
@Eligioo I guess that with autonat v2, [this](https://github.com/nimiq/core-rs-albatross/blob/albatross/lib/src/config/config_file/client.example.toml#L37) is no longer needed.

network-libp2p/src/autonat.rs
Eligioo: Changed it to `self.listen_address_status.entry.or_default()` so it will always exists in `listen_address`

network-libp2p/src/autonat.rs
Eligioo: There will always be an entry now

network-libp2p/src/network_types.rs
Eligioo: It is necessary to keep a state of the all the addresses in order to determine the final NAT state. With `autonatv1`, it would give you a NAT of `Public` or `Private` eventually but v2 only gives you the results of the probes. Thus you have to keep track of everything yourself.

jsdanielh: But who's reading this status? I am only seeing some logging related to this

jsdanielh (approved):

network-libp2p/src/network_types.rs
jsdanielh: We discussed it internally and we decided to keep it for `Nat` status visibility.

PR #2961, opened 2024-10-08
Handel refactor and update

nibhar:
This PR updates Handel.

Among the changes:

* Adds more testing
* It makes the difference between the signers slots of an aggregation in contrast to the signers of an aggregation more clear by using `Identity` for the latter.
* It makes the choice of the first recipients of messages deterministic. It is also pair wise symetric such that the overall coverage of messages should improve.
* Removes the `FinishedAggreation` in favour of a better mechanism to stop sending updates, except for the full one.
  * Polling of updates moved from the `PendingContributionList` to `Aggregation` to facilitate this auto response.
* Deduplicate send messages on a per peer basis. It will not send an identical message twice in a row.
* Cleanup of parts of the validator.

Likely there is a bunch more to say here and will be updated at a later time.

### DIFF
--- a/collections/src/bitset.rs
+++ b/collections/src/bitset.rs
@@ -23,13 +23,7 @@ pub struct BitSet {
 }
 
 impl BitSet {
-    /// An empty BitSet.
-    pub const EMPTY: Self = Self {
-        store: vec![],
-        count: 0,
-    };
-
-    pub fn new() -> Self {
+    pub const fn new() -> Self {
         BitSet {
             store: Vec::new(),
             count: 0,
--- a/collections/src/bitset.rs
+++ b/collections/src/bitset.rs
@@ -23,6 +23,12 @@ pub struct BitSet {
 }
 
 impl BitSet {
+    /// An empty BitSet.
+    pub const EMPTY: Self = Self {
+        store: vec![],
+        count: 0,
+    };
+
     pub fn new() -> Self {
         BitSet {
             store: Vec::new(),
--- a/handel/src/aggregation.rs
+++ b/handel/src/aggregation.rs
@@ -151,7 +151,7 @@ where
             level.start();
 
             debug!(
-                id = ?self.protocol.identify(),
+                id = %self.protocol.identify(),
                 level = level.id,
                 activated_by,
                 "Starting level"
@@ -210,7 +210,7 @@ where
         }
 
         trace!(
-            id = ?self.protocol.identify(),
+            id = %self.protocol.identify(),
             level_id,
             num_peers,
             "Level complete",
@@ -431,9 +431,10 @@ where
         let evaluator = self.protocol.evaluator();
         while let Poll::Ready(Some(update)) = self.input_stream.poll_next_unpin(cx) {
             // Verify the level update.
-            if !evaluator.verify(&update) {
-                trace!(
-                    id = ?self.protocol.identify(),
+            if let Err(error) = evaluator.verify(&update) {
+                warn!(
+                    id = %self.protocol.identify(),
+                    ?error,
                     ?update,
                     "Rejecting invalid level update",
                 );
@@ -538,7 +539,7 @@ where
         // If the best aggregate is a full aggregation, this aggregation is finished.
         if self.is_complete_aggregate(&best_aggregate) {
             debug!(
-                id = ?self.protocol.identify(),
+                id = %self.protocol.identify(),
                 "Aggregation complete"
             );
 
--- a/handel/src/aggregation.rs
+++ b/handel/src/aggregation.rs
@@ -97,6 +97,7 @@ where
         let levels = Arc::new(Level::create_levels(
             protocol.partitioner(),
             protocol.identify(),
+            protocol.node_id(),
         ));
 
         // Create an empty list which can later be polled for the best available pending contribution.
--- a/handel/src/aggregation.rs
+++ b/handel/src/aggregation.rs
@@ -1,6 +1,5 @@
 use std::{
     pin::Pin,
-    sync::Arc,
     task::{Context, Poll},
 };
 
@@ -44,7 +43,7 @@ where
     config: Config,
 
     /// Levels
-    levels: Arc<Vec<Level>>,
+    levels: Vec<Level>,
 
     /// Stream of level updates received from other peers.
     input_stream: LevelUpdateStream<P::Contribution>,
@@ -94,11 +93,11 @@ where
         let sender = NetworkHelper::new(protocol.partitioner().size(), network);
 
         // Invoke the partitioner to create the level structure of peers.
-        let levels = Arc::new(Level::create_levels(
+        let levels = Level::create_levels(
             protocol.partitioner(),
             protocol.identify(),
             protocol.node_id(),
-        ));
+        );
 
         // Create an empty list which can later be polled for the best available pending contribution.
         let mut pending_contributions =
--- a/handel/src/aggregation.rs
+++ b/handel/src/aggregation.rs
@@ -131,46 +131,47 @@ where
         }
     }
 
-    /// Starts level `level`
-    fn start_level(
-        &mut self,
-        level: usize,
-        store: &<P as Protocol<TId>>::Store,
-        activated_by: &'static str,
-    ) {
-        // Try to start the level. `level.start()` returns false if the level was already started.
-        let level = &self.levels[level];
-        if !level.start() {
-            return;
-        }
-
-        debug!(
-            id = ?self.protocol.identify(),
-            level = level.id,
-            activated_by,
-            "Starting level"
-        );
-
-        // Reset the level timeout.
-        self.start_level_interval = interval(self.config.level_timeout);
-
-        // Nothing else to do for level 0 as it only contains this node.
-        if level.id == 0 {
-            return;
-        }
-
-        // Get the current best aggregate for the level.
-        let Some(best) = store.combined(level.id - 1) else {
+    /// Starts all levels up to `level`.
+    fn start_level(&mut self, level: usize, store: &P::Store, activated_by: &'static str) {
+        // Find the first level that has not been started yet.
+        // If all levels are already started, there is nothing to do.
+        let Some(to_start) = self.levels.iter().position(|level| !level.is_started()) else {
             return;
         };
 
-        // Send the best aggregate to the peers on the level.
-        self.send_update(
-            best,
-            level.id,
-            !level.is_complete(),
-            level.select_next_peers(self.config.peer_count),
-        );
+        // If we have already started a higher level, there is nothing to do.
+        if to_start > level {
+            return;
+        }
+
+        // Activate all levels between `to_start` (inclusive) and `level` (inclusive).
+        for lvl in to_start..=level {
+            // Start the level.
+            let level = &self.levels[lvl];
+            level.start();
+
+            debug!(
+                id = ?self.protocol.identify(),
+                level = level.id,
+                activated_by,
+                "Starting level"
+            );
+
+            // Nothing else to do for level 0 as it only contains this node.
+            if level.id == 0 {
+                continue;
+            }
+
+            // Send the best aggregate to the peers on the level.
+            self.send_update(
+                level.id,
+                level.select_next_peers(self.config.peer_count),
+                store,
+            );
+        }
+
+        // Reset the level timeout.
+        self.start_level_interval = interval(self.config.level_timeout);
     }
 
     fn num_contributors(&self, aggregate: &P::Contribution) -> usize {
@@ -185,7 +186,7 @@ where
     }
 
     /// Check if the given level was completed.
-    fn check_completed_level(&mut self, level_id: usize, store: &<P as Protocol<TId>>::Store) {
+    fn check_completed_level(&mut self, level_id: usize, store: &P::Store) {
         // Nothing to do if the level is already complete.
         let level = &self.levels[level_id];
         if level.is_complete() {
@@ -230,12 +231,11 @@ where
 
         if !next_level.is_started() {
             self.start_level(next_level_id, store, "LevelComplete");
-        } else if let Some(best_aggregate) = store.combined(next_level_id - 1) {
+        } else {
             self.send_update(
-                best_aggregate,
                 next_level_id,
-                true,
                 next_level.select_next_peers(self.config.peer_count),
+                store,
             );
         }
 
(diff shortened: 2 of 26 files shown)


Eligioo:
@nibhar it has quite some merge conflicts with the head

collections/src/bitset.rs
```diff
@@ -23,6 +23,12 @@ pub struct BitSet {
 }
 
 impl BitSet {
+    /// An empty BitSet.
+    pub const EMPTY: Self = Self {
+        store: vec![],
+        count: 0,
+    };
+
```
hrxi: Would be easier to make `BitSet::new()` `const` so you can use it in place of the constant. Other containers in Rust don't have such a constant either, you simply use `Vec::new()` or `HashSet::new()`.

handel/src/aggregation.rs
```diff
@@ -43,7 +44,10 @@ where
     config: Config,
 
     /// Levels
-    levels: Vec<Level>,
+    levels: Arc<Vec<Level>>,
```
hrxi: Why is that now inside an `Arc`? I don't see any clones of the `Arc`, so it seems there's only ever a single instance.

hrxi:
> * Deduplicate send messages on a per peer basis. It will not send an identical message twice in a row.

Not having looked the code for this one, how does that work when sending the message fails? Is that accounted for?

**Update:** Yes, it is accounted for by allowing duplicate messages every 5 seconds.

handel/src/network.rs
```diff
@@ -1,132 +1,180 @@
...
 /// It will buffer up to one message per recipient, while maintaining the order of messages.
 /// Messages that do not fit the buffer will be dropped.
-pub struct LevelUpdateSender<TNetwork: Network> {
-    /// Buffer for one message per recipient. Second value of the pair is the index of the next message which needs to
-    /// be sent after this one. In case there is none it will be set to the len of the Vec, which is an OOB index.
+pub struct NetworkHelper<TNetwork: Network> {
+    /// Buffer for one message per recipient. Second value of the pair is the index of the next
+    /// message which needs to be sent after this one. If there is no message buffered, it will
+    /// point to `message_buffer.len()` which is the first OOB index.
     message_buffer: Vec<Option<(LevelUpdate<TNetwork::Contribution>, usize)>>,
```
hrxi: This should be a [`LinkedHashMap`](https://docs.rs/linked-hash-map/0.5.6/linked_hash_map/struct.LinkedHashMap.html) instead, if I understand its usage correctly. It's a hashmap that remembers insertion order.

handel/src/network.rs
```diff
@@ -1,132 +1,180 @@
...
 
-    /// Index of the first message in this buffer that should be sent. If there is no message buffered it will point to
-    /// `message_buffer.len()` which is the first OOB index.
+    /// Information about the last message sent to each recipient, used to deduplicate messages.
+    last_messages: Vec<Option<LastLevelUpdate>>,
```
hrxi: This should probably be a `HashMap`, it's emulating one.

handel/src/network.rs
```diff
@@ -1,132 +1,180 @@
...
+    /// Index of the first message in this buffer that should be sent. If there is no message
+    /// buffered it will point to `message_buffer.len()` which is the first OOB index.
     first: usize,
 
-    /// Index of the last message in this buffer that should be sent. It is used to quickly append messages.
-    /// If there is no message buffered it will point to `message_buffer.len()` which is the first OOB index.
+    /// Index of the last message in this buffer that should be sent. It is used to quickly append
+    /// messages. If there is no message buffered it will point to `message_buffer.len()` which is
+    /// the first OOB index.
     last: usize,
```
hrxi: These two integers `first` and `last` can go if a `LinkedHashMap` is used.

handel/src/network.rs
```diff
@@ -1,132 +1,180 @@
...
             waker: None,
         }
     }
 
-    pub fn send(&mut self, (msg, recipient): (LevelUpdate<TNetwork::Contribution>, usize)) {
+    pub fn send(&mut self, node_id: Identity, msg: LevelUpdate<TNetwork::Contribution>) {
+        // Get the index of the node from the identity. This will fail for any
+        // identity which is empty or has more than one identity within it.
+        // Annoyingly the compiler cannot infer the usize here properly.
+        let Ok(node_id): Result<usize, _> = node_id.try_into() else {
```
hrxi, line 93: We don't expect anything but single-node identities to be passed to this function. I think a `usize` as a function parameter would be better suited to represent this.

If we don't want to change the `Identity` parameter, I'd suggest panicking here, because calling this function with an identity containing more than one node is a programming error.

handel/src/network.rs
```diff
@@ -1,132 +1,180 @@
...
+            return;
+        };
+
+        // `message_buffer` and `last_messages` have the same length, so this check prevents
+        // out-of-bounds access to both of these vectors.
         let len = self.message_buffer.len();
+        if node_id >= len {
+            error!(node_id, len, "Attempted to send to out-of-bounds node_id");
```
hrxi: This should probably be a panic.

handel/src/network.rs
```diff
@@ -1,132 +1,180 @@
...
 
-impl<TNetwork: Network + Unpin> Future for LevelUpdateSender<TNetwork> {
+impl<TNetwork: Network + Unpin> Future for NetworkHelper<TNetwork> {
     type Output = ();
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        // Loop here such that after polling pending futures there is an opportunity to create new ones
-        // before returning.
+        // Drive pending ban futures.
+        while let Poll::Ready(Some(_)) = self.pending_bans.poll_next_unpin(cx) {}
```
hrxi:
```suggestion
        while let Poll::Ready(Some(())) = self.pending_bans.poll_next_unpin(cx) {}
```

handel/src/aggregation.rs
```diff
@@ -54,25 +58,26 @@ where
     /// Our contribution
     contribution: P::Contribution,
 
-    /// Sink used to relay messages
-    sender: LevelUpdateSender<N>,
+    /// Gateway to the network to send updates and ban nodes.
+    network: NetworkHelper<N>,
 
-    /// Interval for starting the next level regardless of previous levels completion
+    /// Timeout for starting the next level regardless of previous level's completion.
```
hrxi:
```suggestion
    /// Interval for starting the next level regardless of previous level's completion.
```
Or also change the comment for `periodic_update_interval` below.

handel/src/aggregation.rs
```diff
@@ -54,25 +58,26 @@ where
...
-    /// the level which needs activation next
-    next_level_timeout: usize,
-
     /// Future of the currently verified pending contribution.
     /// There is only ever one contribution being verified at a time.
     current_verification:
         Option<BoxFuture<'static, (VerificationResult, PendingContribution<P::Contribution>)>>,
+
+    /// The final result of the aggregation once it has been produced.
+    /// A Some(_) value here indicates that the aggregation has finished.
```
hrxi:
```suggestion
    /// A `Some(_)` value here indicates that the aggregation has finished.
```

handel/src/aggregation.rs
```diff
@@ -82,73 +87,91 @@ where
         protocol: P,
         config: Config,
         own_contribution: P::Contribution,
-        input_stream: LevelUpdateStream<P, TId>,
-        sender: LevelUpdateSender<N>,
+        input_stream: LevelUpdateStream<P::Contribution>,
+        network: N,
     ) -> Self {
+        // Create the Sender, buffering a single message per recipient.
```
hrxi:
```suggestion
        // Create the sender, buffering a single message per recipient.
```

handel/src/aggregation.rs
```diff
@@ -82,73 +87,91 @@ where
...
+        ));
 
         // Create an empty list which can later be polled for the best available pending contribution.
         let mut pending_contributions =
-            PendingContributionList::new(protocol.identify(), protocol.evaluator(), input_stream);
+            PendingContributionList::new(protocol.identify(), protocol.evaluator());
 
         // Add our own contribution to the list.
-        pending_contributions.add_contribution(own_contribution.clone(), 0);
+        pending_contributions.add_contribution(own_contribution.clone(), 0, protocol.node_id());
```
hrxi: Theoretically, there's no reason to do the expensive verification for our own contribution, right?

handel/src/aggregation.rs
```diff
@@ -162,220 +185,202 @@ where
...
-        }
-    }
+        // If there are no levels to start, we're done.
+        let Some(to_start) = to_start else {
+            return;
+        };
 
-    fn into_inner(self) -> (LevelUpdateStream<P, TId>, LevelUpdateSender<N>) {
-        (self.pending_contributions.into_stream(), self.sender)
+        // Start the level.
```
hrxi, line 341: The new code seems to start a next level by timeout every 400 ms, but the old code only did it if the targeted level wasn't started already. Is this change intentional? What's the reasoning for this change?

handel/src/aggregation.rs
```diff
@@ -162,220 +185,202 @@ where
...
-        let num_peers = {
-            let level = self
-                .levels
-                .get(level_id)
-                .expect("Attempted to check completeness of invalid level");
-
-            // check if level already is completed
-            if level.state.read().receive_completed {
-                // The level was completed before so nothing more to do.
+    /// Check if the given level was completed.
```
hrxi:
```suggestion
    /// Check if the given level was completed and mark it as such.
    ///
    /// Only if the level was marked as completed by this call, it also start the next level automatically and recursively checks if any further level is also completed.
```

handel/src/aggregation.rs
```diff
@@ -162,220 +185,202 @@ where
...
+        };
 
-            // Create the LevelUpdate
-            let update = LevelUpdate::<P::Contribution>::new(
-                contribution,
-                individual,
-                level_id,
-                self.protocol.node_id(),
-            );
+        // FIXME Comment
```
hrxi: ^^

handel/src/aggregation.rs
```diff
@@ -395,9 +400,23 @@ where
         self.current_verification = Some(fut);
         None
     }
+
+    /// Responds to an update received from `node_id` with our own best aggregate for `level_id`.
+    /// If we already have a final result, we send that result instead.
+    fn respond_to_update(&mut self, node_id: usize, level_id: usize) {
```
hrxi:
```suggestion
    fn respond_to_update(&mut self, level_id: usize, node_id: usize) {
```
Other functions take `level_id` first and `node_id` second. Careful, also needs updating of the call sites!

handel/src/aggregation.rs
```diff
@@ -406,255 +425,137 @@ where
...
-                self.sender.send(response);
-            } else {
-                log::error!("Failed to poll input stream; None returned")
             }
         }
 
-        // Send as many messages as possible.
-        assert!(self.sender.poll_unpin(cx).is_pending());
+        // Drive the network helper future. It always returns Pending.
+        let _ = self.network.poll_unpin(cx);
```
hrxi: The `assert!` can stay, then it's not only documented in the comment but actually verified at runtime.

handel/src/identity.rs
```diff
@@ -6,63 +6,107 @@ use nimiq_collections::bitset::BitSet;
 use crate::contribution::AggregatableContribution;
 
 /// Struct that defines an identity.
-/// An identity is composed of zero, one or more signers of a contribution.
+/// An identity is composed of zero, one or more identities.
```
hrxi, line 9: This comment doesn't really help me. "An X is composed of zero, one or more Xs." I don't have a good idea for a replacement.

Perhaps the struct could be named like `IdentitySet` or `PeerSet` or similar?

nibhar:
> @nibhar it has quite some merge conflicts with the head

I know, I have rebased it.

handel/src/partitioner.rs
```diff
@@ -71,16 +72,24 @@ impl Partitioner for BinomialPartitioner {
     }
 
     fn level_size(&self, level: usize) -> usize {
-        if let Ok(range) = self.range(level) {
-            (range.end() - range.start()) + 1
+        if let Ok(range) = self.identities_on(level) {
```
hrxi:
```suggestion
        if let Ok(identities) = self.identities_on(level) {
```

hrxi:
```suggestion
        if let Ok(identities) = self.identities_on(level) {
```

handel/tests/mod.rs
```diff
@@ -210,227 +208,290 @@ struct NetworkWrapper<N: NetworkInterface<PeerId = MockPeerId>>(Arc<N>);
...
+}
 
-    // The final value needs to be the sum of all contributions.
-    // For instance for `contributor_num = 7: 8 + 7 + 6 + 5 + 4 + 3 + 2 + 1 = 36`
-    let mut exp_agg_value = 0u64;
-    for i in 0..=contributor_num {
-        exp_agg_value += i as u64 + 1u64;
-    }
+#[test(tokio::test)]
+#[ignore]
```
hrxi, line 415: Why is this test `#[ignore]`d?

hrxi, line 415: Why is this test `#[ignore]`d?

hrxi (commented):
Thanks for the huge handel refactor, I think it improved the code quality a lot.

handel/src/network.rs
nibhar: This is done in order to only ever handle `Indentity`'s except for the very beginning where we retrieve them from the signers `BitSet` or before we must send stuff, where we need a `usize`. I could see changing Identity to something like: 
```rust
enum Identity {
    Single(usize),
    Multiple(BitSet),
    None,
}
```
Would you prefer that?

handel/src/identity.rs
nibhar: See other comment related to `Identity`.

handel/src/aggregation.rs
nibhar: True!

Where would you introduce that check? It is rather easy to put a 
```rust 
if level == 0 {
    return; // or continue or break
}
```
`Aggregation::start_verification` would probably be my choice.

handel/src/aggregation.rs
nibhar: If I understand correctly this used to start a level every set duration `d`. Meaning that if every level but the last was started immediately you would have to wait `(level_count - 1) * d`. Now it should start after just one times `d` elapses. Note that `Aggregation::start_level` does also reset this interval.

Basically before the change every `d` a level would attempted to be started, no matter the current circumstances. Now a new level will be started after `d` has elapsed since the previous level was started.

handel/src/aggregation.rs
nibhar: Only one call side. Changed it there, too.

handel/src/partitioner.rs
nibhar: Duplicate?

handel/tests/mod.rs
nibhar: Closed as duplicate

nibhar: So currently the requirement is that validators are somewhat well connected amongst themselves. The `LevelUpdates` are send in a direct fashion. This test connects validators as a straight line, with every validator only be connected to its predecessor. Basically it is a very poorly connected graph. 

As it stands Handel cannot deal with this. The information required cannot be send in such a graph. There is a design for a future upgrade to Handel which will enable such scenarios to resolve. Once we get to implementing that design we will use this test to verify its efficacy. Until then the mitigation is to wait until the network connects better amongst itself, which our network should constantly try to do anyways.

handel/src/aggregation.rs
hrxi: Interesting question. I think I wouldn't just make it dependent on the `level`, as that might lead to bugs later on. Perhaps another `bool` could be added to a `PendingContribution` that specifies whether it's trusted already?

hrxi (approved):

jsdanielh:
Rebasing branch and signing commits to merge it

PR #2969, opened 2024-10-11
pow-migration: Remove support for pre-staking re-delegation

jsdanielh:
Remove support for pre-staking re-delegation (switching validators won't be possible) and only count pre-staking transactions that increase the stake of the previous delegated validator. Otherwise, ignore the transaction.

## Pull request checklist


sisou:
Did you run the tool with this change and found no re-delegation so far?

jsdanielh:
We are not going to remove support to pre-staking re-delegation.

PR #2970, opened 2024-10-11
web-client: Add network connect/disconnect functions

jsdanielh:
Add network functions to disconnect from every connected peer and to be able to re-start connecting after disconnecting from every peer.

## Pull request checklist

### DIFF
--- a/web-client/src/client/lib.rs
+++ b/web-client/src/client/lib.rs
@@ -876,6 +876,27 @@ impl Client {
         Ok(serde_wasm_bindgen::to_value(&txs)?.into())
     }
 
+    /// This function is used to tell the network to disconnect from every connected
+    /// peer and stop trying to connect to other peers.
+    ///
+    /// **Important**: this function returns when the signal to disconnect was sent,
+    /// before all peers actually disconnect. This means that in order to ensure the
+    /// network is disconnected, wait for all peers to disappear after calling.
+    #[wasm_bindgen(js_name = disconnectNetwork)]
+    pub async fn disconnect_network(&self) {
+        let network = self.inner.network();
+        network.disconnect(CloseReason::GoingOffline).await;
+    }
+
+    /// This function is used to tell the network to (re)start connecting to peers.
+    /// This is could be used to tell the network to restart connection operations after
+    /// disconnect network is called.
+    #[wasm_bindgen(js_name = connectNetwork)]
+    pub async fn connect_network(&self) {
+        let network = self.inner.network();
+        network.start_connecting().await;
+    }
+
     fn setup_offline_online_event_handlers(&self) {
         let network = self.inner.network();
 


sisou (approved):
Looks good. Can we directly call `restart_connecting()` after we awaited `disconnect_peers()`?

If so, can we wrap them in a new method called `reset_consensus()` that does exactly that?

web-client/src/client/lib.rs
```diff
@@ -856,6 +856,23 @@ impl Client {
...
+    }
+
+    /// This function is used to tell the network to restart connecting to peers.
+    /// This is intended to tell the network to restart connection operations after
+    /// disconnect peers is called.
+    #[wasm_bindgen(js_name = restartConnecting)]
+    pub async fn restart_connecting(&self) {
+        let network = self.inner.network();
+        network.start_connecting().await;
+    }
```
sisou: Is there any problem with calling this method while the client is connected as usual? Does it break anything?

I could imagine simply calling the methods `disconnect_network()` and `connect_network()`  (as we don't have a `client.network` subclass like the PoW client has).

Albermonte (approved):

jsdanielh:
> Looks good. Can we directly call `restart_connecting()` after we awaited `disconnect_peers()`?
> 
> If so, can we wrap them in a new method called `reset_consensus()` that does exactly that?

I'd check that no peers are still there after `disconnect_peers`. This is an `async` instruction sent to the network that runs in it's own task. So, something like this:
```
await disconnectPeers();
while (!(timeout || num_peers == 0)) {
    console.log("Waiting peers to disconnect");
}
await restartConnecting();
````

File: web-client/src/client/lib.rs (hunk 1 above)
Comment (jsdanielh) [reply web-client/src/client/lib.rs]: Yeah, I think this makes sense. Will make the change

File: web-client/src/client/lib.rs [hunk 2]
```diff
@@ -876,6 +876,23 @@ impl Client {
         Ok(serde_wasm_bindgen::to_value(&txs)?.into())
     }
 
+    /// This function is used to tell the network to disconnect from every connected
+    /// peer and stop trying to connect to other peers.
```
sisou, line 880:
```suggestion
    /// peer and stop trying to connect to other peers.
    ///
    /// **Important**: this function returns when the signal to disconnect was sent, before all peers actually disconnected. This means to ensure the network is disconnected, wait for all peers to disappear after calling. 
```

Eligioo (approved):

onmax:
> I'd check that no peers are still there after disconnect_peers. This is an async instruction sent to the network that runs in it's own task. So, something like this:

For future reference and Vue apps, I have created this code:

```ts
  async function disconnect(timeoutMs = 10000) {
    await client.value?.disconnectNetwork()

    const startTime = Date.now()

    while (!(Date.now() - startTime > timeoutMs || peers.value.length === 0)) {
      await new Promise(resolve => setTimeout(resolve, 1000))
    }

    peers.value = []
  }
```

PR #2976, opened 2024-10-15
Remove inactive validators from activation process

### LINKED ISSUE #2960: Migration: Remove validators who are not ready in the last X activation windows
Currently, the activation period is not bounded, meaning it can take an indefinite amount of activation windows to reach 80% readiness. 

However, we would like this process to end. 

For this reason, we would like to consider validators who were ready during X amount of activation windows and stop considering those who were not ready during that period, meaning, those who were not ready during X amount of activation windows are going to be set as inactive in the first epoch and are no longer considered as part of the process to reach 80% readiness.

Also if a validator signaled readiness during one activation window and is no longer ready in subsequent activation windows, that validator is also removed from the readiness process and is set as inactive.

Note that each activation window lasts for ~24 hours. 
So we consider a validator as ready if during anytime during that window it sent the corresponding ready transaction.

viquezclaudio:
- Starting at ACTIVATION_WINDOW_TRESHOLD, those validators who are not ready, are going to be inactivated in the first epoch and they are no longer considered part of the readiness voting process
    
 - Fixes a bug that always used the first candidate block when creating the genesis block irrespective of the current activation window

...
## Pull request checklist

### DIFF
--- a/pow-migration/src/monitor.rs
+++ b/pow-migration/src/monitor.rs
@@ -151,11 +151,33 @@ pub async fn send_tx(client: &Client, transaction: OutgoingTransaction) -> Resul
     }
 }
 
+/// Returns true if the given validator was ready at the given activation block window
+pub async fn was_validator_ready(
+    pow_client: &Client,
+    validator_address: String,
+    activation_block_window: Range<u32>,
+    genesis_hash: String,
+) -> bool {
+    if let Ok(transactions) =
+        async_retryer(|| pow_client.get_transactions_by_address(&validator_address, 10)).await
+    {
+        // We only keep the ones past the activation window that met the activation criteria
+        let filtered_txns: Vec<TransactionDetails> = transactions
+            .into_iter()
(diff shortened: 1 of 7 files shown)


Eligioo (approved):

pow-migration/src/monitor.rs
```diff
@@ -151,11 +151,33 @@ pub async fn send_tx(client: &Client, transaction: OutgoingTransaction) -> Resul
...
 
+/// Returns true if the given validator was ready at the given activation block window
+pub async fn was_validator_ready(
+    pow_client: &Client,
+    validator_address: String,
+    activation_block_window: Range<u32>,
+    genesis_hash: String,
+) -> bool {
+    if let Ok(transactions) =
+        async_retryer(|| pow_client.get_transactions_by_address(&validator_address, 10)).await
```
jsdanielh: I don't think that 10 would be enough

jsdanielh (approved):

PR #2978, opened 2024-10-16
Separate pre-genesis data into different database

paberr:
This PR separates pre-genesis and PoS transaction history into different databases.
This increases portability.

- The migration client uses `HistoryStore`/`HistoryStoreIndex` as if there was only one database.
- The PoS client uses a `HistoryStoreMerger` to merge both databases.
    - The merger uses the pre-genesis data read-only.
    - Without the pre-genesis data present, the methods have the same behaviour as previously.
    - Methods that write into the database will panic if we write epoch 0 (pre-genesis data).

## Pull request checklist

### DIFF
--- a/blockchain/src/blockchain/blockchain.rs
+++ b/blockchain/src/blockchain/blockchain.rs
@@ -18,9 +18,8 @@ use tokio::sync::broadcast;
 #[cfg(feature = "metrics")]
 use crate::chain_metrics::BlockchainMetrics;
 use crate::{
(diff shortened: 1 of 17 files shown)


paberr:
I have tested some basic things, but I didn't have the time to test the full migration + RPC requests in different scenarios yet.

styppo (approved):
LGTM

paberr:
I also added tests now that check consistency with a single store (there were two odd cases where we don't actually would return `None`).

Eligioo (approved):

PR #2982, opened 2024-10-16
BlockQueue: accept block announcements from any connected peer

styppo:


ii-cruz (approved):
LGTM

nibhar (changes requested):
As this supersedes #2889 I think as stated in that PR the `macro_head_hash` should remain in the response.

styppo:
Added macro and election head hash to the response in c064cc1.

nibhar (approved):
Please create an Issue to track utilizing the macro_heads.

PR #2983, opened 2024-10-17
Fix `DoubleProposalProof`s

### LINKED ISSUE #2981: Double proposal proof signature verification generates the wrong data to be signed
This means that it's essentially impossible to generate double proposal proofs currently.

Currently, the double proposal proofs simply sign the Blake2**b** hash of the macro header, but the correct data can be seen in the following:

https://github.com/nimiq/core-rs-albatross/blob/76cc30b050c1273e1bef0e0915c8d99995c9000b/validator/src/tendermint.rs#L364-L378

hrxi:
They required data unrelated to the signed proposals. It now uses the same signing infrastructure by factoring it out to `TendermintProposal` in a common crate.

Fixes #2981.

### DIFF
--- a/primitives/block/src/equivocation_proof.rs
+++ b/primitives/block/src/equivocation_proof.rs
@@ -372,7 +372,7 @@ impl DoubleProposalProof {
     }
     /// Round of the proposals.
     pub fn round(&self) -> u32 {
-        self.proposal1.proposal.round
+        self.proposal1.round
     }
     /// Hash of header number 1.
     pub fn header1_hash(&self) -> Blake2bHash {
@@ -404,7 +404,6 @@ impl DoubleProposalProof {
         }
 
         if self.proposal1.proposal.block_number != self.proposal2.proposal.block_number
-            || self.proposal1.proposal.round != self.proposal2.proposal.round
             || self.proposal1.round != self.proposal2.round
         {
             return Err(EquivocationProofError::SlotMismatch);
@@ -810,8 +809,8 @@ mod test {
         let proposals: Vec<_> = headers
             .into_iter()
             .map(|header| TendermintProposal {
+                round: header.round,
                 proposal: header,
-                round: 0,
                 valid_round: None,
             })
             .collect();
--- a/primitives/block/src/equivocation_proof.rs
+++ b/primitives/block/src/equivocation_proof.rs
@@ -314,13 +314,9 @@ pub struct DoubleProposalProof {
     /// Address of the offending validator.
     validator_address: Address,
     /// Header number 1.
-    header1: MacroHeader,
+    proposal1: TendermintProposal<MacroHeader>,
     /// Header number 2.
-    header2: MacroHeader,
-    round1: u32,
-    round2: u32,
-    valid_round1: Option<u32>,
-    valid_round2: Option<u32>,
+    proposal2: TendermintProposal<MacroHeader>,
     /// Justification for header number 1.
     justification1: SchnorrSignature,
     /// Justification for header number 2.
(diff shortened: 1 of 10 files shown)


primitives/block/src/equivocation_proof.rs
```diff
@@ -405,13 +419,27 @@ impl DoubleProposalProof {
 
         if self.header1.block_number != self.header2.block_number
             || self.header1.round != self.header2.round
```
hrxi: Is this check still correct or should it go away?

primitives/block/src/equivocation_proof.rs
```diff
@@ -327,20 +331,30 @@ impl DoubleProposalProof {
     pub fn new(
         validator_address: Address,
         mut header1: MacroHeader,
+        mut round1: u32,
+        mut valid_round1: Option<u32>,
         mut justification1: SchnorrSignature,
         mut header2: MacroHeader,
+        mut round2: u32,
+        mut valid_round2: Option<u32>,
```
hrxi: Should I merge the two `round` parameters into one, given that they have to be the same?

primitives/block/src/equivocation_proof.rs
```diff
@@ -317,6 +317,10 @@ pub struct DoubleProposalProof {
     header1: MacroHeader,
     /// Header number 2.
     header2: MacroHeader,
+    round1: u32,
+    round2: u32,
+    valid_round1: Option<u32>,
+    valid_round2: Option<u32>,
```
nibhar: I would prefer
```suggestion
    proposal1: TendermintProposal,
    /// Proposal number 2.
    proposal2: TendermintProposal,
```

primitives/block/src/equivocation_proof.rs
nibhar: I would prefer these be `TendermintProposal`, making the signature much smaller. If that turns out to be annoying or to require unnecessary cloning, then I think one round could be removed yes.

primitives/block/src/equivocation_proof.rs
nibhar: I would say this was not correct before and it should be removed. The round in the header only denotes when a proposal was created, not when it was proposed. Thus it should not be checked here.

nibhar (changes requested):
Looks good to me outside of the `MacroHeader.round` check being incorrect (existed prior to this PR). So that should be removed. Other comments are purely cosmetic .

primitives/block/src/equivocation_proof.rs
```diff
@@ -326,21 +326,21 @@ pub struct DoubleProposalProof {
...
+        mut proposal1: TendermintProposal<MacroHeader>,
         mut justification1: SchnorrSignature,
-        mut header2: MacroHeader,
+        mut proposal2: TendermintProposal<MacroHeader>,
         mut justification2: SchnorrSignature,
     ) -> DoubleProposalProof {
-        let hash1: Blake2bHash = header1.hash();
-        let hash2: Blake2bHash = header2.hash();
+        let hash1: Blake2bHash = proposal1.proposal.hash();
+        let hash2: Blake2bHash = proposal2.proposal.hash();
```
nibhar: Maybe we should add an assert for the equal rounds here. Not strictly necessary but `DoubleProposalProof`s without matching rounds are invalid and should never be created.
```suggestion
        assert_eq!(proposal1.round, proposal2.round);
        let hash1: Blake2bHash = proposal1.proposal.hash();
        let hash2: Blake2bHash = proposal2.proposal.hash();
```

primitives/block/src/equivocation_proof.rs
```diff
@@ -360,36 +360,36 @@ impl DoubleProposalProof {
...
 
     /// Verify the validity of a double proposal proof.
     pub fn verify_excluding_address_and_network(
         &self,
         signing_key: &SchnorrPublicKey,
     ) -> Result<(), EquivocationProofError> {
-        let hash1: Blake2bHash = self.header1.hash();
-        let hash2: Blake2bHash = self.header2.hash();
+        let hash1: Blake2bHash = self.proposal1.proposal.hash();
+        let hash2: Blake2bHash = self.proposal2.proposal.hash();
```
nibhar: As far as I can see these are now only used for sorting. Sorting could also just use the Blake2sHash instead saving one hashing.

nibhar (approved):
LGTM. Left minor comments only.

primitives/block/src/equivocation_proof.rs
hrxi: Done.

primitives/block/src/equivocation_proof.rs
hrxi: Not done for now because we currently don't do that for the creation of any equivocation proof structure. We could automatically call `verify` I guess and `assert!` that it verifies. But maybe we could also do that at the call sites?

hrxi:
Rebased for merging.

PR #2991, opened 2024-10-21
network-libp2p: Fix double increments and missing decrements for peer/IP counters in Connection Pool Behaviour

Eligioo:
- Incrementing the counters happened in both `handle_pending_inbound_connection` and `on_connection_established`.
- Secondly due to early returns, there were scenarios where some counters wouldn't get decremented.
- Lastly decrement the limit counters on a listen failure.

## Pull request checklist

### DIFF
--- a/network-libp2p/src/connection_pool/behaviour.rs
+++ b/network-libp2p/src/connection_pool/behaviour.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{BTreeMap, BTreeSet, HashMap, HashSet, VecDeque},
+    collections::{hash_map::Entry, BTreeMap, BTreeSet, HashMap, HashSet, VecDeque},
     net::IpAddr,
     pin::Pin,
     sync::Arc,
@@ -16,7 +16,7 @@ use libp2p::{
         behaviour::{ConnectionClosed, ConnectionEstablished, DialFailure},
         dial_opts::{DialOpts, PeerCondition},
         dummy, CloseConnection, ConnectionDenied, ConnectionId, DialError, FromSwarm,
-        NetworkBehaviour, THandler, THandlerInEvent, THandlerOutEvent, ToSwarm,
+        ListenFailure, NetworkBehaviour, THandler, THandlerInEvent, THandlerOutEvent, ToSwarm,
     },
     Multiaddr, PeerId, TransportError,
 };
@@ -672,21 +672,6 @@ impl Behaviour {
             return;
         }
 
-        // Get IP from multiaddress if it exists.
-        let ip_info = self.get_ip_info_from_multiaddr(address);
-        if let Some(ip_info) = ip_info {
-            // Increment peer counts per IP
-            if let Some(subnet_ip) = ip_info.subnet_ip {
-                let value = self.limits.ip_subnet_count.entry(subnet_ip).or_insert(0);
-                *value = value.saturating_add(1);
-            }
-
-            let value = self.limits.ip_count.entry(ip_info.ip).or_insert(0);
-            *value = value.saturating_add(1);
-
-            self.limits.peer_count = self.limits.peer_count.saturating_add(1);
-        }
-
         // Peer is connected, mark it as such.
         let peer_services = self
             .contacts
@@ -708,11 +693,6 @@ impl Behaviour {
         endpoint: &ConnectedPoint,
         remaining_established: usize,
     ) {
-        // Check there are no more remaining connections to this peer
-        if remaining_established > 0 {
-            return;
-        }
-
         let address = endpoint.get_remote_address();
 
         // Get IP from multiaddress if it exists.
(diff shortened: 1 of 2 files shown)


network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -821,6 +793,89 @@ impl Behaviour {
...
+        let ip_info = self.get_ip_info_from_multiaddr(send_back_addr);
+
+        // Decrement IP counters if needed
+        if let Some(ip_info) = ip_info {
+            self.decrement_ip_counters(&ip_info);
+        }
+        self.limits.peer_count = self.limits.peer_count.saturating_sub(1);
+    }
+
+    fn reached_peer_limit(&mut self) -> Result<(), ConnectionDenied> {
```
nibhar: I'd rather have it named sth like this. For as it is currently named I would expect `OK(())` to denote that the limit was reached. On first glance I though this was handling the case where the limit was reached already.
```suggestion
    fn increase_and_check_counter(&mut self) -> Result<(), ConnectionDenied> {
```

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -821,6 +793,89 @@ impl Behaviour {
...
+            return Err(ConnectionDenied::new(Error::MaxPeerConnectionsReached));
+        }
+
+        Ok(())
+    }
+
+    fn reached_ip_limit(&mut self, ip_info: &IpInfo) -> Result<(), ConnectionDenied> {
+        // Increment peer IP counter
+        let ip_count = self.limits.ip_count.entry(ip_info.ip).or_insert(0);
+        *ip_count = ip_count.saturating_add(1);
```
nibhar:
```suggestion
        let ip_count = self.limits.ip_count.entry(ip_info.ip).and_modify(|counter| *counter = *counter.saturating_add(1)).or_insert(1);
```

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -821,6 +793,89 @@ impl Behaviour {
...
+
+        // Increment peer subnet IP counter
+        if let Some(subnet_ip) = ip_info.subnet_ip {
+            let subnet_count = self.limits.ip_subnet_count.entry(subnet_ip).or_insert(0);
+            *subnet_count = subnet_count.saturating_add(1);
```
nibhar:
```suggestion
            let subnet_count = self.limits.ip_subnet_count.entry(subnet_ip).amd_modify(|counter| *counter = *counter.saturating_add(1)).or_insert(1);
```

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -821,6 +793,89 @@ impl Behaviour {
...
+        }
+
+        // Check if peer subnet IP limit is reached
+        if let Some(subnet_ip) = ip_info.subnet_ip {
+            if *self
+                .limits
+                .ip_subnet_count
+                .get(&subnet_ip)
+                .expect("must have subnet_ip entry")
+                > self.config.peer_count_per_subnet_max
```
nibhar: The lookup here has already been performed earlier. Your could retain the result from earlier to check here. If order does not matter, meaning a connection may be refused because of MaxSubnet before MaxPeerPerIp then I would just swap this check with the previous one and then this entire thing can go into the condition of line 829.

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -821,6 +793,89 @@ impl Behaviour {
...
+
+        Ok(())
+    }
+
+    fn decrement_ip_counters(&mut self, ip_info: &IpInfo) {
+        // Decrement peer IP counter
+        if let Entry::Occupied(mut ip_count) = self.limits.ip_count.entry(ip_info.ip) {
+            *ip_count.get_mut() = ip_count.get().saturating_sub(1);
+            if *ip_count.get() == 0 {
+                self.limits.ip_count.remove(&ip_info.ip);
```
nibhar:
```suggestion
                ip_count.remove();
```

network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -821,6 +793,89 @@ impl Behaviour {
...
+            }
+        }
+
+        // Decrement peer subnet IP counter
+        if let Some(subnet_ip) = ip_info.subnet_ip {
+            if let Entry::Occupied(mut subnet_count) = self.limits.ip_subnet_count.entry(subnet_ip)
+            {
+                *subnet_count.get_mut() = subnet_count.get().saturating_sub(1);
+                if *subnet_count.get() == 0 {
+                    self.limits.ip_subnet_count.remove(&subnet_ip);
```
nibhar:
```suggestion
                    subnet_count.remove();
```

nibhar (approved):
Nice find. Good to go from my end. Comments are mostly cosmetic.

network-libp2p/src/connection_pool/behaviour.rs
Eligioo: The order does matter as I want to bump both counters before I check the limits. But I introduced a temp var to get rid of the second lookup 👍 

```rust
// Increment peer subnet IP counter
let subnet_ip_count = if let Some(subnet_ip) = ip_info.subnet_ip {
    Some(...)
} else {
    None
};
```

nibhar: Yes I realized that both counters must be increased. Your solution works perfectly.

There is an alternative if the order of increasing counters and the order of returning maybe need not be the same. Like so:
```rust 
let ip_count = ...;

if let Some(subnet_ip) = ip_info.subnet_ip {
    let subnet_ip_count = ...;
    if *subnet_ip_count > self.config.peer_count_per_subnet_max {
        ...
        return Err(...);
    }
}

if *ip_count > self.config.peer_count_per_ip_max {
    ...
    return Err(...);
}
```
Then you dont need to keep it around. But that means it will return with the subnet_ip error before returning with the ip_count error.

nibhar: That being said, I have no preference in this, so feel free to keep it as you have it already.

jsdanielh (approved):

PR #2993, opened 2024-10-21
Hex encode the data field of the online transaction

viquezclaudio:
Hex encode the data field of the online transaction.

## Pull request checklist

### DIFF
--- a/pow-migration/src/monitor.rs
+++ b/pow-migration/src/monitor.rs
@@ -61,7 +61,7 @@ pub fn generate_ready_tx(validator: String, hash: &Blake2bHash) -> OutgoingTrans
 /// - Sender: Validator address
 /// - Recipient: Burn address
 /// - Value: 1 Luna
-/// - Data: "online"
+/// - Data: "online" as hex representation
 pub fn generate_online_tx(validator: String) -> OutgoingTransaction {
     log::debug!(
         validator_address = validator,
@@ -72,7 +72,7 @@ pub fn generate_online_tx(validator: String) -> OutgoingTransaction {
         to: Address::burn_address().to_user_friendly_address(),
         value: 1, //Lunas
         fee: 0,
-        data: Some("online".to_string()),
+        data: Some(hex::encode("online")),
     }
 }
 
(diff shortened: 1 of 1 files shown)


pow-migration/src/monitor.rs
```diff
@@ -72,7 +72,7 @@ pub fn generate_online_tx(validator: String) -> OutgoingTransaction {
         to: Address::burn_address().to_user_friendly_address(),
         value: 1, //Lunas
         fee: 0,
-        data: Some("online".to_string()),
+        data: Some(Blake2bHash::default().to_hex()),
```
sisou: Default hash is 64 0s long. A lot of unnecessary data. You can get the utf8 bytes of a string with `"online".bytes()`, which you can then put as the data.

viquezclaudio: I thought about keep using "online" but then realized that later in the process what we are doing is sending a hash (genesis hash), so I thought it was ok to send the default hash as an indication of "we have not computed a hash yet" 
But since we are not doing anything useful with this field, I guess "online" or anything is fine

hrxi:
What's the motivation for changing it from `"online"`?

pow-migration/src/monitor.rs
sisou: The 0-hash is _actually_ still a valid genesis block hash :D

pow-migration/src/monitor.rs
```diff
@@ -132,7 +132,7 @@ fn is_valid_ready_txn(
 /// Checks if the provided transaction meets the criteria in order to be
 /// considered a valid online-transaction
 fn is_valid_online_txn(txn: &TransactionDetails, block_window: &Range<u32>) -> bool {
-    Some("online".to_string()) == txn.data
+    Some(hex::encode(String::from("online"))) == txn.data
```
sisou: Um guessing this is optimized by the compiler for being static?

hrxi: It's not, but it also doesn't really matter.

```suggestion
    Some(hex::encode("online")) == txn.data
```

```suggestion
    txn.data.as_ref().is_some_and(|hex| hex == "6f6e6c696e65") // "online"
```

jsdanielh (approved):

PR #2997, opened 2024-10-23
Blockchain: add `network_id` to `SkipBlockInfo`

styppo:
Add `network_id` to `SkipBlockInfo` to prevent usage of the skip block signature across networks.

### DIFF
--- a/blockchain/src/block_production/mod.rs
+++ b/blockchain/src/block_production/mod.rs
@@ -135,6 +135,7 @@ impl BlockProducer {
 
         let skip_block_info = if skip_block_proof.is_some() {
             Some(SkipBlockInfo {
+                network_id: blockchain.network_id,
                 block_number,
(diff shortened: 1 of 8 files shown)


hrxi (commented):
The `network_id` probably needs to be checked somewhere, too.

styppo:
The `network_id` is implicitly verified [here](https://github.com/nimiq/core-rs-albatross/commit/669dad9bdba13921c92ddf6abc0eb63d8cef6baa#diff-411d19cd7af1cc906798c571bc8bc7a5c296c22267f9e8763fe2375785b42211R104) as the signed message (`SkipBlockInfo`) is constructed from the `network_id` given in the block header.

jsdanielh (approved):
LGTM

Eligioo (approved):

PR #3005, opened 2024-10-28
Make PeerContact timestamp non optional

nibhar:
Seed nodes never enter the PeerAddressBook as seed nodes specifically, only via discovery. Previously seed node information was encoded as a `None` timestamp. As that will never be Set to None it is removed.

This is a prerequisite to include validator information in the discovered PeerContacts as these also include the timestamp in their authenticated structure. Thus it cannot be `None` for them which this PR guarantees.


lib/src/client.rs
```diff
@@ -260,14 +261,16 @@ impl ClientInner {
                 _ => true,
             }
         });
-        let mut peer_contact = PeerContact::new(
+        let peer_contact = PeerContact::new(
             peer_contact_addresses,
             identity_keypair.public(),
             provided_services,
-            None,
+            SystemTime::now()
```
Eligioo, line 268: Question now is if we even want to keep the `timestamp` as one of the `PeerContact::new` parameters? `timestamp` can now get constructed inside the `new` method.

Eligioo: Hmm, doesn't really hold. It needs to be flexible for the `test_housekeeping()` test.

Eligioo (approved):

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -73,16 +74,15 @@ impl PeerContact {
...
-    pub fn is_seed(&self) -> bool {
-        self.timestamp.is_none()
-    }
-
     /// Derives the peer ID from the public key
     pub fn peer_id(&self) -> PeerId {
         self.public_key.clone().to_peer_id()
     }
 
+    pub fn timestamp(&self) -> u64 {
```
jsdanielh, line 83:
```suggestion
    /// Returns the contact timestamp
    pub fn timestamp(&self) -> u64 {
```

jsdanielh (approved):

PR #3009, opened 2024-10-30
database: Allow full nodes to index history items

Eligioo:
Optionally allow full nodes to also index the history. The default for `index_history` per node type remains untouched: `true` for history nodes, `false` for the other node types. By being able to index the history allows full nodes to use RPC methods where the index is required: `get_transaction_by_hash`, `get_transaction_hashes_by_address` and `get_transactions_by_address`.

Setting `index_history` to `true` for light nodes won't change the behaviour of the node as this property is ignored.

## Pull request checklist


jsdanielh (approved):
LGTM

PR #3011, opened 2024-10-30
DB: Liforeclaim -> true

viquezclaudio:
Setting liforeclaim to true provides better performance when prunning epochs with lots of data

...
## Pull request checklist

### DIFF
--- a/database/src/mdbx/database.rs
+++ b/database/src/mdbx/database.rs
@@ -59,6 +59,7 @@ impl From<DatabaseConfig> for libmdbx::DatabaseOptions {
                 max_size: value.size.map(|r| r.end),
                 ..Default::default()
             }),
+            liforeclaim: true,
             ..Default::default()
         }
     }


jsdanielh (approved):

Eligioo:
@viquezclaudio does this flag have drawbacks? Otherwise it would have a default of true?

sisou:
Documentation of that flag I found is here: https://gitflic.ru/project/erthink/libmdbx/blob/?file=mdbx.h&branch=master#line-num-1324

Further down it also says this flag can cause database corruption in case of system crash, e.g. when the filesystem cannot finish a write.

Eligioo:
Database corruptions would be very bad, especially for history nodes as this is also the category of nodes we will see least often. They are the backbone of the chain so to say where full and light can do a simple re-sync.

We maybe instead should bump the hardware requirements for history nodes (if even necessary) to compensate for the extra work when we had a full epoch. And have this flag only enabled for the block producers (full nodes) so that these nodes are less longer busy with pruning and not get slowed down by this process (not affect block production).

viquezclaudio:
The database corruption is mentioned in the context of using the MDBX_UTTERLY_NOSYNC sync mode, which we are not using

Eligioo:
Thanks for the clarification

Eligioo (approved):

sisou:
Ah, I didn't notice the talking about corruption was within the docblock for `MDBX_UTTERLY_NOSYNC`. Then seems all good!

PR #3012, opened 2024-10-30
Network swarm: Split functions and reduce indentation.

nibhar:
This PR does not change any functionality. It only restructures the `handle_event` function and some other minor locations in that file. I was doing these on the fly as I went and the changes would have been very hard to track, so I decided to bundle them up and make a separate PR for them to ease reviewing.

### DIFF
--- a/network-libp2p/src/swarm.rs
+++ b/network-libp2p/src/swarm.rs
@@ -6,7 +6,7 @@ use instant::Instant;
 #[cfg(all(target_family = "wasm", not(feature = "tokio-websocket")))]
 use libp2p::websocket_websys;
 use libp2p::{
-    autonat::OutboundFailure,
+    autonat::{self, InboundFailure, OutboundFailure},
     core::{
         self,
         muxing::StreamMuxerBox,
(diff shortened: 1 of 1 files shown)


hrxi:
[Diff without whitespace](https://github.com/nimiq/core-rs-albatross/pull/3012/files?w=1).

network-libp2p/src/swarm.rs
```diff
@@ -408,517 +411,579 @@ fn handle_event(
...
+                    .unwrap()
+                    .finish();
+            }
+        }
+        Ok(GetRecordOk::FinishedWithNoAdditionalRecord { cache_candidates }) => {
+            // Remove the query, send the best result to the application layer
+            // and push the best result to the cache candidates
+
+            let Some(results) = event_info.state.dht_get_results.remove(&id) else {
+                panic!("DHT inconsistent state, query_id: {:?}", id);
```
Eligioo, line 567: Do we need to panic here?

Eligioo (commented):
Nice effort! Added one question.

Decoupling these events from its handlers also already prepares how we can eventually split our current `Behaviour` and introduce a 'light' `Behaviour` 👍

network-libp2p/src/swarm.rs
nibhar: The panic was there before. I was considering to remove it, but choose not to do it in this pr, even if it should change, as it is supposed to not alter how the code works.

Eligioo (approved):

jsdanielh (approved):

PR #3014, opened 2024-11-01
Validator address as key for the Dht

nibhar:
Previous to this PR The Dht used the voting public key as the key in the dht. The record was also signed by this key. In this PR that changes to the record using the validator address as key and signing it using the signing key. Obviously the verifying key now needs to be obtained elsewhere as it is no longer part of the record itself.

The validator address is unambiguous. There exists only one validator with that address. The public key (signing or voting) could, in theory, be used by multiple validators.

When rotating keys this also provided a challenge, since the public key in the current validators set will not change as it is obtained from the previous election block. So the lookup key (old public key) must still be retained until the epoch ends to renew these entries (which was also not happening).

Lastly the change of the signature makes sense as there is no benefit in using the voting key over the signing key. Also semantically it makes far more sense to use the signing key here.

## Remaining todos:
- [ ] Light client support: Light clients do not have access to a staking contract. A different means of acquiring the verification key is required.
- [ ] Triggering an immediate Dht put with the updated signature in case the signing key is changed. The old entry will no longer verify as the current staking contracts keys are used.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ members = [
   "database",
   "database/database-value",
   "database/database-value-derive",
+  "dht",
   "fuzz",
   "genesis",
   "genesis-builder",
@@ -189,6 +190,7 @@ nimiq-consensus = { path = "consensus", default-features = false }
 nimiq-database = { path = "database", default-features = false }
 nimiq-database-value = { path = "database/database-value", default-features = false }
 nimiq-database-value-derive = { path = "database/database-value-derive", default-features = false }
+nimiq-dht = { path = "dht", default-features = false }
 nimiq-genesis = { path = "genesis", default-features = false }
 nimiq-genesis-builder = { path = "genesis-builder", default-features = false }
 nimiq-handel = { path = "handel", default-features = false }
--- a/lib/src/client.rs
+++ b/lib/src/client.rs
@@ -13,6 +13,7 @@ use nimiq_consensus::{
     sync::syncer_proxy::SyncerProxy, Consensus as AbstractConsensus,
     ConsensusProxy as AbstractConsensusProxy,
 };
+use nimiq_dht::Verifier;
 #[cfg(feature = "zkp-prover")]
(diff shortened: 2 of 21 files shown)


sisou:
Signing key signatures (Ed25519) are also smaller than BLS signatures, making this more efficient.

Cargo.toml
```diff
@@ -12,7 +12,8 @@ members = [
   "consensus",
   "database",
   "database/database-value",
-  "database/database-value-derive",
+  "database/database-value-derive",
```
hrxi:
```suggestion
  "database/database-value-derive",
```

hrxi (approved):
Looks good to me.

lib/src/client.rs
```diff
@@ -413,8 +410,41 @@ impl ClientInner {
...
+        // Create the Dht verifier
+        #[cfg(feature = "full-consensus")]
+        let dht_verifier = Verifier::new(blockchain_proxy.clone());
+
+        // Create the network.
+        let network = Arc::new(
+            Network::new(
+                network_config,
+                #[cfg(feature = "full-consensus")]
+                dht_verifier,
```
hrxi, line 429: Even if we have the `full-consensus` feature enabled at compile time, we might still start a light client.

network-libp2p/src/network.rs
```diff
@@ -74,8 +75,12 @@ impl Network {
     /// # Arguments
     ///
     ///  - `config`: The network configuration, containing key pair, and other behavior-specific configuration.
+    ///  - `dht_verifier`: The verifier used to verify all Dht records.
     ///
-    pub async fn new(config: Config) -> Self {
+    pub async fn new(
+        config: Config,
+        #[cfg(feature = "kad")] dht_verifier: impl dht::Verifier + 'static,
```
hrxi, line 82: This is problematic, the cargo feature `kad` is no longer purely additive. It should probably always take an `Option<impl …>`.

PR #3018, opened 2024-11-04
Consensus: add limits to various request parameters

### LINKED ISSUE #3010: Limit the number of blocks to prove via `RequestBlocksProof`

styppo:
This is labeled `testnet-restart` because it changes the request error enums that are sent over the network.

Much easier to review [without whitespace](https://github.com/nimiq/core-rs-albatross/pull/3018/files?w=1).

Closes #3010

### DIFF
--- a/consensus/src/messages/handlers.rs
+++ b/consensus/src/messages/handlers.rs
@@ -28,7 +28,7 @@ use crate::sync::live::{
 };
 
 impl RequestMacroChain {
-    const MAX_LOCATORS: usize = 255;
+    const MAX_LOCATORS: usize = 100;
     const MAX_EPOCHS: u32 = 255;
 }
 impl<N: Network> Handle<N, BlockchainProxy> for RequestMacroChain {
@@ -213,7 +213,7 @@ impl<N: Network> Handle<N, BlockchainProxy> for RequestMissingBlocks {
     }
 }
 impl RequestMissingBlocks {
-    const MAX_LOCATORS: usize = 255;
+    const MAX_LOCATORS: usize = 100;
 
     fn handle_backward<N: Network>(
         &self,
--- a/consensus/src/messages/handlers.rs
+++ b/consensus/src/messages/handlers.rs
@@ -27,16 +27,24 @@ use crate::sync::live::{
     state_queue::{Chunk, RequestChunk, ResponseChunk},
 };
 
+impl RequestMacroChain {
+    const MAX_LOCATORS: usize = 255;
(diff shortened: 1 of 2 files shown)


consensus/src/messages/handlers.rs
```diff
@@ -193,13 +201,20 @@ impl<N: Network> Handle<N, BlockchainProxy> for RequestMissingBlocks {
...
+        }
+
         match self.direction {
             Direction::Forward => self.handle_forward::<N>(peer_id, blockchain),
             Direction::Backward => self.handle_backward::<N>(peer_id, blockchain),
         }
     }
 }
 impl RequestMissingBlocks {
+    const MAX_LOCATORS: usize = 255;
```
ii-cruz: ATM we send maximum blocks per batch as locators (60), so this could be smaller.

consensus/src/messages/handlers.rs
```diff
@@ -593,20 +616,28 @@ impl<N: Network> Handle<N, Arc<RwLock<Blockchain>>> for RequestTransactionsProof
...
+            Self::prove_txns_with_block_number(blockchain, &self.hashes, block_number)
         } else {
-            prove_transaction(blockchain, &self.hashes[0])
+            Self::prove_transaction(blockchain, &self.hashes[0])
         }
     }
 }
 
+impl RequestTransactionReceiptsByAddress {
+    const MAX_RECEIPTS: u16 = 500;
```
ii-cruz, line 639: Since this is the same as `MAX_TRANSACTIONS_BY_ADDRESS` defined in the web client, it would be better to use the same constant on both places

consensus/src/messages/handlers.rs
```diff
@@ -27,16 +27,24 @@ use crate::sync::live::{
     state_queue::{Chunk, RequestChunk, ResponseChunk},
 };
 
+impl RequestMacroChain {
+    const MAX_LOCATORS: usize = 255;
```
ii-cruz: For the locators, we are usually putting the election block and one more macro block max. I think this can be significantly smaller.

ii-cruz (approved):
I think some params can be smaller but this is also a start, we can progressively adjust then later with more testing

consensus/src/messages/handlers.rs
sisou: I would say that because the data sent as a response to tx-by-address and receipts-by-address is different (receipts-by-address is much less data per tx), different constants are fine. The work done by the node processing the request is probably the same, though.

PR #3023, opened 2024-11-06
pow-migration: Revert history when needed

jsdanielh:
Add procedure to the history migration tool to be able to compute history tree roots of blocks that has its history already migrated. For this, it detects the current history tree block height and reverts the history associated to blocks up until the required block.

## Pull request checklist

### DIFF
--- a/pow-migration/src/history.rs
+++ b/pow-migration/src/history.rs
@@ -79,6 +79,38 @@ fn from_pow_transaction(pow_transaction: &PoWTransaction) -> Result<Transaction,
     Ok(tx)
 }
 
+async fn remove_history(
+    env: &MdbxDatabase,
+    pow_client: &Client,
+    history_store: &HistoryStoreProxy,
+    history_store_height: u32,
+    wanted_history_store_height: u32,
+) {
+    let mut num_txns_to_revert = 0usize;
+    for block_height in wanted_history_store_height + 1..=history_store_height {
+        let block = async_retryer(|| pow_client.get_block_by_number(block_height, false))
+            .await
+            .unwrap();
+
+        match block.transactions {
+            PoWTransactionSequence::BlockHashes(transactions) => {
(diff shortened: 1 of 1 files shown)


pow-migration/src/history.rs
```diff
@@ -79,6 +79,38 @@ fn from_pow_transaction(pow_transaction: &PoWTransaction) -> Result<Transaction,
...
+    for block_height in wanted_history_store_height + 1..=history_store_height {
+        let block = async_retryer(|| pow_client.get_block_by_number(block_height, false))
+            .await
+            .unwrap();
+
+        match block.transactions {
+            PoWTransactionSequence::BlockHashes(transactions) => {
+                num_txns_to_revert += transactions.len()
+            }
+            PoWTransactionSequence::Transactions(_) => todo!(),
```
Eligioo, line 99: Handle the `PoWTransactionSequence::Transactions` case?

Eligioo: Handle the `PoWTransactionSequence::Transactions` case?

pow-migration/src/history.rs
```diff
@@ -79,6 +79,38 @@ fn from_pow_transaction(pow_transaction: &PoWTransaction) -> Result<Transaction,
...
+        }
+    }
+
+    log::info!(
+        wanted_history_store_height,
+        history_store_height,
+        num_txns_to_revert,
+        "Reverting transactions from history store since more history than the requested was already migrated"
```
Eligioo:
```suggestion
        "Reverting transactions from history store since more than the necessary history was already migrated"
```

pow-migration/src/history.rs
jsdanielh: This case shouldn't happen since we are passing `false` to the `full_transactions` parameter in `pow_client.get_block_by_number`. So added a panic instead.

Eligioo (approved):

PR #3008, opened 2024-10-29
Network: add gossipsub rate limiting

styppo:
Closes #2570

PR #3027, opened 2024-11-07
Decouple block queue pruning from block announcements

### LINKED ISSUE #2954: Full node cannot achieve consensus
I have set up a new full node on bare metal (12x5.1GHz, 32GB RAM and 1TB NVME), this node does not seem able to achieve consensus. 

You will see in the logs that I restarted it several times during the course of 9hrs, at one point even deleting the db. 

This is not the first time I experienced this issue. 

Log: https://www.transfernow.net/dl/20241002PGuWWsd9  (available for 7 days)

ii-cruz:
The live sync has an invariant that all blocks come with no gaps from the gossip sub.
Before #2982 we had a bug where peers (active validators) were being miscategorized as not synced. Thus we would ignore their blocks from the gossip sub, creating gaps during the syncing process. Every gap would generate a missing blocks request from the head of our node to the block announced in the gossip sub, thus creating too many overlapping requests across multiple batches, leading to the issue reported by community #2954.

This PR changes the following:
- Moves the processing of block announcements to the end of the poll function, prioritizing finishing the ongoing missing block request results.
- Detaches pruning from the block announcements, so we prune before adding more data to our structs and so we always prune regardless of an announced block arriving.

This PR no longer restricts the request missing blocks to the current batch. The following changes were dropped in favor of tighter rate limiting on the gossipsub #3008:
  - Only allow concurrent missing block requests within the current batch, thus diminishing the DoS surface significantly. 
  - Any block on a future batch gets buffered instead, but no request missing blocks is sent over the network.

## Pull request checklist

### DIFF
--- a/consensus/src/sync/live/block_queue/block_request_component.rs
+++ b/consensus/src/sync/live/block_queue/block_request_component.rs
@@ -318,10 +318,6 @@ impl<N: Network> BlockRequestComponent<N> {
         self.pending_requests.contains(target_block_hash)
     }
 
-    pub fn has_pending_requests(&self) -> bool {
-        !self.pending_requests.is_empty()
-    }
-
     pub fn add_peer(&self, peer_id: N::PeerId) {
         self.peers.write().add_peer(peer_id);
     }
--- a/consensus/src/sync/live/block_queue/block_request_component.rs
+++ b/consensus/src/sync/live/block_queue/block_request_component.rs
@@ -318,6 +318,10 @@ impl<N: Network> BlockRequestComponent<N> {
         self.pending_requests.contains(target_block_hash)
     }
 
+    pub fn has_pending_requests(&self) -> bool {
+        !self.pending_requests.is_empty()
+    }
+
     pub fn add_peer(&self, peer_id: N::PeerId) {
         self.peers.write().add_peer(peer_id);
     }
--- a/consensus/src/sync/live/block_queue/queue.rs
+++ b/consensus/src/sync/live/block_queue/queue.rs
@@ -176,15 +176,7 @@ impl<N: Network> BlockQueue<N> {
 
         let parent_known = blockchain.contains(block.parent_hash(), true);
         drop(blockchain);
-
-        // Check if a macro block boundary was passed.
-        // If so prune the block buffer as well as pending requests.
         let macro_height = Policy::last_macro_block(head_height);
-        if macro_height > self.current_macro_height {
-            self.current_macro_height = macro_height;
-            self.prune_pending_requests();
-            self.prune_buffer();
-        }
 
         if block_number < head_height.saturating_sub(self.config.tolerate_past_max) {
             block_source.ignore_block(&self.network);
@@ -224,9 +216,24 @@ impl<N: Network> BlockQueue<N> {
                 macro_height
             );
             block_source.ignore_block(&self.network);
-        } else {
-            // Block is inside the buffer window, put it in the buffer.
+        } else if !self.request_component.has_pending_requests()
+            || macro_height == Policy::last_macro_block(block_number)
+        {
+            // We only allow a new request missing blocks to start if the block is from the
+            // current batch or if there are no ongoing request.
             self.buffer_and_request_missing_blocks(block, block_source);
+        } else {
+            // If we are on not within the same batch or we already are requesting blocks,
+            // we just buffer it without requesting for blocks.
+            // Any potential gaps will be filled after we sync up to the batch.
+            if self.insert_block_into_buffer(block, block_source) {
+                log::trace!(block_number, "Buffering block");
+            } else {
+                log::trace!(
+                    block_number,
+                    "Not buffering block - already known or exceeded the per peer limit",
+                );
+            }
         }
 
         None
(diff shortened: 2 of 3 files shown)


consensus/src/messages/handlers.rs
```diff
@@ -206,7 +206,7 @@ impl RequestMissingBlocks {
         blockchain: &BlockchainProxy,
     ) -> Result<ResponseBlocks, ResponseBlocksError> {
         // TODO We might want to do a sanity check on the locator hashes and reject the request if
-        //  they they don't match up with the given target hash.
+        // they they don't match up with the given target hash.
```
styppo: This leading space indicates that the line belongs to the TODO above, making JetBrains IDEs highlight the whole comment as a TODO.

consensus/src/sync/live/block_queue/queue.rs
```diff
@@ -216,17 +207,32 @@ impl<N: Network> BlockQueue<N> {
                 );
                 return Some(QueuedBlock::TooFarAhead(peer_id));
             }
-        } else if block_number <= macro_height {
+        } else if block_number <= self.current_macro_height {
```
styppo: It's safer to use `macro_height` here (computed as before: `Policy::last_macro_block(head_height)`), as otherwise there is a race condition between `check_and_prune()` (which updates `self.current_macro_height`) and `check_announced_block()`, because the blockchain read lock is dropped between those calls.

consensus/src/sync/live/block_queue/queue.rs
```diff
@@ -216,17 +207,32 @@ impl<N: Network> BlockQueue<N> {
...
                 "Discarding block {} - we're already at macro block #{}",
                 block,
-                macro_height
+                self.current_macro_height
             );
             block_source.ignore_block(&self.network);
-        } else {
-            // Block is inside the buffer window, put it in the buffer.
+        } else if self.request_component.has_no_pending_requests()
+            || self.current_macro_height == Policy::last_macro_block(block.block_number())
```
styppo:
```suggestion
            || self.current_macro_height == Policy::last_macro_block(block_number)
```

consensus/src/sync/live/block_queue/queue.rs
```diff
@@ -216,17 +207,32 @@ impl<N: Network> BlockQueue<N> {
...
                 return Some(QueuedBlock::TooFarAhead(peer_id));
             }
-        } else if block_number <= macro_height {
+        } else if block_number <= self.current_macro_height {
             // Block is from a previous batch/epoch, discard it.
             log::warn!(
                 "Discarding block {} - we're already at macro block #{}",
                 block,
-                macro_height
+                self.current_macro_height
```
styppo: `macro_height` as above

consensus/src/sync/live/block_queue/queue.rs
```diff
@@ -216,17 +207,32 @@ impl<N: Network> BlockQueue<N> {
...
+            || self.current_macro_height == Policy::last_macro_block(block.block_number())
+        {
+            // We only allow a new request missing blocks to start if the block is from the
+            // current batch or if there are no ongoing request.
             self.buffer_and_request_missing_blocks(block, block_source);
+        } else {
+            // If we are on not within the same batch or we already are requesting blocks,
+            // we just buffer it without requesting for blocks.
+            // Any potential gaps will be filled after we sync up to the batch.
+            if self.insert_block_into_buffer(block, block_source.clone()) {
```
styppo:
```suggestion
            if self.insert_block_into_buffer(block, block_source) {
```
Looks like this `clone()` is unnecessary.

consensus/src/sync/live/block_queue/queue.rs
```diff
@@ -216,17 +207,32 @@ impl<N: Network> BlockQueue<N> {
...
             log::warn!(
                 "Discarding block {} - we're already at macro block #{}",
                 block,
-                macro_height
+                self.current_macro_height
             );
             block_source.ignore_block(&self.network);
-        } else {
-            // Block is inside the buffer window, put it in the buffer.
+        } else if self.request_component.has_no_pending_requests()
```
styppo:
```suggestion
        } else if !self.request_component.has_pending_requests()
```

consensus/src/sync/live/block_queue/block_request_component.rs
```diff
@@ -318,6 +318,10 @@ impl<N: Network> BlockRequestComponent<N> {
         self.pending_requests.contains(target_block_hash)
     }
 
+    pub fn has_no_pending_requests(&self) -> bool {
+        self.pending_requests.is_empty()
```
styppo, line 322:
```suggestion
        !self.pending_requests.is_empty()
```

styppo:
```suggestion
    pub fn has_pending_requests(&self) -> bool {
        !self.pending_requests.is_empty()
```
Avoid negative function names as they lead to double negation if we need to invert them.

paberr (approved):
lgtm

Eligioo (approved):

hrxi (approved):
Looks good to me.

PR #3038, opened 2024-11-12
Network: enable DHT server mode only while consensus is established

styppo:
With #3014, we can only properly verify DHT records once we have the staking contract. Therefore, this PR defers the switch to DHT server mode to when consensus is established. The DHT switches back to client mode when consensus is lost.

### DIFF
--- a/network-libp2p/src/swarm.rs
+++ b/network-libp2p/src/swarm.rs
@@ -31,7 +31,7 @@ use libp2p::{
 use libp2p::{dns, tcp, websocket};
 use log::Instrument;
 use nimiq_network_interface::{
-    network::{CloseReason, NetworkEvent},
+    network::{CloseReason, DhtMode, NetworkEvent},
     peer_info::PeerInfo,
     request::{peek_type, InboundRequestError, OutboundRequestError, RequestError},
 };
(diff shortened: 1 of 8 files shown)


network-mock/src/network.rs
```diff
@@ -534,6 +535,10 @@ impl Network for MockNetwork {
         // TODO implement
     }
 
+    async fn dht_set_mode(&self, _mode: DhtMode) {
+        // TODO implement
```
Eligioo, line 539: Set an `unimplemented!()` here? Same for `validate_message`. Otherwise the caller has no idea this still needs an actual impl

network-libp2p/src/swarm.rs
```diff
@@ -1041,6 +1052,27 @@ fn perform_action(action: NetworkAction, swarm: &mut NimiqSwarm, state: &mut Tas
             let result = swarm.dial(dial_opts).map_err(Into::into);
             output.send(result).ok();
         }
+        NetworkAction::DhtSetMode { mode } => {
+            #[cfg(feature = "kad")]
```
Eligioo, line 1056: Feature annotations support `{ }` to encapsulate a block of code:

```
#[cfg(feature = "kad")]
{
    ...
}
```

nibhar (changes requested):
Nothing to add to @Eligioo comments. Otherwise looks good to me. Requesting changes on his behalf.

PR #3040, opened 2024-11-12
Force rust toolchain for web-client NPM build

### LINKED ISSUE #3022: Failed to build npm (wasm) package in CI
CI fails to build the npm wasm package, retry also failed with same error:

https://github.com/nimiq/core-rs-albatross/actions/runs/11676122517/job/32564694138

sisou:
## Pull request checklist

- [ ] ~~All tests pass. The project builds and runs.~~

### DIFF
--- a/.github/workflows/npm_publish.yml
+++ b/.github/workflows/npm_publish.yml
@@ -14,19 +14,15 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
         with:
+          toolchain: 1.81.0
           targets: wasm32-unknown-unknown
       - uses: Swatinem/rust-cache@v2
-      - uses: jetli/wasm-bindgen-action@v0.2.0
-        with:
-          version: 'latest'
       - name: Build package
         working-directory: ./web-client
         run: ./scripts/build.sh
       - uses: actions/setup-node@v4
         with:
-          node-version: '18.x'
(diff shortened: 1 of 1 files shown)


sisou:
Manual CI run with this branch passed: https://github.com/nimiq/core-rs-albatross/actions/runs/11801453470

jsdanielh (approved):

.github/workflows/npm_publish.yml
```diff
@@ -14,19 +14,15 @@ jobs:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
```
Eligioo, line 15: When you want to use a different toolchain other than the current stable, you are supposed to get it by: `dtolnay/rust-toolchain@1.81.0`.

_The selection of Rust toolchain is made based on the particular @ rev of this Action being requested. For example "dtolnay/rust-toolchain@nightly" pulls in the nightly Rust toolchain, while "dtolnay/rust-toolchain@1.42.0" pulls in 1.42.0._

_Important: the default is to match the @ rev as described above. When passing an explicit toolchain as an input instead of @ rev, you'll want to use "dtolnay/rust-toolchain@master" as the revision of the action._

sisou: Welp, it still works. If they want to enforce that, they should prevent the other ways from working :D

PR #3042, opened 2024-11-12
Clarify `BlockBodyMessage::header_hash`

hrxi:
Refactor `BlockAssembler::poll_next` a bit: Explicitly name the body cache key members and destructure tuples with explicit names.


jsdanielh (approved):
LGTM

PR #3047, opened 2024-11-13
Upgrade to `ark-* 0.5.0`

hrxi:
This allows us to drop our local patches. :)


paberr (approved):
You removed `ToBeBytesGadget`, did we not use that?
Apart from that, LGTM

Regarding the current check failures:
- The failing web-client looks a bit like the loop unrolling problem if I'm not mistaken, but it should be fixed in 0.5 through a conditional compilation flag.
- The tests fail in the check whether we have enough signers. I already looked into it but couldn't find an obvious issue.

paberr:
Also, while reviewing the failing tests, I stumbled across the following code from Bruno (`check_signers` function):
```rust
        // num_signers >= min_signers
        num_signers.is_cmp(&min_signers, Ordering::Greater, true)
```
It looks to me that there's a mismatch between the comment and the code.
We should ensure consistency between the circuit code and our regular consensus condition though. If the consensus accepts blocks with `num_signers == min_signers` but the ZKP does not, we can run into an issue where we can not create ZKPs anymore.

paberr (commented):
Also, we need to check that the upgrade did not modify any constraints that invalidate our generated ZKP keys.

hrxi:
There's also this issue I created wrt. the test failure: https://github.com/arkworks-rs/r1cs-std/issues/151. We can probably just copy that function to our codebase without the panic.

> You removed `ToBeBytesGadget`, did we not use that?

This feature was added by 0.5.0 with the same name and same functionality, so I could just remove our `ToBeBytesGadget`.

> * The failing web-client looks a bit like the loop unrolling problem if I'm not mistaken, but it should be fixed in 0.5 through a conditional compilation flag.

Yup on both, but maybe another loop unrolling problem appeared since then. I'll look into it.

> We should ensure consistency between the circuit code and our regular consensus condition though. If the consensus accepts blocks with `num_signers == min_signers` but the ZKP does not, we can run into an issue where we can not create ZKPs anymore.

Does that mean we need to adjust the consensus code because we can't adjust the ZKP code anymore?

hrxi:
Created a PR for a bug fix: https://github.com/arkworks-rs/r1cs-std/pull/156.

hrxi:
Still has the "too many locals" problem, even when removing all the loop unrolling:
```
     Running unittests src/lib.rs (core-rs-albatross/target/wasm32-unknown-unknown/debug/deps/nimiq_web_client-4c40cda6b2d8e318.wasm)
Error: failed to deserialize Wasm module          

Caused by:
    0: failed to parse code section
    1: too many locals: locals exceed maximum (at offset 0x126495d)
```

hrxi:
The functions exceeding the locals limit of 50000 are:
```
95904 ark_ec::models::short_weierstrass::affine::Affine<P>::is_on_curve::h420bc5919b8bf3a1
96003 ark_ec::models::short_weierstrass::affine::Affine<P>::get_ys_from_x_unchecked::hcaaab6c84882627d
96131 <ark_ec::models::short_weierstrass::group::Projective<P> as ark_ff::fields::AdditiveGroup>::double_in_place::h022f34def0e6a2c2
```

See also https://github.com/rustwasm/wasm-pack/issues/981.

hrxi:
Observation: `wasm-pack test --chrome --headless` fails with the aforementioned error, `wasm-pack test --chrome --headless --release` passes.

paberr:
> It looks to me that there's a mismatch between the comment and the code. We should ensure consistency between the circuit code and our regular consensus condition though. If the consensus accepts blocks with `num_signers == min_signers` but the ZKP does not, we can run into an issue where we can not create ZKPs anymore.

There actually isn't a mismatch. The `true` parameter in the ZKP is for allowing equality.

One ToDo:
ℹ️ We should also test creating a ZKP using our existing keys with the new version and verifying it.

hrxi:
> One ToDo:
> ℹ️ We should also test creating a ZKP using our existing keys with the new version and verifying it.

How do I do this?

PR #3053, opened 2024-11-13
Adding a debug log when we discard a skip block message on a competing chain

ii-cruz:
Log when we discard a skip block message on a competing chain.

## Pull request checklist

### DIFF
--- a/validator/src/aggregation/skip_block.rs
+++ b/validator/src/aggregation/skip_block.rs
@@ -254,6 +254,7 @@ impl SkipBlockAggregation {
             .filter_map(move |(item, validator_id)| {
                 // Check that the update is for the current skip block aggregation.
                 if item.info != current_skip_block {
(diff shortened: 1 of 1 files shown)


validator/src/aggregation/skip_block.rs
```diff
@@ -254,6 +254,7 @@ impl SkipBlockAggregation {
             .filter_map(move |(item, validator_id)| {
                 // Check that the update is for the current skip block aggregation.
                 if item.info != current_skip_block {
+                    debug!(%current_skip_block, received_skip_block =% item.info,"Discarding skip block update as it comes from a different skip block aggregation.");
```
jsdanielh:
```suggestion
                    debug!(%current_skip_block, received_skip_block =% item.info,"Discarding skip block update as it comes from a different skip block aggregation");
```

jsdanielh (approved):

PR #3060, opened 2024-11-14
Web-Client: Fix docs for network, expose ConsensusState to wasm

blouflashdb:
- Updated docs to better reflect which networks are possible
- Added wasm_bindgen to ConsensusState to have better typings

## Pull request checklist

### DIFF
--- a/web-client/scripts/build-lib.sh
+++ b/web-client/scripts/build-lib.sh
@@ -5,16 +5,17 @@ yarn install
 
 # Build for bundler
 yarn tsup lib/index.ts --format esm --platform browser --out-dir ../dist/lib/bundler *.ts
-sed --in-place='' --expression='s/"@nimiq\/core"/"..\/..\/bundler\/index.js"/g' ../dist/lib/bundler/index.mjs
+sed --in-place='' --expression=s/"@nimiq\/core"/"..\/..\/bundler\/index.js"/g ../dist/lib/bundler/index.mjs
 
 # Build for web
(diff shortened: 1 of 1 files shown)


jsdanielh (approved):
LGTM

sisou (changes requested):
Unfortunately it's not so simple.

Adding `#[wasm_bindgen]` makes the compiler output an additional `ConsensusState` type in the type definitions, so now it's doubled (which is in itself a type error). To fix this, we would have to remove the `Tsify` and `serde` derives so the other definition is not emitted.

However, enums in Rust cannot have string values. They are always numbers. So `ConsensusState.Connecting` is `0`, `ConsensusState.Syncing` is `1`, etc:

```ts
export enum ConsensusState {
  Connecting = 0,
  Syncing = 1,
  Established = 2,
}
```

That changes the developer experience from being able to compare the consensus state as strings, to having to know which number corresponds to which state. Yes, people can then use the exported enum, but they also _have_ to. 

I would like the API to be user-friendly and emit strings for states, not numbers.

(On the other hand, if we continue to emit strings from the listener, the new `ConsensusState` enum is useless as it cannot be used to compare with the strings.)

blouflashdb:
My goal was to have a type in typescript like this: 
```
export type ConsensusState = 'connecting' | 'syncing' | 'established'
```
that i can then import from the @nimiq/core package

Seems like its not that simple as I hoped. 

EDIT after looking into the latest release of the @nimiq/core@next i saw that its already how i want it to be. My IDE was just not showing that it can be imported...

PR #3061, opened 2024-11-14
metrics-server: Restrict task monitor names

jsdanielh:
Restrict task monitor names such that the resulting metric name is compelling with the OpenMetrics spec (snake case naming).

## Pull request checklist

### DIFF
--- a/metrics-server/src/lib.rs
+++ b/metrics-server/src/lib.rs
@@ -31,10 +31,23 @@ mod server;
 mod tokio_runtime;
 mod tokio_task;
 
+/// Monitor (metrics) for a Nimiq task
 #[derive(Clone)]
 pub struct NimiqTaskMonitor {
-    pub name: String,
-    pub monitor: TaskMonitor,
+    name: String,
+    monitor: TaskMonitor,
+}
+
+impl NimiqTaskMonitor {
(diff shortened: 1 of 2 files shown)


metrics-server/src/lib.rs
```diff
@@ -31,10 +32,19 @@ mod server;
...
-    pub name: String,
-    pub monitor: TaskMonitor,
+    name: String,
+    monitor: TaskMonitor,
+}
+
+impl NimiqTaskMonitor {
+    pub fn new(name: String, monitor: TaskMonitor) -> Self {
+        // Metrics name **must** be snake case according to OpenMetrics spec.
+        let name = name.to_case(Case::Snake);
```
hrxi: Instead of fixing up the input, I'd suggest to panic if the input contains a non-alphanumeric non-underscore character.

This would not require the new dependency.

hrxi (commented):
In https://github.com/prometheus/OpenMetrics/blob/296468bc2359ebac83f24301b54a0871f2268016/specification/OpenMetrics.md, I only see:

> Names SHOULD be in snake_case.

But a good idea to comply with SHOULDs, too.

hrxi (approved):
Looks good to me, thanks!

hrxi:
Rebasing for merge.

PR #3062, opened 2024-11-14
Restrict Block topics from gossipsub to two blocks per second

ii-cruz:
The live sync relies on this constant being tight so that the peers do not trigger too many concurrent requests that ultimately could prevent the peer from syncing.

This PR, restricts block topics of gossipsub to 2 blocks per second or more precisely 20 blocks every 10 seconds. The rate limiting works on a per progataion source and message type.

## Pull request checklist

### DIFF
--- a/consensus/src/messages/mod.rs
+++ b/consensus/src/messages/mod.rs
@@ -143,7 +143,7 @@ impl Topic for BlockHeaderTopic {
     const BUFFER_SIZE: usize = 16;
     const NAME: &'static str = "block-header";
     const VALIDATE: bool = true;
-    const MAX_MESSAGES: u32 = 50;
+    const MAX_MESSAGES: u32 = 20;
 }
 
 #[derive(Clone, Debug, Default)]
@@ -155,7 +155,7 @@ impl Topic for BlockBodyTopic {
     const BUFFER_SIZE: usize = 16;
     const NAME: &'static str = "block-body";
     const VALIDATE: bool = true;
-    const MAX_MESSAGES: u32 = 50;
+    const MAX_MESSAGES: u32 = BlockHeaderTopic::MAX_MESSAGES;
 }
 
 /*


hrxi:
We expect one block per second via the gossipsub topic. This means that blocks will get dropped due to normal fluctuations, if I understand it correctly. Maybe we should set it to 20 instead?

sisou:
Yes, especially after macro blocks which take bit longer, or after skip blocks that have 4s delay, micro blocks are produced and broadcast faster.

consensus/src/messages/mod.rs
```diff
@@ -138,7 +138,7 @@ impl Topic for BlockHeaderTopic {
     const BUFFER_SIZE: usize = 16;
     const NAME: &'static str = "block-header";
     const VALIDATE: bool = true;
-    const MAX_MESSAGES: u32 = 50;
+    const MAX_MESSAGES: u32 = 10;
```
sisou: Why does `10` mean one per second?

nibhar: `TIME_WINDOW` is 10 seconds.

nibhar: So this is not 1 per second it is 10 per 10 seconds. After a macro block or a skip block this should still work out decently okay. I would still think it should not be as close a constant as it is, but it does not need to be 50. 20 as @hrxi suggested seems good to me.

ii-cruz: This is per-peer ID, so the chances of the same validator being chosen to produce multiple blocks within a 10-second time window are rather low, so the limit still allows for a bit of a buffer.

hrxi: > This is per-peer ID

As discussed directly with you, this is not the peer ID of who transmitted the message initially, but the peer ID of who propagated it to you. So it's still likely to be 10 blocks per 10 seconds.

ii-cruz: After checking with @hrxi, the limiting is not per peer id it is a propagation source. We are better off relaxing the max messages to 20 as suggested. Regarding the concerns about the requested missing blocks, issue #3071 was created as simple rate-limiting won't fully mitigate the issue.

_updated the PR description accordingly._

PR #3065, opened 2024-11-14
web-client: Add SignatureProof.SINGLE_SIG_SIZE constant

sisou:


jsdanielh (commented):
Could you add rustdocs?

jsdanielh (approved):

PR #3072, opened 2024-11-16
Block: remove unnecessary `clone` when accessing block body

styppo:
Remove unused `BlockJustification` as well.


hrxi (approved):
Looks good to me.

PR #3077, opened 2024-11-16
As validator, check keys per epoch

nibhar:
Since keys can change, this PR adds a basic check to the validator which makes sure its keys are consistent with the ones the election block expects. Similar checks are introduced to signing Level updates and Micro blocks such that a validator will not produce blocks if the keys do not match and not sign any LevelUpdates.

### DIFF
--- a/validator/src/validator.rs
+++ b/validator/src/validator.rs
@@ -355,6 +355,11 @@ where
             {
                 panic!("Invalid validator configuration: None of the voting keys match the one expected from this validator in the current epoch")
             }
+
+            // Compare configured validator signing key to the one in the current epoch to make sure it is the same.
+            if epoch_validator.signing_key != self.signing_key().public {
+                panic!("Invalid validator configuration: Configured signing key does not match signing key in this epoch");
+            }
         } else {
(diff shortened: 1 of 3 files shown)


validator/src/validator.rs
```diff
@@ -342,6 +342,19 @@ where
...
+
+            // Compare configured validator voting key to the one in the current epoch to make sure it is the same.
+            if *epoch_validator.voting_key.compressed() != self.voting_key().public_key.compress() {
+                panic!("Invalid validator configuration: Configured voting key does not match voting key in this epoch");
+            }
+
+            // Compare configured validator signing key to the one in the current epoch to make sure it is the same.
+            if epoch_validator.signing_key != self.signing_key().public {
+                panic!("Invalid validator configuration: Configured signing key does not match signing key in this epoch");
+            }
```
sisou, line 362: As discussed in-person just now: the check below that checks against the staking contract can probably be removed.

nibhar: done

Eligioo:
@nibhar can this PR get rebased?

PR #3083, opened 2024-11-17
Include the version of the migration tool in the online tx

hrxi:


pow-migration/src/monitor.rs
```diff
@@ -72,7 +74,7 @@ pub fn generate_online_tx(validator: String) -> OutgoingTransaction {
         to: Address::burn_address().to_user_friendly_address(),
         value: 1, //Lunas
         fee: 0,
-        data: Some(hex::encode("online")),
+        data: Some(hex::encode(String::from(ONLINE_TX_DATA))),
```
nibhar, line 77: There is a comment above that could be updated, but it is not of great concern.

nibhar (approved):

pow-migration/src/monitor.rs
hrxi: Done.

viquezclaudio (approved):

PR #3086, opened 2024-11-17
Rename the mainnet genesis override env variable

jsdanielh:
Rename the mainnet genesis override env variable from
    `NIMIQ_OVERRIDE_MAINET_CONFIG` to `NIMIQ_OVERRIDE_MAINNET_CONFIG`.

## Pull request checklist

### DIFF
--- a/genesis/src/networks.rs
+++ b/genesis/src/networks.rs
@@ -166,12 +166,12 @@ fn network_impl(network_id: NetworkId) -> Option<&'static NetworkInfo> {
                 use std::sync::OnceLock;
                 static OVERRIDE: OnceLock<Option<NetworkInfo>> = OnceLock::new();
                 if let Some(info) = OVERRIDE.get_or_init(|| {
-                    let override_path = env::var_os("NIMIQ_OVERRIDE_MAINET_CONFIG");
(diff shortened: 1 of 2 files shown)


blockchain-interface/src/error.rs
```diff
@@ -51,7 +51,7 @@ pub enum BlockchainError {
     #[error(
         "A full genesis config is required for initializing a history node on \
         the mainnet for the first time. Obtain the full genesis config (it's \
-        not in the repository) and set the `NIMIQ_OVERRIDE_MAINNET_CONFIG` \
+        not in the repository) and set the `NIMIQ_OVERRIDE_MAINET_CONFIG` \
```
hrxi: I think it should rather be the environment variable that should get fixed, optionally with a fallback for the typo.

sisou: Why? Now with the v1 release version we have the chance to make it right also in the nimiq-client.

sisou: I agree. Now with the 1.0 release we have the chance to make it right.

jsdanielh: Done

viquezclaudio (approved):

PR #3093, opened 2024-11-18
Fix TS types of web-client transaction sender & recipient types

sisou:
Somewhere the conversion from number to string got lost, but the type annotations for the resulting object didn't get updated.


jsdanielh (approved):
LGTM

PR #3094, opened 2024-11-18
Add RPC call and config option to rotate keys without restarting

fiaxh:
## Pull request checklist

### DIFF
--- a/validator/src/key_utils.rs
+++ b/validator/src/key_utils.rs
@@ -1,0 +1,40 @@
+use std::collections::HashMap;
+
+use nimiq_bls::{CompressedPublicKey, KeyPair as BlsKeyPair};
+
+pub struct VotingKeys {
+    keys: HashMap<CompressedPublicKey, BlsKeyPair>,
+    current_key: BlsKeyPair,
+}
+
+impl VotingKeys {
+    pub fn new(keys: Vec<BlsKeyPair>) -> Self {
+        assert!(!keys.is_empty());
+        let mut key_hm = HashMap::new();
+        for key in &keys {
+            key_hm.insert(key.public_key.compress(), key.clone());
+        }
+        VotingKeys {
+            keys: key_hm,
+            current_key: keys.first().unwrap().clone(),
+        }
+    }
+
+    pub fn add_key(&mut self, key: BlsKeyPair) {
+        self.keys.insert(key.public_key.compress(), key);
+    }
+
+    pub fn get_current_key(&self) -> BlsKeyPair {
+        self.current_key.clone()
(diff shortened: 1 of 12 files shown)


validator/src/key_utils.rs
```diff
@@ -0,0 +1,42 @@
...
+
+    pub fn get_keys(&self) -> Vec<BlsKeyPair> {
+        self.keys.values().cloned().collect()
+    }
+
+    pub fn update_current_key(&mut self, public_key: &CompressedPublicKey) {
+        self.current_key = self
+            .keys
+            .get(public_key)
+            .expect("Invalid validator configuration: Expected voting key not available")
```
hrxi: This is the user-visible panic when the user starts the client without the correct key, but also when the key is rotated on-chain and an election block happens.

I'm not sure what the wanted behavior is here. It's probably good to quit the client if the correct signing key isn't available on startup, but I'm not sure what to do when the client is already running and has hit an election block.

validator/src/validator.rs
```diff
@@ -70,7 +71,7 @@ struct InactivityState {
 pub struct ValidatorProxy {
     pub validator_address: Arc<RwLock<Address>>,
     pub signing_key: Arc<RwLock<SchnorrKeyPair>>,
-    pub voting_key: Arc<RwLock<BlsKeyPair>>,
+    pub voting_key: Arc<RwLock<VotingKeys>>,
```
hrxi:
```suggestion
    pub voting_keys: Arc<RwLock<VotingKeys>>,
```
Here and below.

hrxi (commented):
It would be nice if the same was available for signing keys as these can also be rotated.

fiaxh:
> It would be nice if the same was available for signing keys as these can also be rotated.

Yes, I agree. The implementations will look very similar. Thus, I wanted to first get this reviewed, such that I'm not doing every mistake twice, and then have to fix everything twice :slightly_smiling_face:

validator/src/validator.rs
```diff
@@ -153,7 +154,7 @@ where
         validator_address: Address,
         automatic_reactivate: bool,
         signing_key: SchnorrKeyPair,
-        voting_key: BlsKeyPair,
+        voting_key: VotingKeys,
```
hrxi:
```suggestion
        voting_keys: VotingKeys,
```

hrxi (commented):
Looks good to me.

hrxi:
Rebased for merging.

sisou:
@fiaxh Ping on the version of this for signing keys, as talked about above.

PR #3095, opened 2024-11-18
Changing the max connections per peer so that the seed nodes become public

### LINKED ISSUE #3069: DHT server mode never reached in CI runs
There are several CI failures where we fail to aggregate signatures on a skip block when we kill one validator.

https://github.com/nimiq/core-rs-albatross/actions/runs/11844877331

ii-cruz:
Changes the max connections per peer. The seed nodes have many connections, and this constant is being exceeded. This leads them to get errors when connecting to peers; thus, we mistakenly assume we are not public.

## Pull request checklist

### DIFF
--- a/network-libp2p/src/behaviour.rs
+++ b/network-libp2p/src/behaviour.rs
@@ -20,7 +20,7 @@ use crate::{
 };
 
 /// Maximum simultaneous libp2p connections per peer
-const MAX_CONNECTIONS_PER_PEER: u32 = 3;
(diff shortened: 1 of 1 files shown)


hrxi:
> The seed nodes have many connections

If I understand the constants correctly, they're about connections between the exactly same peers. Why does the seed node have more than 3 connections to the same peer?

Is it because the node got shut down but the connection is still there? They are all on the same node, right? Can a TCP connection to localhost persist in Linux if the process has died?

jsdanielh (approved):

PR #3096, opened 2024-11-18
Implement Version Upgrade Mechanism

paberr:
This PR implements the mechanism to carry out breaking changes (version upgrades) after launch.

It mostly follows the specification in the [Albatross paper](https://arxiv.org/pdf/1903.01589v5) with one exception:
We will not only check for the proportion of **stake supporting the upgrade**, but also for the proportion of **slots supporting the upgrade**. This makes sure we don't propose an upgrade in an epoch where less than 2f+1 elected validators support the upgrade.

> ⚠️ All changes in this PR that affect the consensus **pre-version upgrade** are backwards compatible in that the current nodes enforce **stricter** rules. All consensus breaking changes would only ever be performed **post-version upgrade** (e.g., deactivating validators).

### Mechanism Overview
The version upgrade mechanism introduces the following key changes, discussed in a bit more detail below:

1. Modified block validity (i.e., when upgrades are possible)
2. Tools for checking supporting stake/slots
3. Validators producing blocks with new version
4. New version upgrade inherent (deactivating unsupporting validators)
5. Moving validator selection in election blocks from _before the accounts commit_ to _after the accounts commit_

#### Block validity
The version number is stored in every block. We allow version upgrades only in election blocks.
Instead of checking against a constant version from the `Policy`, block-level verification of the version is now split into two parts:
- `Block::verify_header`: Making sure that the version is always below or equal to the maximum supported version `Policy::max_supported_version(network_id)`. The maximum supported version depends on the network, so that we can roll out changes in the testnet first.
- `Block:: verify_immediate_successor` and `Block::verify_macro_successor`: Making sure that the version number can only increase in election blocks and only in steps of one. For all other scenarios, the version needs to match the predecessor's version. We test this in both functions as these are called in different scenarios.

#### Helper functions
We only perform version upgrades when at least `Policy:: UPGRADE_MIN_SUPPORT`% of the active stake signals support for the upgrade and when at least `Policy::TWO_F_PLUS_ONE` slots of the currently elected validators signals support.
We add helper functions to determine the supporting stake and slots.

The `get_supporting_stake` function is computationally expensive in that it iterates over all validators in the staking contract. This number is bounded by the validator deposit and we only call the function when proposing an election block and not being on the maximum version already.

Currently, the support for a version is signalled by encoding the supported version in the first two bytes of the validator's signal data (in big-endian). This can still be changed later and be implemented version-specific in the `Policy::supports_upgrade` function.

We also add a function to deactivate unsupportive stake, which is discussed in the section about the new inherent.

#### Producing an upgrade
Validators who updated their client and have a maximum supported version higher than the current blockchain's version (determined from the head block) will determine support for this upgrade whenever they are the ones to produce a macro proposal. Only when the conditions (`supporting_stake >= Policy:: UPGRADE_MIN_SUPPORT * active stake && supporting_slots > Policy::TWO_F_PLUS_ONE`) are met, the proposal will contain the increased version number. If the proposal is accepted, the upgrade is successfully active from this block on (including this election block).

There is no explicit check for validators to only vote for a block if they support the upgrade. If they updated their client and have a new maximum supported version, the proposal will be deemed valid and voted for. Validators not supporting the upgrade won't update their clients and will reject the proposal due to an invalid version.

#### Version upgrade inherent
The mechanism adds a new inherent `VersionUpgrade` that also has the new version. It has the side effect of deactivating all validators that did not explicitly signal support for the new version with immediate effect. This also means that these unsupporting validators should not be selected for the next epoch (thus preventing possible stalls).

Full nodes will apply this side-effect through trie-diffs. The inherent is not stored in the history and thus does not require any special handling.

#### Switching validator selection and accounts commit
In the current model, we first select the validators for the next epoch (based on the current staking contract) and then apply the accounts changes of an election block. The only accounts changes possible in an election block right now are: reward distribution and preparing the bitsets in the staking contract for the next epoch.

Since none of these have an effect on the validator selection, we switch the order of these two. This is required so that unsupporting validators are deactivated **before** the validator selection.

### Tests
This PR includes unit tests in the following crates:
- `validator`: Testing that the validator will propose an election block with a new version if the conditions are met.
- `block`: Testing the new validity rules.
- `staking_contract`: Testing correct computation of stake/slot support and correct handling of inherent.
- `blockchain`: Testing that the blockchain correctly emits the new inherent.

## Pull request checklist


primitives/src/policy.rs
```diff
@@ -156,6 +156,37 @@ impl Policy {
...
+        match network_id {
+            NetworkId::TestAlbatross => 1,
+            NetworkId::DevAlbatross => 1,
+            NetworkId::MainAlbatross => 1,
+
+            NetworkId::UnitAlbatross => 2,
+
+            _ => 1,
+        }
+    }
```
hrxi: This should probably be a member variable of `Policy` instead.

validator/src/tendermint.rs
```diff
@@ -228,11 +228,62 @@ where
...
+                    %supporting_stake, %total_active_stake, "Version upgrade does not have enough support"
+                );
+                None
+            }
+        } else {
+            None
+        };
+
         // Create the proposal.
+        // TODO: Version
```
hrxi: Outdated?

validator/src/tendermint.rs
```diff
@@ -228,11 +228,62 @@ where
...
+                data_store_read,
+                blockchain
+                    .current_validators()
+                    .expect("There need to be validators present"),
+                support_check,
+            );
+
+            // We propose an upgraded block version only if:
+            // - `Policy::UPGRADE_MIN_SUPPORT` of the stake supports the upgrade.
+            // - `Policy::TWO_F_PLUS_ONE` slots support the upgrade.
```
hrxi: Maybe also only upgrade when 80% of the slots support the upgrade? Otherwise, a few slots can successfully reject this proposal.

primitives/src/policy.rs
```diff
@@ -156,6 +156,37 @@ impl Policy {
     /// 25 MB.
     pub const HISTORY_CHUNKS_MAX_SIZE: u64 = 25 * 1024 * 1024;
 
+    /// Percentage of minimum supporting stake to perform upgrade.
+    /// A value of `80` means 80% of the stake needs to support the upgrade.
+    pub const UPGRADE_MIN_SUPPORT: u8 = 80;
+
+    /// This function is used to determine if a validator signalled for a specific upgrade.
+    /// For now, this is checking the first two bytes of the signal data.
```
hrxi:
```suggestion
    /// This is checking the first two bytes of the signal data.
```

"For now" suggests to me that we're currently already planning something else.

validator/src/tendermint.rs
```diff
@@ -228,11 +228,62 @@ where
             return Err(ProtocolError::Abort);
         }
 
+        // Check if we want to upgrade the version.
+        // This assumes we only ever upgrade to the latest version
+        // and don't queue multiple upgrades.
```
hrxi, line 233: This is not a problem, I think. We can fix this if we ever want to do multiple upgrades at once.

blockchain/tests/inherents.rs
```diff
@@ -699,3 +699,119 @@ async fn create_fork_proof() {
...
+        seed: VrfSeed::default(),
+        extra_data: vec![],
+        state_root: Blake2bHash::default(),
+        body_root: Blake2sHash::default(),
+        diff_root: Blake2bHash::default(),
+        history_root: Blake2bHash::default(),
+        validators: None,
+        next_batch_initial_punished_set,
+        ..Default::default()
+    };
```
hrxi, line 712: Most of this struct can probably be omitted since you specify `..Default::default()` anyway.

hrxi (commented):
I wonder if we should somehow support the case of someone not wanting an upgrade. In that case, they should probably not signal support for it — however, this means that they'll get deactivated on a version upgrade. This puts them in a bad spot. Could we maybe add a "I have technical support for this upgrade but I don't like it" flag that allows a validator to vote against an upgrade while not getting deactivated if it does get activated?

validator/src/tendermint.rs
nibhar: I disagree. I think this is a problem.
In the scenario of a proposed upgrade never being adopted there is no good way of having another update being adopted. Basically this means that every update must be accepted before another can be accepted or to introduce ambiguity.

What version would another update have? In order for it to be able to be adopted it needs to be `current_version + 1`, but that is already occupied by the first (not adopted) upgrade. This ambiguity is not good I think.

blockchain/src/blockchain/inherents.rs
```diff
@@ -321,8 +321,15 @@ impl Blockchain {
     }
 
     /// Creates the inherent to finalize an epoch. The inherent is for updating the StakingContract.
-    pub fn finalize_previous_epoch(&self) -> Inherent {
+    pub fn finalize_previous_epoch(&self, version: u16) -> Vec<Inherent> {
         // Create the FinalizeEpoch inherent.
-        Inherent::FinalizeEpoch
+        let mut inherents = vec![Inherent::FinalizeEpoch];
+        // Add version upgrade inherent if needed.
+        if version != self.state.current_version() {
```
jsdanielh: What about:
```suggestion
        if version > self.state.current_version() {
```
?

blockchain/src/blockchain/slots.rs
```diff
@@ -44,10 +44,12 @@ impl Blockchain {
     }
 
     /// Calculates the next validators from a given seed.
-    pub fn next_validators(&self, seed: &VrfSeed) -> Validators {
-        let staking_contract = self.get_staking_contract();
+    pub fn next_validators(&self, seed: &VrfSeed, txn: Option<&MdbxReadTransaction>) -> Validators {
+        let txn = txn.or_new(&self.db);
+        let staking_contract = self
+            .get_staking_contract_if_complete(Some(&txn))
+            .expect("We should always have the staking contract.");
```
jsdanielh:
```suggestion
            .expect("We should always have the staking contract");
```

primitives/Cargo.toml
```diff
@@ -61,7 +61,7 @@ coin = ["hex", "nimiq-serde", "regex", "thiserror"]
 key-nibbles = ["hex", "nimiq-keys", "nimiq-database-value", "nimiq-database-value-derive", "nimiq-serde"]
 networks = ["thiserror"]
 parallel = ["rayon", "ark-ec/parallel"]
-policy = ["nimiq-keys", "nimiq-utils", "parking_lot"]
+policy = ["nimiq-keys", "nimiq-utils", "parking_lot", "networks"]
```
jsdanielh:
```suggestion
policy = [ "networks", "nimiq-keys", "nimiq-utils", "parking_lot"]
```

primitives/block/src/block.rs
```diff
@@ -406,13 +406,20 @@ impl Block {
             return Err(BlockError::NetworkMismatch);
         }
 
-        // Check the version
-        if self.version() != Policy::VERSION {
+        error!(
```
jsdanielh: Why is this an error log?

validator/src/tendermint.rs
hrxi: After one update is rejected, the next update (with version + 2) can include the mechanism that updates with + 2 can be accepted as well. We're not setting anything in stone here, I think. Just like our current code doesn't support updates at all but we can still allow the update with version + 1 to work.

blockchain/src/blockchain/history_sync.rs
```diff
@@ -287,7 +287,7 @@ impl Blockchain {
                 "Inserting final macro block"
             );
             block_numbers.push(block.block_number());
-            block_timestamps.push(0); // FIXME
+            block_timestamps.push(block.timestamp());
```
nibhar, line 290: A bunch of lines up, there is a similar `FIXME`. Could that one also be fixed?

blockchain/src/blockchain/inherents.rs
nibhar: Or even, as that is currently the case as far as I understand it
```suggestion
        if version == self.state.current_version().checked_add(1) {
```

blockchain/src/blockchain/history_sync.rs
```diff
@@ -303,10 +303,25 @@ impl Blockchain {
...
+                        block.block_number(),
+                        "Only the last block can be an election block"
+                    );
                     block_inherents
                         .get_mut(i)
                         .unwrap()
                         .push(Inherent::FinalizeEpoch);
+
+                    // If the version changed, add the upgrade inherent.
+                    if this.state.current_version() < block.version() {
```
nibhar, line 317: This does not reflect the criteria that the version can only change by one here. Maybe it does so in a different spot in the call stack, but I would prefer to have it here too, to be more explicit.

blockchain/src/blockchain/slots.rs
nibhar: Is this actually guaranteed? Is there no way for `next_validators` to be called during state sync?

It feels overly restrictive and obfuscated for this function to require the staking contract to be present this way. I would prefer to have the staking contract as a parameter, making this requirement explicit. All call sites of `next_validators` have already called `get_staking_contract_if_complete` before calling `next_validators` anyways.

blockchain/src/blockchain/verify.rs
```diff
@@ -143,6 +143,22 @@ impl Blockchain {
         if let Block::Macro(macro_block) = block {
             // If we don't have the staking contract, there is nothing we can check.
             if let Some(staking_contract) = self.get_staking_contract_if_complete(Some(txn)) {
+                // Verify validators.
+                let validators = match macro_block.is_election() {
+                    true => Some(self.next_validators(&macro_block.header.seed, Some(txn))),
```
nibhar, line 149: See other comment about the staking contract requirement.

blockchain/tests/inherents.rs
```diff
@@ -699,3 +699,119 @@ async fn create_fork_proof() {
...
+        transactions: reward_transactions,
+    };
+
+    let macro_block = MacroBlock {
+        header: macro_header.clone(),
+        body: Some(body.clone()),
+        justification: None,
+    };
+
+    // Simple case. Expect 1x FinalizeBatch, 1x FinalizeEpoch, 1x Reward to validator, 2x Version Upgrade
```
nibhar: This sums up to 5 inherents but 4 are asserted.

blockchain/tests/inherents.rs
```diff
@@ -699,3 +699,119 @@ async fn create_fork_proof() {
...
+    // Downgrading version will remove version upgrade.
+    macro_header.version = 1;
+
+    let macro_block = MacroBlock {
+        header: macro_header.clone(),
+        body: Some(body),
+        justification: None,
+    };
+
+    // Simple case. Expect 1x FinalizeBatch, 1x FinalizeEpoch, 1x Reward to validator, 2x Version Upgrade
```
nibhar: This still adds up to 5 but 3 are asserted.

blockchain/tests/push.rs
```diff
@@ -283,7 +283,9 @@ fn it_validates_network() {
 fn it_validates_version() {
     expect_push_micro_block(
         BlockConfig {
-            version: Some(Policy::VERSION - 1),
+            version: Some(Policy::max_supported_version(NetworkId::UnitAlbatross) + 1),
```
nibhar: For this it would probably also be nice to reject a version jump of 2 as well as downgrading the version in an election block.

light-blockchain/tests/push.rs
```diff
@@ -308,7 +308,7 @@ fn it_works_with_valid_blocks() {
 fn it_validates_version() {
     expect_push_micro_block(
         BlockConfig {
-            version: Some(Policy::VERSION - 1),
```
nibhar, line 311: See other comment about additional test cases.

primitives/account/src/account/staking_contract/mod.rs
```diff
@@ -177,4 +182,97 @@ impl StakingContract {
...
+        StakingContractStoreRead::new(data_store)
+            .iter_validators()
+            .filter_map(|validator| {
+                if validator.is_active() && support_check(validator.signal_data) {
+                    Some(validator.total_stake)
+                } else {
+                    None
+                }
+            })
+            .sum()
```
nibhar: Alternatively:
```suggestion
        let staking_Contract_read = StakingContractStoreRead::new(data_store);
        self.active_validators
            .iter()
            .filter_map(|(validator_address, stake)| {
                let validator_signal_data = staking_contract_read
                    .get_validator(address)
                    .expect("Active Validators must be present in the Staking Contract.")
                    .signal_data;
                support_check(validator_signal_data).then_some(*stake)
            })
            .sum()
```
Should also work. It shifts from going over all validator into going over only active validators and querying their signal data individually. May even be worse, just left it here for consideration.

rpc-interface/src/types.rs
```diff
@@ -340,7 +340,6 @@ pub struct PolicyConstants {
     pub coinbase_address: String,
     pub transaction_validity_window: u32,
     pub max_size_micro_body: usize,
-    pub version: u16,
```
nibhar, line 365: This seems useful, why was it removed? Maybe rename it to `max_supported_version`?

validator/src/tendermint.rs
```diff
@@ -228,11 +228,62 @@ where
...
+                    .expect("There need to be validators present"),
+                support_check,
+            );
+
+            // We propose an upgraded block version only if:
+            // - `Policy::UPGRADE_MIN_SUPPORT` of the stake supports the upgrade.
+            // - `Policy::TWO_F_PLUS_ONE` slots support the upgrade.
+            if supporting_slots >= Policy::TWO_F_PLUS_ONE
+                && u64::from(supporting_stake) * 100 / u64::from(total_active_stake)
+                    >= Policy::UPGRADE_MIN_SUPPORT as u64
```
nibhar: Maybe that constant should be a u64 to begin with.

nibhar (commented):
Left some comments mostly for clarity. Functionally this looks good to me.

validator/src/tendermint.rs
paberr: I agree with @hrxi . I left the comment in the code so that the (current) assumption is explicit. We can still change that in the future.

primitives/block/src/block.rs
paberr: Good question :D

blockchain/src/blockchain/history_sync.rs
paberr: No, because we don't have a history item associated with that macro block

blockchain/tests/inherents.rs
paberr: Comment is wrong

primitives/account/src/account/staking_contract/mod.rs
paberr: Depends on the number of validators. It's `log(n) + n` for the current vs. `m * log(n)` in the other case. Not sure what is better in practice.

blockchain/src/blockchain/verify.rs
paberr: I don't think the two are equivalent.
`next_validators` will fetch the staking contract again, which means that inherents have already been applied.

This should be especially important in the upgrade: Some validators just got deactivated.

paberr:
I implemented all suggested changes except the one about `next_validator` assuming the staking contract is there (this is the same behavior as previously, and I am unsure if passing in the staking contract is semantically equivalent here).

In particular, I moved the `max_supported_version` into the `NetworkInfo` similar to the genesis block number.
That successfully removed the network dependency from the policy again.

jsdanielh (approved):
LGTM

viquezclaudio (approved):

PR #3105, opened 2024-11-20
Instead of `active/inactive` validators log `elected/not elected`

nibhar:
As active/inactive is overloaded and has already led to some confusion, this changes these printouts to elected/not elected. For elected validators is also includes the slots.

### DIFF
--- a/validator/src/validator.rs
+++ b/validator/src/validator.rs
@@ -337,15 +337,16 @@ where
         *self.slot_band.write() = validators.get_slot_band_by_address(&self.validator_address());
 
         if let Some(slot_band) = *self.slot_band.read() {
+            let epoch_validator = validators.get_validator_by_slot_band(slot_band);
             log::info!(
                 validator_address = %self.validator_address(),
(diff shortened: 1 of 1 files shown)


validator/src/validator.rs
```diff
@@ -337,15 +337,16 @@ where
...
 
         if let Some(slot_band) = *self.slot_band.read() {
+            let epoch_validator = validators.get_validator_by_slot_band(slot_band);
             log::info!(
                 validator_address = %self.validator_address(),
                 validator_slot_band = slot_band,
                 epoch_number = blockchain.epoch_number(),
-                "We are ACTIVE in this epoch"
+                slots = ?epoch_validator.slots,
+                "We are ELECTED this epoch"
```
sisou:
```suggestion
                "We are ELECTED in this epoch"
```

For consistency.

sisou (approved):

ii-cruz (approved):

jsdanielh (approved):

PR #3109, opened 2024-11-21
consensus: Fix transaction proofs for non finalized blocks

jsdanielh:
Fix transaction proofs for non finalized blocks in `prove_transaction` which is called when a transaction proof is requested without specifying a block. In this case if the transaction is found to be within the current batch (not finalized) it was incorrectly generating the proof against the head of the blockchain instead of the proving block which in this case is the block that included the transaction.
This was a condition that was overseen in #1517.

## Pull request checklist

### DIFF
--- a/consensus/src/messages/handlers.rs
+++ b/consensus/src/messages/handlers.rs
@@ -559,8 +559,8 @@ impl RequestTransactionsProof {
             let block_number = hist_txn.block_number;
 
             let proving_block_number = if block_number <= election_head {
-                // If the txn is in a finalized epoch, we use the last election block
-                election_head
+                // If the txn is in a finalized epoch, we use the election block of that epoch
+                Policy::election_block_after(block_number)
             } else if block_number <= macro_head {
                 // If the txn is in a finalized batch in the current epoch, we use the last checkpoint block
                 macro_head
--- a/consensus/src/messages/handlers.rs
+++ b/consensus/src/messages/handlers.rs
@@ -546,6 +546,7 @@ impl RequestTransactionsProof {
         let mut verifier_state = None;
         let election_head = blockchain.election_head().block_number();
         let macro_head = blockchain.macro_head().block_number();
+        let current_head = blockchain.head().block_number();
 
         // Get the historic transaction from the history store
         let historic_transaction = blockchain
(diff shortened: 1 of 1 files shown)


sisou:
Aha. Does this explain the "Block not found" errors in the wallet? Because the proof is against a block the light node doesn't yet have? Or did this mean the proof was invalid?

jsdanielh:
> Aha. Does this explain the "Block not found" errors in the wallet? Because the proof is against a block the light node doesn't yet have? Or did this mean the proof was invalid?

We were getting invalid proofs when requesting transactions that are in the current non finalized batch

consensus/src/messages/handlers.rs
```diff
@@ -578,10 +579,8 @@ impl RequestTransactionsProof {
                 return Err(ResponseTransactionProofError::BlockNotFound);
             };
 
-            // If it is a checkpoint block, we have some extra work to do
-            if Policy::is_macro_block_at(proving_block_number)
-                && !Policy::is_election_block_at(proving_block_number)
-            {
+            // We have some extra work in the current epoch, if the block we are proving is not our current head
+            if proving_block_number > election_head && proving_block_number < current_head {
```
nibhar, line 583: What makes the `current_head` not applicable here?

nibhar (changes requested):
While we were talking here we discovered that there is likely another issue, with blocks from epochs earlier then the last. I have pushed a commit to `nibhar/fix` which sits on top of this branch. Feel free to use it as you see fit.

consensus/src/messages/handlers.rs
jsdanielh: In that case, `verifier_state` is allowed to be `None` as it would be without entering this `if` condition

jsdanielh:
> While we were talking here we discovered that there is likely another issue, with blocks from epochs earlier then the last. I have pushed a commit to `nibhar/fix` which sits on top of this branch. Feel free to use it as you see fit.

Ah, nice catch. Added it to this PR

sisou (approved):

nibhar (approved):

viquezclaudio (approved):

PR #3111, opened 2024-11-21
Only ever prune beyond `max_epochs_stored`

### LINKED ISSUE #3110: `max_epochs_stored` should affect history store
Currently, the `max_epochs_stored` config option only affects the chain store, but not the history.
Also, the `saturating_sub` always deletes the first epoch, even if not intended.

See: https://github.com/nimiq/core-rs-albatross/blob/d39203fa1d7e53fd158dd2368ec733792d895976/blockchain/src/blockchain/push.rs#L370

### LINKED ISSUE #3112: Config option `consensus.max_epochs_stored` is ignored
On both history and full nodes, the generated config logged in DEBUG at node startup always shows `max_epochs_stored: 1`, no matter what is set in the config file.

Maybe other settings have the same issue.

paberr:
- Use `max_epochs_stored` to identify history to be pruned
- Do not use saturating sub to avoid pruning epochs too early
- Add missing call in config builder to read setting from config file

## Pull request checklist

### DIFF
--- a/blockchain/src/blockchain/push.rs
+++ b/blockchain/src/blockchain/push.rs
@@ -358,19 +358,19 @@ impl Blockchain {
             let max_epochs_stored =
                 cmp::max(this.config.max_epochs_stored, Policy::MIN_EPOCHS_STORED);
 
-            // Calculate the epoch to be pruned. Saturate at zero.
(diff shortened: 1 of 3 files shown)


jsdanielh (approved):
LGTM

sisou (approved):

viquezclaudio (approved):

PR #3134, opened 2024-11-22
Implement deserializers and defaults for PlainTransactionDetails

### LINKED ISSUE #227: History item missing
After using the wallet live and witnessing transactions coming in, one of them was missing for an unknown reason. It was one transaction in the middle of others, which could be observed properly. After that the history item was never retrieved. Only after enabling a debugger and removing the `knownTxDetails` used in the `client.getTransactionsByAddress(...)` it was able to be resolved.

sisou:
So it can handle PoW-style plain TransactionDetails.

#### This fixes transaction history requests in the wallet with an existing PoW history.

Fixes https://github.com/nimiq/wallet/issues/227.

## Pull request checklist


jsdanielh (approved):
LGTM

PR #3137, opened 2024-11-22
consensus: Also require the pre-genesis flag for tx receipts

jsdanielh:
Require the pre-genesis flag for transaction receipts when the request doesn't have a block number or when any of the receipts is for a block number that precedes the genesis.

## Pull request checklist


paberr (approved):
LGTM

sisou (approved):

PR #3139, opened 2024-11-23
Set pre-genesis flag before using it

paberr:
The pre-genesis flag was set only after we created our peer contact.
That was possible as it implements `Copy`.

## Pull request checklist


jsdanielh (approved):
LGTM

PR #3140, opened 2024-11-23
light-blockchain: Always provide some blocks from ChainStore

fiaxh:
Currently, when a HeadChangedListener is called on a macro block, calling get_block right afterwards doesn't work, as the ChainStore is cleared even before the event gets signaled.

## Pull request checklist

### DIFF
--- a/light-blockchain/src/chain_store.rs
+++ b/light-blockchain/src/chain_store.rs
@@ -351,11 +351,25 @@ impl ChainStore {
             .insert(Policy::epoch_at(header.block_number), header);
     }
 
-    /// Clears the ChainStore of all blocks (except the election blocks). This can be used at the
-    /// end of each batch, so that we don't keep unnecessary micro blocks.
-    pub fn clear(&mut self) {
-        self.chain_db.clear();
-        self.height_idx.clear();
+    /// Removes old blocks. It retains the latest election block, and removes all other blocks if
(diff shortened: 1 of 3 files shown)


light-blockchain/src/chain_store.rs
```diff
@@ -351,11 +351,25 @@ impl ChainStore {
             .insert(Policy::epoch_at(header.block_number), header);
     }
 
-    /// Clears the ChainStore of all blocks (except the election blocks). This can be used at the
-    /// end of each batch, so that we don't keep unnecessary micro blocks.
-    pub fn clear(&mut self) {
-        self.chain_db.clear();
-        self.height_idx.clear();
+    /// Removes old blocks. It retains the latest election block, and removes all other blocks if
+    /// they are older than BLOCKS_PER_EPOCH. The light blockchain would only need micro blocks of
```
jsdanielh:
```suggestion
    /// they are older than BLOCKS_PER_BATCH. The light blockchain would only need micro blocks of
```

jsdanielh (approved):
LGTM

jsdanielh:
Signing commit...

hrxi:
Can you drop the "A setting for the number of batches to keep can be added in a subsequent PR." if we don't intend to do it, please?

hrxi:
Thanks!

PR #3141, opened 2024-11-23
Intern BLS lazy public keys

hrxi:
This ensures that for each compressed public key there's at most one lazy public key, which means everyone profits when it is uncompressed.

This drops the (unused: never read from) BLS public key cache.


utils/src/interner.rs
```diff
@@ -0,0 +1,209 @@
...
+    borrow::Borrow,
+    collections::HashSet,
+    hash::{Hash, Hasher},
+    ops::Deref,
+    sync::{Arc, Weak},
+};
+
+use parking_lot::RwLock;
+
+// TODO: document that you need to call notify_deletion in the `Drop` impl
```
paberr: Document what you promised

paberr (commented):
Think about whether we can drop the requirement of the user of the interner to implement a specific Drop by wrapping the type.
Also fix a couple of Clippy warnings.

utils/src/interner.rs
hrxi: No longer necessary.

hrxi:
> Think about whether we can drop the requirement of the user of the interner to implement a specific Drop by wrapping the type.

Did so, resulted in a much nicer API. :)

bls/src/lazy.rs
```diff
@@ -1,141 +1,110 @@
...
-    pub(crate) compressed: CompressedPublicKey,
-    pub(crate) cache: RwLock<Option<PublicKey>>,
+fn cache() -> &'static Interner<CompressedPublicKey, OnceLock<Option<PublicKey>>> {
+    static CACHE: OnceLock<Interner<CompressedPublicKey, OnceLock<Option<PublicKey>>>> =
+        OnceLock::new();
+    CACHE.get_or_init(Default::default)
 }
 
+#[derive(Clone)]
+pub struct LazyPublicKey(Interned<CompressedPublicKey, OnceLock<Option<PublicKey>>>);
```
jsdanielh, line 20: Could you add `rustdoc`?

lib/src/client.rs
```diff
@@ -377,9 +376,7 @@ impl ClientInner {
             config.database,
         )?;
 
-        let bls_cache = Arc::new(Mutex::new(PublicKeyCache::new(
-            Policy::BLS_CACHE_MAX_CAPACITY,
-        )));
+        let bls_cache = Arc::new(Mutex::new(BlsCache::default()));
```
jsdanielh, line 379:
```suggestion
        let bls_cache = Arc::new(Mutex::new(BlsCache));
```

consensus/src/bls_cache.rs
```diff
@@ -0,0 +1,25 @@
+use nimiq_bls::LazyPublicKey as BlsLazyPublicKey;
+use nimiq_primitives::policy::Policy;
+
+// TODO: implement some caching strategy
+pub struct BlsCache;
```
jsdanielh: Could you add some `rustdoc`?

consensus/src/bls_cache.rs
```diff
@@ -0,0 +1,30 @@
+use hashlink::LruCache;
+use nimiq_bls::LazyPublicKey as BlsLazyPublicKey;
+use nimiq_primitives::policy::Policy;
+
+pub struct BlsCache {
```
jsdanielh, line 6: Could you add some rustdoc?

jsdanielh:
Also it looks like there are untouched filed with clippy warnings

viquezclaudio:
With this changes, in Safari we get consensus in ~12secs

Previously, in Safari, we used to get consensus in ~23secs

lib/src/client.rs
hrxi: Was intentionally not done because I wanted to add some `struct` members later.

bls/src/lazy.rs
hrxi: Done.

consensus/src/bls_cache.rs
hrxi: Done.

hrxi:
> Also it looks like there are untouched filed with clippy warnings

Fixed.

jsdanielh (approved):
LGTM

Issue #2986, opened 2024-10-18
Metrics server reports topic as unknown


sisou:
![Image](https://github.com/user-attachments/assets/1d633b3d-0908-4830-83e1-67b58b79bfa3)

In the above graph I am plotting the received gossipsub messages by topic. The line that hovers around 100 (which is the TPS at the moment in the network) is reported as `unknown`. This node is a non-elected validator.

I would _imagine_ these messages are transactions and blocks.

PR #3142, opened 2024-11-24
Add a plain list of `services` to web-client peer info structs

sisou:
## Pull request checklist

### DIFF
--- a/web-client/example/module.js
+++ b/web-client/example/module.js
@@ -5,6 +5,13 @@ window.Nimiq = Nimiq;
 init().then(async () => {
     const config = new Nimiq.ClientConfiguration();
     config.logLevel('debug');
+    // config.network('testalbatross');
+    // config.seedNodes([
+    //     "/dns4/seed1.pos.nimiq-testnet.com/tcp/8443/wss",
+    //     "/dns4/seed2.pos.nimiq-testnet.com/tcp/8443/wss",
+    //     "/dns4/seed3.pos.nimiq-testnet.com/tcp/8443/wss",
+    //     "/dns4/seed4.pos.nimiq-testnet.com/tcp/8443/wss",
+    // ]);
 
     const client = await Nimiq.Client.create(config.build());
     window.client = client; // Prevent garbage collection and for playing around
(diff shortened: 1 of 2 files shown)


web-client/src/client/peer_info.rs
```diff
@@ -6,13 +6,49 @@ use wasm_bindgen::prelude::*;
...
+    History,
+    AccountsProof,
+    AccountsChunk,
+    Mempool,
+    TransactionIndex,
+    Validator,
+    PreGenesisTransactions,
+
+    // Catch-all to not have to panic when new services are added
+    Unknown,
```
hrxi, line 35: Is `Unknown` useful info for the web client? If not, I'd prefer to simply skip these.

The `from` function would return an `Option<PlainService>` and the `PlainPeerInfo` could then just ignore the `None`s.

sisou: I added it because I wanted to avoid silent failing. If we ever add more services, but forget to add them here, noone would ever find out.

Panicking or erroring is bad for backward-compatibility, so I think 'Unknown' is the next best thing.

That's also how I found https://github.com/nimiq/core-rs-albatross/issues/2986, because the new topics were not hidden.

web-client/example/module.js
```diff
@@ -5,6 +5,13 @@ window.Nimiq = Nimiq;
 init().then(async () => {
     const config = new Nimiq.ClientConfiguration();
     config.logLevel('debug');
+    // config.network('testalbatross');
+    // config.seedNodes([
+    //     "/dns4/seed1.pos.nimiq-testnet.com/tcp/8443/wss",
+    //     "/dns4/seed2.pos.nimiq-testnet.com/tcp/8443/wss",
+    //     "/dns4/seed3.pos.nimiq-testnet.com/tcp/8443/wss",
+    //     "/dns4/seed4.pos.nimiq-testnet.com/tcp/8443/wss",
+    // ]);
```
jsdanielh, line 14:
```suggestion
```

web-client/example/module.js
```diff
@@ -46,7 +53,17 @@ init().then(async () => {
 
     document.querySelector('#address-book').addEventListener("click", async () => {
         let contacts = await client.getAddressBook();
-        console.table(contacts);
+        console.log(contacts
+            .sort((a, b) => {
+                // Sort seeds first
+                const aIsSeed = a.address.includes('seed');
+                const bIsSeed = b.address.includes('seed');
```
jsdanielh, line 60: Mmm, this might not hold true forever

hrxi: What might not hold true forever?

jsdanielh: That seeds always include `seed` in their addresses

hrxi: Ah true, and also other nodes might just include `seed` in their hostnames.

sisou: It's just a sorting for better scannability in the demo page, nothing for production.

web-client/example/module.js
sisou: It's nice to have this around in the demo page to quickly test in testnet, without having to look up these URLs somewhere.

jsdanielh (approved):

PR #3158, opened 2024-11-26
Wrap `futures::stream::SelectAll` for easier waker behavior

hrxi:


jsdanielh (approved):
LGTM but could you apply `cargo +nightly fmt`?

PR #3159, opened 2024-11-26
Wait for Account proofs blocks if they are not available yet

nibhar:
When getting the trie in the RemoteDataStore previous to this PR the block which is referenced by the returned proof must already exist locally. That is by no means guaranteed and will result in discarding the proof. 

This PR thus added a waiting window within which the RemoteDataStore will wait for the block to arrive. It will wait a specified amount of blockchain events which have added blocks. 

Additionally whenever a proof would not verify another peer will be asked to proof an address, whereas before this PR the get_trie would simply fail at this point.

A button was added to the web client example to exercise the behaviour.

### DIFF
--- a/consensus/src/consensus/remote_data_store.rs
+++ b/consensus/src/consensus/remote_data_store.rs
@@ -3,11 +3,14 @@ use std::{
     sync::Arc,
 };
 
+use futures::StreamExt;
 use nimiq_account::{
     Account, DataStoreReadOps, Staker, StakingContract, StakingContractStore, Tombstone, Validator,
 };
+use nimiq_block::Block;
 use nimiq_blockchain_interface::AbstractBlockchain;
 use nimiq_blockchain_proxy::BlockchainProxy;
+use nimiq_hash::Blake2bHash;
 use nimiq_keys::Address;
 use nimiq_network_interface::{
     network::{CloseReason, Network},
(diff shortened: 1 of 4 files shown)


consensus/src/consensus/remote_data_store.rs
```diff
@@ -86,35 +145,31 @@ impl<N: Network> RemoteDataStore<N> {
...
             match response {
                 Ok(Ok(response)) => {
-                    let blockchain = blockchain.read();
-                    // First try to obtain, from our chain store, the block that was used to generate the proof
-                    let block = blockchain.get_block(&response.block_hash, false).ok();
+                    // Get the block referenced by the proof, or discard the proof as it cannot be verified
+                    let Some(block) =
+                        Self::get_or_await_block(&blockchain, &response.block_hash).await
+                    else {
+                        continue;
```
jsdanielh, line 160:
```suggestion
                        // If the block for the proof couldn't be found, use another peer 
                        continue;
```

web-client/example/module.js
```diff
@@ -49,6 +49,11 @@ init().then(async () => {
         console.table(contacts);
     });
 
+    document.querySelector('#query-account').addEventListener("click", async () => {
```
jsdanielh: Can we use like a simple text/submit button instead that receives an `Address`?

consensus/src/consensus/remote_data_store.rs
```diff
@@ -48,6 +51,62 @@ enum RemoteDataStoreOps {
...
+            }
+            blockchain.notifier_as_stream()
+        };
+
+        // Wait for MAX_BLOCK_DRIFT events that add blocks, while checking if the added blocks contain the
+        // block hash in question
+        let item = blockchain_event_stream
+            .filter(|event| std::future::ready(!event.added_hashes().is_empty()))
+            .take(Self::MAX_BLOCK_DRIFT)
+            .filter(|event| std::future::ready(event.added_hashes().contains(&block_hash)))
```
paberr: Avoid calling `added_hashes` twice:
```suggestion
            .filter_map(async |event|  { let hashes = event.added_hashes(); if hashes.is_empty() { None } else { Some(hashes) }))
            .take(Self::MAX_BLOCK_DRIFT)
            .filter(|hashes| hashes.contains(block_hash)))
```

paberr:
Added a comment, but generally looks good.

jsdanielh (approved):
LGTM

jsdanielh:
Signing commit

sisou:
~~@nibhar turns out there was already this while loop in the client. Why wasn't it working, and can it be removed?~~
https://github.com/nimiq/core-rs-albatross/blob/c801676208b640d53d8b8f83c5d9a8f2a6be306b/web-client/src/client/lib.rs#L1177-L1195

sisou:
Ah, stupid me. That while loop is for the tx listener, not for account proofs 👍

PR #3163, opened 2024-11-27
Do the locking around `Validator` state instead of on each member

hrxi:
This makes it easier to add new members, also, the individual locking did not really make sense because these shouldn't go out of sync.

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -103,7 +103,10 @@ async fn main_inner() -> Result<(), Error> {
     // Start validator
     let val_metric_monitor = tokio_metrics::TaskMonitor::new();
     if let Some(validator) = client.take_validator() {
-        info!("Initializing validator {}", validator.validator_address());
+        info!(
+            "Initializing validator {}",
+            validator.state().read().validator_address
+        );
 
         if metrics_enabled {
             let mp_metrics_monitor = validator.get_mempool_monitor();
--- a/lib/src/client.rs
+++ b/lib/src/client.rs
@@ -39,7 +39,7 @@ use nimiq_validator::key_utils::VotingKeys;
 #[cfg(feature = "validator")]
 use nimiq_validator::validator::Validator as AbstractValidator;
 #[cfg(feature = "validator")]
-use nimiq_validator::validator::ValidatorProxy as AbstractValidatorProxy;
+use nimiq_validator::validator::ValidatorState as AbstractValidatorState;
 #[cfg(feature = "validator")]
 use nimiq_validator_network::network_impl::ValidatorNetworkImpl;
 #[cfg(feature = "wallet")]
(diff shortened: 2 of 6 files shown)


client/src/main.rs
```diff
@@ -101,7 +101,10 @@ async fn main_inner() -> Result<(), Error> {
     // Start validator
     let val_metric_monitor = tokio_metrics::TaskMonitor::new();
     if let Some(validator) = client.take_validator() {
-        info!("Initializing validator {}", validator.validator_address());
+        info!(
+            "Initializing validator {}",
+            validator.state().read().validator_address
```
hrxi, line 108:
```suggestion
            validator.state().validator_address()
```
This `read()` is maybe not so nice, should I encapsulate that?

sisou:
Can you add an explanation of _why_ to the PR?

client/src/main.rs
nibhar: I do not mind the `.read()` here all that much, so from my end a change would not be necessary.

validator/src/validator.rs
```diff
@@ -105,22 +87,15 @@ where
...
     network_event_rx: SubscribeEvents<<TValidatorNetwork::NetworkType as Network>::PeerId>,
     fork_event_rx: BroadcastStream<ForkEvent>,
 
-    slot_band: Arc<RwLock<Option<u16>>>,
-    consensus_state: Arc<RwLock<ConsensusState>>,
-    validator_state: Option<InactivityState>,
-    automatic_reactivate: Arc<AtomicBool>,
-
     macro_producer: Option<ProduceMacroBlock<TValidatorNetwork>>,
     macro_state: Arc<RwLock<Option<MacroState>>>,
```
nibhar, line 100: This could be added to the `ValidatorState`. Additionally that would open up a Rpc call to see the macro state which could be a very useful debugging tool.

nibhar (approved):
LGTM, just added a wish list item.

lib/src/client.rs
```diff
@@ -768,7 +768,7 @@ impl Client {
 
     #[cfg(feature = "validator")]
     /// Returns a reference to the *Validator proxy*.
-    pub fn validator_proxy(&self) -> Option<ValidatorProxy> {
+    pub fn validator_proxy(&self) -> Option<Arc<RwLock<ValidatorState>>> {
```
Eligioo: Update Rustdoc and change fn name to `validator_state`?

rpc-server/src/dispatchers/validator.rs
```diff
@@ -31,18 +32,19 @@ impl ValidatorInterface for ValidatorDispatcher {
     type Error = Error;
 
     async fn get_address(&mut self) -> RPCResult<Address, (), Self::Error> {
-        Ok(self.validator.validator_address.read().clone().into())
+        Ok(self.validator.read().validator_address.clone().into())
     }
 
+    // TODO: why do we give out secret keys via RPC?
```
Eligioo, line 38: Action this?

hrxi: Added an issue: #3270.

Theoretically a breaking change so I didn't want to do it here.

lib/src/client.rs
hrxi: Done.

PR #3165, opened 2024-11-28
Fix clippy warning, clean up `nimiq-time` crate and use it everywhere

hrxi:
- Make sure to only export API supported on all platforms.
- Make sure not to leak implementation details.
- All return values are named types now.
- Also make sure we continue to use `nimiq-time` everywhere.


jsdanielh (approved):
LGTM

PR #3169, opened 2024-11-29
Changing basic transaction data input from hexcode to plain string

ii-cruz:
Requested by frontend. Changes data input from hexcode to plain string for the rpc method basic transaction with data. These are the only functions with the expectation of a hex string as input.

## Pull request checklist


ii-cruz:
we don't actually want this.

PR #3174, opened 2024-12-02
Store decompressed BlsPublicKeys in IndexedDB

fiaxh:


sisou:
Without understanding/thoroughly reading the code: is this Indexeddb cache optional? Will it fail without crashing the client?

Indexeddb might not be available in all browser contexts nor in NodeJS, so this needs to be purely optional and not affect the client if not available or unwritable/unreadable for any other reason.

web-client/Cargo.toml
```diff
@@ -22,13 +22,16 @@ workspace = true
...
 
 [dependencies]
+ark-serialize = "0.4"
 futures = { workspace = true }
 gloo-timers = { version = "0.3", features = ["futures"] }
 hex = "0.4"
 js-sys = "0.3"
 log = { workspace = true }
 rand_core = "0.6.4"
+idb = "0.6.4"
```
hrxi: Alphabetical.

hrxi:
> Without understanding/thoroughly reading the code: is this Indexeddb cache optional? Will it fail without crashing the client?

Yes, it's in an `Option` and just does nothing when it's not available.

web-client/src/client/bls_cache.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// Add the given keys into IndexedDB
+    pub async fn add_keys(&self, keys: Vec<LazyPublicKey>) -> Result<(), Error> {
+        if let Some(db) = &self.db {
+            let transaction = db.transaction(&["bls_keys"], TransactionMode::ReadWrite)?;
+            let bls_keys_store = transaction.object_store("bls_keys")?;
+
+            for key in keys {
+                let mut result = Vec::new();
+                key.uncompress()
+                    .unwrap()
```
hrxi: Is it the function's contract that you may only pass valid keys? If so, we should specify that in the documentation. If not, we should not `unwrap` here, but skip the iteration instead.

web-client/src/client/bls_cache.rs
```diff
@@ -0,0 +1,89 @@
...
+
+    /// Add the given keys into IndexedDB
+    pub async fn add_keys(&self, keys: Vec<LazyPublicKey>) -> Result<(), Error> {
+        if let Some(db) = &self.db {
+            let transaction = db.transaction(&["bls_keys"], TransactionMode::ReadWrite)?;
+            let bls_keys_store = transaction.object_store("bls_keys")?;
+
+            for key in keys {
+                let mut result = Vec::new();
+                key.uncompress()
```
hrxi: This is a potentially expensive operation. We should probably tell the caller that the keys should have been uncompressed before already.

web-client/src/client/bls_cache.rs
```diff
@@ -0,0 +1,89 @@
...
+                    .unwrap()
+                    .public_key
+                    .serialize_with_mode(&mut result, Compress::No)
+                    .unwrap();
+                let public_key = hex::encode(&result);
+                let compressed_key = hex::encode(key.compressed().serialize_to_vec());
```
hrxi: Can the data also be stored without the hex encoding, i.e. directly as bytes in the IndexedDB?

web-client/src/client/lib.rs
```diff
@@ -1009,34 +1021,23 @@ impl Client {
         let mut blockchain_events = blockchain.read().notifier_as_stream();
 
         let block_listeners = Rc::clone(&self.head_changed_listeners);
+        let bls_cache = Rc::clone(&self.bls_cache);
 
         spawn_local(async move {
             loop {
-                let (hash, reason, reverted_blocks, adopted_blocks) =
+                let (hash, reason, reverted_blocks, adopted_blocks_hashs) =
```
hrxi:
```suggestion
                let (hash, reason, reverted_blocks, adopted_block_hashes) =
```

web-client/src/client/lib.rs
```diff
@@ -1059,23 +1060,45 @@ impl Client {
...
+                for hash in &adopted_blocks_hashs {
+                    adopted_blocks.push(&hash.to_hex().into());
+                }
+
                 let args = Array::new();
                 args.push(&hash.to_hex().into());
                 args.push(&reason.into());
                 args.push(&reverted_blocks);
-                args.push(&adopted_blocks);
+                args.push(&adopted_blocks.clone());
```
hrxi:
```suggestion
                args.push(&adopted_blocks);
```

web-client/src/client/bls_cache.rs
```diff
@@ -0,0 +1,89 @@
...
+/// Caches decompressed BlsPublicKeys in an IndexedDB
+pub(crate) struct BlsCache {
+    db: Option<Database>,
+    keys: Vec<LazyPublicKey>,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+struct BlsKeyEntry {
+    compressed_key: String,
+    public_key: String,
```
hrxi: We talked about adding a "last used" field so we can apply some sort of cache eviction in the future.

hrxi (commented):
There's still [an error in the web client compilation](https://github.com/nimiq/core-rs-albatross/actions/runs/12126731599/job/33809448345):
```
Generating main WASM for target nodejs...
error: 

it looks like the Rust project used to create this Wasm file was linked against
version of wasm-bindgen that uses a different bindgen format than this binary:

  rust Wasm file schema version: 0.2.95
     this binary schema version: 0.2.97

Currently the bindgen format is unstable enough that these two schema versions
must exactly match. You can accomplish this by either updating this binary or
the wasm-bindgen dependency in the Rust project.

You should be able to update the wasm-bindgen dependency with:

    cargo update -p wasm-bindgen --precise 0.2.97

don't forget to recompile your Wasm file! Alternatively, you can update the
binary with:

    cargo install -f wasm-bindgen-cli --version 0.2.95

if this warning fails to go away though and you're not sure what to do feel free
to open an issue at https://github.com/rustwasm/wasm-bindgen/issues!
```

sisou:
The wasm-bindgen version missmatch will go away when rebasing on top of `albatross`, as that dependency was updated recently to match.

hrxi:
Rebased on top of albatross.

web-client/Cargo.toml
hrxi: Done.

web-client/src/client/bls_cache.rs
hrxi: Specified.

web-client/src/client/bls_cache.rs
hrxi: Told the caller.

web-client/src/client/bls_cache.rs
hrxi: Yes. Done.

bls/src/types/public_key.rs
```diff
@@ -19,6 +19,8 @@ pub struct PublicKey {
 }
 
 impl PublicKey {
+    pub const TRUSTED_SERIALIZATION_SIZE: usize = 570;
```
nibhar, line 25: Is this the same as `PUBLIC_KEY_SIZE`? If so I would prefer to name this `SIZE`, similar to `CompressedPublicKey::SIZE`.

```suggestion
    /// The size of a serialized PublicKey in bytes.
    pub const SIZE: usize = 570;
```

bls/src/types/public_key.rs
```diff
@@ -60,8 +62,9 @@ impl PublicKey {
     /// and one bit indicating if it is the "point-at-infinity".
     pub fn compress(&self) -> CompressedPublicKey {
         let mut buffer = [0u8; CompressedPublicKey::SIZE];
-        CanonicalSerialize::serialize_compressed(&self.public_key.into_affine(), &mut buffer[..])
-            .unwrap();
+        let affine = self.public_key.into_affine();
+        assert_eq!(affine.compressed_size(), buffer.len());
+        affine.serialize_compressed(&mut buffer[..]).unwrap();
```
nibhar, line 70: More an out of interest question:
When does this return `Err(_)`? I realize the `.unwrap()` was there before, but obviously you took care of non fitting buffers. Is this now no longer failing for any reason?

bls/src/types/public_key.rs
```diff
@@ -72,6 +75,24 @@ impl PublicKey {
         let public_key = self.public_key.mul_bigint([x as u64]);
         PublicKey { public_key }
     }
+
+    pub fn trusted_serialize(&self) -> [u8; PublicKey::TRUSTED_SERIALIZATION_SIZE] {
```
nibhar, line 88: I would prefer to see a rustdoc, explaining what makes this trusted and how it can fail.

bls/src/types/public_key.rs
```diff
@@ -72,6 +75,24 @@ impl PublicKey {
...
+        let mut result = [0u8; PublicKey::TRUSTED_SERIALIZATION_SIZE];
+        assert_eq!(self.public_key.uncompressed_size(), result.len());
+        self.public_key
+            .serialize_uncompressed(&mut result[..])
+            .unwrap();
+        result
+    }
+
+    pub fn trusted_deserialize(
```
nibhar, line 105: I would prefer to see a rustdoc, explaining what makes this trusted and how it can fail.

genesis-builder/src/lib.rs
```diff
@@ -658,25 +656,34 @@ impl GenesisBuilder {
...
+        }) = &block
+        {
+            let mut decompressed = Vec::new();
+            for key in validators.voting_keys() {
+                decompressed.extend_from_slice(&key.trusted_serialize());
+            }
+            fs::write(decompressed_path, decompressed)?;
+        } else {
+            unreachable!("genesis block must be an election block");
+        }
```
nibhar: You may be able to use `block.validators().expect()` here as that will give you the `Validators` as well and is a bit shorter and potentially better readable.

nibhar (changes requested):
Generally looks good to me. Left some minor comments only. It however does still have 2 clippy warnings for a `RefCell` which is held across an `.await`. Those should be reseolved.

I would also like to see some more rustdoc added, where it is missing.

bls/src/types/public_key.rs
hrxi: The error type is [`ark_serialize::SerializationError`](https://docs.rs/ark-serialize/0.5.0/ark_serialize/enum.SerializationError.html):
- `NotEnoughSpace` — not applicable, we give a correctly sized buffer (also verified by the assert).
- `InvalidData` — not sure how the incoming data can be incorrect, but we'd like to know about it.
- `UnexpectedFlags` — programming error, should panic.
- `IoError` — can't happen, we don't serialize with IO, should be unreachable.

bls/src/types/public_key.rs
hrxi: Added some docs. No, it's specific to the trusted serialization, and should generally not be used, unless you know that you can trust the place where you got the serialization from. The normal public key serialization has 285 bytes.

bls/src/types/public_key.rs
hrxi: Done. Cannot fail.

bls/src/types/public_key.rs
hrxi: Done. Cannot fail (unless you give it untrusted data).

genesis-builder/src/lib.rs
hrxi: Done.

hrxi:
> Generally looks good to me. Left some minor comments only. It however does still have 2 clippy warnings for a `RefCell` which is held across an `.await`. Those should be reseolved.

Done.

bls/src/types/compressed_public_key.rs
```diff
@@ -35,6 +35,7 @@ impl CompressedPublicKey {
 
     /// Transforms the compressed form back into the projective form.
     pub fn uncompress(&self) -> Result<PublicKey, Error> {
+        log::info!("decompressing {}", &self.to_hex()[..16]);
```
Eligioo: Tracing style logging, should it be `INFO`?

web-client/src/client/bls_cache.rs
```diff
@@ -0,0 +1,97 @@
...
+            .version(1)
+            .add_object_store(
+                ObjectStore::builder(BLS_KEYS).key_path(Some(KeyPath::new_single(PUBLIC_KEY))),
+            )
+            .build()
+            .await
+        {
+            Ok(db) => Some(db),
+            Err(err) => {
+                log::warn!("idb: Couldn't create database {}", err);
```
Eligioo: Tracing style log

web-client/src/client/lib.rs
```diff
@@ -1106,6 +1112,23 @@ impl Client {
...
+                            let bls_keys = validators
+                                .validators
+                                .iter()
+                                .map(|validator| validator.voting_key.clone())
+                                .collect::<Vec<LazyPublicKey>>();
+                            if let Err(err) = bls_cache.add_keys(bls_keys).await {
+                                log::warn!("Failed caching bls keys {}", err);
+                            }
+                        }
+                    }
```
Eligioo:
```suggestion
                  let Ok(Block::Macro(macro_block)) = blockchain.read().get_block(hash, false)
                    else {
                        continue;
                    };

                    let Some(validators) = macro_block.header.validators else {
                        continue;
                    };

                    let bls_keys = validators
                        .validators
                        .iter()
                        .map(|validator| validator.voting_key.clone())
                        .collect::<Vec<LazyPublicKey>>();
                    if let Err(err) = bls_cache.add_keys(bls_keys).await {
                        log::warn!("Failed caching bls keys {}", err);
                    }
```

genesis/src/networks.rs
```diff
@@ -81,16 +83,40 @@ fn read_genesis_config(config: &Path) -> Result<GenesisData, GenesisBuilderError
...
+        let keys = match network_id {
+            NetworkId::DevAlbatross => &KEYS_DEV,
+            NetworkId::TestAlbatross => &KEYS_TEST,
+            NetworkId::UnitAlbatross => &KEYS_UNIT,
+            NetworkId::MainAlbatross => &KEYS_MAIN,
+            _ => unreachable!(),
+        };
+        keys.get_or_init(|| {
+            info.genesis
+                .decompressed_keys
```
Eligioo, line 111: What would happen if you provide a genesis config without the `decompressed_keys` section? This object is added later on

bls/src/types/compressed_public_key.rs
hrxi: Done. I think it should be `INFO` or louder since it is a ~100 ms action and it's interesting to the person viewing the logs that an expensive action happened.

web-client/src/client/bls_cache.rs
hrxi: Done.

web-client/src/client/lib.rs
hrxi: Done.

genesis/src/networks.rs
hrxi: In this case, the keys in the genesis block would be decompressed at startup time. This `decompressed_keys` array would be empty and no decompressed keys would be extracted.

nibhar (approved):
LGTM

hrxi:
Rebased for merging.

PR #3177, opened 2024-12-04
Correct web-client vesting contract data parsing

sisou:
Now takes the transaction's actual recipient data, instead of the internal representation of a vesting contract's `CreationTransactionData`. The internal representation uses `varints` to serialize the contract's `startTime` and `timeStep`, but the on-chain transaction recipient data doesn't. Since the transaction recipient data also has shorter formats that don't include the contract's `totalAmount`, this field is required now to be passed to the parsing method.

## Pull request checklist


jsdanielh (approved):
LGTM

PR #3181, opened 2024-12-05
Fix skip block test

viquezclaudio:
- Modify skip block test to actually produce a skip block

...
## Pull request checklist

### DIFF
--- a/validator/tests/mock.rs
+++ b/validator/tests/mock.rs
@@ -139,28 +139,20 @@ async fn four_validators_can_create_micro_blocks() {
 }
 
 #[test(tokio::test)]
-async fn four_validators_can_do_skip_block() {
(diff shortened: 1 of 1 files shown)


validator/tests/mock.rs
```diff
@@ -174,10 +174,17 @@ async fn four_validators_can_do_skip_block() {
         spawn(validator);
     }
 
-    // Wait for the new block producer to create a block.
+    // Wait for the new block producer to create a micro block.
```
jsdanielh: I don't get the change. I guess that next validator is popped when calling `pop_validator_for_slot` and such validator is getting paused when calling `tokio::time::pause` so the rest should produce a skip block right after. Or am I missing something?

jsdanielh (approved):
LGTM

PR #3186, opened 2024-12-09
RPC Server: `pushTransaction` and `pushHighPriorityTransaction` methods broadcast the transaction

### LINKED ISSUE #3131: Transactions broadcasted with `pushTransaction` are not included in the chain
## Description
While implementing transaction broadcasting to the chain, I encountered an issue where transactions sent via the pushTransaction RPC method were not included in the chain. Although the response includes the transaction hash, this hash cannot be found in the explorer.

## Sample data

RPC request with `pushTransaction` method:
```
{"id":0,"method":"pushTransaction","params":["0000d40e2da392a93f524b6c60f9c0879702c80a9662ede05453c7398fdb0aa25df953b859d551a98166eec8f4464d11bd92dd5be4140000000000000001000000000000012c003890a5185e9424ae16f0c4be1388745313dbaa7f840bfaa891190e008529e1e99965586c94a0469f3b0d82d2cc063a96f074bffb489429c896dc6b1d72cb29fbb88ae708"],"jsonrpc":"2.0"}
```

RPC request with `sendRawTransaction` method:
```
{"id":0,"method":"sendRawTransaction","params":["0000d40e2da392a93f524b6c60f9c0879702c80a9662ede05453c7398fdb0aa25df953b859d551a98166eec8f4464d11bd92dd5be4140000000000000001000000000000012c003890a5185e9424ae16f0c4be1388745313dbaa7f840bfaa891190e008529e1e99965586c94a0469f3b0d82d2cc063a96f074bffb489429c896dc6b1d72cb29fbb88ae708"],"jsonrpc":"2.0"}
```

## Workaround
Since the pushTransaction method performs a validity check on the transaction, I will continue using it. However, upon receiving a successful response with the transaction hash, I will follow up with an additional RPC request using the sendRawTransaction method with the same data. This ensures the transaction is included in the chain.

## Assumptions
Both RPC requests will generate and return the same transaction hash.

Eligioo:
By only pushing the transaction into the local mempool, the transaction never gets adopted if the node itself isn't an elected block producer.

## Pull request checklist

### DIFF
--- a/rpc-interface/src/mempool.rs
+++ b/rpc-interface/src/mempool.rs
@@ -9,11 +9,11 @@ use crate::types::{HashOrTx, MempoolInfo, RPCResult};
 pub trait MempoolInterface {
     type Error;
 
-    /// Pushes a raw transaction into the mempool, it will be assigned a default priority.
+    /// Pushes a raw transaction with a default priority assigned into the mempool and broadcast it to the network.
     async fn push_transaction(&mut self, raw_tx: String)
         -> RPCResult<Blake2bHash, (), Self::Error>;
 
-    /// Pushes a raw transaction into the mempool with high priority.
+    /// Pushes a raw transaction with a high priority assigned into the mempool and broadcast it to the network.
     async fn push_high_priority_transaction(
         &mut self,
         raw_tx: String,
--- a/rpc-server/src/dispatchers/mempool.rs
+++ b/rpc-server/src/dispatchers/mempool.rs
@@ -1,8 +1,10 @@
 use std::sync::Arc;
 
 use async_trait::async_trait;
+use nimiq_consensus::ConsensusProxy;
 use nimiq_hash::{Blake2bHash, Hash};
 use nimiq_mempool::{mempool::Mempool, mempool_transactions::TxPriority};
+use nimiq_network_libp2p::Network;
 use nimiq_rpc_interface::{
     mempool::MempoolInterface,
     types::{HashOrTx, MempoolInfo, RPCResult},
@@ -14,12 +16,13 @@ use crate::error::Error;
 
 #[allow(dead_code)]
 pub struct MempoolDispatcher {
+    consensus: ConsensusProxy<Network>,
     mempool: Arc<Mempool>,
 }
 
 impl MempoolDispatcher {
-    pub fn new(mempool: Arc<Mempool>) -> Self {
-        MempoolDispatcher { mempool }
+    pub fn new(consensus: ConsensusProxy<Network>, mempool: Arc<Mempool>) -> Self {
+        MempoolDispatcher { consensus, mempool }
     }
 }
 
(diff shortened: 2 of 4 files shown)


rpc-server/src/dispatchers/mempool.rs
```diff
@@ -48,9 +51,9 @@ impl MempoolInterface for MempoolDispatcher {
         let tx = Transaction::deserialize_from_vec(&hex::decode(&raw_tx)?)?;
         let txid = tx.hash::<Blake2bHash>();
 
-        match self.mempool.add_transaction(tx, Some(TxPriority::High)) {
+        match self.consensus.send_transaction(tx).await {
```
jsdanielh: Shouldn't it be both, add to own mempool and send the transaction over the network?

rpc-server/src/dispatchers/mempool.rs
```diff
@@ -35,9 +38,9 @@ impl MempoolInterface for MempoolDispatcher {
         let tx = Transaction::deserialize_from_vec(&hex::decode(&raw_tx)?)?;
         let txid = tx.hash::<Blake2bHash>();
 
-        match self.mempool.add_transaction(tx, None) {
+        match self.consensus.send_transaction(tx).await {
```
jsdanielh: Shouldn't it be both, add to own mempool and send the transaction over the network?

rpc-interface/src/mempool.rs
```diff
@@ -9,11 +9,11 @@ use crate::types::{HashOrTx, MempoolInfo, RPCResult};
 pub trait MempoolInterface {
     type Error;
 
-    /// Pushes a raw transaction into the mempool, it will be assigned a default priority.
+    /// Pushes a raw transaction with a default priority assigned and broadcast it to the network.
     async fn push_transaction(&mut self, raw_tx: String)
         -> RPCResult<Blake2bHash, (), Self::Error>;
 
-    /// Pushes a raw transaction into the mempool with high priority.
+    /// Pushes a raw transaction with a high priority assigned and broadcast it to the network.
```
jsdanielh: But with the current implementation it is only broadcasted to the network

rpc-interface/src/mempool.rs
```diff
@@ -9,11 +9,11 @@ use crate::types::{HashOrTx, MempoolInfo, RPCResult};
 pub trait MempoolInterface {
     type Error;
 
-    /// Pushes a raw transaction into the mempool, it will be assigned a default priority.
+    /// Pushes a raw transaction with a default priority assigned and broadcast it to the network.
```
jsdanielh: But with the current implementation it is only broadcasted to the network

sisou:
I'd rather this get fixed with mempool syncing (#2832).

Eligioo:
> I'd rather this get fixed with mempool syncing (#2832).

The goal of PR #2832 is for nodes that just reached consensus get an idea of what the current mempool landscape is it in that moment in time and try to acquire as much of that state. It's not a mechanism to keep mempools in sync in between nodes forever as the mempool in general is an unsynced component. Therefore that PR isn't going to fix #3131.

As stated in the PR description, the Mempool Syncer terminates itself after a fixed amount of time and from there on the node relies on broadcasts done over the network in order to acquire new transactions.

sisou:
Then I have the same feedback as JD: The `push*` methods should push to local mempool first, and then **only if it succeeded**, broadcast to other peers, too. Because the `push*` methods are the only way to see if a transaction is currently valid - just broadcasting does not provide _any_ feedback on tx validity.

Eligioo:
@sisou yes I realised the distinction later on too. Basically now in this PR sendRawTransaction and push* have become exactly the same. Will update it.

jsdanielh (approved):
LGTM

PR #3187, opened 2024-12-10
Transaction validity and to-plain conversion for PoW transactions

sisou:
Ensure PoW transactions can be properly converted to their plain representation, which is primarily done during address transaction history fetching in the web-client. Before this change, PoW transactions could not be returned and the process panicked, due to different serializations of contract creation data, HTLC proofs and signature proofs.

I have tried to document all changes and differences in code comments, let me know if you require more info, here or in the code.

## Pull request checklist

- [x] All tests pass. The project builds and runs.
- [x] I have resolved any merge conflicts.
- [x] I have resolved all `clippy` and `rustfmt` warnings.

PR #3192, opened 2024-12-12
web-client: Fix proofs being always set to empty string in Transaction.fromPlain

sisou:
#3134 (released in v1.0.2) forced plain transaction proof deserialization to always deserialize a `PlainTransactionProof::Raw` variant to be compatible with PoW plain transactions, which didn't have a `proof.type` field.

Unfortunately I forgot to remove the empty string default for `PlainTransactionProof::Raw` when I renamed it from `PlainTransactionProof::Empty`.

## Pull request checklist

### DIFF
--- a/web-client/extras/tests/Transaction.test.ts
+++ b/web-client/extras/tests/Transaction.test.ts
@@ -88,7 +88,13 @@ describe('Transaction', () => {
             },
         };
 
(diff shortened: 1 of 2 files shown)


jsdanielh (commented):
LGTM. We just need to wait for #3187 to be merged.

jsdanielh (approved):
LGTM

PR #3196, opened 2024-12-13
Changing add stake operation

ii-cruz:
- Adds the Minimum stake to the `add_stake` operation. Add stake transactions have an intrinsic check for their `value => min_stake` and thus are rejected before they can break the invariant.
> - The idea is that this minimum prevents indefinitely small value add_stake transactions from hindering the retirement. 
- `add_stake` assumes that the staker wants to have auto-staking as long as there is an active stake. Once no active stake exists, the funds are credited directly to the inactive balance.
> - The idea is to prevent DoS of the redelegation of stake.
- Adds more unit tests for legacy add stake and the new add stake policy
- Creates the protocol version gate for the add_stake and intrinsic tx verification

This PR needs #3800 merged first

## Pull request checklist


primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -283,41 +285,65 @@ impl StakingContract {
...
+        // Create the receipt.
+        let receipt = AddStakeReceipt {
+            credited_balance: credited_balance.clone(),
+        };
+
+        if credited_balance == BalanceType::Active {
+            // If we are delegating to a validator, we need to update it.
+            if let Some(validator_address) = &staker.delegation {
+                // Check that the delegation is still valid, i.e. the validator hasn't been deleted.
+                store.expect_validator(validator_address)?;
```
paberr: There might be a problem with the ordering. We usually want to do changes only after all potentially failing conditions are checked.
In practice, I don't think it is a problem, as all previous changes are in the staker and that one needs to be put explicitly back.

paberr (approved):
Looks good to me

sisou:
When `add_stake` adds to the inactive balance (when there's no active balance anymore and the inactive one is the biggest), I assume it _doesn't_ reset the `inactiveFrom` block number?

paberr:
It doesn't reset it.

Also, while discussing with @ii-cruz , one thing that is missing is a version gate for the upgrade.
The new behavior should only be active when we have a new version.

primitives/account/tests/staking_contract/add_stake_policy.rs
```diff
@@ -0,0 +1,906 @@
...
+        .unwrap();
+
+    assert_eq!(validator.num_stakers, 1);
+    assert_eq!(
+        validator.total_stake,
+        Coin::from_u64_unchecked(Policy::VALIDATOR_DEPOSIT)
+    );
+}
+
+/* Legacy Add Stake testing */
```
paberr, line 682: Very good. That would have been my next question: do we still test the old methods? ✅

primitives/transaction/src/account/staking_contract/structs.rs
```diff
@@ -176,10 +176,10 @@ impl IncomingStakingTransactionData {
...
             }
             IncomingStakingTransactionData::AddStake { .. } => {
-                // Adding stake should be at least greater than 0.
-                if transaction.value.is_zero() {
-                    warn!("Add stake transactions must have positive value. The offending transaction is the following:\n{:?}", transaction);
-                    return Err(TransactionError::ZeroValue);
+                // Adding stake should be greater than 0.
+                if transaction.value < Coin::from_u64_unchecked(Policy::MINIMUM_STAKE) {
+                    warn!("Add stake must increment stake by at least minimum stake. The offending transaction is the following:\n{:?}", transaction);
+                    return Err(TransactionError::InvalidValue);
```
paberr, line 193: Is this not a problem for older transactions?

paberr (commented):
Generally looks good to me

primitives/transaction/src/account/staking_contract/structs.rs
sisou: Yeah. This needs to still allow syncing txs that were valid before the upgrade.

ii-cruz: Correct, I'll fix it and push the changes
Thanks!

primitives/account/src/account/staking_contract/staker.rs
ii-cruz: It was a problem, yes. ~~We either allow adding stake when there is a delegation to a tombstone, and then I remove this expect, or we don't allow it and the expect goes to the beginning. WDYT?~~

edit: I decided to forbid adding stake to a tombstone and added the expect and added a test for this scenario.

paberr (approved):
Looks good to me now.

primitives/account/src/account/staking_contract/traits.rs
```diff
@@ -169,8 +169,14 @@ impl AccountTransactionInteraction for StakingContract {
...
-                .add_stake(&mut store, &staker_address, transaction.value, tx_logger)
-                .map(|_| None),
+                .add_stake(
+                    &mut store,
+                    &staker_address,
+                    transaction.value,
+                    block_state.protocol_version,
+                    tx_logger,
+                )
+                .map(|receipt| Some(receipt.into())),
```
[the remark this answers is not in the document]
ii-cruz: i just changed the previous code to have a receipt for the add_stake. Since receipts only live until the next macroblock and then they are pruned, we don't need retroactive compatibility for this part.

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -276,75 +281,127 @@ impl StakingContract {
...
+        // because the intrinsic tx checks that value is >= min stake.
+        assert!(
+            Staker::enforce_min_stake(
+                staker.active_balance + value,
+                staker.inactive_balance,
+                staker.retired_balance,
+            )
+            .is_ok(),
+            "Add stake should never violate the min stake invariants"
+        );
```
paberr: I believe there might be a potential path to this assertion through history sync (where we don't re-run the intrinsic checks). Of course, that only would happen if some other conditions of history sync do not work as expected. Given that the graceful `enforce_min_stake(...)?` is not much more work, should we do that?

primitives/account/tests/staking_contract/staker_add_stake_policy.rs
```diff
@@ -138,13 +151,558 @@ fn add_stake(protocol_version: u16) {
...
+        .accounts
+        .data_store(&Policy::STAKING_CONTRACT_ADDRESS);
+    let mut db_txn = staker_setup.env.write_transaction();
+    let mut db_txn = (&mut db_txn).into();
+    let staker_keypair = ed25519_key_pair(STAKER_PRIVATE_KEY);
+
+    // -----------------------------------
+    // Test execution:
+    // -----------------------------------
+    // Add stake operation credits to retired stake.
```
paberr: Copy-paste error

```suggestion
    // Add stake operation credits to inactive stake.
```

primitives/account/tests/staking_contract/staker_add_stake_policy.rs
```diff
@@ -138,13 +151,558 @@ fn add_stake(protocol_version: u16) {
...
+        validator.total_stake,
+        Coin::from_u64_unchecked(Policy::VALIDATOR_DEPOSIT)
+    );
+}
+
+/* Legacy Add Stake testing */
+
+#[test]
+fn add_stake_legacy_works_before_v2() {
+    for v in 1..=upgrades::v3::STAKING_CHANGE_ADD_STAKE_POLICY {
```
paberr: This range is inclusive, so the last iteration (`v == STAKING_CHANGE_ADD_STAKE_POLICY`) runs the *new* code path under the legacy test name (and overlaps `add_stake_policy_works`). Nothing fails because the shared helper's assertions hold for both paths, but for a clean legacy/new split this should be exclusive:

```suggestion
    for v in 1..upgrades::v3::STAKING_CHANGE_ADD_STAKE_POLICY {
```

primitives/account/src/account/staking_contract/receipts.rs
```diff
@@ -111,6 +111,22 @@ pub struct DeleteValidatorReceipt {
...
 
+#[derive(Debug, PartialEq, Eq, Clone, Hash, Serialize, Deserialize)]
+pub enum BalanceType {
+    Active,
+    Inactive,
+    Retired,
+}
+
+/// Receipt for most staker-related transactions. This is necessary to be able to revert
+/// these transactions.
```
paberr: Doc comment is copied from `StakerReceipt`.

```suggestion
/// Receipt for add stake transactions. Stores which balance the stake was
/// credited to, which is necessary to be able to revert the transaction.
```

paberr (changes requested):
Some minor comments to be addressed. Overall looks good.

primitives/account/src/account/staking_contract/staker.rs
ii-cruz: Hum that's a fair point, it would make the code more robust. i'll do that

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -1015,4 +1071,60 @@ impl StakingContract {
...
+    // IMPORTANT: DO NOT REMOVE THIS CODE!
+    // It is needed for history nodes to sync.
+
+    /// Legacy add stake logic, it adds more Coins to a staker's active balance.
+    /// It will be directly added to the staker's balance.
+    /// Anyone can add stake for a staker. The staker must already exist.
+    /// This action is only possible if (a) the resulting non-retired funds respect
+    /// the invariant 1 - minimum stake for non-retired funds. This was not part of the
+    /// intrinsic add stake transaction checks on old versions, thus must be enforced at this level.
+    pub fn legacy_add_stake_before_v2(
```
viquezclaudio: I think it should be: 
legacy_add_stake_before_v3

ii-cruz: yes you are correct!

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -276,19 +281,33 @@ impl StakingContract {
...
         &mut self,
         store: &mut StakingContractStoreWrite,
         staker_address: &Address,
         value: Coin,
+        protocol_version: u16,
         tx_logger: &mut TransactionLog,
-    ) -> Result<(), AccountError> {
+    ) -> Result<AddStakeReceipt, AccountError> {
+        if protocol_version < upgrades::v3::STAKING_CHANGE_ADD_STAKE_POLICY {
+            log::warn!(%protocol_version, "Adding stake using old protocol version");
```
paberr: This will be very loud: history sync replays every historic add-stake through this branch, so we'd emit a `warn!` per transaction for what is completely expected behaviour. Can we drop this to `debug!`, or just remove it? Also the rest of the file uses the unqualified macros via the crate's `#[macro_use]`, so the `log::` prefix isn't needed here.

primitives/account/src/account/staking_contract/receipts.rs
```diff
@@ -111,6 +111,22 @@ pub struct DeleteValidatorReceipt {
 }
 convert_receipt!(DeleteValidatorReceipt);
 
+#[derive(Debug, PartialEq, Eq, Clone, Hash, Serialize, Deserialize)]
+pub enum BalanceType {
```
paberr: Two things on this enum:

1. There's no `rename_all`, so `creditedBalance` will serialize as `"Active"` / `"Inactive"` / `"Retired"`, whereas `Log` uses kebab-case tags with camelCase fields and the comparable `AccountType` carries `serde(rename_all = "lowercase")`. `Log::Stake` is part of the RPC surface via `rpc-interface/src/types.rs`, so once this is released changing the casing breaks consumers. Should we add `#[serde(rename_all = "lowercase")]` here or so? I think that only affects the human-readable format and not the receipt's binary encoding, but please double-check that.
2. `Retired` can never be produced: neither `add_stake` nor `legacy_add_stake_before_v3` credits the retired balance, so the `BalanceType::Retired` arm in `revert_add_stake` is unreachable.

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -276,19 +281,33 @@ impl StakingContract {
...
 
-        // Fail if the minimum stake would be violated for the non-retired funds (invariant 1).
+        // Check that the delegation is still valid, i.e. the validator hasn't been deleted.
+        if let Some(validator_address) = &staker.delegation {
+            store.expect_validator(validator_address)?;
+        }
+
+        // Intrinsic checks normally ensure that add-stake transactions preserve the minimum non-retired stake
+        // (invariant 1). History sync replays transactions without re-running those checks, so we enforce
+        // the invariant here too as defense in depth.
```
paberr, line 307: we pass `staker.active_balance + value` here (line below, cannot add the comment in the correct line) even when the value ends up on the inactive balance. It's equivalent because invariant 1 only looks at the sum, but it reads as if we were only checking the active path.

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -26,8 +28,10 @@ use crate::{
 ///
 /// Actions concerning a staker are:
 /// 1. Create:           Creates a staker.
-/// 2. AddStake:         Adds coins from any outside address to the staker's active balance.
-///                      This action is only possible if:
+/// 2. AddStake:         Adds coins from any outside address to the staker.
+///                      The balance to accredit these funds is preferentially the active balance. If there is no
+///                      positive active balance, it credits the inactive balance instead.
+///                      This operation has a minimum value of minimum stake, this it always respects the invariant:
```
paberr: typo: `this it always respects` -> `thus it always respects`

primitives/src/policy.rs
```diff
@@ -34,7 +34,7 @@ pub struct Policy {
     #[cfg_attr(feature = "ts-types", wasm_bindgen(skip))]
     pub genesis_block_number: u32,
     /// Maximum supported protocol version.
-    // This should only be increased when a new hard fork code is released.
+    // This should only be increased when a new hard fork code is released
```
paberr: this only drops the trailing period, and it now disagrees with the identical comment further down (l. 711, 724, 737). unrelated to this PR, maybe just revert it?

primitives/transaction/src/account/staking_contract/structs.rs
```diff
@@ -171,8 +174,8 @@ impl IncomingStakingTransactionData {
                 verify_transaction_signature(transaction, proof)?
             }
             IncomingStakingTransactionData::CreateStaker { proof, .. } => {
-                // Check that stake is at least minimum stake.
                 if transaction.value < Coin::from_u64_unchecked(Policy::MINIMUM_STAKE) {
+                    // Check that stake is at least minimum stake.
```
paberr: the comment moved inside the `if`, so it now sits with the failure handling rather than with the condition it describes. also unrelated to this PR, can we put it back above the `if`?

primitives/src/policy.rs
paberr: or fix the others

paberr (commented):
some minor comments from my side

primitives/account/src/account/staking_contract/staker.rs
ii-cruz: you are right, dropped it all together

primitives/account/src/account/staking_contract/staker.rs
ii-cruz: i see, i'll work on the comments on it a bit more

primitives/account/src/account/staking_contract/receipts.rs
ii-cruz: 1. great catch. i ended up just removing the enum all together since it can be a simple boolean
2. Removed the retired. Initially, we were thinking of crediting the retired if there was no inactive and non-retired balance, but we concluded it couldn't be safely done. This is a remaining piece of code from that version.

primitives/account/src/account/staking_contract/traits.rs
```diff
@@ -282,7 +288,15 @@ impl AccountTransactionInteraction for StakingContract {
                 self.revert_create_staker(&mut store, &staker_address, transaction.value, tx_logger)
             }
             IncomingStakingTransactionData::AddStake { staker_address } => {
-                self.revert_add_stake(&mut store, &staker_address, transaction.value, tx_logger)
+                let receipt = receipt.ok_or(AccountError::InvalidReceipt)?.try_into()?;
```
paberr: **Blocking:** I think this breaks reverts across a binary upgrade. `AddStake` used to produce no receipt, and we now require one unconditionally, including for pre-v3 blocks, since `legacy_add_stake_before_v3` returns a receipt too.

1. A node applies a block containing an `AddStake` tx with the current release and stores the revert info in `RevertTable` with `None` for that transaction.
2. The operator upgrades to the v3 binary, which will usually happen well before the fork actually activates.
3. Something triggers a rebranch over that block. We hit `receipt.ok_or(AccountError::InvalidReceipt)?` here, and `revert_block` turns every error into `panic!("Failed to revert {block} - {e:?}")`.

Possible fix: keep the legacy path receipt-less, so `add_stake` returns `Option<AddStakeReceipt>`, `legacy_add_stake_before_v3` returns `None`, and `revert_add_stake` treats `None` as "credited to the active balance". Then only post-fork blocks carry a receipt, and those are always produced by the new binary. (AI-found, I have not reproduced it against an actual rebranch.)

primitives/account/src/account/staking_contract/receipts.rs
```diff
@@ -111,6 +111,15 @@ pub struct DeleteValidatorReceipt {
 }
 convert_receipt!(DeleteValidatorReceipt);
 
+/// Receipt for add stake transactions. Stores which balance the stake was
+/// credited to, which is necessary to be able to revert the transaction.
+#[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
+pub struct AddStakeReceipt {
+    /// true if add stake credited to active balance, false if it was credited to inactive balance.
+    pub credited_active_balance: bool,
```
paberr: Taking a step back, do we even need this receipt? `value` is always > 0 because of the intrinsic check, so after applying we credited the active balance exactly iff `staker.active_balance` is non-zero: we only credit the inactive balance when the active balance is zero, and that leaves it at zero. So `revert_add_stake` could derive this with `!staker.active_balance.is_zero()` and we would not change the revert info format at all. I'm a bit torn, the explicit receipt is easier to follow, but it would sidestep my comment in `traits.rs`.

primitives/account/src/account/staking_contract/receipts.rs
```diff
@@ -111,6 +111,15 @@ pub struct DeleteValidatorReceipt {
 }
 convert_receipt!(DeleteValidatorReceipt);
 
+/// Receipt for add stake transactions. Stores which balance the stake was
+/// credited to, which is necessary to be able to revert the transaction.
+#[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
```
paberr: maybe derive `Copy` here too, like `PenalizeReceipt`? not essential.

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -297,54 +317,81 @@ impl StakingContract {
...
+            if staker.inactive_from.is_none() {
+                staker.inactive_from = Some(0); // To uphold the (invariant 3)
+            }
+        };
         self.balance += value;
 
-        // Build the return logs
+        // Create the receipt.
+        let receipt = AddStakeReceipt {
+            credited_active_balance: credited_active_balance.clone(),
```
paberr: `.clone()` on a `bool` — clippy's `clone_on_copy` should complain about this.

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -297,54 +317,81 @@ impl StakingContract {
...
+        // then it will attribute the funds to the inactive balance.
+        let credited_active_balance = !staker.active_balance.is_zero();
+        if credited_active_balance {
+            staker.active_balance += value;
+        } else {
+            staker.inactive_balance += value;
+            if staker.inactive_from.is_none() {
+                staker.inactive_from = Some(0); // To uphold the (invariant 3)
+            }
+        };
```
paberr: leftover `;` from when this was a `let` binding.

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -276,19 +280,35 @@ impl StakingContract {
...
+        // Check that the delegation is still valid, i.e. the validator hasn't been deleted.
+        if let Some(validator_address) = &staker.delegation {
+            store.expect_validator(validator_address)?;
+        }
+
+        // Intrinsic checks normally ensure that add-stake transactions preserve the minimum non-retired stake
+        // (invariant 1). History sync replays transactions without re-running those checks, so we enforce
+        // the invariant here too as defense in depth.
+        // Add stake may be credited to either the active or inactive balance below. Both minimum-stake
+        // invariants only use these balances through their sum, so adding `value` to the either active or inactive
```
paberr: (typo: "to the either active or inactive")

primitives/account/src/logs.rs
```diff
@@ -105,6 +105,7 @@ pub enum Log {
         staker_address: Address,
         validator_address: Option<Address>,
         value: Coin,
+        credited_active_balance: bool,
```
paberr: `credited_active_balance` reads like it holds an amount rather than a flag, shouldn't it be `credited_to_active` or so?

primitives/account/src/account/staking_contract/staker.rs
paberr: actually it does right below my comment 😆 .

paberr (changes requested):
Mostly minor things, mostly typos. Except one bigger item in the receipts.

primitives/account/src/account/staking_contract/receipts.rs
ii-cruz: meh, penalize is the only receipt deriving copy; all the others don't. I would actually just remove it from the Penalize receipt since it's not used anywhere

primitives/account/src/account/staking_contract/traits.rs
ii-cruz: Very true; it would be a problem in the window of time between the upgrade deployment on the validator and the next macro block. Afterwards, it would work as intended.

primitives/account/src/account/staking_contract/receipts.rs
ii-cruz: yep yep right. It's better to do this and avoid the creation of the receipt. I'll document the behaviour so that it's clear for future changes.

primitives/account/src/account/staking_contract/traits.rs
ii-cruz: I removed the receipt as we discussed in the comment on receipts.rs, so the problem is no longer applicable

sisou:
@ii-cruz this PR apparently does two things: 
1. set a minimum add-stake value the tx must transfer
2. Add the stake to inactive stake once active stake is 0

Do we actually need the first change, when the second change already prevents any DoS? The description says the first is to help retirement, the second change helps redelegation. But I don't understand the difference? Both require the inactive balance to have cleared its lock time, so for me they are the same in that regard?

My fear is that limiting add-stake to transfer at least 100 NIM, no matter what the staker balance already is, limits pool payouts and even users themselves.

ii-cruz:
@sisou The minimum number of add-stake operations prevents cheap dusting with the goal of a DoS on the retirement operation. To retire a stake, the staker must have all the stake inactive and released. After that cooldown, the retirement fails only if invariant 1 is violated (0 < non-retired stake (active + inactive) < Min stake). By only allowing add stake operations with at least min stake, nobody can DoS the retirement of the stake via the invariant violation path.

primitives/transaction/tests/staking_contract_verify.rs
```diff
@@ -742,6 +742,22 @@ fn stake() {
...
+    // Deposit too small (post upgrade restriction).
+    tx.value = Coin::from_u64_unchecked(Policy::MINIMUM_STAKE - 1);
+    for_each_protocol_version(|v| {
+        if v < upgrades::v3::STAKING_CHANGE_ADD_STAKE_POLICY {
+            assert_eq!(AccountType::verify_incoming_transaction(&tx, v), Ok(()));
+        } else {
+            assert_eq!(
+                AccountType::verify_incoming_transaction(
+                    &tx,
+                    upgrades::v3::STAKING_CHANGE_ADD_STAKE_POLICY
```
sisou: The else branch passes the constant instead of the loop variable, so versions 4 and 5 just re-test version 3.
```suggestion
            assert_eq!(AccountType::verify_incoming_transaction(&tx, v), Ok(()));
        } else {
            assert_eq!(
                AccountType::verify_incoming_transaction(&tx, v)
```

primitives/account/src/account/staking_contract/staker.rs
```diff
@@ -1015,4 +1063,58 @@ impl StakingContract {
...
+    // IMPORTANT: DO NOT REMOVE THIS CODE!
+    // It is needed for history nodes to sync.
+
+    /// Legacy add stake logic, it adds more Coins to a staker's active balance.
+    /// It will be directly added to the staker's balance.
+    /// Anyone can add stake for a staker. The staker must already exist.
+    /// This action is only possible if (a) the resulting non-retired funds respect
+    /// the invariant 1 - minimum stake for non-retired funds. This was not part of the
+    /// intrinsic add stake transaction checks on old versions, thus must be enforced at this level.
+    pub fn legacy_add_stake_before_v3(
```
sisou: Does this need to be `pub`?

primitives/src/policy/upgrades/v3.rs
```diff
@@ -9,3 +9,5 @@ pub const VERSION: u16 = 3;
 
 // --- Changes shipping in v3 -------------------------------------------------
 // Add one `pub const MY_CHANGE: u16 = VERSION;` line per change below.
+
+pub const STAKING_CHANGE_ADD_STAKE_POLICY: u16 = VERSION;
```
sisou, line 17: Please add a doc comment to give more context to this gate. See [`v2.rs`](https://github.com/nimiq/core-rs-albatross/blob/7cd241e512a8f635670352e6aced22e39446d122/primitives/src/policy/upgrades/v2.rs) for examples.

primitives/account/src/account/staking_contract/receipts.rs
```diff
@@ -11,7 +11,7 @@ use crate::{convert_receipt, AccountReceipt};
 
 /// Penalize receipt for the inherent. This is necessary to be able to revert
 /// these inherents.
-#[derive(Clone, Copy, Debug, Serialize, Deserialize, Eq, PartialEq)]
+#[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
```
sisou, line 14: Any reason to remove `Copy`?

sisou (changes requested):
Claude found a potential issue during the upgrade mempool re-validation, which apparently re-validates the cached txs async, which means an add-stake tx which adds less than the MINIMUM_STAKE can be in there, and will potentially only be evicted after the first micro block of the upgraded epoch has already passed. This can also lead to a chain split, between nodes that already evicted the txs vs nodes that have not (and produced the block).

In the last upgrade this was not an issue, because the rules _widened_. But now, they are _tightening_, so previously valid txs become invalid, not the other way around.

Here it is in Claude's own words:

## The minimum-stake floor is only enforced at intrinsic verification

`add_stake` (`staker.rs:311`) never checks `value >= Policy::MINIMUM_STAKE`. The comment above the `enforce_min_stake` call claims it re-establishes the intrinsic check as defense in depth, but it doesn't — `enforce_min_stake` only fails when the *sum* of non-retired balances drops below the minimum. A staker with `active = 100 * MINIMUM_STAKE` accepts an add-stake of 1 Luna there without complaint.

So the floor lives exclusively in `IncomingStakingTransactionData::verify`, and there are two paths that apply transactions without running it:

**1. `verify_transactions` skips the mempool-cached hash** (`blockchain/src/blockchain/verify.rs:117-127`):

```rust
if !self.tx_verification_cache.is_known(&transaction.hash()) {
    transaction.verify(self.network_id, version)?;
}
```

The cache is the mempool and it isn't keyed by protocol version. This PR adds the first version-gated rule that *tightens* validity (`v2::WARM_KEY_SIGNALING` only loosens, so this was unreachable before). Eviction is async — `MempoolTask` handles `BlockchainEvent::ProtocolUpgrade` → `resync_protocol_version` → `revalidate_transactions`. The upgrade lands on an election block and the *next* micro block can already carry a sub-minimum add-stake tx. Nodes still holding it in their mempool skip verification and accept; nodes that don't (already revalidated, freshly started, syncing) reject. Chain split.

**2. History sync never verifies transactions at all.** `push_history_sync` (`blockchain/src/blockchain/history_sync.rs:41`) rebuilds blocks from historic transactions and commits accounts directly.

These compose: a tx that slips in via (1) becomes canonical history, and (2) then replays it everywhere.

One line in the v3 branch closes both at the state-transition layer, where the rule belongs:

```rust
if value < Coin::from_u64_unchecked(Policy::MINIMUM_STAKE) {
    return Err(AccountError::InvalidCoinValue);
}
```

It also fixes a smaller thing: the `inactive_from` assignment at `staker.rs:327-329` isn't conditioned on `value` being non-zero, so a zero-value add-stake flips `inactive_from` to `Some(0)` while `inactive_balance` stays `0` — a permanent invariant-4 violation. Not reachable from a valid block today (the `ZeroValue` check isn't version-gated), but the guard is unconditional and `add_stake` is `pub`.

The cache issue is arguably a #3800 framework bug rather than this PR's, but this PR is what makes it live. Fix options: key the cache by `(hash, version)`, bypass it when `block.version() != mempool.protocol_version()`, or revalidate synchronously inside the push that carries the upgrade.

primitives/account/src/account/staking_contract/staker.rs
ii-cruz: nope, you are right, removed it

primitives/account/src/account/staking_contract/receipts.rs
ii-cruz: It's the only receipt deriving copy; it seems like it's not used anywhere thus I removed it.

ii-cruz:
@sisou, the findings you have are not correct in their entirety, but I added the defense-in-depth check of the min stake to the add stake tx since it doesn't hurt.
There is a potential race condition in the mempool where a tx to be added to the mempool can read the version, and then the blockchain lock can be taken for committing the upgrade election block. That can indeed lead to having a tx that is no longer valid, but that is out of scope for this PR. Regardless of that, the block would be rejected by the other upgraded validators, so the consequence wouldn't result in a chain split.
There's the other aspect of the chain upgrade where it's rather common to socially suspend tx submissions during the upgrade, making this unlikely to happen in the happy case.

PR #3202, opened 2024-12-17
Ensure transactions in past election blocks can be proven

sisou:
Before this change, transactions in election blocks where tried to be proven with the block number of the following election block, because `Policy::election_block_after()` returns the following election block if given an election block number.

## Pull request checklist


viquezclaudio (approved):

viquezclaudio:
Since election blocks do not contain regular transactions, you mean the problem was related to prove inherents/rewards in election blocks?

sisou:
Yes, but they are handled and logged like any other transaction as far as the web-client is concerned.

jsdanielh (approved):
LGTM

PR #3203, opened 2024-12-17
RPC Server: impl custom deserializer for ValidityStartHeight

Eligioo:

This makes the input accepts both a `string` and a `u32` whereas before only a `string` was accepted:
- "1000" (Absolute)
- "+1000" (Relative)
- 1000 (Absolute, newly added)

## Pull request checklist

### DIFF
--- a/rpc-interface/src/types.rs
+++ b/rpc-interface/src/types.rs
@@ -27,8 +27,8 @@ use nimiq_transaction::{
     },
 };
 use nimiq_vrf::VrfSeed;
-use serde::{Deserialize, Serialize};
-use serde_with::{serde_as, DeserializeFromStr, SerializeDisplay};
+use serde::{de::Error, Deserialize, Serialize};
+use serde_with::{serde_as, SerializeDisplay};
 
 #[derive(Clone, Debug, Serialize, Deserialize)]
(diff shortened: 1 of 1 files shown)


rpc-interface/src/types.rs
```diff
@@ -79,6 +79,34 @@ impl Display for ValidityStartHeight {
...
+                .trim()
+                .strip_prefix('+')
+                .map_or_else(
+                    || height.parse::<u32>().map(Self::Absolute), // In the case string is not prefixed by '+'
+                    |stripped| stripped.parse::<u32>().map(Self::Relative), // In the case string is prefixed by '+'
+                )
+                .map_err(|_| {
+                    Error::custom("provided ValidityStartHeight cannot be converted into a `u32`")
+                }),
+        }
```
sisou: Instead of re-implementing the string parsing, you could just use the existing `FromStr` implementation below? Or otherwise remove that I guess?

sisou: This could already work?
```suggestion
            StrOrU32::String(height) => Self::from_str(&height),
        }
```

sisou (approved):

jsdanielh (approved):
LGTM

PR #3207, opened 2024-12-18
Validator health status

viquezclaudio:
The validator health is used to track the overall validator health based on the number of consecutive deactivations of a validator on a per epoch basis.

The number of consecutive deactivations would cause a delay in terms of when the reactivation transaction is sent. 

## Pull request checklist

### DIFF
--- a/primitives/account/src/account/staking_contract/validator.rs
+++ b/primitives/account/src/account/staking_contract/validator.rs
@@ -46,7 +46,9 @@ use crate::{
 ///      in the first place.
 /// (**) The validator may be set to automatically reactivate itself upon inactivation.
 ///      If this setting is not enabled the state change can only be triggered manually.
-///
+///      However, there is a delay incurred if the validator is deactivated multiple consecutive times
+///      in the current epoch. The delay is a function of the number of deactivations and is reduced
+///      if the validator starts producing blocks in time.
 /// Create, Update, Deactivate, Retire and Re-activate are incoming transactions to the staking contract.
 /// Delete is an outgoing transaction from the staking contract.
 /// To Create, Update or Delete, the cold key must be used (the one corresponding to the validator
--- a/test-utils/src/validator.rs
+++ b/test-utils/src/validator.rs
@@ -66,6 +66,7 @@ pub async fn build_validators<N: TestNetwork + NetworkInterface>(
     peer_ids: &[u64],
     hub: &mut Option<MockHub>,
     is_prover_active: bool,
+    automatic_reactivate: bool,
 ) -> Vec<Validator<ValidatorNetworkImpl<N>>>
 where
     N::Error: Send + Sync,
@@ -113,7 +114,7 @@ where
         let (v, c) = build_validator(
             peer_ids[i],
             Address::from(&validator_keys[i]),
-            false,
+            automatic_reactivate,
             signing_keys[i].clone(),
             voting_keys[i].clone(),
             fee_keys[i].clone(),
--- a/validator/src/health.rs
+++ b/validator/src/health.rs
@@ -1,0 +1,159 @@
+use std::cmp;
+
+use nimiq_keys::Address;
+
+/// Enum that represents the overall health of a validator
+/// - Green means the Validator is working as expected.
+/// - Yellow means there has been more than 'VALIDATOR_YELLOW_HEALTH_INACTIVATIONS' consecutive inactivations
+///   in the current epoch.
+/// - Red means there has been more than 'VALIDATOR_RED_HEALTH_INACTIVATIONS' consecutive inactivations
+///   in the current epoch.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum ValidatorHealthState {
+    Green,
+    Yellow,
+    Red,
+}
+
+/// Struct that represents the overall Validator Health
+pub struct ValidatorHealth {
+    /// The current validator health
+    health: ValidatorHealthState,
+    /// Validator address
+    address: Address,
+    /// Only used for testing purposes controls whether blocks are published by the validator
+    publish: bool,
+    /// Number of consecutive inactivations that have occurred in the current epoch
+    inactivations: u32,
+    /// Next block number indicating when the re-activate transaction should be sent
+    reactivate_block_number: u32,
+    /// Flag that indicates if we are currently inactive
+    inactive: bool,
+}
+
+impl ValidatorHealth {
+    /// The number of consecutive inactivations from which a validator is considered with yellow health
+    const VALIDATOR_YELLOW_HEALTH_INACTIVATIONS: u32 = 2;
+    /// The number of consecutive inactivations from which a validator is considered with red health
+    const VALIDATOR_RED_HEALTH_INACTIVATIONS: u32 = 4;
+    // The maximum number of blocks the reactivate transaction can be delayed
+    const MAX_REACTIVATE_DELAY: u32 = 10_000;
+
+    /// Creates a new instance of the validator health structure
+    pub fn new(validator_address: &Address) -> ValidatorHealth {
+        ValidatorHealth {
+            health: ValidatorHealthState::Green,
+            publish: true,
+            inactivations: 0,
+            reactivate_block_number: 0,
+            inactive: false,
+            address: validator_address.clone(),
+        }
+    }
+
+    /// Computes the associated delay based on the current number of inactivations
+    fn get_reactivate_delay(&self) -> u32 {
+        cmp::min(self.inactivations.pow(2), Self::MAX_REACTIVATE_DELAY)
+    }
+
+    /// Returns the current health state of the validator
+    pub fn health(&self) -> ValidatorHealthState {
+        self.health
+    }
+
+    /// This flag should only be set in testing environments.
+    pub fn _set_publish_flag(&mut self, publish: bool) {
+        self.publish = publish
+    }
+
+    /// Decides if blocks should be published by the validator (only used for testing)
+    pub fn publish_block(&self) -> bool {
+        self.publish
+    }
+
+    /// Recomputes the current validator health state
+    pub fn refresh_validator_health_status(&mut self) {
+        log::trace!(
+            address=%self.address,
+            self.inactivations,
+            "Current validator Inactivations counter"
+        );
+
+        match self.health {
+            ValidatorHealthState::Green => {}
+            ValidatorHealthState::Yellow => {
+                if self.inactivations < Self::VALIDATOR_YELLOW_HEALTH_INACTIVATIONS {
+                    log::info!(self.inactivations, "Changed validator health to green");
+                    self.health = ValidatorHealthState::Green;
+                }
+            }
+            ValidatorHealthState::Red => {
+                if self.inactivations < Self::VALIDATOR_RED_HEALTH_INACTIVATIONS {
+                    log::info!(self.inactivations, "Changed validator health to yellow");
+                    self.health = ValidatorHealthState::Yellow;
+                }
+            }
+        }
+    }
+
+    /// Decreases the number of consecutive deactivations
+    pub fn block_produced(&mut self) {
+        self.inactivations = self.inactivations.saturating_sub(1);
+    }
+
+    /// Increases the number of consecutive deactivations
+    pub fn inactivate(&mut self, block_number: u32) {
+        if self.inactive {
+            // If we are currently inactive, then we dont have anything to do.
+            return;
+        }
+
+        self.inactivations = self.inactivations.saturating_add(1);
+        self.reactivate_block_number = block_number + self.get_reactivate_delay();
+        self.inactive = true;
+
+        match self.health {
+            ValidatorHealthState::Green => {
+                if self.inactivations >= Self::VALIDATOR_YELLOW_HEALTH_INACTIVATIONS {
+                    log::warn!(self.inactivations, "Changed validator health to yellow");
+                    self.health = ValidatorHealthState::Yellow;
+                }
+            }
+            ValidatorHealthState::Yellow => {
+                if self.inactivations >= Self::VALIDATOR_RED_HEALTH_INACTIVATIONS {
+                    log::warn!(self.inactivations, "Changed validator health to red");
+                    self.health = ValidatorHealthState::Red;
+                }
+            }
+            ValidatorHealthState::Red => {
+                log::warn!("Validator health is still red")
+            }
+        }
+
+        log::debug!(
+            address=%self.address,
+            self.inactivations,
+            self.reactivate_block_number,
+            block_number,
+            "New inactivation, current status",
+        );
+    }
+
+    /// Resets the internal counters and validator health state for a new epoch
+    pub fn reset_epoch(&mut self) {
+        // Reset the inactivations counter
+        self.inactivations = 0;
+        // Reset the validator health every epoch
+        self.health = ValidatorHealthState::Green;
+    }
+
+    /// Resets the pending reactivation flag
+    pub fn reactivate(&mut self) {
+        self.inactive = false;
+    }
+
+    /// Returns the block number for the next re-activate transaction
+    pub fn get_reactivate_block_number(&self) -> u32 {
+        self.reactivate_block_number
+    }
+}
--- a/validator/src/micro.rs
+++ b/validator/src/micro.rs
@@ -18,7 +18,10 @@ use nimiq_validator_network::ValidatorNetwork;
 use nimiq_vrf::VrfSeed;
 use parking_lot::RwLock;
 
-use crate::{aggregation::skip_block::SkipBlockAggregation, validator::Validator};
+use crate::{
+    aggregation::skip_block::SkipBlockAggregation,
+    validator::{Validator, ValidatorState},
+};
 
 pub(crate) enum ProduceMicroBlockEvent {
     MicroBlock,
@@ -36,6 +39,7 @@ struct NextProduceMicroBlockEvent<TValidatorNetwork> {
     block_number: u32,
     producer_timeout: Duration,
     block_separation_time: Duration,
+    state: Arc<RwLock<ValidatorState>>,
 }
 
 impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<TValidatorNetwork> {
@@ -53,6 +57,7 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
         block_number: u32,
         producer_timeout: Duration,
         block_separation_time: Duration,
+        state: Arc<RwLock<ValidatorState>>,
     ) -> Self {
         Self {
             blockchain,
@@ -65,6 +70,7 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
             block_number,
             producer_timeout,
             block_separation_time,
+            state,
         }
     }
 
@@ -117,6 +123,7 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
             info!(
                 block_number = self.block_number,
                 slot_band = self.validator_slot_band,
+                address = %self.state.read().validator_address,
                 "Our turn, producing micro block #{}",
                 self.block_number,
             );
@@ -153,6 +160,11 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
                 num_transactions
             );
 
+            if !self.state.read().validator_health.publish_block() {
+                log::warn!(block = block.block_number(), "Not publishing block");
+                break Some(Some(ProduceMicroBlockEvent::MicroBlock));
+            }
+
             // Publish the block. It is valid as we have just created it.
             Validator::publish_block(Arc::clone(&self.network), block.clone());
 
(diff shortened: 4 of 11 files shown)


test-utils/src/validator.rs
```diff
@@ -113,7 +113,7 @@ where
         let (v, c) = build_validator(
             peer_ids[i],
             Address::from(&validator_keys[i]),
-            false,
+            true,
```
Eligioo: Make `automatic_reactivate` a fn param for `build_validators` so it doesn't influence other tests?

validator/src/validator.rs
```diff
@@ -62,6 +65,30 @@ pub struct ConsensusState {
...
+    Yellow(u32),
+    Red(u32),
+}
+
+/// Struct that represents the overall Validator Health
+pub struct HealthState {
+    /// The current validator health
+    pub health: ValidatorHealth,
+    /// Number of blocks that we have produced in time(without being inactivated)
+    pub blk_cnt: u32,
```
Eligioo:
```suggestion
    pub block_counter: u32,
```

validator/src/micro.rs
```diff
@@ -181,6 +198,8 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
                 continue;
             }
 
+            self.health_state.write().blk_cnt += 1;
```
Eligioo:
```suggestion
            self.health_state.write().blk_cnt.saturating_add(1);
```
Prevent overflow

validator/src/micro.rs
```diff
@@ -153,6 +164,12 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
                 num_transactions
             );
 
+            if !self.health_state.read().publish {
+                log::warn!(block = block.block_number(), "Not publishing block");
+                let event = ProduceMicroBlockEvent::MicroBlock;
+                break Some(Some(event));
```
Eligioo:
```suggestion
                log::warn!(block = block.block_number(), "Not publishing block");
                break Some(Some(ProduceMicroBlockEvent::MicroBlock));
```

validator/src/validator.rs
```diff
@@ -62,6 +65,30 @@ pub struct ConsensusState {
...
+    /// For testing/debug purposes control wether produced blocks are published by the validator
```
Eligioo:
```suggestion
    /// For testing/debug purposes control whether produced blocks are published by the validator
```

validator/src/validator.rs
```diff
@@ -833,13 +875,45 @@ where
...
             match self.get_staking_state(&blockchain) {
                 ValidatorStakingState::Active => {
                     drop(blockchain);
                     if self.validator_state.is_some() {
                         self.validator_state = None;
                         info!("Automatically reactivated.");
                     }
+
+                    let validator_health = self.health_state.read().health;
+                    match validator_health {
```
Eligioo:
```suggestion
                    match self.health_state.read().health {
```

validator/src/validator.rs
```diff
@@ -62,6 +65,30 @@ pub struct ConsensusState {
...
+    Green,
+    Yellow(u32),
+    Red(u32),
+}
+
+/// Struct that represents the overall Validator Health
+pub struct HealthState {
+    /// The current validator health
+    pub health: ValidatorHealth,
+    /// Number of blocks that we have produced in time(without being inactivated)
```
Eligioo:
```suggestion
    /// Number of subsequent blocks that we have been produced without being inactivated
```

validator/src/validator.rs
```diff
@@ -833,13 +875,45 @@ where
...
+                    match validator_health {
+                        ValidatorHealth::Green => {}
+                        ValidatorHealth::Yellow(yellow_block_number) => {
+                            debug!(
+                                address = %self.validator_address.read(),
+                                inactivated = yellow_block_number,
+                                good_blocks = %self.health_state.read().blk_cnt,
+                                "Current validator health is yellow",
+                            );
+                            if self.health_state.read().blk_cnt >= VALIDATOR_HEALTH_THRESHOLD {
```
Eligioo: This logic doesn't seem to match with the original description: `If the validator is Yellow and is not deactivated in a quarter of an epoch, we change its status to Green.`

validator/src/validator.rs
```diff
@@ -833,13 +875,45 @@ where
...
+                                self.health_state.write().blk_cnt = 0;
+                            }
+                        }
+                        ValidatorHealth::Red(red_block_number) => {
+                            debug!(
+                                address = %self.validator_address.read(),
+                                inactivated = red_block_number,
+                                "Current validator health is red",
+                            );
+                            if self.health_state.read().blk_cnt >= VALIDATOR_HEALTH_THRESHOLD {
```
Eligioo: This logic doesn't seem to match with the original description: `If the validator is Red and is not deactivated in one epoch, we change its status to Yellow.`

validator/src/validator.rs
```diff
@@ -850,9 +924,38 @@ where
                             .unwrap_or(true)
                         && self.automatic_reactivate.load(Ordering::Acquire)
                     {
-                        let inactivity_state = self.reactivate(&blockchain);
-                        drop(blockchain);
-                        self.validator_state = Some(inactivity_state);
+                        let validator_health = self.health_state.read().health;
+                        match validator_health {
```
Eligioo:
```suggestion
                        match self.health_state.read().health {
```

validator/src/validator.rs
```diff
@@ -850,9 +924,38 @@ where
...
+                                let inactivity_state = self.reactivate(&blockchain);
+                                drop(blockchain);
+                                self.validator_state = Some(inactivity_state);
+                                self.health_state.write().health =
+                                    ValidatorHealth::Red(block_number);
+                                self.health_state.write().blk_cnt = 0;
+                            }
+                            ValidatorHealth::Red(_) => {
+                                log::warn!(
+                                    "The validator needs human intervention, no automatic reactivate"
```
Eligioo:
```suggestion
                                    "The validator needs human intervention, no automatic reactivation"
```

validator/src/health.rs
```diff
@@ -0,0 +1,153 @@
...
+    pub fn health(&self) -> ValidatorHealthState {
+        self.health
+    }
+
+    pub fn publish_block(&self) -> bool {
+        self.publish
+    }
+
+    /// This flag should only be used in tests context.
+    pub fn _set_publish_flag(&mut self, publish: bool) {
```
Eligioo, line 65: The method can be annotated with `#[cfg(test)]` such that it's only included in `cargo test` and not in `cargo build`.

primitives/account/src/account/staking_contract/validator.rs
```diff
@@ -46,6 +46,9 @@ use crate::{
 ///      in the first place.
 /// (**) The validator may be set to automatically reactivate itself upon inactivation.
 ///      If this setting is not enabled the state change can only be triggered manually.
+///      However, there is a delay incurred if the validator is deactivated multiple consecutive times
+///      in the current epoch. The delay is function of the number of deactivations and is reduced
```
Eligioo:
```suggestion
///      in the current epoch. The delay is a function of the number of deactivations and is reduced
```

validator/src/health.rs
```diff
@@ -0,0 +1,156 @@
+use std::cmp;
+
+use nimiq_keys::Address;
+
+/// Struct that represents the overall health of a validator
```
Eligioo: `ValidatorHealthState` is an enum not a struct

validator/src/health.rs
```diff
@@ -0,0 +1,156 @@
...
+
+/// Struct that represents the overall Validator Health
+pub struct ValidatorHealth {
+    /// The current validator health
+    health: ValidatorHealthState,
+    /// Validator address
+    address: Address,
+    /// Only used for testing purposes controls whether blocks are published by the validator
+    publish: bool,
+    /// Number of consecutive inactivations that have ocurred in the current epoch
```
Eligioo:
```suggestion
    /// Number of consecutive inactivations that have occurred in the current epoch
```

validator/src/health.rs
```diff
@@ -0,0 +1,156 @@
...
+    inactivations: u32,
+    /// Next block number where the re-activate transaction should be sent
```
Eligioo:
```suggestion
    /// Next block number indicating when the re-activate transaction should be sent
```

validator/src/validator.rs
```diff
@@ -467,6 +478,9 @@ where
                 self.on_blockchain_extended(hash);
             }
             BlockchainEvent::EpochFinalized(ref hash) => {
+                // Reset the validator health for the new epoch
+                self.validator_health.write().reset_epoch();
```
Eligioo: Why would everything get reset once an epoch finalises? The 'health' of a validator doesn't suddenly improve if a new epoch starts.

validator/src/micro.rs
```diff
@@ -181,6 +196,9 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
                 continue;
             }
 
+            // Each successfull block will decrease the number of inactivations
```
Eligioo:
```suggestion
            // Each successful block will decrease the number of inactivations
```

validator/src/health.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    /// Decreases the number of consecutive deactivations
+    pub fn block_produced(&mut self) {
+        self.inactivations = self.inactivations.saturating_sub(1);
+    }
+
+    /// Increases the number of consecutive deactivations
+    pub fn inactivate(&mut self, block_number: u32) {
+        if !self.pending_reactivate {
+            self.inactivations += 1;
```
Eligioo: saturating add?

validator/tests/mock.rs
```diff
@@ -15,7 +15,7 @@ use nimiq_network_interface::{
 };
 use nimiq_network_libp2p::Network;
 use nimiq_network_mock::{MockHub, MockNetwork};
-use nimiq_primitives::{networks::NetworkId, policy::Policy};
+use nimiq_primitives::{coin::Coin, networks::NetworkId, policy::Policy};
```
Eligioo: Unused imports?

validator/tests/mock.rs
```diff
@@ -24,9 +24,11 @@ use nimiq_test_utils::{
     },
 };
 use nimiq_time::{sleep, timeout};
+use nimiq_transaction_builder::TransactionBuilder;
```
Eligioo: Same as above

validator/tests/mock.rs
```diff
@@ -176,6 +184,204 @@ async fn validators_can_do_skip_block() {
...
+        log::info!("Spawning validator: {}", validator.validator_address());
+        spawn(validator);
+    }
+
+    validator_proxy
+        .validator_health
+        .write()
+        ._set_publish_flag(false);
+
+    events.take(30).for_each(|_| future::ready(())).await;
```
Eligioo: I guess you can take 50 here and get in the red state immediately. Testing if the validator got the yellow status is already covered in another test.

validator/src/health.rs
viquezclaudio: `#[cfg(test)]` is applied only in unittests and not integration tests, so I would leave it like this

validator/src/health.rs
```diff
@@ -0,0 +1,156 @@
...
+pub enum ValidatorHealthState {
+    Green,
+    Yellow,
+    Red,
+}
+
+/// Struct that represents the overall Validator Health
+pub struct ValidatorHealth {
+    /// The current validator health
+    health: ValidatorHealthState,
```
nibhar, line 21: As I understand it this is a derived value over the other fields of this struct. I would prefer then for this to not be stored explicitly, but instead be retrieved on a function call like `health(&self)` below.

It actually is only used in a single test, and there it is just enough check if it is `healthy() -> bool`. If there was a purpose to Yellow (or Red), it has been lost during one of the changes. 

I suggest to remove it entirely and use `healthy(&self) -> bool` instead. The enum can then also go.

validator/src/health.rs
```diff
@@ -0,0 +1,156 @@
...
+    /// Validator address
+    address: Address,
+    /// Only used for testing purposes controls whether blocks are published by the validator
+    publish: bool,
```
nibhar, line 25: I think this does not belong here. It should be done in the validator or in the network even.

If you insist on having it here, add `#[cfg(test)]` where appropriate to not include this in non testing builds.

validator/src/health.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub fn block_produced(&mut self) {
+        self.inactivations = self.inactivations.saturating_sub(1);
+    }
+
+    /// Increases the number of consecutive deactivations
+    pub fn inactivate(&mut self, block_number: u32) {
+        if !self.pending_reactivate {
+            self.inactivations = self.inactivations.saturating_add(1);
+            self.reactivate_block_number = block_number + self.get_reactivate_delay();
+            self.pending_reactivate = true;
```
nibhar: Ok so this is a somewhat obscure way of referencing that some other piece of code is going to try and re-try for a reactivate and that in the meantime the node will remain inactive and thus not produce blocks.  

If my assessment here is correct then I think this should be asserted instead.
```
assert!(!self.pending_reactivate, "Cannot be inactivated when not active");
```
If not I prefer this be inverted:
```rust
if self.pending_reactivate {
  return;
}
...
```

nibhar: What does `pending_reactivate` actually mean here? I do not see a reactivate being sent here, so it probably means something else.

I would rename this to `inactive`, or `currently_inactive`. If there is a reactivate currently pending is unknown, but also irrelevant for what this does.

validator/src/health.rs
```diff
@@ -0,0 +1,156 @@
...
+    pub fn reset_epoch(&mut self) {
+        // Reset the inactivations counter
+        self.inactivations = 0;
+        // Reset the validator health every epoch
+        self.health = ValidatorHealthState::Green;
+    }
+
+    /// Resets the pending reactivation flag
+    pub fn reset_reactivation(&mut self) {
+        self.pending_reactivate = false;
```
nibhar: What about the `reactivate_block_number`?

validator/src/health.rs
```diff
@@ -0,0 +1,156 @@
...
+                self.inactivations,
+                self.reactivate_block_number,
+                block_number,
+                "New inactivation, current status",
+            );
+        }
+    }
+
+    /// Resets the internal counters and validator health state for a new epoch
+    pub fn reset_epoch(&mut self) {
```
nibhar, line 143: Generally I would argue that maybe resetting on epoch is not the best, but if it does, should it not reset `reactivate_block_number` and `pending_reactivate` as well? Does an inactivation count of zero with `pending_reactivate = true` make any sense? As I see it you will then not only be inactive, but not even electable and thus should not have any slots at all..?

validator/src/micro.rs
```diff
@@ -181,6 +196,9 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
                 continue;
             }
 
+            // Each successful block will decrease the number of inactivations
+            self.validator_health.write().block_produced();
```
nibhar: I feel like this could be done in the validator instead and then we could get rid of the `Arc<RwLock<..>>`, as only the Validator hold the instance. 

The disabling of the publish would not be covered, but I would argue that is unrelated to the validator health anyways and should probably be implemented somewhere else.

validator/src/micro.rs
```diff
@@ -153,6 +163,11 @@ impl<TValidatorNetwork: ValidatorNetwork + 'static> NextProduceMicroBlockEvent<T
                 num_transactions
             );
 
+            if !self.validator_health.read().publish_block() {
+                log::warn!(block = block.block_number(), "Not publishing block");
+                break Some(Some(ProduceMicroBlockEvent::MicroBlock));
+            }
```
nibhar: I would want this to be moved into `Validator::publish_block`. There it can also be used for macro blocks at the same time.

validator/src/validator.rs
```diff
@@ -88,6 +90,7 @@ impl Clone for ValidatorProxy {
             automatic_reactivate: Arc::clone(&self.automatic_reactivate),
             slot_band: Arc::clone(&self.slot_band),
             consensus_state: Arc::clone(&self.consensus_state),
+            validator_health: Arc::clone(&self.validator_health),
```
nibhar: Why does the proxy need the validator health? If we remove this here and in the micro producer then the Arc and the RwLock can be removed. If this is only for testing then we might want to think about better alternatives for the testing purposes.

validator/src/health.rs
```diff
@@ -0,0 +1,156 @@
...
+    /// The current validator health
+    health: ValidatorHealthState,
+    /// Validator address
+    address: Address,
+    /// Only used for testing purposes controls whether blocks are published by the validator
+    publish: bool,
+    /// Number of consecutive inactivations that have occurred in the current epoch
+    inactivations: u32,
+    /// Next block number indicating when the re-activate transaction should be sent
+    reactivate_block_number: u32,
```
nibhar, line 29: Is this related to Validator health? Might be a better fit where the reactivation actually happens.

nibhar (changes requested):
I think some changes made the distinction between red and yellow health go away. I would like to see some structural changes to the Validator Health struct and potentially removing the heap allocation as I think it is unnecessary.

validator/src/health.rs
viquezclaudio: I think the Green, Yellow, Red distinction is to convey that meaning to the validator operators in the logs (and I think it would be valuable to provide an RPC function as well to get the validator health) rather than just providing the number of consecutive deactivations a validator has experienced. 

But correct, this could be simplified to just keep track the number of consecutive deactivations and internally computing a delay based on this number without any notion of Good (green), Warning (yellow), Bad (red)

validator/src/health.rs
viquezclaudio: #[cfg(test)] is applied only in unittests and not integration tests so it cannot be used to filter out testing only fields.
another option would be to create a feature flag only for this purpose but i think thats unnecessary. 

I agree that this should be moved to the validator and not validator health. But then we wont be able to tweak this flag in runtime from the test, because we can only tweak from a test (or externally) whatever is exposed in the validator proxy 

Im open to suggestions

validator/src/health.rs
viquezclaudio: The reactivate block number would be computed the next time we are de-activated. 
So it is not necessary to compute it here, although i could compute it here just for completeness

validator/src/health.rs
viquezclaudio: Same as above `reactivate_block_number`, is computed when we are deactivated so there is no need to compute it here.
We can decide if we want to reset every epoch or if we reset every X number of blocks, or we dont reset and we just stay with the current validator health. who makes this decision ?

validator/src/micro.rs
viquezclaudio: How can we do this in the validator? because the validator only polls the micro producer but AFAIK it is not aware if the block was successfully produced and published

validator/src/micro.rs
viquezclaudio: I agree
But I would like to control the publish_flag externally from a test, so if I move it to the validator, how can I do that?

validator/src/validator.rs
viquezclaudio: Yup, this is only for testing.
if theres a better alternative to control wether blocks are published or not, I would remove this Arc

validator/src/health.rs
viquezclaudio: The reactivation happens within the validator. I put it here to update it when the `validator_health::inactivate` is called to keep everything together rather than having this number in the validator and updating it from the validator health.

validator/src/health.rs
nibhar: Not storing `health` in the struct I would say we should do regardless. There is no benefit in doing so and calculating it on the fly is very cheap. If this will only be used for logging, and the logging is done internally, then i think the `fn healthy(&self) -> bool` is actually sufficient. Logging could then be done whenever a threshold is reached and the 'state' changes. 
While I would prefer that I think it would also be fine to keep the enum and have `fn health(&self) -> ValidatorHealthState`.

validator/src/micro.rs
nibhar: I am not 100% sure about it, but in `Validator::poll_micro` the polling of the micro producer should give you the event, which should contain the micro block. You can check if it is a skip block or not, and that is going to be just as good as doing it here I would think. I would have to go look though.

viquezclaudio: Currently the block itself is not part of the event returned in the ProduceMicroBlock stream (which is the one that is being polled in `Validator::poll_micro` )

nibhar: Maybe you could add the skip blocks as distinct events from micro blocks.
```rust
enum ProduceMicroBlockEvent {
  MicroBlock,
  SkipBlock,
}
```
and use it then in `Validator::poll_micro`?

validator/src/health.rs
nibhar: As it is not at all guaranteed that this number is not getting used in between a call to `reset_activations` and `inactivate` there is potential for inconsistent state. This should not make that assumption on the caller, it should present consistent state, no matter what the caller does.

validator/src/health.rs
nibhar: Same as above, the `reactivate_block_number` as well as `pending_reactivate` (now `inactive`) are accessible at all times. The presented state should be consistent. 

Do we need to reset this at all? It should be quasi self resetting in a way as soon as a validators actual health improves?

PR #3222, opened 2025-01-07
CI: bump fixed Rust toolchains to 1.82.0

Eligioo:
Fix the CI and update `libmdbx`. Rust 1.84 is planned to be released on the 9th of Jan meaning that testing 2 versions older than the current Rust stable will uphold.

1.82.0 introduces `unsafe extern blocks` (RFC3484) which are used by `libmdbx` 0.5.3

## Pull request checklist

### DIFF
--- a/.github/workflows/build+test.yml
+++ b/.github/workflows/build+test.yml
@@ -26,7 +26,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
-    - uses: dtolnay/rust-toolchain@1.79.0 # Also test our Rust MSRV here.
+    - uses: dtolnay/rust-toolchain@1.82.0 # Also test our Rust MSRV here.
     - uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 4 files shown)


jsdanielh (approved):
LGTM

Eligioo:
@sisou the npm release action is still fixed to 1.81.0. Can this one also get increased?

jsdanielh:
> @sisou the npm release action is still fixed to 1.81.0. Can this one also get increased?

I think it is safe to bump it as well. The reason why we had to pin it, was because the combination of rust 1.82.0 and `wasm-bindgen` 0.2.[94,95] made `wasm-opt` to fail per [this issue](https://github.com/rustwasm/wasm-pack/issues/1441).

Now that we are using a newer version of `wasm-bindgen`, it doesn't fail. I ran it locally and built fine for me.

jsdanielh (approved):

PR #3226, opened 2025-01-07
RPC client `create_validator` uses the configuration as input

### LINKED ISSUE #3225: RPC Client: take signing key from node config when creating a validator
Use configured signing key when creating a `CreateValidator` transaction with the RPC client instead of passing the signing key via RPC.

ii-cruz:
The RPC client's operations for the validator were designed to be applied to the local validator. The create validator was the exception. This PR standardizes it.
To submit txs regarding other validators one can use the send raw transaction.

## Pull request checklist

### DIFF
--- a/rpc-client/src/main.rs
+++ b/rpc-client/src/main.rs
@@ -56,6 +56,7 @@ enum Command {
 
     /// Shows and modifies validator information.
     /// Create, signs and send transactions referring to the local validator.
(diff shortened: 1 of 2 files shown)


sisou (approved):
Ah! If this only changes the _client_ implementation and not the RPC server interface then I'm OK with this.

Eligioo (approved):

jsdanielh (approved):

PR #3246, opened 2025-01-16
Web-client multisig primitives

### LINKED ISSUE #1578: Add multisig support for the web client
This is a follow-up issue of #1339 for the multisig support in the web client.
It was decided to close the aforementioned issue and continue tracking the multisig support in this issue.

sisou:

Expose basic primitives for multisig operations through the web-client API.

The primitves are:

- RandomSecret
- Commitment
- CommitmentPair
- PartialSignature

Also add a `Address.fromPublicKeys(public_keys, num_signers)` method to calculate a multisig address from the participating public keys.

These primitives enable the client-side operations of the multisig app with the @nimiq/core package alone and thus remove the need for custom Rust-compiled-to-WASM in both the Multisig App and Keyguard.

This PR does not include what are currently server-side operations, such as aggregating commitments, summing partial signatures or calculating merkle paths. That might get added later in a separate PR (tracked by a separate upcoming issue).

## Pull request checklist

### DIFF
--- a/keys/src/multisig/commitment.rs
+++ b/keys/src/multisig/commitment.rs
@@ -10,7 +10,7 @@ use zeroize::Zeroize;
 use super::{error::InvalidScalarError, MUSIG2_PARAMETER_V};
 
 /// A nonce is a "number only used once" and it is supposed to be kept secret (similar to a private key).
-#[derive(PartialEq, Eq, Debug, Clone, Copy, Zeroize)]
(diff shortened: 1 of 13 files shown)


sisou:
I was not able to bundle the multisig primitives into a separate entry point, unfortunately, as the multisig primitives also use common primitives like `PublicKey` and `PrivateKey`. When loaded in two separate WASM modules, they cannot access each other's memory (pointers).

jsdanielh (approved):
LGTM

PR #3247, opened 2025-01-16
Omit autonat and kad features from the web-client

sisou:

Improve usage of feature flags throughout different crates to avoid bundling `autonat` and parts of `kad` functionality into the web-client (and I believe thus light client, too)

## Pull request checklist

### DIFF
--- a/lib/Cargo.toml
+++ b/lib/Cargo.toml
@@ -125,6 +125,7 @@ full-consensus = [
     "nimiq-blockchain",
     "nimiq-consensus/full",
     "nimiq-dht",
+    "nimiq-network-libp2p/autonat",
     "nimiq-network-libp2p/kad",
 ]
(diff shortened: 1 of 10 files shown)


Eligioo:
I would prefer to fix this by creating a new 'light' behaviour and gate that based on the target platform [here](https://github.com/nimiq/core-rs-albatross/blob/albatross/network-libp2p/src/swarm.rs#L86-L87) instead of adding all these feature flags. It makes everything quite messy and difficult to maintain now.

This new light behaviour would omit all the stuff(autonat, kad, etc.) that are not necessary for a light client.

PR https://github.com/nimiq/core-rs-albatross/pull/3012 prepped this effort by detaching the events from the handlers.

sisou:
OK fair enough, but I don't know anything about how that works, so someone else will have to tackle it. Is it a huge effort?

Eligioo:
> Is it a huge effort?

No I don't think so.

PR #3258, opened 2025-01-29
RPC server: add option to configure CORS origins

### LINKED ISSUE #2225: [RPC / Websocket] - Way to add CORS allowlist so we can allow all or specific website to directly connect.
## New issue checklist
- [X] I have read all of the [`README`](https://github.com/nimiq-network/core/blob/master/README.md) 
- [X] I have searched [existing issues](https://github.com/nimiq-network/core/issues?q=is%3Aissue+sort%3Acreated-desc) and **this is not a duplicate**.

### General information

- Library version(s): -
- Browser version(s): -
- Devices/Simulators/Machine affected: -
- Reproducible in the testnet? (Yes/No):  Yes
- Related issues: - 

Nimiq 0.20.0+

#### Expected behavior

I would expect to connect a browser directly to websocket/rpc interface with nimiq but thats not possible due to blocking in CORS, there is currently no way to add cors allowlist to all for example.

#### Actual behavior

Blocked by cors in browser.

#### Steps to reproduce

Connect via website to a rpc node directly.

--

Add support in config to add CORS allowlist.

Eligioo:
Add option to configure CORS for the RPC server

## Pull request checklist

### DIFF
--- a/lib/src/extras/rpc_server.rs
+++ b/lib/src/extras/rpc_server.rs
@@ -1,7 +1,7 @@
 use std::{collections::HashSet, iter::FromIterator, sync::Arc};
 
 use nimiq_jsonrpc_server::{
-    AllowListDispatcher, Config, Credentials, ModularDispatcher, Server as _Server,
+    AllowListDispatcher, Config, Cors, Credentials, ModularDispatcher, Server as _Server,
 };
 use nimiq_rpc_server::dispatchers::*;
 use nimiq_wallet::WalletStore;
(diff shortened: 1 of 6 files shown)


lib/src/extras/rpc_server.rs
```diff
@@ -33,8 +33,13 @@ pub fn initialize_rpc_server(
         Some(HashSet::from_iter(allowed_methods))
     };
 
-    // TODO: Pass this to the rpc server config
-    let _corsdomain = config.corsdomain.unwrap_or_default();
+    let cors_domains = config.cors_domains.unwrap_or_default();
+    let is_cors_wildcard = cors_domains.iter().any(|origin| origin.trim() == "*");
+    let cors_config = if is_cors_wildcard {
+        Cors::new().with_any_origin()
```
jsdanielh, line 39: But if it's empty (the default), shouldn't it allow any origin as well?

Eligioo: No, origins only explicitly provided should pass the preflight. So if none are provided, all should fail. This is the same behaviour as before you weren't able to configure CORS.

jsdanielh (approved):
LGTM

PR #3266, opened 2025-02-04
Remove nimiq-genesis as dependency for test-log

### LINKED ISSUE #2558: `nimiq-test-log` should not transitively depend on `nimiq-genesis`
This makes compiling tests very slow. `nimiq-test-log` is intentionally kept small so that all tests can depend on it.

viquezclaudio:
Fixes #2558

...
## Pull request checklist

### DIFF
--- a/primitives/src/policy.rs
+++ b/primitives/src/policy.rs
@@ -659,7 +659,8 @@ pub const TEST_POLICY: Policy = Policy {
     batches_per_epoch: 4,
     state_chunks_max_size: 3,
     transaction_validity_window: 2,
-    genesis_block_number: 0,
+    // This number should match the one that is defined in the `unit` network genesis file which is the genesis used for unit testing
+    genesis_block_number: 200,
 };
 
 #[cfg(test)]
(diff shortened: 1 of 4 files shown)


sisou:
I think you need to remove `nimiq-genesis` also from the `Cargo.toml` to really prevent compilation when this crate is used.

sisou (approved):

primitives/src/policy.rs
```diff
@@ -659,7 +659,8 @@ pub const TEST_POLICY: Policy = Policy {
     batches_per_epoch: 4,
     state_chunks_max_size: 3,
     transaction_validity_window: 2,
-    genesis_block_number: 0,
+    // This number should match the one that is defined in the unittest genesis file
```
jsdanielh:
```suggestion
    // This number should match the one that is defined in the `unit` network genesis file which is the genesis used for unit testing
```

hrxi (approved):
Amazing. This dependency has been a thorn in my side for a long time already.

hrxi:
Rebased for merging.

PR #3278, opened 2025-02-08
chore(web-client): fix default comlink import in node

onmax:
## Summary
@nimiq/core Node.js build uses incorrect default imports for comlink in 3 files. Comlink has no default export, violating ESM spec.

## Live Demo

**Bug Demo:** https://nimiq-core-3278-bug.vercel.app

**Instructions:**
1. Open the demo page
2. Press F12 to open browser DevTools console
3. Click "Import @nimiq/core" button
4. Watch console for runtime import errors

The page imports @nimiq/core@2.2.1 via esm.sh CDN, which will show the comlink import error in browser console.

## Verify Bug Locally

```bash
git clone --depth 1 --filter=blob:none --sparse --branch repro/nimiq-core-3278 https://github.com/onmax/repros.git
cd repros && git sparse-checkout set nimiq-core-3278
cd nimiq-core-3278 && pnpm i

# Show 3 files with incorrect default import
grep "import Comlink from" node_modules/@nimiq/core/nodejs/*.mjs

# Verify comlink has no default export
tail -5 node_modules/.pnpm/comlink@*/node_modules/comlink/dist/esm/comlink.mjs
# Shows: export { createEndpoint, expose, ... } - NO default
```

## Verify Fix

```bash
git sparse-checkout add nimiq-core-3278-fixed
cd ../nimiq-core-3278-fixed && pnpm i

# Verify correct namespace import
grep "import \* as Comlink from" node_modules/@nimiq/core/nodejs/*.mjs
```

The `-fixed` folder uses [pnpm patch](https://pnpm.io/cli/patch) to apply the fix locally.

## Real-World Impact

Production apps need patches to work around this:
- [nimiq/starter patches/@nimiq__core@2.2.0.patch](https://github.com/nimiq/starter/blob/main/patches/%40nimiq__core%402.2.0.patch) - Cloudflare D1 example requires patching
- Correct pattern already used in bundler build: [bundler/crypto.js:4](https://github.com/nimiq/core-rs-albatross/blob/albatross/web-client/dist/bundler/crypto.js#L4)

## Changes
3 files updated to use namespace imports:
- `nodejs/crypto.mjs`
- `nodejs/index.mjs`
- `nodejs/worker.mjs`

### DIFF
--- a/web-client/dist/nodejs/crypto.mjs
+++ b/web-client/dist/nodejs/crypto.mjs
@@ -2,7 +2,7 @@
 // global scope. We import the generated JS file to make `wasm_bindgen`
 // available which we need to initialize our WASM code.
 import { parentPort } from 'node:worker_threads';
-import Comlink from 'comlink';
+import * as Comlink from 'comlink';
 import nodeEndpoint from 'comlink/dist/esm/node-adapter.mjs';
 import { CryptoUtils } from './crypto-wasm/index.js';
 
--- a/web-client/dist/nodejs/index.mjs
+++ b/web-client/dist/nodejs/index.mjs
@@ -1,6 +1,6 @@
 import { webcrypto } from 'node:crypto';
 import { Worker } from 'node:worker_threads';
-import Comlink from 'comlink';
+import * as Comlink from 'comlink';
 import nodeEndpoint from 'comlink/dist/esm/node-adapter.min.mjs';
 import { Address, CryptoUtils, Transaction } from './main-wasm/index.js';
 import { clientFactory } from '../launcher/node/client-proxy.mjs';
--- a/web-client/dist/nodejs/worker.mjs
+++ b/web-client/dist/nodejs/worker.mjs
@@ -2,7 +2,7 @@
 // global scope. We import the generated JS file to make `wasm_bindgen`
 // available which we need to initialize our WASM code.
 import { parentPort } from 'node:worker_threads';
-import Comlink from 'comlink';
+import * as Comlink from 'comlink';
 import nodeEndpoint from 'comlink/dist/esm/node-adapter.mjs';
 import websocket from 'websocket';
 import { Client } from './worker-wasm/index.js';


sisou:
I'm using NodeJS v20, and for me the example in `web-client/example/node` works as is:
- Apply the revised build command from #3282
- If trying on latest `albatross` branch, apply fix from #3281 - not needed if trying with latest tag or before

Can you please try out the example (run `index.mjs`) and report if it works with your Node version?

If it does, let me know in which context and how you are using the lib in Node that doesn't work.

onmax:
I get this error when running `maxi@Maxi:~/nimiq/core-rs-albatross/web-client$ ./scripts/build.sh --only nodejs,types`.

Do I need to compile something before running `./script/build.sh`?

```
error[E0463]: can't find crate for `core`
  |
  = note: the `wasm32-unknown-unknown` target may not be installed
  = help: consider downloading the target with `rustup target add wasm32-unknown-unknown`

For more information about this error, try `rustc --explain E0463`.
error: could not compile `subtle` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
error[E0463]: can't find crate for `std`
  |
  = note: the `wasm32-unknown-unknown` target may not be installed
  = help: consider downloading the target with `rustup target add wasm32-unknown-unknown`

error: could not compile `cfg-if` (lib) due to 1 previous error
error: could not compile `once_cell` (lib) due to 1 previous error
error: could not compile `byteorder` (lib) due to 1 previous error
error: could not compile `const-oid` (lib) due to 1 previous error
error: could not compile `typenum` (lib) due to 1 previous error
error: could not compile `pin-project-lite` (lib) due to 1 previous error
error: could not compile `memchr` (lib) due to 1 previous error
```

sisou:
Well, tells you right there in the message.

jsdanielh:
@onmax Were you able to verify this?

Issue #3277, opened 2025-02-08
Unable to use WebClient in cloudflare


onmax:
I want to deploy a project I have been working on [Cloudflare Pages runtime](https://developers.cloudflare.com/pages/functions/), which is different from a normal node environment and has fewer features.

The way we import the `.wasm` file is currently completely incompatible with Cloudflare.

For example, I tried this:

```
const publicKey = '82d80b86d9bf7906832e9f0ba4fa69018792f59190075c396b0e85aeac444e55'
const key = PublicKey.fromHex(publicKey)
```

but I get the following error:

```
[wrangler:err] TypeError: Cannot read properties of undefined (reading '__wbindgen_malloc')
    at PublicKey.fromHex (file:///home/maxi/personal/local-cloudflare/node_modules/.pnpm/@nimiq+core@2.0.6/node_modules/@nimiq/core/bundler/main-wasm/index_bg.js:2453:50)
    at Object.fetch (file:///home/maxi/personal/local-cloudflare/index.ts:6:27)
    at fetchDispatcher (file:///home/maxi/personal/local-cloudflare/.wrangler/tmp/bundle-2TP1X7/middleware-loader.entry.ts:54:17)
    at __facade_invokeChain__ (file:///home/maxi/personal/local-cloudflare/node_modules/.pnpm/wrangler@3.107.3_bufferutil@4.0.9_utf-8-validate@5.0.10/node_modules/wrangler/templates/middleware/common.ts:53:9)
    at Object.next (file:///home/maxi/personal/local-cloudflare/node_modules/.pnpm/wrangler@3.107.3_bufferutil@4.0.9_utf-8-validate@5.0.10/node_modules/wrangler/templates/middleware/common.ts:50:11)
    at jsonError (file:///home/maxi/personal/local-cloudflare/node_modules/.pnpm/wrangler@3.107.3_bufferutil@4.0.9_utf-8-validate@5.0.10/node_modules/wrangler/templates/middleware/middleware-miniflare3-json-error.ts:22:30)
    at __facade_invokeChain__ (file:///home/maxi/personal/local-cloudflare/node_modules/.pnpm/wrangler@3.107.3_bufferutil@4.0.9_utf-8-validate@5.0.10/node_modules/wrangler/templates/middleware/common.ts:53:9)
[...]

PR #3279, opened 2025-02-10
chore(web-client): added esm import for non-browser environments

onmax:

After this message: https://github.com/nimiq/core-rs-albatross/issues/3277#issuecomment-2646614887, I need to import the `web` submodule from a non-browser runtime.

I get the following error:

```
[error] Package subpath './web' is not defined by "exports" in ./node_modules/@nimiq/core/package.json imported from ./.nuxt/prerender/index.mjs
```

Not sure if this is right solution but it works in my `pnpm patch` so I think is ok

### DIFF
--- a/web-client/dist/package.json
+++ b/web-client/dist/package.json
@@ -24,6 +24,7 @@
     },
     "./web": {
       "browser": "./web/index.js",
+      "import": "./web/index.js",
       "types": "./types/web.d.ts"
     }
   },


sisou (approved):

jsdanielh (approved):

onmax:
I am not sure if it makes sense to have this PR...

Thinking it more, maybe makes more sense if we have the import for ESM from the bundler instead from the `web` folder 🤔

Something like this:

```
"exports": {
    ".": {
      "node": {
        "import": "./nodejs/index.mjs",
        "require": "./nodejs/index.js"
      },
      "browser": "./bundler/index.js",
      "import": "./web/index.js", /// This line instead
      "types": "./types/bundler.d.ts"
    },
    "./web": {
      "browser": "./web/index.js",
      "types": "./types/web.d.ts"
    }
  }, 
```

hrxi (approved):
Thanks.

hrxi:
Rebased for merging.

onmax:
@sisou is possible that the PR https://github.com/Albermonte/nimiq-svelte-demo/pull/1 is related to this PR?

sisou:
Hm. The problem in that PR was that it was using a bundler (vite), but not the bundler version of the package.

Here you explicitly need a non-bundler version in Cloudflare, because it handles wasm files differently.

So I'd say this change here is still warranted.

PR #3245, opened 2025-01-16
Web-client: Make client worker initialization more resilient

sisou:
## What's in this pull request?

Change client web-worker initialization signal from push-based to pull-based, i.e. change from the worker sending a signal that it's ready (which can be missed (race-condition) if the worker initializes too quickly) to polling the worker from the main thread, which is more resilient. 

## Pull request checklist

- [x] All tests pass. The project builds and runs.
- [x] I have resolved any merge conflicts.
- [x] I have resolved all `clippy` and `rustfmt` warnings.

PR #3281, opened 2025-02-10
Fix initialization in NodeJS environments

sisou:
In Node, the callback argument _is_ the `data`. The broken state (#3245) is not yet released, so this must go in before the next release.


jsdanielh (approved):
LGTM

PR #3285, opened 2025-02-11
Pico consensus

viquezclaudio:
Pico Consensus Implementation
    
The PicoMacroSync is a macro sync mechanism that pushes the latest election block
that is obtained from a peer, if subsequent (sufficient) peers also have the same latest election
(and latest checkpoint blocks) then we move all the peers to live sync and we establish consensus.
If some peer has a different state than the initial state we obtained, we need to fallback to the
regular light macro sync.

To use it or try it in the web client:
web-client/src/client/lib.rs

`
let mut config = builder.volatile().pico().build().expect("Build configuration failed");
`

You can also set consensus sync mode to 'pico' if you want to test it in a light client

## Pull request checklist

### DIFF
--- a/.github/devnet_topologies/validators_stress.toml
+++ b/.github/devnet_topologies/validators_stress.toml
@@ -42,3 +42,7 @@ sync_mode = "history"
 [[node]]
 restartable = false
 sync_mode = "light"
+
+[[node]]
+restartable = false
+sync_mode = "pico"
--- a/consensus/src/sync/light/mod.rs
+++ b/consensus/src/sync/light/mod.rs
@@ -4,4 +4,4 @@ mod sync_stream;
 #[cfg(feature = "full")]
 mod validity_window;
 
-pub use sync::LightMacroSync;
+pub use sync::{EpochIds, LightMacroSync, PeerMacroRequests};
--- a/consensus/src/sync/light/mod.rs
+++ b/consensus/src/sync/light/mod.rs
@@ -4,4 +4,4 @@ mod sync_stream;
 #[cfg(feature = "full")]
 mod validity_window;
 
-pub use sync::{EpochIds, LightMacroSync, PeerMacroRequests};
+pub use sync::LightMacroSync;
--- a/consensus/src/sync/light/sync.rs
+++ b/consensus/src/sync/light/sync.rs
@@ -197,7 +197,7 @@ impl<TNetwork: Network> MacroSync<TNetwork::PeerId> for LightMacroSync<TNetwork>
         unimplemented!()
     }
 
-    fn collect_peers(&mut self) -> Vec<TNetwork::PeerId> {
+    fn drain_peers(&mut self) -> Vec<TNetwork::PeerId> {
         unimplemented!()
     }
 }
--- a/consensus/src/sync/light/sync.rs
+++ b/consensus/src/sync/light/sync.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{HashMap, HashSet, VecDeque},
+    collections::{HashMap, HashSet},
     sync::Arc,
 };
 
@@ -20,42 +20,14 @@ use nimiq_zkp_component::{
 use parking_lot::RwLock;
 
 use crate::{
-    messages::{BlockError, Checkpoint},
-    sync::syncer::MacroSync,
+    messages::BlockError,
+    sync::sync_interface::{EpochIds, MacroSync, PeerMacroRequests},
 };
 #[cfg(feature = "full")]
 use crate::{
     messages::{HistoryChunk, HistoryChunkError, RequestHistoryChunk},
     sync::{peer_list::PeerList, sync_queue::SyncQueue},
 };
-
-#[derive(Clone)]
-/// This struct is used to request Epochs IDs (hashes) from other peers
-/// in order to determine their macro chain state relative to us
-pub struct EpochIds<T> {
-    /// Indicates if the latest epoch id that was queried was found in the peer's chain
-    pub locator_found: bool,
-    /// The most recent epoch ids (hashes)
-    pub ids: Vec<Blake2bHash>,
-    /// The most recent checkpoint block in the latest epoch (if any)
-    pub checkpoint: Option<Checkpoint>,
-    /// Epoch number corresponding to the first hash in ids
-    pub first_epoch_number: usize,
-    /// The sender that created this struct
-    pub sender: T,
-}
-
-impl<T> EpochIds<T> {
-    #[inline]
-    pub(crate) fn checkpoint_epoch_number(&self) -> usize {
-        self.first_epoch_number + self.ids.len()
-    }
-
-    #[inline]
-    pub(crate) fn last_epoch_number(&self) -> usize {
-        self.checkpoint_epoch_number().saturating_sub(1)
-    }
-}
 
 #[cfg(feature = "full")]
 /// Struct used to track the progress of the validity window chunk process.
@@ -70,68 +42,6 @@ pub struct ValidityChunkRequest {
     pub election_in_window: bool,
     /// Number of items in the previous requested chunk (for cases where we adopt a new macro head)
     pub last_chunk_items: Option<usize>,
-}
-
-/// This struct is used to track all the macro requests sent to a particular peer
-pub struct PeerMacroRequests {
-    /// Number of requests that have been fulfilled
-    completed_requests: usize,
-    /// A Queue used to track the requests that have been sent, and their respective result
-    queued_requests: VecDeque<(Blake2bHash, Option<Block>)>,
-}
-
-impl Default for PeerMacroRequests {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-impl PeerMacroRequests {
-    pub fn new() -> Self {
-        Self {
-            completed_requests: 0,
-            queued_requests: VecDeque::new(),
-        }
-    }
-
-    // Pushes a new request into the queue
-    pub fn push_request(&mut self, block_hash: Blake2bHash) {
-        self.queued_requests.push_back((block_hash, None))
-    }
-
-    // Pops a request from the queue
-    pub fn pop_request(&mut self) -> Option<(Blake2bHash, Option<Block>)> {
-        self.queued_requests.pop_front()
-    }
-
-    // Returns true if the request was updated, false in case the request was not found
-    pub fn update_request(&mut self, mut block: Block) -> bool {
-        let position = self
-            .queued_requests
-            .iter()
-            .position(|(hash, _)| *hash == block.hash_cached());
-
-        if let Some(position) = position {
-            if self.queued_requests[position].1.is_none() {
-                // A fulfilled request is only count once
-                self.completed_requests += 1;
-            }
-            // We update our block request.
-            // Note: If we receive a response more than once, we use the latest
-            let block_hash = block.hash_cached();
-            log::trace!(%block_hash, "Updating block request");
-            self.queued_requests[position] = (block_hash, Some(block));
-
-            true
-        } else {
-            log::trace!("Received a response for a block that we didn't expect");
-            false
-        }
-    }
-
-    // Returns true if all the requests have been completed
-    pub fn is_ready(&self) -> bool {
-        self.queued_requests.len() == self.completed_requests
-    }
 }
 
 #[cfg(feature = "full")]
--- a/consensus/src/sync/light/sync.rs
+++ b/consensus/src/sync/light/sync.rs
@@ -1,7 +1,5 @@
-#[cfg(feature = "full")]
-use std::collections::HashSet;
 use std::{
-    collections::{HashMap, VecDeque},
+    collections::{HashMap, HashSet, VecDeque},
     sync::Arc,
 };
 
@@ -34,7 +32,7 @@ use crate::{
 #[derive(Clone)]
 /// This struct is used to request Epochs IDs (hashes) from other peers
 /// in order to determine their macro chain state relative to us
-pub(crate) struct EpochIds<T> {
+pub struct EpochIds<T> {
     /// Indicates if the latest epoch id that was queried was found in the peer's chain
     pub locator_found: bool,
     /// The most recent epoch ids (hashes)
@@ -80,6 +78,12 @@ pub struct PeerMacroRequests {
     completed_requests: usize,
     /// A Queue used to track the requests that have been sent, and their respective result
     queued_requests: VecDeque<(Blake2bHash, Option<Block>)>,
+}
+
+impl Default for PeerMacroRequests {
+    fn default() -> Self {
+        Self::new()
+    }
 }
 impl PeerMacroRequests {
     pub fn new() -> Self {
@@ -275,8 +279,15 @@ impl<TNetwork: Network> MacroSync<TNetwork::PeerId> for LightMacroSync<TNetwork>
 
     fn add_peer(&mut self, peer_id: TNetwork::PeerId) {
         info!(%peer_id, "Requesting zkp from peer");
-
         self.zkp_requests
             .push(Self::request_zkps(self.zkp_component_proxy.clone(), peer_id).boxed());
+    }
+
+    fn fallback(&mut self, _peers: Vec<TNetwork::PeerId>) {
+        unimplemented!()
+    }
+
+    fn collect_peers(&mut self) -> Vec<TNetwork::PeerId> {
+        unimplemented!()
     }
 }
--- a/consensus/src/sync/pico/sync.rs
+++ b/consensus/src/sync/pico/sync.rs
@@ -15,10 +15,7 @@ use nimiq_zkp_component::zkp_component::ZKPComponentProxy;
 
 use crate::{
     messages::BlockError,
-    sync::{
-        light::{EpochIds, PeerMacroRequests},
-        syncer::MacroSync,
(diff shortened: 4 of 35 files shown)


light-blockchain/src/blockchain.rs
```diff
@@ -79,6 +80,23 @@ impl LightBlockchain {
         }
     }
 
+    // Resets the light blockchain to its initial state
```
jsdanielh:
```suggestion
    /// Resets the light blockchain to its initial state
```

lib/src/config/config_file/mod.rs
```diff
@@ -271,6 +271,9 @@ pub enum SyncMode {
     Full,
     /// Light nodes use LightMacroSync + BlockLiveSync to reach consensus
     Light,
+    /// Pico nodes: They use PicoMacroSync optimistically
+    ///  if something fails, they fallback to LightMacroSync
```
```suggestion
    /// if something fails, they fallback to LightMacroSync
```

lib/src/config/config.rs
```diff
@@ -59,6 +59,9 @@ pub enum SyncMode {
     Full,
     /// Light nodes: They use LightMacroSync + BlockLiveSync
     Light,
+    /// Pico nodes: They use PicoMacroSync optimistically
+    ///  if something fails, they fallback to LightMacroSync
```
jsdanielh:
```suggestion
    /// if something fails, they fallback to LightMacroSync
```

light-blockchain/src/sync.rs
```diff
@@ -190,4 +191,71 @@ impl LightBlockchain {
 
         Ok(PushResult::Extended)
     }
+
+    // Pushes an election block for the pico sync, bypassing many checks...
```
jsdanielh:
```suggestion
    /// Pushes an election block for the pico sync, bypassing many checks
```

consensus/src/sync/light/mod.rs
```diff
@@ -4,4 +4,4 @@ mod sync_stream;
 #[cfg(feature = "full")]
 mod validity_window;
 
-pub use sync::LightMacroSync;
+pub use sync::{EpochIds, LightMacroSync, PeerMacroRequests};
```
jsdanielh: Why do we need to make it public?

consensus/src/sync/light/sync.rs
```diff
@@ -34,7 +32,7 @@ use crate::{
 #[derive(Clone)]
 /// This struct is used to request Epochs IDs (hashes) from other peers
 /// in order to determine their macro chain state relative to us
-pub(crate) struct EpochIds<T> {
+pub struct EpochIds<T> {
```
jsdanielh: Why do we need to make it public?

consensus/src/sync/light/sync.rs
```diff
@@ -254,6 +257,7 @@ impl<TNetwork: Network> LightMacroSync<TNetwork> {
     }
 
     pub fn remove_peer_requests(&mut self, peer_id: TNetwork::PeerId) {
+        self.syncing_peers.remove(&peer_id);
```
jsdanielh: I believe semantically, this function now maybe needs to be called `remove_peer` instead?

consensus/src/sync/pico/sync.rs
```diff
@@ -0,0 +1,116 @@
...
+            'static,
+            (
+                Result<Result<Block, BlockError>, RequestError>,
+                TNetwork::PeerId,
+            ),
+        >,
+    >,
+    /// Collection of peers we are currently pico macro syncing with
+    pub(crate) syncing_peers: HashSet<TNetwork::PeerId>,
+    /// Reference to the ZKP proxy used to interact with the ZKP componen
```
jsdanielh:
```suggestion
    /// Reference to the ZKP proxy used to interact with the ZKP component
```

consensus/src/sync/pico/sync.rs
```diff
@@ -0,0 +1,116 @@
...
+    pub(crate) zkp_component_proxy: ZKPComponentProxy<TNetwork>,
```
jsdanielh, line 55: Do we need the it for pico sync?

consensus/src/sync/pico/sync.rs
```diff
@@ -0,0 +1,116 @@
...
+            network_event_rx,
+            peer_requests: HashMap::new(),
+            epoch_ids_stream: FuturesUnordered::new(),
+            block_headers: Default::default(),
+            syncing_peers: HashSet::new(),
+            zkp_component_proxy,
+        }
+    }
+
+    pub fn remove_peer_requests(&mut self, peer_id: TNetwork::PeerId) {
```
jsdanielh, line 78: Could you add some rustdoc?

consensus/src/sync/pico/sync.rs
```diff
@@ -0,0 +1,116 @@
...
+        self.peer_requests.remove(&peer_id);
+        self.syncing_peers.remove(&peer_id);
+    }
+
+    pub fn disconnect_peer(&mut self, peer_id: TNetwork::PeerId, reason: CloseReason) {
```
jsdanielh, line 84: Could you add some rustdoc?

consensus/src/sync/light/sync.rs
```diff
@@ -275,8 +279,16 @@ impl<TNetwork: Network> MacroSync<TNetwork::PeerId> for LightMacroSync<TNetwork>
...
         self.zkp_requests
             .push(Self::request_zkps(self.zkp_component_proxy.clone(), peer_id).boxed());
     }
+
+    fn fallback(&mut self, _peers: Vec<TNetwork::PeerId>) {
+        unimplemented!()
+    }
+
+    fn peers(&mut self) -> Vec<TNetwork::PeerId> {
+        self.syncing_peers.drain().collect()
```
jsdanielh: With drain we're clearing the set. Why?

consensus/src/sync/pico/sync.rs
```diff
@@ -0,0 +1,116 @@
...
+                .boxed();
+        self.epoch_ids_stream.push(future);
+    }
+
+    fn fallback(&mut self, _peers: Vec<TNetwork::PeerId>) {
+        unimplemented!()
+    }
+
+    fn peers(&mut self) -> Vec<TNetwork::PeerId> {
+        self.syncing_peers.drain().collect()
```
jsdanielh, line 115: With drain we're clearing the set, why?

consensus/src/sync/pico/sync_requests.rs
```diff
@@ -0,0 +1,271 @@
...
+    messages::{BlockError, MacroChain, MacroChainError, RequestBlock, RequestMacroChain},
+    sync::{
+        light::{EpochIds, PeerMacroRequests},
+        pico::PicoMacroSync,
+        syncer::MacroSync,
+    },
+};
+
+impl<TNetwork: Network> PicoMacroSync<TNetwork> {
+    pub(crate) async fn request_epoch_ids(
```
jsdanielh, line 24: Could you add some `rustdoc`?

consensus/src/sync/pico/sync_requests.rs
```diff
@@ -0,0 +1,271 @@
...
+            }
+            Err(error) => {
+                log::warn!(%error, %peer_id, "Request macro chain failed");
+                network.disconnect_peer(peer_id, CloseReason::Error).await;
+                None
+            }
+        }
+    }
+
+    pub(crate) fn request_macro_headers(
```
jsdanielh, line 129: Could you add some `rustdoc`?

consensus/src/sync/pico/sync_requests.rs
```diff
@@ -0,0 +1,271 @@
...
+                debug!(
+                    our_epoch_number,
+                    peers_epoch_number,
+                    peer = %epoch_ids.sender,
+                    "Peer is behind"
+                );
+                return Some(epoch_ids.sender);
+            } else {
+                // Check that the epoch_id sent by the peer at our current epoch number corresponds to
+                // our accepted state. If it doesn't, the peer is on a "permanent" fork, so we ban it.
```
jsdanielh, line 159: Are we banning it?

consensus/src/sync/pico/sync_requests.rs
```diff
@@ -0,0 +1,271 @@
...
+                .boxed(),
+            );
+        }
+
+        self.peer_requests.insert(epoch_ids.sender, peer_requests);
+
+        None
+    }
+
+    pub async fn request_macro_block(
```
jsdanielh, line 247: Could you add some `rustdoc`?

consensus/src/sync/pico/sync_requests.rs
```diff
@@ -0,0 +1,271 @@
...
+                RequestBlock {
+                    hash,
+                    include_body: false,
+                },
+                peer_id,
+            )
+            .await
+    }
+
+    pub async fn request_macro_chain(
```
jsdanielh, line 265: Could you add some `rustdoc`?

consensus/src/sync/pico/sync_stream.rs
```diff
@@ -0,0 +1,253 @@
...
+                                            if Policy::genesis_block_number() == latest_block_number
+                                            {
+                                                log::info!("PicoSync: Pushing latest election");
+                                                // If we are currently at the genesis block we blindly push the election block we obtained
+                                                LightBlockchain::push_pico_election(
+                                                    blockchain,
+                                                    block.clone(),
+                                                )
+                                            } else {
+                                                // We dont accept other election blocks if we are not in the genesis.
```
jsdanielh:
```suggestion
                                                // We don't accept other election blocks if we are not in the genesis.
```

consensus/src/sync/pico/sync_stream.rs
```diff
@@ -0,0 +1,253 @@
...
+
+                                // Check if the block is still valid for us or if it is outdated before trying to apply it
+                                let push_result = match self.blockchain {
+                                    #[cfg(feature = "full")]
+                                    BlockchainProxy::Full(_) => {
+                                        panic!("Pico macro sync is not supported in the full blockchain")
+                                    }
+                                    BlockchainProxy::Light(ref light_blockchain) => {
+                                        let blockchain = light_blockchain.upgradable_read();
+
```
jsdanielh:
```suggestion
```

consensus/src/sync/pico/sync_stream.rs
```diff
@@ -0,0 +1,253 @@
...
+                                            } else {
+                                                // We dont accept other election blocks if we are not in the genesis.
+                                                // Although we could tolerate +/- 1 block differences
+                                                self.syncing_peers.remove(&peer_id);
+                                                return Poll::Ready(Some(
+                                                    MacroSyncReturn::Conflicting(peer_id),
+                                                ));
+                                            }
+                                        } else {
+                                            // We only accept macro blocks from the current epoch
```
jsdanielh, line 164: Shouldn't we check this is a macro block?

consensus/src/sync/pico/sync_stream.rs
```diff
@@ -0,0 +1,253 @@
...
+                                        );
+                                    }
+                                    Err(error) => {
+                                        log::warn!(
+                                            block_number = block.block_number(),
+                                            ?error,
+                                            %peer_id,
+                                            "Failed to push macro block",
+                                        );
+                                        // Since we cannot differentiate between a malicious peer and someone whos behind and
```
jsdanielh:
```suggestion
                                        // Since we cannot differentiate between a malicious peer and someone who's behind and
```

consensus/src/sync/pico/sync_stream.rs
```diff
@@ -0,0 +1,253 @@
...
+                                        // we dont know if the initial state (election block) that we obtained was good
```
jsdanielh:
```suggestion
                                        // we don't know if the initial state (election block) that we obtained was good
```

consensus/src/sync/pico/sync_stream.rs
```diff
@@ -0,0 +1,253 @@
...
+                                        // then we need to fallback to light macro sync.
+                                        self.syncing_peers.remove(&peer_id);
+                                        return Poll::Ready(Some(MacroSyncReturn::Conflicting(
+                                            peer_id,
+                                        )));
+                                    }
+                                }
+                            }
+                            //  At this point we applied all the pending requests from this peer
```
jsdanielh:
```suggestion
                            // At this point we applied all the pending requests from this peer
```

consensus/src/sync/syncer.rs
```diff
@@ -324,7 +331,15 @@ impl<N: Network, M: MacroSync<N::PeerId>, L: LiveSync<N>> Stream for Syncer<N, M
...
-                None => {}
+                Some(MacroSyncReturn::Conflicting(peer_id)) => {
+                    debug!(%peer_id, "Macro sync returned conflicting peer");
+                    let mut syncing_peers = self.live_sync.peers();
+                    syncing_peers.push(peer_id);
+                    syncing_peers.append(&mut self.macro_sync.peers());
+
+                    self.macro_sync.fallback(syncing_peers);
+                }
+                _ => {}
```
jsdanielh: Isn't it better to keep the `None` to make the compiler alert us if we add another `enum` variant?

jsdanielh:
Could you also consider adding a pico node to the devnet scenarios? I think it would be valuable to test it as well

consensus/src/sync/light/mod.rs
viquezclaudio: Because they need to be accesible from the pico variant, otherwise:

```
rror[E0432]: unresolved imports `crate::sync::light::EpochIds`, `crate::sync::light::PeerMacroRequests`
  --> consensus/src/sync/pico/sync_requests.rs:17:17
   |
17 |         light::{EpochIds, PeerMacroRequests},
   |                 ^^^^^^^^  ^^^^^^^^^^^^^^^^^ no `PeerMacroRequests` in `sync::light`
   |                 |
   |                 no `EpochIds` in `sync::light`
   |
note: these structs exist but are inaccessible
```

consensus/src/sync/light/sync.rs
viquezclaudio: Same as above

consensus/src/sync/light/sync.rs
viquezclaudio: Syncing peers is a HashSet, so the 'remove' function comes from the HashSet function name.
To change the name I would need to create a wrapper just for this purpose, and I dont think thats worth it

consensus/src/sync/light/sync.rs
viquezclaudio: Because when we call this function we want to obtain the peers to fallback to the other macro sync mechanism, so we dont want or need to keep any peer inside

consensus/src/sync/pico/sync.rs
viquezclaudio: Not for the pico sync, but when we fallback to the light macro sync we need to supply this reference

consensus/src/sync/pico/sync.rs
viquezclaudio: Because when we call this function we want to obtain the peers to fallback to the other macro sync mechanism, so we dont want or need to keep any peer inside

consensus/src/sync/pico/sync_requests.rs
viquezclaudio: No, we cannot ban it, because we dont know if the peer is outdated or if it is being malicious.
We can only know this once we fallback to light macro sync

consensus/src/sync/light/mod.rs
viquezclaudio: Another option would be to move those structs to some shared file between light and macro sync

jsdanielh: Yeah, to me it means they maybe don't belong here

consensus/src/sync/light/sync.rs
jsdanielh: But the whole crate should be able to access it

consensus/src/sync/pico/sync.rs
jsdanielh: Could you add it as part of the `rustdoc`?

consensus/src/sync/light/sync.rs
jsdanielh: So maybe this function should be name `drain_peers` instead

consensus/src/sync/pico/sync.rs
jsdanielh: So maybe this function should be name `drain_peers` instead. Right now from the name it looks like a getter but it's not

consensus/src/sync/pico/sync_requests.rs
jsdanielh: So maybe we need to update the comment?

.github/devnet_topologies/validators_stress.toml
```diff
@@ -42,3 +42,7 @@ sync_mode = "history"
 [[node]]
 restartable = false
 sync_mode = "light"
+
+[[node]]
+restartable = false
+sync_mode = "pico"
```
jsdanielh, line 48: I think you also need to add it to at least [here](https://github.com/nimiq/core-rs-albatross/blob/906088071bafb69266cd33d01602e69e0810d0bd/scripts/devnet/topology.py#L108).

jsdanielh (approved):
LGTM

PR #3286, opened 2025-02-12
Rollup of 2 pull requests

hrxi:
#3163, #3287.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -54,9 +54,9 @@ dependencies = [
 
 [[package]]
 name = "afl"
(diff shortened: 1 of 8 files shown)


jsdanielh:
@hrxi #3284 is now closed in favor of #3287. Could you maybe include that one instead?

jsdanielh:
Also because of that, GH believes #3163 is closed but it is not.

hrxi:
Thanks for explaining the situation. I guess I always need to put dependabot updates last.

PR #3299, opened 2025-02-18
Fix issue when theres nothing new to learn from a peer

viquezclaudio:
When processing epoch ids it could happen that by the time we requested the epoch ids and the time we received those we no longer have anything new to learn from the peer, so we need to emit the peer as good.
This caused the peer to hang within the macro-sync process, increasing the time to consensus

## Pull request checklist


hrxi (approved):
Looks good to me. :)

hrxi:
Rebased for merging.

PR #3302, opened 2025-02-24
build(deps): avoid duplicate crates

rex4539:


hrxi:
What's the motivation for this pull request?

It seems nothing in the dependencies changed as there was no update to Cargo.lock. Do you want to guard against future potential duplicate crate dependencies?

rex4539:
Correct :)

hrxi:
I think this makes it less obvious what crates (/crate versions) we depend on. Since this only guards against potential future duplicate crate dependencies, I'd be inclined to close this PR.

hrxi:
> I think this makes it less obvious what crates (/crate versions) we depend on. Since this only guards against potential future duplicate crate dependencies, I'd be inclined to close this PR.

Closing for this reason. Still thanks for your pull request. :)

PR #3305, opened 2025-02-27
RPC: Make blocks' body hash non-optional, as it's always provided

sisou:
This doesn't change anything in the RPC response, as before this field was always a `string` (never `null`), and that's the same now.


jsdanielh (approved):
LGTM

hrxi (approved):
Looks good to me, thanks.

hrxi:
Rebased for merging.

PR #3313, opened 2025-03-11
Update to Rust Edition 2024

sisou:
Requires Rust v1.85.0, so can only be merged when that version is 2 versions old, as per MSRV policy (this PR sets the MSRV in the CI test to 1.85.0).

- Remove now unnecessary `ref` and `ref mut` keywords in `match` patterns
- Makes sure the new keyword `gen` is properly escaped where unavoidable
- Satisfy new lifetime declaration rules

## Pull request checklist

### DIFF
--- a/.github/workflows/build+test.yml
+++ b/.github/workflows/build+test.yml
@@ -127,7 +127,7 @@ jobs:
         tool-cache: true
         large-packages: false
     - uses: actions/checkout@v4
-    - uses: dtolnay/rust-toolchain@stable
+    - uses: dtolnay/rust-toolchain@1.86.0
       with:
         targets: wasm32-unknown-unknown
     - uses: Swatinem/rust-cache@v2
--- a/.github/workflows/build+test.yml
+++ b/.github/workflows/build+test.yml
@@ -26,7 +26,7 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
-    - uses: dtolnay/rust-toolchain@1.83.0 # Also test our Rust MSRV here.
+    - uses: dtolnay/rust-toolchain@1.85.0 # Also test our Rust MSRV here.
     - uses: Swatinem/rust-cache@v2
     - run: cargo check --all-features --tests --benches
 
--- a/.github/workflows/npm_publish.yml
+++ b/.github/workflows/npm_publish.yml
@@ -13,7 +13,7 @@ jobs:
       id-token: write
     steps:
       - uses: actions/checkout@v4
-      - uses: dtolnay/rust-toolchain@stable
+      - uses: dtolnay/rust-toolchain@1.86.0
         with:
           targets: wasm32-unknown-unknown
       - uses: Swatinem/rust-cache@v2
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -158,7 +158,7 @@ ark-r1cs-std = { git = "https://github.com/paberr/r1cs-std", branch = "pb/fix-pe
 [workspace.package]
 version = "1.1.0"
 authors = ["The Nimiq Core Development Team <info@nimiq.com>"]
-edition = "2021"
+edition = "2024"
 homepage = "https://nimiq.com"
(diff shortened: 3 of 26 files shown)


nibhar:
Cool :+1:

sisou:
Apparently the formatter has some new rules now, @hrxi. Uppercase is now sorted first, for example.

Should we apply those new formatting rules, or can you configure them to not reorder all the imports?

hrxi:
> or can you configure them to not reorder all the imports?

You can [set `style_edition = "2021"` in rustfmt.toml](https://doc.rust-lang.org/edition-guide/rust-2024/rustfmt-style-edition.html), so we can upgrade these as separate steps at our convenience. I can open a PR.

rustfmt.toml
```diff
@@ -1,2 +1,3 @@
+style_edition = "2021"
```
jsdanielh: Already included in #3321.

rpc-interface/src/types.rs
```diff
@@ -319,7 +319,7 @@ impl Block {
             additional_fields: BlockAdditionalFields::Micro {
                 producer: Slot::from_block_hash(blockchain, &block_hash)?,
                 equivocation_proofs,
-                justification: micro_block.justification.map(Into::into),
+                justification: micro_block.justification,
```
jsdanielh: Might also get included by #3319.

jsdanielh (commented):
Missing the change of the MSRV check in [`.github/workflows/build%2Btest.yml#L29`](https://github.com/nimiq/core-rs-albatross/blob/albatross/.github/workflows/build%2Btest.yml#L29).

rustfmt.toml
hrxi: Will automatically go away when rebasing.

database/benches/hash_keys.rs
```diff
@@ -48,13 +48,13 @@ struct Address([u8; 20]);
 
 impl Distribution<Blake2bHash> for Standard {
     fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> Blake2bHash {
-        Blake2bHash(rng.gen())
+        Blake2bHash(rng.r#gen())
```
hrxi: Probably better to change to rand 0.9, where this is called `rng.random()`.

sisou: I started doing that on branch `rand-0.9` based on `albatross`, but there are some trait changes that I don't understand. If you could look at the compiler errors, that would be great! :pray:

sisou: #3335

rpc-interface/src/types.rs
sisou: #3324

hrxi:
> Requires Rust v1.85.0, so can only be merged when that version is 2 versions old, as per MSRV policy (this PR sets the MSRV in the CI test to 1.85.0).

This will happen on 2025-05-15, when Rust 1.87 becomes stable.

jsdanielh:
Rebased branch. Now that stable is 1.87, we can merge this PR

jsdanielh (approved):

sisou:
New WASM incompatibility. Seems similar to what we had before: https://github.com/rustwasm/wasm-pack/issues/1441#issuecomment-2886314925

jsdanielh:
> New WASM incompatibility. Seems similar to what we had before: [rustwasm/wasm-pack#1441 (comment)](https://github.com/rustwasm/wasm-pack/issues/1441#issuecomment-2886314925)

Yeah, let me try using the workaround described there

jsdanielh:
> > New WASM incompatibility. Seems similar to what we had before: [rustwasm/wasm-pack#1441 (comment)](https://github.com/rustwasm/wasm-pack/issues/1441#issuecomment-2886314925)
> 
> Yeah, let me try using the workaround described there

The workaround described there didn't work. There is probably more incompatibilities than the ones described so I ended up setting the rust toolchain to 1.86.0 in the meantime.

PR #3319, opened 2025-03-12
More initial seed node connections

nibhar:
Previous to this PR the client will only attempt one single seed node connection in order to have a better discovered peer set as the next connections. With this PR a new configuration parameter is introduced to set a specific count. 
The default is set to the default of the required peers to establish consensus + 1 such that consensus should be established easier and more quickly.

### DIFF
--- a/network-libp2p/src/connection_pool/behaviour.rs
+++ b/network-libp2p/src/connection_pool/behaviour.rs
@@ -46,6 +46,8 @@ struct Limits {
 pub(crate) struct Config {
     /// Desired count of peers
     pub desired_peer_count: usize,
+    /// The number of initial attempted connections. These will connect to seed nodes.
+    pub num_initial_connections: usize,
     /// Maximum count of peers
     pub peer_count_max: usize,
     /// Maximum peer count per IP
@@ -83,6 +85,7 @@ impl Default for Config {
             ipv4_subnet_prefix_len: 24,
             ipv6_subnet_prefix_len: 96,
             dialing_count_max: 3,
+            num_initial_connections: 1,
             retry_down_after: Duration::from_secs(60 * 10), // 10 minutes
             housekeeping_interval: Duration::from_secs(60 * 2), // 2 minutes
         }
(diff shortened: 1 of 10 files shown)


network-libp2p/src/connection_pool/behaviour.rs
```diff
@@ -83,6 +85,7 @@ impl Default for Config {
             ipv4_subnet_prefix_len: 24,
             ipv6_subnet_prefix_len: 96,
             dialing_count_max: 3,
+            num_initial_connections: 1,
```
sisou, line 88: Why `1` here, when the default is `4`?

nibhar: This is the structures default. It is different from the config file default. If I would put a 4 here I would have to change it if the default changes here as well. 

This number is overwritten when we use it here: https://github.com/nimiq/core-rs-albatross/blob/nibhar/more-seed-nodes/network-libp2p/src/behaviour.rs#L90-L97

1 as the structures default did make sense to me as it is the minimum viable number and will also work for example in a test scenario where there is not even 3 total nodes to begin with.

It could be changed I guess. The default implementation could probably removed for example making only one spot the actual default.

jsdanielh (approved):
LGTM

hrxi:
This sounds like it'd make the network more centralized since every node will now have 4 seed connections, also quadrupling the amount of connections a seed node has to handle. How does this interact with maximum connection limits, does this mean seed nodes might start refusing connections more because they're above the limit?

nibhar:
> This sounds like it'd make the network more centralized since every node will now have 4 seed connections, also quadrupling the amount of connections a seed node has to handle. How does this interact with maximum connection limits, does this mean seed nodes might start refusing connections more because they're above the limit?

Yes. Getting to consensus a fair bit quicker was deemed worth this tradeof. I would think however that we can make this more economic by cycling seed nodes out of the peer connections later in the process, increasing the decentralization again once the address book has been populated.

jsdanielh:
Rebasing branch to merge it

PR #3324, opened 2025-03-13
Fix a small clippy warning

hrxi:
Split off from #3319.


hrxi (commented):
Since it's not my commit, I'm going to review it.

Looks good to me. :)

PR #3330, opened 2025-03-18
bls: Decrease log level of public key decompression

jsdanielh:
Decrease log level (from info to debug) of the BLS public key decompression.

## Pull request checklist


viquezclaudio (approved):

hrxi:
They're `info` because they're somewhat important on the web client. Decompressing a BLS public key is an action that takes ~100 ms so it's nice to see a log message for it on the web client, probably even on the normal client. Most actions that get logged take far less time.

PR #3331, opened 2025-03-18
web-client: Set num_initial_connections to default 4

sisou:
Fix an issue with the config defaults where a config setting introduced in #3319 didn't use it's builder-default value, but is instead initialized to `0`, which means the client does not attempt to connect to any seed.

@nibhar will be working on a proper fix of the builder system to make this change unnecessary.


jsdanielh (approved):
LGTM

PR #3333, opened 2025-03-18
Make release workflows able to run when a tag is pushed

jsdanielh:
Make release workflows able to run when a tag is pushed instead of when the GH release is created. The creation event as it was before is not performed for instance when a draft release is published.

## Pull request checklist

### DIFF
--- a/.github/workflows/npm_publish.yml
+++ b/.github/workflows/npm_publish.yml
@@ -1,7 +1,8 @@
 name: Publish web-client to npm
 on:
-  release:
-    types: [created]
+  push:
+    tags:
+      - 'v*'
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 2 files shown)


sisou (changes requested):
The RPC docs workflow has the same issue. In the docker workflow, the condition is actually 

```
on:
  push:
    tags:
      - 'v*'
  workflow_dispatch:
```

It would be good to standardize on one style I think.

jsdanielh:
> The RPC docs workflow has the same issue. In the docker workflow, the condition is actually
> 
> ```
> on:
>   push:
>     tags:
>       - 'v*'
>   workflow_dispatch:
> ```
> 
> It would be good to standardize on one style I think.

Yeah, agree. I like more the tag approach.

sisou (approved):

hrxi:
#3333 🎉

PR #3335, opened 2025-03-19
Upgrade to rand 0.9

hrxi:
Since `ark-*` hasn't upgraded to rand 0.9 yet, use a compatibility wrapper for code interacting with it.


hrxi:
Ready for review. :)

transaction-builder/Cargo.toml
```diff
@@ -30,7 +30,7 @@ nimiq-transaction = { workspace = true }
 
 [dev-dependencies]
 hex = "0.4"
-rand = "0.8"
+rand = "0.9"
```
paberr: Just noticed that some of the crates only have changes in the Cargo.toml.
Do they actually use the dependency at all?

paberr (commented):
Generally looks good to me.
I tried to make sure that we don't rely on any rng in consensus relevant code (we have this VRF seed based rng) but it seems all is good.

Just double-checked: we implement our own Rng trait for that. So, no dependency on the rng crate. ✅

transaction-builder/Cargo.toml
hrxi: No, apparently not. Removed them.

jsdanielh (approved):
LGTM

jsdanielh:
Rebasing branch for merge

PR #3338, opened 2025-03-21
Dont request zkp when in pico mode

viquezclaudio:
When we reach consensus, we request ZKP proofs to our peers.
This is not necessary for Pico nodes since the Pico syncing mechanism
does not rely on ZKP proofs.
This could cause problems since the Pico nodes do not always have
the election block associated to the latest ZKP proof

...
## Pull request checklist


jsdanielh (approved):
LGTM

PR #3339, opened 2025-03-21
Avoid constant zkp requests

ii-cruz:
Before this PR we were constantly requesting zkp proofs on the consensus poll function. This behavior was delaying web nodes, so in #3338 we removed it completely from the pico nodes.
This PR changes the zkp request to be only executed for history nodes when performing head requests. It also avoids requesting proofs if we have a sufficiently close proof (latest election block or its predecessor).

## Pull request checklist


viquezclaudio (commented):
LGTM
Tested with light and pico sync (web nodes) using mainnet.

jsdanielh (approved):
LGTM

PR #3343, opened 2025-03-28
Use binary search from std, remove dead code

hrxi:


paberr (approved):

primitives/src/slots_allocation.rs
```diff
@@ -142,24 +142,21 @@ impl Validators {
...
+                if validator.slots.end <= slot {
+                    Ordering::Less
+                } else if validator.slots.start > slot {
+                    Ordering::Greater
+                } else {
+                    Ordering::Equal
+                }
+            })
+            .ok()
+            .expect("validator slot band not found")
```
jsdanielh, line 157: Could you fix the clippy warning?

hrxi: Fixed by ignoring it, it was a false-positive.

jsdanielh (approved):
LGTM

PR #3350, opened 2025-04-03
Fix failed to publish asset for OpenRPC workflow

Eligioo:
This commit partially reverts #3333. Specifically the workflow for publishing the OpenRPC document on a new release. The problem now is that with trigger `on.push.tags` there is a possibility that the actual release on Github isn't yet created. Therefore, after generating the JSON file, uploading files to this non-existent release fails.

To mitigate the problem that #3333 tried to solve, I've changed the trigger back to `on.release.types` for the OpenRPC workflow. However, before the PR it was `[created]` and now I set this to `[published]`. This makes sure that even if the release initially was published as a draft, the workflow is still triggered.

...
## Pull request checklist

### DIFF
--- a/.github/workflows/publish_openrpc_document.yml
+++ b/.github/workflows/publish_openrpc_document.yml
@@ -1,10 +1,11 @@
 name: Publish OpenRPC document
 
+# This workflow only works with the 'on.release'-trigger.
+# On.release ensures a new release exists when artifacts from this workflow are being uploaded to the release.
+# On.tags can get triggered without the creation of a new release thus causing uploads to fail.
 on:
-  push:
-    tags:
-      - 'v*'
-  workflow_dispatch:
+  release:
+    types: [published]
 
 jobs:
   build_release_binary_and_publish_document:


.github/workflows/publish_openrpc_document.yml
```diff
@@ -1,10 +1,8 @@
 name: Publish OpenRPC document
 
 on:
-  push:
-    tags:
-      - 'v*'
-  workflow_dispatch:
+  release:
+    types: [published]
```
jsdanielh, line 8: I'd keep the `workflow_dispatch` just in case we need to manually trigger it

.github/workflows/publish_openrpc_document.yml
```diff
@@ -1,10 +1,8 @@
 name: Publish OpenRPC document
 
 on:
-  push:
-    tags:
-      - 'v*'
-  workflow_dispatch:
+  release:
```
jsdanielh, line 7: I'd add a comment that it needs to be in release bc the artifact uploading section.

.github/workflows/publish_openrpc_document.yml
Eligioo: Other workflow triggers other than `on.release` will fail because the action relies on specific webhook data that's only available with `on.release`. This is why I also removed the `workflow_dispatch`.

.github/workflows/publish_openrpc_document.yml
Eligioo: Could you clarify this?

jsdanielh: This workflow can never work with tags since tags are created before having a release and since the last step of this workflow uploads artifacts to the release, the tag approach won't ever work

jsdanielh (approved):

PR #3362, opened 2025-04-10
Fix zkp binary for proof generation tests

ii-cruz:
With the change of zkp meta data to json formatting and the introduction of the genesis data from the PoW migration the zkp binary for generating 2 consecutive proofs for mainnet no longer worked.

This PR adapts the code to work for the zkp binary by adapting the `Main` Network ID. This network id now has the same genesis data as the unit tests but it uses the zkp setup from mainnet. This allows us to test the zk proof generation without needing the migrated PoW db.

## Pull request checklist


viquezclaudio:
Closing because ZKP was removed

PR #3382, opened 2025-05-06
Ensure macro blocks adhere to the block separation time of 1s

### LINKED ISSUE #3296: Investigate artificial block delay for macro blocks

sisou:
## Pull request checklist

### DIFF
--- a/validator/src/tendermint.rs
+++ b/validator/src/tendermint.rs
@@ -277,8 +277,11 @@ where
             None
         };
 
-        // Create the proposal.
-        let time = blockchain.time.now();
+        // Create the proposal. Ensure macro block observes the block separation time.
+        let time = u64::max(
+            blockchain.timestamp() + Policy::BLOCK_SEPARATION_TIME,
+            blockchain.time.now(),
+        );
         let block = self
             .block_producer
             .next_macro_block_proposal(&blockchain, time, round, vec![], version)


sisou:
Devnet Release Scenario run: https://github.com/nimiq/core-rs-albatross/actions/runs/14704410843

jsdanielh (commented):
Shouldn't we add a check for this in `verify_macro_block_proposal`?

sisou:
No that would make it non-backward compatible. The beauty of this change is that old and new nodes will accept it (I think). Adding that check would make it a hard-fork.

jsdanielh:
> No that would make it non-backward compatible. The beauty of this change is that old and new nodes will accept it (I think). Adding that check would make it a hard-fork.

Mmm, then we could check for an acceptable range. Otherwise, we could accept any value there that could make the opposite effect.

hrxi:
Have you tested that this has the desired effect? By what mechanism does this introduce a one second delay?

Is it by making the next micro block producer wait two block separation times? Is the macro block not accepted/produced for another second due to this header change?

sisou:
I have not verified it. Would we worth a private testnet run to see if it works.

I _believe_ it produces the macro block directly (there's no waiting loop there like we have for the micro blocks) and then the next micro block waits 2 block times.

If that's really how it works, can you please check if there is a limit for what the macro block timestamp can be in a proposal? Can it be set to 1 year in the future?

nibhar:
I was looking into this again, and I believe this to be the actual easiest and smallest addition that can be made to make honest nodes adhere to the 1 second separation time floor.

> > No that would make it non-backward compatible. The beauty of this change is that old and new nodes will accept it (I think). Adding that check would make it a hard-fork.
> 
> Mmm, then we could check for an acceptable range. Otherwise, we could accept any value there that could make the opposite effect.

This is already in effect. Timestamps within `TIMESTAMP_MAX_DRIFT` are acceptable, and thus `max(parent+1s, now())` is an acceptable value. There is also little incentive other than malice to do this, as there is a batch delay penalty. Putting this into validation is a hard fork and I don't think it is needed. Could be folded into a future hard fork as a nice to have maybe.

> I _believe_ it produces the macro block directly (there's no waiting loop there like we have for the micro blocks) and then the next micro block waits 2 block times.

That would be my assessment as well
 
> If that's really how it works, can you please check if there is a limit for what the macro block timestamp can be in a proposal? Can it be set to 1 year in the future?

There is a `TIMESTAMP_MAX_DRIFT` constant that dictates this, but only for wall clock, as we do not have a network clock. currently this is 10 minutes which admittedly, is not exactly small, but also not huge either.

If there are no objections I will resurrect this PR and add a regression test to it.

viquezclaudio (approved):

sisou:
Rebased on top of `albatross`, thereby signing the commits.

validator/src/tendermint.rs
```diff
@@ -270,8 +270,19 @@ where
...
+        // micro block anchor in `expected_next_timestamp`, so an on-time batch keeps the block
+        // separation time without over-counting a late one. `now` is the fallback for a stalled
+        // chain, capped at the drift allowance so verifiers do not reject the proposal.
+        let now = blockchain.time.now();
+        let scheduled = blockchain.macro_head().header.timestamp.saturating_add(
+            Policy::BLOCK_SEPARATION_TIME.saturating_mul(Policy::blocks_per_batch() as u64),
+        );
+        let time = scheduled
+            .max(now)
+            .min(now.saturating_add(Policy::TIMESTAMP_MAX_DRIFT));
```
sisou, line 285: While reviewing this I was confused as to why you would reset the scheduled time back to `now` with `max(now)`. But the methods are just named counterintuitively. So this is logically right, but looks wrong to me.

sisou (commented):
Please sign your commits, @nibhar.

validator/src/tendermint.rs
jsdanielh: I don't understand the comment. How is `max(now)` counterintuitive?

jsdanielh (approved):
LGTM

validator/src/tendermint.rs
sisou: It reads to me like this code takes the `scheduled` time and sets it to maximally `now`. Which resets the 1-second-in-the-future `scheduled` back to the current time.

sisou: It's the difference between writing `i64::max(scheduled, now)` vs `scheduled.max(now)`. Intuitively, they have opposite effects. But I know now that it's the same.

PR #3384, opened 2025-05-07
Network: limit discovery message size to 1 MB

styppo:
* Limit discovery message size to 1 MB
* Resize message buffer incrementally as data is read


jsdanielh (approved):
LGTM

PR #3366, opened 2025-04-18
Switch back to upstream `tsify`

jsdanielh:
Switch back to upstream `tsify` now that the changes to copy over rust docs to TS are now merged. See
[this PR](https://github.com/madonoharu/tsify/pull/32).

## Pull request checklist

- [X] All tests pass. The project builds and runs.
- [X] I have resolved any merge conflicts.
- [X] I have resolved all `clippy` and `rustfmt` warnings.

PR #3387, opened 2025-05-08
Validate and fix web-client type declarations

sisou:
The latest npm release v2.1.0 has broken web-client type declarations. I tracked it down to a [change in the upstream `tsify` update](https://github.com/madonoharu/tsify/pull/44) which was not part of my fork. When we switched to the upstream version from my fork in #3366, this change was included and broke the types.

## Pull request checklist

### DIFF
--- a/.github/workflows/build+test.yml
+++ b/.github/workflows/build+test.yml
@@ -134,6 +134,9 @@ jobs:
     - name: Compile to wasm and generate bindings
       working-directory: ./web-client
       run: ./scripts/build.sh --only nodejs,types
+    - name: Validate type declarations
+      working-directory: ./web-client/extras
+      run: yarn tsc ../dist/types/wasm/bundler.d.ts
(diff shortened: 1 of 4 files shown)


jsdanielh (commented):
And question, why is that we don't use `AccountType` and `TransactionFormat`? i.e. why do we need something called `Plain...`?

sisou:
Because traditionally a `Transaction` object has the numbers as the account type and tx format, and the plain (human-readable) version (e.g. `transaction.toPlain()`) has the strings as the account type and format.

So I need both, although internally they are the same rust type.

The solution in this PR is shit, I know. Let me try tomorrow to come up with a better solution.

jsdanielh (approved):
LGTM

sisou:
@jsdanielh I fixed the test failure. Accidentally removed the serde rename_all.

jsdanielh (approved):

Issue #3376, opened 2025-04-30
`Address::from_user_friendly_address` can panic


hrxi:
It slices strings which panics if the byte indices don't fall between codepoints.

PR #3395, opened 2025-05-13
Add a fuzzer for user friendly addresses

hrxi:
This instantly finds crashes due to #3376.


fuzz/src/bin/00_testcases.rs
```diff
@@ -56,5 +57,19 @@ fn main() -> io::Result<()> {
     write("with_child", trie_node_with_child())?;
     write("with_value", trie_node_with_value())?;
 
+    let write = |name: &str, object: String| {
+        fs::write(
+            &format!("in/user_friendly_address/{name}"),
```
jsdanielh: Clippy warning:
```suggestion
            format!("in/user_friendly_address/{name}"),
```

hrxi: Fixed.

jsdanielh (approved):
LGTM

PR #3429, opened 2025-07-08
Make the `nimiq-client` version available via RPC and as CLI argument

### LINKED ISSUE #3390: Get `nimiq-client` version via CLI and RPC
Currently you have to go back to the final config log line when the `nimiq-client` starts or open the `Cargo.toml` and look for the package version. Being able to easily get the version via CLI and as RPC method makes things easier to find.

Eligioo:
#### This fixes #3390.

Retrieving the version now becomes available as:
- `nimiq-client --version`
- `getClientVersion` RPC method

## Pull request checklist

### DIFF
--- a/client/src/main.rs
+++ b/client/src/main.rs
@@ -15,7 +15,7 @@ pub use nimiq::{
     },
 };
 use nimiq_time::interval;
-use nimiq_utils::spawn;
+use nimiq_utils::{spawn, CLIENT_VERSION};
 
 async fn main_inner() -> Result<(), Error> {
     // Keep for potential future reactivation
@@ -38,6 +38,12 @@ async fn main_inner() -> Result<(), Error> {
             Some(&config_file.log)
         },
     )?;
+
+    // Simply log the Cargo package version a.k.a version of the client and return early
+    if command_line.version {
+        log::info!("Client version {}", CLIENT_VERSION);
+        return Ok(());
+    }
 
     // Initialize panic hook.
     initialize_panic_reporting();
--- a/lib/src/config/command_line.rs
+++ b/lib/src/config/command_line.rs
@@ -8,6 +8,7 @@ use thiserror::Error;
 use crate::config::config_file::SyncMode;
 
 #[derive(Debug, Parser)]
+#[command(name = "nimiq-client", version = nimiq_utils::CARGO_VERSION)]
 pub struct CommandLine {
     /// Use a custom configuration file.
     ///
--- a/lib/src/config/command_line.rs
+++ b/lib/src/config/command_line.rs
@@ -67,6 +67,10 @@ pub struct CommandLine {
     /// Internally used flag to start a zero-knowledge prover process.
     #[clap(long, action)]
     pub prove: bool,
+
+    /// Output the version of the client and exit immediately afterwards.
+    #[clap(long, short = 'v')]
+    pub version: bool,
 }
 
 impl CommandLine {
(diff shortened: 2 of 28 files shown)


utils/src/lib.rs
```diff
@@ -28,3 +28,6 @@ mod waker;
 #[cfg(feature = "spawn")]
 pub use self::spawn::{spawn, spawn_local};
 pub use self::{sensitive::Sensitive, waker::WakerExt};
+
+/// The version of the client based on the Cargo workspace package
+pub const CLIENT_VERSION: &str = env!("CARGO_PKG_VERSION");
```
hrxi, line 33: It's a bit weird to put this into `utils`, currently `nimiq-utils` doesn't have any concept of a client.

lib/src/config/command_line.rs
```diff
@@ -67,6 +67,10 @@ pub struct CommandLine {
     /// Internally used flag to start a zero-knowledge prover process.
     #[clap(long, action)]
     pub prove: bool,
+
+    /// Output the version of the client and exit immediately afterwards.
+    #[clap(long, short = 'v')]
```
hrxi, line 72: I'd remove the short option, `-v` is usually used for verbose logging.

client/src/main.rs
```diff
@@ -39,6 +39,12 @@ async fn main_inner() -> Result<(), Error> {
         },
     )?;
 
+    // Simply log the Cargo package version a.k.a version of the client and return early
+    if command_line.version {
+        log::info!("Client version {}", CLIENT_VERSION);
```
hrxi, line 44:
```suggestion
        log::info!("nimiq-client {}", CLIENT_VERSION);
```
I checked a couple of `--version` flags on my machine and most have the output `<PROGRAM> <VERSION>`.

lib/src/config/command_line.rs
jsdanielh: I'd second that

rpc-interface/src/network.rs
```diff
@@ -15,4 +15,7 @@ pub trait NetworkInterface {
 
     /// Returns a list with the IDs of all our peers.
     async fn get_peer_list(&self) -> RPCResult<Vec<String>, (), Self::Error>;
+
+    /// Returns the version of the local client.
```
jsdanielh, line 19: To me it's more the remote, but we can make it clear by saying:
```suggestion
    /// Returns the version of the client the RPC server is running.
```

utils/src/lib.rs
jsdanielh: Yeah, maybe `lib` instead?

jsdanielh:
I'd also add a versioning to the `rpc-client` and potentially any other binary we have

lib/src/config/command_line.rs
Eligioo: Alright, it depends on what you are used to I guess. `gcc`, `node` and `docker` from the top of my head use `-v` for the client version. But agree that being more explicit here doesn't hurt.

utils/src/lib.rs
Eligioo: I initially created the `const` in `lib` too but didn't want to introduce a new crate dependency for `rpc-server` just for this constant. So other suggestions are welcome or else `lib` works for me.

jsdanielh (approved):
LGTM

sisou:
Can it be made available on the web-client, too?

jsdanielh:
> Can it be made available on the web-client, too?

Will add it in #3806.

PR #3436, opened 2025-07-21
Fixes #3415

### LINKED ISSUE #3415: Cannot get network ID via RPC
The RPC server has no method to query the network ID of the network the node is configured for.

The only way currently is via `getLatestBlock`, which includes a `network` property that is either `MainAlbatross` or `TestAlbatross`.

But would be good to have a dedicated method.

viquezclaudio:
Directly exposes the network-id via RPC
Note: In the current albatross code the network-id is stored as one of the blockchain properties. This commit does not change that.

...
## Pull request checklist


jsdanielh (approved):
LGTM

PR #3445, opened 2025-07-30
Fix compilation and clippy warnings

ii-cruz:
Fixes the clippy warnings (elided lifetimes, removing nested if let clauses).
Increases the MSRV.

## Pull request checklist


hrxi:
The MSRV should not be increased to 1.88 before 2025-09-18, the day when 1.90 reaches stable.

PR #3454, opened 2025-08-12
web-client: Allow unpacking empty arrays

sisou:

It's currently not possible to create a `PartialSignature` from the web-client without any other public keys or commitments, e.g. for 1-of-X multisig wallets. The `unpack_*` methods in the web-client that convert a Javascript `Array` into a Rust `Vec` require(d) a non-empty list.

This requirement is not necessary and removing these checks fixes the 1-of-X use case.

## Pull request checklist


jsdanielh (approved):
LGTM

PR #3457, opened 2025-08-13
Create a new consensus sync tracker

### LINKED ISSUE #2815: Provide syncing state via RPC
The PoW node had a [`syncing` RPC method](https://github.com/nimiq/core-js/blob/master/clients/nodejs/modules/JsonRpcServer.js#L189-L193) that returned

```ts
{
    startingBlock: number,
    currentBlock: number,
    highestBlock: number
}
```

while syncing, and `false` when consensus is established.

Would it be possible to have this in PoS?

viquezclaudio:
Add a method that returns the status of the sync progress within consensus. Currently, it tracks:
 - Validity window synced
 - State sync progress
 - Current block/remaning blocks for live sync
 - Consensus established

Fixes #2815

...
## Pull request checklist

### DIFF
--- a/consensus/src/consensus/consensus_proxy.rs
+++ b/consensus/src/consensus/consensus_proxy.rs
@@ -23,6 +23,7 @@ use nimiq_network_interface::{
     request::{OutboundRequestError, RequestError},
 };
 use nimiq_primitives::{key_nibbles::KeyNibbles, policy::Policy};
+use nimiq_serde::{Deserialize, Serialize};
 use nimiq_transaction::{
     historic_transaction::HistoricTransaction, ControlTransaction, ControlTransactionTopic,
     Transaction, TransactionTopic,
@@ -32,7 +33,7 @@ use tokio_stream::wrappers::BroadcastStream;
 
 use super::{ConsensusRequest, ResolveBlockError, ResolveBlockRequest};
 use crate::{
-    consensus::remote_data_store::RemoteDataStore,
+    consensus::{remote_data_store::RemoteDataStore, SyncProgress},
     messages::{
         AddressNotification, AddressSubscriptionOperation, AddressSubscriptionTopic,
         RequestBlocksProof, RequestSubscribeToAddress, RequestTransactionReceiptsByAddress,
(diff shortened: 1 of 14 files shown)


consensus/src/consensus/consensus_proxy.rs
```diff
@@ -32,14 +33,27 @@ use tokio_stream::wrappers::BroadcastStream;
...
+
+/// This struct is used to track the progress of the consensus sync
+/// and return it to the user.
+#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+#[serde(rename_all = "camelCase")]
+pub struct ConsensusSyncStatus {
+    pub is_established: bool,
+    pub synced_validity_window: bool,
+    pub current_block: u32,
+    pub remaning_blocks: u32,
```
paberr: typo, please refactor
```suggestion
    pub remaining_blocks: u32,
```

consensus/src/consensus/mod.rs
```diff
@@ -132,6 +132,14 @@ pub struct ResolveBlockRequest<N: Network> {
 pub enum ConsensusRequest<N: Network> {
     ResolveBlock(ResolveBlockRequest<N>),
 }
+
+/// This struct is used to track the sync progress to reach consensus.
+pub struct SyncProgress {
+    pub current_block_number: AtomicU32,
+    pub remaning_blocks_to_sync: AtomicU32,
```
paberr: typo
```suggestion
    pub remaining_blocks_to_sync: AtomicU32,
```

paberr (changes requested):
only small typos, otherwise good

consensus/src/consensus/consensus_proxy.rs
```diff
@@ -78,6 +94,38 @@ impl<N: Network> ConsensusProxy<N> {
...
     }
 
+    /// Returns the current sync status of the consensus.
+    pub fn get_sync_status(&self) -> ConsensusSyncStatus {
+        let is_established = self.is_established();
+        let synced_validity_window = self.blockchain.read().can_enforce_validity_window();
+        let current_block = self
+            .sync_progress
+            .current_block_number
+            .load(Ordering::Acquire);
```
ii-cruz, line 104: These are matching a relaxed store op. Effectively, this acquisition is downgraded to relaxed. We either have Relaxed on both ends or an Acquire and Release for stronger consistency.
~~I would just have everything set to relaxed since this is only for statistical/informational purposes.~~

edit: After looking at the other progress variables, they are all release/acquire. This will force the relaxed to behave like release acquire, so I would keep the block tracking vars consistent. This means changing Relaxed to Release.

PR #3461, opened 2025-08-18
RPC: getTransactionReceiptsByAddress

### LINKED ISSUE #3208: RPC is missing getTransactionReceiptsByAddress
Could be a nice convenience solution if the user doesn't need all transactions, but just receipts.

viquezclaudio:
Fixes #3208

...
## Pull request checklist

### DIFF
--- a/rpc-server/src/dispatchers/blockchain.rs
+++ b/rpc-server/src/dispatchers/blockchain.rs
@@ -367,6 +367,45 @@ impl BlockchainInterface for BlockchainDispatcher {
         }
     }
 
+    async fn get_transactions_receipts_by_address(
+        &self,
+        address: Address,
+        max: Option<u16>,
+        start_at: Option<Blake2bHash>,
+    ) -> RPCResult<Vec<(Blake2bHash, u32)>, (), Self::Error> {
+        if let BlockchainReadProxy::Full(blockchain) = self.blockchain.read() {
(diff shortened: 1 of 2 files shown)


paberr (approved):
LGTM

rpc-server/src/dispatchers/blockchain.rs
```diff
@@ -367,6 +367,45 @@ impl BlockchainInterface for BlockchainDispatcher {
         }
     }
 
+    async fn get_transactions_receipts_by_address(
```
ii-cruz: Initially, I thought we wanted to get the internal receipts to revert txs due to the name of this function. Could we instead name this something like get last included txs or something similar?

rpc-server/src/dispatchers/blockchain.rs
```diff
@@ -367,6 +367,45 @@ impl BlockchainInterface for BlockchainDispatcher {
...
+                        .history_store
+                        .history_index()
+                        .unwrap()
+                        .get_hist_tx_by_hash(&hash, None)
+                        .iter()
+                        .map(|hist_tx| (hist_tx.tx_hash().into(), hist_tx.block_number)),
+                );
+            }
+
+            let mut receipts: Vec<_> = receipts.into_iter().collect();
```
ii-cruz:
```suggestion
```
I think this is not doing anything

PR #3466, opened 2025-08-22
web-client: check if txn is already included

### LINKED ISSUE #3273: web-client: send_web_transaction doesn't check if transaction has already been included
When sending the same transaction twice, `send_web_transaction` should check if it's already included and return without waiting for any timeout or address subscription.

When sending the same tx from the same client, it logs `Not publishing a message that has already been published` into the console but still waits for the 10 seconds timeout.

When sending the same tx from two different clients, the last one doesn't know that the tx was already included and waits for the 10 seconds timeout anyway.

viquezclaudio:
Fixes #3273

...
## Pull request checklist

### DIFF
--- a/web-client/src/client/lib.rs
+++ b/web-client/src/client/lib.rs
@@ -561,6 +561,16 @@ impl Client {
     ) -> Result<PlainTransactionDetails, JsError> {
         tx.verify(Some(self.network_id))?;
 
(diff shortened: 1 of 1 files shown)


jsdanielh (approved):
LGTM

ii-cruz (approved):
LGTM

PR #3479, opened 2025-09-10
Fix history sync on testnet & Simplifying history sync artificial checkpoint blocks

ii-cruz:
The extend history for syncing history nodes adds synthetic macro blocks if those are missing, so that it can recreate the `FinalizeBatch` inherents before pushing the history of an epoch. However, since we added the reward transactions to the body of the macro blocks, there is only one checkpoint block that can be empty - the first checkpoint block. This is because we pay rewards with a batch delay.
The pr #3758 did some hardening for the original assertion. However, it introduced a bug by rejecting the blocks when we have a gap of 2 batches. This shouldn't be the behaviour since we can have a valid code path for a gap, the case where we are in genesis block (batch 0) and batch 1 has no history, so the first item is on batch 2.

This PR:
- Fixes the introduced bug in 3758 
- Removes and fixes `extend_history_sync` FIX MEs and outdated comments
- Simplifies the addition of the synthetic missing checkpoint blocks to only do that when it's for checkpoint 1.
- Adds unit tests for this edge cases

## Pull request checklist

### DIFF
--- a/blockchain/src/blockchain/history_sync.rs
+++ b/blockchain/src/blockchain/history_sync.rs
@@ -199,54 +199,55 @@ impl Blockchain {
         let mut block_state = vec![];
         let mut block_transactions = vec![];
         let mut block_inherents = vec![];
-
-        let mut prev_batch = 0;
-        let mut prev_block = 0;
+        let mut prev_batch = this
+            .chain_store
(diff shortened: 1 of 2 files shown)


viquezclaudio (approved):
LGTM

jsdanielh (approved):
LGTM

Issue #3288, opened 2025-02-15
Burn skip block rewards.


ii-cruz:
These rewards are not emitted instead of being emitted and burned immediately.

ii-cruz:
This is not a real issue, rewards are burned.

PR #3494, opened 2025-10-02
Add more reward burn related testing

ii-cruz:
Investigated #3288 and concluded it wasn't a bug. The rewards are effectively emitted and burned; however, the comments were misleading, and there was a lack of unit tests.

This PR:
- Clarifies that `Inherent::Reward` and historic reward events represent both normal validator rewards and protocol-side reward burns.
- adds tests for the existing punished-slot burn behavior
- adds tests for validators that deliberately set their reward address to the burn address
- adds tests for validators whose configured reward address cannot accept reward inherents, causing rewards to be burned instead
- adds tests for the combined case where rewards are burned both because of penalized slots and because the configured reward address is not payable
- adds account-level coverage showing that burned rewards are credited to the burn address rather than to validator accounts

## Pull request checklist

### DIFF
--- a/blockchain/tests/inherents.rs
+++ b/blockchain/tests/inherents.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use nimiq_account::{BlockLogger, BlockState};
+use nimiq_account::{BlockLogger, BlockState, StakingContractStoreWrite, TransactionLog};
 use nimiq_block::{
     Block, DoubleProposalProof, DoubleVoteProof, ForkProof, MacroBlock, MacroBody, MacroHeader,
     SkipBlockInfo,
@@ -94,7 +94,7 @@ fn it_can_create_batch_finalization_inherents() {
             _ => panic!(),
         }
     }
-    assert!(got_reward && got_finalize_batch);
+    assert!(got_reward && got_finalize_batch && inherents.len() == 2);
 
     // Penalize one slot. Expect 1x FinalizeBatch, 1x Reward to validator, 1x Reward burn
     let penalize_inherent = Inherent::Penalize {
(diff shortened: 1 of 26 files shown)


blockchain/tests/inherents.rs
```diff
@@ -177,6 +177,391 @@ fn it_can_create_batch_finalization_inherents() {
...
+        .state
+        .accounts
+        .commit(
+            &mut (&mut txn).into(),
+            &[],
+            &[penalize_inherent],
+            &BlockState::new(
+                Policy::blocks_per_batch() + 1 + Policy::genesis_block_number(),
+                1,
+                Policy::max_supported_version()
```
nibhar: Macro header above was set to 1. What happens if this is 5?

`it_can_create_batch_finalization_inherents()` and `it_burns_all_rewards_when_all_slots_penalized()` uses `blockchain.protocol_version()` which I think is correct.

blockchain/tests/inherents.rs
```diff
@@ -177,6 +177,391 @@ fn it_can_create_batch_finalization_inherents() {
...
```
nibhar: Same here.

blockchain/tests/inherents.rs
nibhar, line 1: Looks to me these 3 added tests all share the same setup and they also share it with 2 existing tests. Can probably be extracted and thus simplified. Presentation only, but i think it would be worthwhile.

blockchain/tests/inherents.rs
```diff
@@ -177,6 +177,391 @@ fn it_can_create_batch_finalization_inherents() {
...
+                target: Address::burn_address(),
+                value: Coin::from_u64_unchecked(one_slot_reward),
+            },
+            Inherent::FinalizeBatch
+        ]
+    );
+}
+
+#[test]
+fn validator_reward_address_that_cannot_accept_inherents_burns_rewards() {
```
nibhar, line 410: Excellent addition!

blockchain/tests/inherents.rs
```diff
@@ -177,6 +177,391 @@ fn it_can_create_batch_finalization_inherents() {
     assert!(got_reward && got_penalize && got_finalize_batch);
 }
 
+#[test]
+fn validator_deliberately_burning_rewards_works() {
```
nibhar, line 357: Also nice!

blockchain/tests/inherents.rs
nibhar, line 1: Another really cool tests to add maybe as a follow up would be:

Multiple Validators A, B, C. A can be payed out to, B cannot, C is penalized. Test all of that works  as intended.

nibhar, line 1: Extract constants for re-use: `EXPECTED_REWARD`, `ONE_SLOT_REWARD`.

primitives/account/tests/accounts.rs
nibhar, line 1: Quite a bit of unrelated renaming inflating the diff. It does read better though.

primitives/account/tests/accounts.rs
```diff
@@ -228,24 +228,131 @@ fn it_correctly_rewards_validators() {
...
+    let value2 = Coin::from_u64_unchecked(7);
+    let fee2 = Coin::from_u64_unchecked(11);
+
+    let tx1 = Transaction::new_basic(
+        Address::burn_address(),
+        recipient_address_1.clone(),
+        value1,
+        fee1,
+        2,
+        NetworkId::MainAlbatross,
```
nibhar: Likely not the correct NetworkId. Probably not consequential.

primitives/transaction/src/historic_transaction.rs
```diff
@@ -320,7 +322,7 @@ impl MMRHash<Blake2bHash> for HistoricTransaction {
 pub enum HistoricTransactionData {
     /// A basic transaction. It simply contains the transaction as contained in the block.
     Basic(ExecutedTransaction),
-    /// A reward for an active validator.
+    /// A reward for an active validator or a reward burn caused by punishments.
```
nibhar, line 325: 👍

primitives/account/tests/accounts.rs
```diff
@@ -228,24 +228,131 @@ fn it_correctly_rewards_validators() {
...
+        Coin::from_u64_unchecked(0)
+    );
+
+    let value1 = Coin::from_u64_unchecked(5);
+    let fee1 = Coin::from_u64_unchecked(3);
+    let value2 = Coin::from_u64_unchecked(7);
+    let fee2 = Coin::from_u64_unchecked(11);
+
+    let tx1 = Transaction::new_basic(
+        Address::burn_address(),
```
nibhar: It would probably be better to not have these originate from the burn address.

primitives/account/tests/accounts.rs
nibhar: should this not get a similar treatment to the change made in `primitives/account/tests/staking_contract/validator.rs`?

nibhar (changes requested):
Generally good additions. Mostly small findings only. I think the version is probably the most important one to fix, and maybe originating transaction not from the burn address, but i would also like to see a small refactor of the test setup to make future test additions easier.

blockchain/tests/inherents.rs
ii-cruz: Cool! Added that test but also included a validator D with the reward address to the burn address since it's treated differently (normal reward payout to the burn address)

viquezclaudio (approved):

PR #3514, opened 2025-10-20
ci: Use npm trusted publishers instead of tokens

jsdanielh:
Use `npm` trusted publishers instead of tokens.

## Pull request checklist

### DIFF
--- a/.github/workflows/npm_publish.yml
+++ b/.github/workflows/npm_publish.yml
@@ -5,12 +5,13 @@ on:
       - 'v*'
   workflow_dispatch:
 
+permissions:
+  id-token: write  # Required for OIDC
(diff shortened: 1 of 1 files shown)


.github/workflows/npm_publish.yml
```diff
@@ -22,9 +26,11 @@ jobs:
         run: ./scripts/build.sh
       - uses: actions/setup-node@v4
         with:
+          node-version: '20'
```
sisou: Why this old version? LTS is 22, and 25 is current.

sisou: Maybe we can use `node-version: 'lts'`?

PR #3548, opened 2025-11-27
Add pkg.pr.new preview for web-client

onmax:
Adds preview packages via pkg.pr.new on PRs affecting web-client.

Requires installing GitHub App: https://github.com/apps/pkg-pr-new

Closes T404-78

### DIFF
--- a/.github/workflows/web_client_preview.yml
+++ b/.github/workflows/web_client_preview.yml
@@ -1,0 +1,28 @@
+name: Web Client Preview
+
+on:
+  pull_request:
+    paths:
+      - 'web-client/**'
+      - 'primitives/**'
+      - 'lib/**'
+  workflow_dispatch:
+
+jobs:
+  preview:
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dtolnay/rust-toolchain@stable
+        with:
+          targets: wasm32-unknown-unknown
+      - uses: Swatinem/rust-cache@v2
+      - name: Build package
+        working-directory: ./web-client
+        run: ./scripts/build.sh
+      - uses: actions/setup-node@v4
+      - name: Publish preview
+        run: npx pkg-pr-new publish './web-client/dist' --comment=update


hrxi:
Hmm. Does this allow anyone to publish preview packages in our name, just by opening a pull request?

onmax:
It does not public anything in our name. It will publish it in the registry of pkg.pr.new. Other projects like [vite](https://github.com/vitejs/vite/blob/3f344b4f13f867fa646e0562f62c7bed06549cb0/.github/workflows/preview-release.yml#L51) do this by default. Probably other projects do the same.

Alternavely, this can be also exploitable by anyone by just pushing the `dist` folder in a github repo and then using the gh link to install the package; we don't have any control over it.

---

To be clear, users in package.json will see something like:

```
{
  "name: "my-project",
  "scripts": {
   },
  "dependencies": {
      "@nimiq/core": "https://pkg.pr.new/XXXX" // `@nimiq/core` can be replaced with any name..
  }
}
```

---

Nonetheless, I do see your point (although i don't see it as an issue!), and I can make the Action to be triggered by any Nimiq Team members OR a whitelist of users like me, Soeren… People most likely to work on the web-client.

jsdanielh:
> It does not public anything in our name. It will publish it in the registry of pkg.pr.new. Other projects like [vite](https://github.com/vitejs/vite/blob/3f344b4f13f867fa646e0562f62c7bed06549cb0/.github/workflows/preview-release.yml#L51) do this by default. Probably other projects do the same.
> 
> Alternavely, this can be also exploitable by anyone by just pushing the `dist` folder in a github repo and then using the gh link to install the package; we don't have any control over it.
> 
> To be clear, users in package.json will see something like:
> 
> ```
> {
>   "name: "my-project",
>   "scripts": {
>    },
>   "dependencies": {
>       "@nimiq/core": "https://pkg.pr.new/XXXX" // `@nimiq/core` can be replaced with any name..
>   }
> }
> ```
> 
> Nonetheless, I do see your point (although i don't see it as an issue!), and I can make the Action to be triggered by any Nimiq Team members OR a whitelist of users like me, Soeren… People most likely to work on the web-client.

Mmm, I don't see an issue with this. At the end this will be package previews and not official package releases.

jsdanielh (approved):

PR #3551, opened 2025-12-01
Fix KeyPair.fromHex rejecting output from toHex

onmax:
## Bug

In `web-client/src/primitives/key_pair.rs`:
- [`serialize()`](https://github.com/nimiq/core-rs-albatross/blob/albatross/web-client/src/primitives/key_pair.rs#L66) adds extra `0x00` byte for "unlocked state" (not implemented)
- [`toHex()`](https://github.com/nimiq/core-rs-albatross/blob/albatross/web-client/src/primitives/key_pair.rs#L101) encodes this → 130 hex chars
- [`fromHex()`](https://github.com/nimiq/core-rs-albatross/blob/albatross/web-client/src/primitives/key_pair.rs#L39) parses `&hex[64..]` for public key → gets 66 chars instead of 64 → fails

https://t.me/c/1328779526/31057

## Fix

Ignore trailing bytes to have backward compatibility.

<details>
<summary>Reproduction steps</summary>

### 1. Create test script

```js
// test-keypair.js
const { KeyPair } = require('./dist/nodejs/main-wasm/index.js')

const kp1 = KeyPair.generate()
const hex1 = kp1.toHex()
console.log(`Generated KeyPair hex (${hex1.length} chars): ${hex1}`)

try {
  const kp2 = KeyPair.fromHex(hex1)
  console.log(`Roundtrip: PASS`)
  console.log(`Private keys match: ${kp1.privateKey.toHex() === kp2.privateKey.toHex() ? 'PASS' : 'FAIL'}`)
  console.log(`Public keys match: ${kp1.publicKey.toHex() === kp2.publicKey.toHex() ? 'PASS' : 'FAIL'}`)
} catch (e) {
  console.log(`FAIL - fromHex threw: ${e.message}`)
}
```

### 2. Reproduce bug on `albatross`

```bash
git checkout albatross
cd web-client
./scripts/build.sh --only nodejs --skip-wasm-opt --skip-launcher --skip-lib
node test-keypair.js
```

Expected output:
```
Generated KeyPair hex (130 chars): bdd3a8d5...c200
FAIL - fromHex threw: Invalid length when parsing byte slice.
```

### 3. Verify fix on this branch

```bash
git checkout fix/keypair-fromhex-padding
./scripts/build.sh --only nodejs --skip-wasm-opt --skip-launcher --skip-lib
node test-keypair.js
```

Expected output:
```
Generated KeyPair hex (130 chars): 547b9a70...1700
Roundtrip: PASS
Private keys match: PASS
Public keys match: PASS
```

</details>

### DIFF
--- a/web-client/extras/tests/KeyPair.test.ts
+++ b/web-client/extras/tests/KeyPair.test.ts
@@ -1,0 +1,32 @@
+import { KeyPair, PrivateKey } from '@nimiq/core';
+import { describe, expect, it } from 'vitest';
+
+describe('KeyPair', () => {
+    it('can roundtrip toHex/fromHex', () => {
+        const kp1 = KeyPair.generate();
+        const hex = kp1.toHex();
+        const kp2 = KeyPair.fromHex(hex);
+
+        expect(kp2.privateKey.toHex()).toBe(kp1.privateKey.toHex());
+        expect(kp2.publicKey.toHex()).toBe(kp1.publicKey.toHex());
+    });
+
+    it('accepts 128-char hex (no padding)', () => {
+        const kp1 = KeyPair.generate();
+        const hex = kp1.privateKey.toHex() + kp1.publicKey.toHex();
+
+        expect(hex.length).toBe(128);
+        const kp2 = KeyPair.fromHex(hex);
+        expect(kp2.privateKey.toHex()).toBe(kp1.privateKey.toHex());
+    });
+
+    it('rejects short hex strings', () => {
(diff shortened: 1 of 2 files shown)


sisou (approved):
I compared with the [JS implementation from PoW](https://github.com/nimiq/core-js/blob/300bad88b464391adf9c1e78135cae5204d8b69c/src/main/generic/consensus/base/primitive/KeyPair.js#L56), and that also only reads the expected length of the public key (then checks if more bytes are remaining). So this change is correct.

jsdanielh (approved):

PR #3557, opened 2025-12-10
Add keccak256 hash support

jsdanielh:
- Implement Keccak256 hash support in hash module.
- Implement keccak256 hash algorithm in HTLCs and in the rest of the subcrates.

## Pull request checklist


web-client/src/primitives/hash.rs
```diff
@@ -34,6 +34,14 @@ impl Hash {
             .serialize_to_vec()
     }
 
+    /// Computes a 32-byte [Keccak256] hash from the input data.
+    ///
+    /// [Keccak256]: https://github.com/ethereum/eth-hash
+    #[wasm_bindgen(js_name = computeKeccak256)]
+    pub fn compute_keccak256256(data: &[u8]) -> Vec<u8> {
```
sisou: Any reason for the double `256` in the method name?

sisou: But thanks for thinking of the web client! ❤️

jsdanielh: > Any reason for the double `256` in the method name?

Typo 😓

viquezclaudio (approved):
LGTM

primitives/transaction/tests/htlc_contract_verify.rs
```diff
@@ -159,6 +159,16 @@ fn it_can_verify_regular_transfer() {
...
 
+    // regular: valid Keccak-256
+    let proof = OutgoingHTLCTransactionProof::RegularTransfer {
+        hash_depth: 1,
+        hash_root: AnyHash::from(Keccak256Hasher::default().digest(&[0u8; 32])),
+        pre_image: PreImage::PreImage32(AnyHash32::from([0u8; 32])),
+        signature_proof: recipient_signature_proof.clone(),
+    };
+    tx.proof = proof.serialize_to_vec();
+    assert_eq!(AccountType::verify_outgoing_transaction(&tx), Ok(()));
```
paberr, line 170: I would have thought that this checks the hash but it doesn't seem like it.

paberr (approved):
I wonder whether it makes sense to include a `keccak_commutative(&[u8], &[u8])` variant or so. Depends a bit how we integrate it into the rest of the code.

primitives/transaction/src/account/htlc_contract.rs
```diff
@@ -378,6 +393,11 @@ impl OutgoingHTLCTransactionProof {
                             tmp_hash =
                                 PreImage::from(Sha512Hasher::default().digest(tmp_hash.as_bytes()));
                         }
+                        AnyHash::Keccak256(_) => {
+                            tmp_hash = PreImage::from(
+                                Keccak256Hasher::default().digest(tmp_hash.as_bytes()),
+                            )
```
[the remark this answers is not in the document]
jsdanielh: Although this is syntactically correct, for consistency it makes sense to apply it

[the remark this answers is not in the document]
jsdanielh: Actually this made me realize that for `PoW` we shouldn't be allowing keccak256 hashes and thus, this match arm should return an error instead

PR #3559, opened 2025-12-17
Add keccak256 history root computation and proof generation

jsdanielh:
Add keccak256 history root computation and proof generation. This functionality is gated by a config option disabled by default.
This implementation sorts leaves lexicographically at each level before hashing for compatibility with OpenZeppelin implementation.

## Pull request checklist

### DIFF
--- a/blockchain/src/history/history_store.rs
+++ b/blockchain/src/history/history_store.rs
@@ -10,7 +10,9 @@ use nimiq_database::{
     },
 };
 use nimiq_genesis::NetworkId;
-use nimiq_hash::{Blake2bHash, Hash, Hasher as HashHasher, Keccak256Hash, Keccak256Hasher};
+use nimiq_hash::{
+    Blake2bHash, Hash, HashOutput, Hasher as HashHasher, Keccak256Hash, Keccak256Hasher,
+};
 use nimiq_mmr::{
     error::Error as MMRError,
     mmr::{
@@ -919,6 +921,12 @@ impl HistoryInterface for HistoryStore {
         (first, last)
     }
 
+    /// Gets the Keccak256 history root for a macro block (election or checkpoint).
+    /// Returns None if the block is not a macro block or if there are no transactions.
+    ///
+    /// The Merkle tree is constructed from sorted transaction hashes to create a canonical
+    /// structure that doesn't require left/right metadata in proofs, making it more compatible
+    /// with Ethereum tooling.
     fn get_keccak256_history_root(
         &self,
         block_number: u32,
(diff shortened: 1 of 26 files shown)


blockchain/src/history/history_store.rs
```diff
@@ -917,10 +923,277 @@ impl HistoryInterface for HistoryStore {
...
+    ///
+    /// Proof generation requires building the entire Merkle tree:
+    /// - Small epochs (< 1,000 txs): ~2-10ms
+    /// - Medium epochs (1,000-10,000 txs): ~10-50ms
+    /// - Large epochs (> 10,000 txs): ~50ms+
+    ///
+    /// # Example Verification (Conceptual)
+    ///
+    /// ```ignore
+    /// let proof = history_store.get_keccak256_proof(block_number, tx_index, None)?;
```
viquezclaudio: isnt this commentary too verbose?
maybe remove the example as this can be essentially reviewed in an unit test..

lib/src/config/config_file/client.example.toml
```diff
@@ -155,6 +155,10 @@ sync_mode = "full"
 # Default: true when the sync_mode is "history" and false when the sync_mode is "full".
 #index_history = true
 
+# Enable or disable keccak256 transaction merkle tree root computation and proofs.
+# Default: false. To properly work, it requires index_history enabled.
+#keccak256_proofs = true
```
viquezclaudio, line 160: I think we should check when parsing the config file that if this is true, then index_history must be true as well, otherwise fail

rpc-interface/src/blockchain.rs
```diff
@@ -210,4 +216,209 @@ pub trait BlockchainInterface {
...
+    /// * `block_number` - The block number (must be a macro block - election or checkpoint)
+    ///
+    /// # Returns
+    /// * Hex-encoded Keccak256 root hash with "0x" prefix
+    ///
+    /// # Errors
+    /// * Returns error if block_number is not a macro block
+    /// * Returns error if no history exists for the block
+    ///
+    /// Validates: Requirements 6.1, 6.3, 6.4
```
viquezclaudio: I think this comment is too verbose compared to the other rpc comments

rpc-interface/src/blockchain.rs
```diff
@@ -210,4 +216,209 @@ pub trait BlockchainInterface {
...
+    ///       "recipient": "NQ...",
+    ///       "value": 1000000,
+    ///       ...
+    ///     }
+    ///   },
+    ///   "id": 1
+    /// }
+    /// ```
+    ///
+    /// # Verification Example (JavaScript/ethers.js)
```
viquezclaudio: same as before

viquezclaudio (approved):

PR #3580, opened 2026-01-16
Add keccak services/requests/messages

viquezclaudio:
Functionality to request and relay keccak256 proofs

## Pull request checklist

### DIFF
--- a/blockchain/src/history/history_store_index.rs
+++ b/blockchain/src/history/history_store_index.rs
@@ -654,6 +654,15 @@ impl HistoryIndexInterface for HistoryStoreIndex {
             .get_historic_tx(leaf.epoch_number, leaf.index, Some(&txn))
     }
 
+    /// Returns the epoch number and leaf index corresponding to the given transaction hash.
+    fn get_leaf_index_by_tx_hash(
+        &self,
+        raw_tx_hash: &Blake2bHash,
+        txn_option: Option<&MdbxReadTransaction>,
+    ) -> Option<EpochBasedIndex> {
+        self.get_leaf_indices_by_tx_hash(raw_tx_hash, txn_option)
+    }
+
     /// Returns a vector containing all transaction (and reward inherents) hashes corresponding to the given
     /// address. It fetches the transactions from most recent to least recent up to the maximum
     /// number given. It allows to give a starting point to fetch the transactions from (exclusive). If this hash is given
--- a/blockchain/src/history/interface.rs
+++ b/blockchain/src/history/interface.rs
@@ -16,6 +16,7 @@ use nimiq_transaction::{
 };
 use nimiq_utils::merkle::MerklePath;
 
+use super::utils::EpochBasedIndex;
 use crate::HistoryTreeChunk;
 
 /// Defines several methods to interact with a history store.
@@ -236,6 +237,16 @@ pub trait HistoryIndexInterface: HistoryInterface {
         raw_tx_hash: &Blake2bHash,
         txn_option: Option<&MdbxReadTransaction>,
     ) -> Option<HistoricTransaction>;
+
+    /// Returns the epoch number and leaf index corresponding to the given transaction hash.
+    ///
+    /// This is the original (unsorted) leaf index within the epoch, which can be used for
+    /// Keccak256 proof generation via [`HistoryInterface::get_keccak256_proof`].
+    fn get_leaf_index_by_tx_hash(
+        &self,
+        raw_tx_hash: &Blake2bHash,
+        txn_option: Option<&MdbxReadTransaction>,
+    ) -> Option<EpochBasedIndex>;
 
     /// Returns a vector containing all transaction (and reward inherents) hashes corresponding to the given
     /// address. It fetches the transactions from most recent to least recent up to the maximum
--- a/consensus/src/consensus/consensus_proxy.rs
+++ b/consensus/src/consensus/consensus_proxy.rs
@@ -36,8 +36,9 @@ use crate::{
     consensus::{remote_data_store::RemoteDataStore, SyncProgress},
     messages::{
         AddressNotification, AddressSubscriptionOperation, AddressSubscriptionTopic,
-        RequestBlocksProof, RequestSubscribeToAddress, RequestTransactionReceiptsByAddress,
-        RequestTransactionsProof, ResponseBlocksProof,
+        RequestBlocksProof, RequestKeccak256TransactionProof, RequestSubscribeToAddress,
+        RequestTransactionReceiptsByAddress, RequestTransactionsProof, ResponseBlocksProof,
+        ResponseKeccak256TransactionProof,
     },
     ConsensusEvent,
 };
@@ -257,6 +258,85 @@ impl<N: Network> ConsensusProxy<N> {
                 OutboundRequestError::NoReceiver,
             )),
         }
+    }
+
+    /// Requests a Keccak256-based Merkle proof for a transaction from multiple peers.
+    ///
+    /// This function requests a Keccak256 transaction proof for the given transaction hash
+    /// at the specified macro block number. The proof can be used to verify transaction
+    /// inclusion using Ethereum-compatible tools and libraries.
+    pub async fn request_keccak256_transaction_proof(
+        &self,
+        tx_hash: Blake2bHash,
+        macro_block_number: u32,
+        min_peers: usize,
+    ) -> Result<ResponseKeccak256TransactionProof, RequestError> {
+        // Verify that the block number corresponds to a macro block
+        if !Policy::is_macro_block_at(macro_block_number) {
+            return Err(RequestError::OutboundRequest(OutboundRequestError::Other(
+                format!(
+                    "Block number {} is not a macro block number",
+                    macro_block_number
+                ),
+            )));
+        }
+
+        // Get peers that support Keccak256 history
+        let peers = self
+            .get_peers_for_service(Services::KECCAK256_PROOFS, min_peers)
+            .await?;
+
+        // Try each peer until we get a valid response
+        for peer_id in peers {
+            log::debug!(
+                peer_id = %peer_id,
+                tx_hash = %tx_hash,
+                macro_block_number = %macro_block_number,
+                "Requesting Keccak256 transaction proof from peer",
+            );
+
+            let response = self
+                .network
+                .request::<RequestKeccak256TransactionProof>(
+                    RequestKeccak256TransactionProof {
+                        hash: tx_hash.clone(),
+                        block_number: macro_block_number,
+                    },
+                    peer_id,
+                )
+                .await;
+
+            match response {
+                Ok(Ok(proof_response)) => {
+                    log::debug!(
+                        peer_id = %peer_id,
+                        "Successfully obtained Keccak256 transaction proof from peer",
+                    );
+                    return Ok(proof_response);
+                }
+                Ok(Err(error)) => {
+                    log::debug!(
+                        peer_id = %peer_id,
+                        %error,
+                        "Peer could not provide Keccak256 transaction proof",
+                    );
+                    // Continue to next peer
+                }
+                Err(error) => {
+                    log::debug!(
+                        peer_id = %peer_id,
+                        %error,
+                        "Error requesting Keccak256 transaction proof from peer",
+                    );
+                    // Continue to next peer
+                }
+            }
+        }
+
+        // If we get here, all peers failed
+        Err(RequestError::OutboundRequest(
+            OutboundRequestError::NoReceiver,
+        ))
     }
 
     /// Returns peers that support all requested services (e.g. full blocks, history, mempool).
--- a/consensus/src/consensus/mod.rs
+++ b/consensus/src/consensus/mod.rs
@@ -50,7 +50,7 @@ use crate::{
 #[cfg(feature = "full")]
 use crate::{
     messages::{
-        RequestBatchSet, RequestBlocksProof, RequestHistoryChunk,
+        RequestBatchSet, RequestBlocksProof, RequestHistoryChunk, RequestKeccak256TransactionProof,
         RequestTransactionReceiptsByAddress, RequestTransactionsProof, RequestTrieProof,
     },
     sync::live::{diff_queue::RequestTrieDiff, state_queue::RequestChunk},
@@ -304,7 +304,13 @@ impl<N: Network> Consensus<N> {
                 let stream = network.receive_requests::<RequestChunk>();
                 spawn(Box::pin(request_handler(network, stream, blockchain)));
 
-                let supports_history_index = blockchain.read().history_store.supports_index();
+                let (supports_history_index, supports_keccak256_proofs) = {
+                    let blockchain = blockchain.read();
+                    (
+                        blockchain.history_store.supports_index(),
+                        blockchain.config.keccak256_proofs,
+                    )
+                };
 
                 // Only spawn these handlers if the history index is enabled.
                 if supports_history_index {
@@ -312,6 +318,11 @@ impl<N: Network> Consensus<N> {
                     spawn(Box::pin(request_handler(network, stream, blockchain)));
 
                     let stream = network.receive_requests::<RequestTransactionReceiptsByAddress>();
+                    spawn(Box::pin(request_handler(network, stream, blockchain)));
+                }
+
+                if supports_keccak256_proofs {
+                    let stream = network.receive_requests::<RequestKeccak256TransactionProof>();
                     spawn(Box::pin(request_handler(network, stream, blockchain)));
                 }
 
--- a/consensus/src/messages/handlers.rs
+++ b/consensus/src/messages/handlers.rs
@@ -664,6 +664,54 @@ impl<N: Network> Handle<N, Arc<RwLock<Blockchain>>> for RequestTransactionsProof
     }
 }
 
+#[cfg(feature = "full")]
+impl<N: Network> Handle<N, Arc<RwLock<Blockchain>>> for RequestKeccak256TransactionProof {
+    fn handle(
+        &self,
+        _peer_id: N::PeerId,
+        blockchain: &Arc<RwLock<Blockchain>>,
+    ) -> Result<ResponseKeccak256TransactionProof, ResponseKeccak256TransactionProofError> {
+        // Keccak256 proof generation requires a macro block context.
+        if !Policy::is_macro_block_at(self.block_number) {
+            return Err(ResponseKeccak256TransactionProofError::NotAMacroBlock(
+                self.block_number,
+            ));
+        }
+
+        let blockchain = blockchain.read();
+
+        // Keccak256 proof generation requires the history index (to map tx hash -> leaf index).
+        let history_index = blockchain
+            .history_store
+            .history_index()
+            .ok_or(ResponseKeccak256TransactionProofError::NotSupported)?;
+
+        let leaf = history_index
+            .get_leaf_index_by_tx_hash(&self.hash, None)
+            .ok_or(ResponseKeccak256TransactionProofError::TransactionNotFound)?;
+
+        let block = blockchain
+            .chain_store
+            .get_block_at(self.block_number, false, None)
+            .map_err(|_| {
+                ResponseKeccak256TransactionProofError::BlockNotFound(self.block_number)
+            })?;
+
+        let proof = blockchain
+            .history_store
+            .get_keccak256_proof(self.block_number, leaf.index as usize, None)
+            .ok_or(ResponseKeccak256TransactionProofError::CouldntProveInclusion)?;
+
+        let root = blockchain
+            .history_store
+            .get_keccak256_history_root(self.block_number, None)
+            .ok_or(ResponseKeccak256TransactionProofError::CouldntProveInclusion)?;
+
+        Ok(ResponseKeccak256TransactionProof { proof, root, block })
+    }
+}
+
+#[cfg(feature = "full")]
 impl RequestTransactionReceiptsByAddress {
     const MAX_RECEIPTS: u16 = 500;
 }
--- a/consensus/src/messages/mod.rs
+++ b/consensus/src/messages/mod.rs
@@ -17,7 +17,7 @@ use nimiq_block::{
 #[cfg(feature = "full")]
 use nimiq_blockchain::HistoryTreeChunk;
 use nimiq_blockchain_interface::Direction;
-use nimiq_hash::{Blake2bHash, Blake2sHash, Hash, HashOutput, SerializeContent};
+use nimiq_hash::{Blake2bHash, Blake2sHash, Hash, HashOutput, Keccak256Hash, SerializeContent};
 use nimiq_keys::Address;
 use nimiq_macros::test_max_req_size;
 use nimiq_mmr::mmr::proof::SizeProof;
@@ -30,6 +30,7 @@ use nimiq_serde::{Deserialize, Serialize, SerializedMaxSize};
 use nimiq_transaction::{
     historic_transaction::HistoricTransaction, history_proof::HistoryTreeProof,
 };
+use nimiq_utils::merkle::MerklePath;
 use thiserror::Error;
 
 use crate::error::SubscribeToAddressesError;
@@ -549,6 +550,7 @@ impl RequestCommon for RequestTransactionsProof {
     const MAX_REQUESTS: u32 = 1000;
 }
 
+/// Response containing the Keccak256 history root for a macro block.
 /// Returns the latest transactions for a given address. All the transactions
 /// where the given address is listed as a recipient or as a sender are considered. Reward
 /// transactions are also returned. It has an option to specify the maximum number of transactions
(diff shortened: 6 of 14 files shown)


network-interface/src/peer_info.rs
```diff
@@ -44,6 +44,9 @@ bitflags! {
 
         /// This node provides pre-genesis information.
         const PRE_GENESIS_TRANSACTIONS = 1 << 8;
+
+        /// This node provides keccak256 history root and proofs.
+        const KECCAK256_HISTORY = 1 << 9;
```
jsdanielh: I think it's better to call it proofs since we don't provide proofs for the whole history
```suggestion
        const KECCAK256_PROOFS = 1 << 9;
```

consensus/src/consensus/consensus_proxy.rs
```diff
@@ -259,6 +260,90 @@ impl<N: Network> ConsensusProxy<N> {
...
+    ///
+    /// # Returns
+    /// * `Ok(ResponseKeccak256TransactionProof)` - The proof, root, and block from a peer
+    /// * `Err(RequestError)` - If not enough peers are available or no valid response is received
+    ///
+    /// # Errors
+    /// This function will fail if:
+    /// - Not enough peers support `Services::KECCAK256_HISTORY`
+    /// - All peers return errors or cannot provide the proof
+    /// - The block number is not a macro block
```
jsdanielh: Comment is too long compared to the others we have

consensus/src/consensus/consensus_proxy.rs
```diff
@@ -259,6 +260,90 @@ impl<N: Network> ConsensusProxy<N> {
...
+    pub async fn request_keccak256_transaction_proof(
+        &self,
+        tx_hash: Blake2bHash,
+        block_number: u32,
```
jsdanielh:
```suggestion
        macro_block_number: u32,
```

consensus/src/consensus/consensus_proxy.rs
```diff
@@ -259,6 +260,90 @@ impl<N: Network> ConsensusProxy<N> {
...
+        min_peers: usize,
+    ) -> Result<ResponseKeccak256TransactionProof, RequestError> {
+        // Get peers that support Keccak256 history
+        let peers = self
+            .get_peers_for_service(Services::KECCAK256_HISTORY, min_peers)
+            .await?;
+
```
jsdanielh, line 288: Shouldn't we check that `block_number` is a macro block?

consensus/src/consensus/mod.rs
```diff
@@ -315,6 +317,14 @@ impl<N: Network> Consensus<N> {
                     spawn(Box::pin(request_handler(network, stream, blockchain)));
                 }
 
+                if supports_keccak256_proofs {
+                    let stream = network.receive_requests::<RequestKeccak256TransactionProof>();
+                    spawn(Box::pin(request_handler(network, stream, blockchain)));
+
+                    let stream = network.receive_requests::<RequestKeccak256HistoryRoot>();
+                    spawn(Box::pin(request_handler(network, stream, blockchain)));
```
jsdanielh: But I'm not seeing who calls it 🤔

consensus/src/consensus/mod.rs
```diff
@@ -305,6 +306,7 @@ impl<N: Network> Consensus<N> {
                 spawn(Box::pin(request_handler(network, stream, blockchain)));
 
                 let supports_history_index = blockchain.read().history_store.supports_index();
+                let supports_keccak256_proofs = blockchain.read().config.keccak256_proofs;
```
jsdanielh: Maybe it's worth to take the lock once:

```suggestion
let (supports_history_index, supports_keccak256_proofs) = { 
    let blockchain = blockchain.read();
   (blockchain.history_store.supports_index(), blockchain..config.keccak256_proofs)
};
```

blockchain/src/history/history_store_index.rs
```diff
@@ -654,6 +654,14 @@ impl HistoryIndexInterface for HistoryStoreIndex {
             .get_historic_tx(leaf.epoch_number, leaf.index, Some(&txn))
     }
 
+    fn get_leaf_index_by_tx_hash(
```
jsdanielh, line 658: Missing `rustdoc`

consensus/src/messages/mod.rs
```diff
@@ -549,6 +550,91 @@ impl RequestCommon for RequestTransactionsProof {
     const MAX_REQUESTS: u32 = 1000;
 }
 
+/// Response containing a Keccak256-based Merkle proof for a transaction.
+#[derive(Serialize, Deserialize)]
+pub struct ResponseKeccak256TransactionProof {
+    /// The Keccak256 Merkle proof (sorted tree, OpenZeppelin-compatible).
+    pub proof: MerklePath<Keccak256Hash>,
+    /// The Keccak256 root that this proof is expected to match.
+    pub root: Keccak256Hash,
```
jsdanielh: Why are we including the root in the proof response?

consensus/src/messages/mod.rs
```diff
@@ -549,6 +550,91 @@ impl RequestCommon for RequestTransactionsProof {
...
+/// epoch/batch up to that macro block.
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct RequestKeccak256TransactionProof {
+    pub hash: Blake2bHash,
+    pub block_number: u32,
+}
+
+impl RequestCommon for RequestKeccak256TransactionProof {
+    type Kind = RequestMarker;
+    const TYPE_ID: u16 = 219;
```
jsdanielh: Because of this, I'd place this code in the very end of this file

consensus/src/messages/mod.rs
```diff
@@ -549,6 +550,91 @@ impl RequestCommon for RequestTransactionsProof {
...
+#[derive(Clone, Debug, Deserialize, Error, Serialize)]
+pub enum ResponseKeccak256HistoryRootError {
+    #[error("keccak256 proofs are disabled on this node")]
+    NotSupported,
+    #[error("block {0} is not a macro block")]
+    NotAMacroBlock(u32),
+    #[error("Block not found: {0}")]
+    BlockNotFound(u32),
+    #[error("history root not found")]
+    HistoryNotFound,
```
jsdanielh: Is it possible to not find a history root? How?

consensus/src/messages/mod.rs
```diff
@@ -549,6 +550,91 @@ impl RequestCommon for RequestTransactionsProof {
...
+    #[error("unknown error")]
+    #[serde(other)]
+    Other,
+}
+
+/// Request the Keccak256 history root for a macro block (checkpoint or election).
```
jsdanielh: Theoretically we could do it for any block, couldn't we?

consensus/src/messages/mod.rs
```diff
@@ -549,6 +550,91 @@ impl RequestCommon for RequestTransactionsProof {
...
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct RequestKeccak256HistoryRoot {
+    pub block_number: u32,
+}
+
+impl RequestCommon for RequestKeccak256HistoryRoot {
+    type Kind = RequestMarker;
+    const TYPE_ID: u16 = 220;
```
jsdanielh: Same as above. Because of this, I'd place this at the end of the file

consensus/src/messages/handlers.rs
```diff
@@ -664,6 +664,82 @@ impl<N: Network> Handle<N, Arc<RwLock<Blockchain>>> for RequestTransactionsProof
...
+
+        let proof = blockchain
+            .history_store
+            .get_keccak256_proof(self.block_number, leaf.index as usize, None)
+            .ok_or(ResponseKeccak256TransactionProofError::CouldntProveInclusion)?;
+
+        let root = blockchain
+            .history_store
+            .get_keccak256_history_root(self.block_number, None)
+            .ok_or(ResponseKeccak256TransactionProofError::CouldntProveInclusion)?;
```
jsdanielh, line 708: I think the root is not necessary

consensus/src/messages/handlers.rs
```diff
@@ -664,6 +664,82 @@ impl<N: Network> Handle<N, Arc<RwLock<Blockchain>>> for RequestTransactionsProof
...
+}
+
+#[cfg(feature = "full")]
+impl<N: Network> Handle<N, Arc<RwLock<Blockchain>>> for RequestKeccak256HistoryRoot {
+    fn handle(
+        &self,
+        _peer_id: N::PeerId,
+        blockchain: &Arc<RwLock<Blockchain>>,
+    ) -> Result<ResponseKeccak256HistoryRoot, ResponseKeccak256HistoryRootError> {
+        if !Policy::is_macro_block_at(self.block_number) {
```
jsdanielh: I think theoretically we could provide it for any block

primitives/account/src/account/oracle_contract.rs
```diff
@@ -222,19 +223,71 @@ impl AccountTransactionInteraction for OracleContract {
...
-                    self.hashes.extend(hashes.clone());
+                    // Store the previous hash state for revert
+                    let previous_hash_state = if removed_hashes.is_empty() {
+                        self.hashes.last().cloned()
+                    } else {
+                        // If we removed hashes, the previous hash is the last one before removal
+                        // But since we already drained, we need to track it differently
+                        // Actually, if we removed hashes, the new "previous" is the last remaining hash
+                        self.hashes.last().cloned()
+                    };
```
jsdanielh: This `if` seems redundant. They both return the same thing

primitives/account/src/account/oracle_contract.rs
```diff
@@ -222,19 +223,71 @@ impl AccountTransactionInteraction for OracleContract {
...
+
+                    // Add the new hashes with chaining: hash_i = hash((hash_{i-1}, hash_i))
+                    use nimiq_transaction::account::htlc_contract::{AnyHash32, AnyHash64};
```
jsdanielh: Move this import to the beginning of the file

primitives/account/src/account/oracle_contract.rs
```diff
@@ -222,19 +223,71 @@ impl AccountTransactionInteraction for OracleContract {
...
+                    for new_hash in &hashes {
+                        // Get the previous hash (last in the list, or zero hash if empty)
+                        let previous_hash = self.hashes.last().cloned().unwrap_or_else(|| {
+                            // Use a zero hash of the same type as the new hash
+                            match new_hash {
+                                AnyHash::Blake2b(_) => AnyHash::Blake2b(AnyHash32::default()),
```
jsdanielh: Maybe it's worth to derive `default` in `AnyHash` to be able to call `default()`

primitives/account/src/account/oracle_contract.rs
```diff
@@ -222,19 +223,71 @@ impl AccountTransactionInteraction for OracleContract {
...
+                            )),
+                            AnyHash::Sha256(_) => AnyHash::Sha256(AnyHash32(
+                                Sha256Hasher::default().digest(&combined).into(),
+                            )),
+                            AnyHash::Sha512(_) => AnyHash::Sha512(AnyHash64(
+                                Sha512Hasher::default().digest(&combined).into(),
+                            )),
+                            AnyHash::Keccak256(_) => AnyHash::Keccak256(AnyHash32(
+                                Keccak256Hasher::default().digest(&combined).into(),
+                            )),
```
jsdanielh: Maybe implement a `digest` for `AnyHash`?

primitives/account/src/account/oracle_contract.rs
```diff
@@ -222,19 +223,71 @@ impl AccountTransactionInteraction for OracleContract {
...
+                                AnyHash::Blake2b(_) => AnyHash::Blake2b(AnyHash32::default()),
+                                AnyHash::Sha256(_) => AnyHash::Sha256(AnyHash32::default()),
+                                AnyHash::Sha512(_) => AnyHash::Sha512(AnyHash64::default()),
+                                AnyHash::Keccak256(_) => AnyHash::Keccak256(AnyHash32::default()),
+                            }
+                        });
+
+                        // Concatenate previous hash bytes with new hash bytes
+                        let mut combined = previous_hash.as_bytes().to_vec();
+                        combined.extend_from_slice(new_hash.as_bytes());
```
jsdanielh: Why are we doing this? I think we should be able to call `digest` on each hash.

primitives/account/tests/oracle_contract.rs
```diff
@@ -88,6 +88,48 @@ fn make_hash_keccak256(value: u8) -> AnyHash {
     AnyHash::from(Keccak256Hasher::default().digest(&[value; 32]))
 }
 
+fn compute_chained_hash(previous_hash: Option<&AnyHash>, new_hash: &AnyHash) -> AnyHash {
+    use nimiq_hash::{sha512::Sha512Hasher, Blake2bHasher, Hasher, Keccak256Hasher, Sha256Hasher};
+    use nimiq_transaction::account::htlc_contract::{AnyHash32, AnyHash64};
```
jsdanielh: Move this imports to the beginning of the file

primitives/account/tests/oracle_contract.rs
```diff
@@ -88,6 +88,48 @@ fn make_hash_keccak256(value: u8) -> AnyHash {
...
+
+    // Get the previous hash (or zero hash if None)
+    let prev = previous_hash.cloned().unwrap_or_else(|| match new_hash {
+        AnyHash::Blake2b(_) => AnyHash::Blake2b(AnyHash32::default()),
+        AnyHash::Sha256(_) => AnyHash::Sha256(AnyHash32::default()),
+        AnyHash::Sha512(_) => AnyHash::Sha512(AnyHash64::default()),
+        AnyHash::Keccak256(_) => AnyHash::Keccak256(AnyHash32::default()),
```
jsdanielh: If we derive `default` for `AnyHash`, we could use it here too

primitives/account/tests/oracle_contract.rs
```diff
@@ -88,6 +88,48 @@ fn make_hash_keccak256(value: u8) -> AnyHash {
...
+    });
+
+    // Concatenate previous hash bytes with new hash bytes
+    let mut combined = prev.as_bytes().to_vec();
+    combined.extend_from_slice(new_hash.as_bytes());
```
jsdanielh: I don't understand why are we doing this. I think we should be able to call `digest` on both hashes

primitives/account/tests/oracle_contract.rs
```diff
@@ -88,6 +88,48 @@ fn make_hash_keccak256(value: u8) -> AnyHash {
...
+        AnyHash::Sha256(_) => {
+            AnyHash::Sha256(AnyHash32(Sha256Hasher::default().digest(&combined).into()))
+        }
+        AnyHash::Sha512(_) => {
+            AnyHash::Sha512(AnyHash64(Sha512Hasher::default().digest(&combined).into()))
+        }
+        AnyHash::Keccak256(_) => AnyHash::Keccak256(AnyHash32(
+            Keccak256Hasher::default().digest(&combined).into(),
+        )),
+    }
```
jsdanielh: If we implement `digest` for `AnyHash` we should be able to call it here too

blockchain/src/history/interface.rs
```diff
@@ -237,6 +238,16 @@ pub trait HistoryIndexInterface: HistoryInterface {
         txn_option: Option<&MdbxReadTransaction>,
     ) -> Option<HistoricTransaction>;
 
+    /// Returns the epoch number and leaf index corresponding to the given transaction hash.
+    ///
+    /// This is the original (unsorted) leaf index within the epoch, which can be used for
+    /// Keccak256 proof generation via [`HistoryInterface::get_keccak256_proof`].
+    fn get_leaf_index_by_tx_hash(
```
jsdanielh, line 245: Now that you implemented this, we could also use it in `get_keccak256_transaction_proof` in `rpc-server/src/dispatchers/blockhain.rs`

consensus/src/consensus/mod.rs
viquezclaudio: who calls what?

consensus/src/messages/handlers.rs
viquezclaudio: to return the root?

consensus/src/messages/handlers.rs
viquezclaudio: isnt this inconsistent with the comment above that request_keccak256_transaction_proof should only be used with macro blocks?

consensus/src/messages/mod.rs
viquezclaudio: to verify that the proof we provide verifies against that root but i guess we could remove it

consensus/src/messages/mod.rs
viquezclaudio: HistoryNotFound can occur when an epoch has no transactions, even though a root could be computed.

consensus/src/messages/mod.rs
viquezclaudio: then why your comment above about restricting to just macro blocks?

consensus/src/consensus/mod.rs
jsdanielh: Where are we using this request?

jsdanielh: Where are we using this request?

consensus/src/messages/mod.rs
jsdanielh: I'd say either restrict to macro blocks but then check as early as possible that the block number is a macro block or allow it for any block. The latter is theoretically possible so I'd say we could just generically provide such access

consensus/src/messages/handlers.rs
jsdanielh: Yeah but to me it's strange to have it as part of the proof request. Since the root is a verification artifact, right now this is a proof + root request instead.

consensus/src/messages/mod.rs
jsdanielh: I'd remove it

consensus/src/messages/mod.rs
```diff
@@ -669,6 +700,24 @@ impl RequestCommon for RequestSubscribeToAddress {
...
+}
+
+impl RequestCommon for RequestKeccak256TransactionProof {
+    type Kind = RequestMarker;
+    const TYPE_ID: u16 = 219;
+    type Response =
+        Result<ResponseKeccak256TransactionProof, ResponseKeccak256TransactionProofError>;
+    const MAX_REQUESTS: u32 = 1000;
+}
+
```
jsdanielh: I'd have placed this at the very end of the file. Here it is cutting the `RequestSubscribeToAddress` related declarations

consensus/src/messages/mod.rs
```diff
@@ -549,6 +550,36 @@ impl RequestCommon for RequestTransactionsProof {
     const MAX_REQUESTS: u32 = 1000;
 }
 
+/// Response containing a Keccak256-based Merkle proof for a transaction.
```
jsdanielh: I'd have placed this above of the request declaration to be consistent with the rest of the file

primitives/transaction/src/account/htlc_contract.rs
```diff
@@ -122,6 +122,47 @@ impl AnyHash {
             AnyHash::Keccak256(hash) => &hash.0,
         }
     }
+
+    /// Returns a zero hash of the same type as this hash
+    pub fn zero(&self) -> Self {
```
jsdanielh, line 132: Wasn't it possible to derive `Default`?

primitives/transaction/src/account/htlc_contract.rs
```diff
@@ -122,6 +122,47 @@ impl AnyHash {
...
+    }
+
+    /// Hashes this hash together with another hash using the same algorithm as this hash.
+    /// Returns a new hash of the same type.
+    pub fn digest(&self, other: &Self) -> Self {
+        match self {
+            AnyHash::Blake2b(_) => {
+                let mut hasher = Blake2bHasher::default();
+                hasher.write_all(other.as_bytes()).unwrap();
+                hasher.write_all(self.as_bytes()).unwrap();
```
jsdanielh: The order matters: Shouldn't it be first `self` and then `other`?

primitives/transaction/src/account/htlc_contract.rs
```diff
@@ -122,6 +122,47 @@ impl AnyHash {
...
+        }
+    }
+
+    /// Hashes this hash together with another hash using the same algorithm as this hash.
+    /// Returns a new hash of the same type.
+    pub fn digest(&self, other: &Self) -> Self {
+        match self {
+            AnyHash::Blake2b(_) => {
+                let mut hasher = Blake2bHasher::default();
+                hasher.write_all(other.as_bytes()).unwrap();
```
jsdanielh: Why aren't we using `digest`?

primitives/transaction/src/account/htlc_contract.rs
```diff
@@ -122,6 +122,47 @@ impl AnyHash {
...
+                hasher.write_all(self.as_bytes()).unwrap();
+                AnyHash::Blake2b(AnyHash32(hasher.finish().into()))
+            }
+            AnyHash::Sha256(_) => {
+                let mut hasher = Sha256Hasher::default();
+                hasher.write_all(other.as_bytes()).unwrap();
+                hasher.write_all(self.as_bytes()).unwrap();
+                AnyHash::Sha256(AnyHash32(hasher.finish().into()))
```
jsdanielh: Same as above

primitives/transaction/src/account/htlc_contract.rs
```diff
@@ -122,6 +122,47 @@ impl AnyHash {
...
+            }
+            AnyHash::Sha512(_) => {
+                let mut hasher = Sha512Hasher::default();
+                hasher.write_all(other.as_bytes()).unwrap();
+                hasher.write_all(self.as_bytes()).unwrap();
+                AnyHash::Sha512(AnyHash64(hasher.finish().into()))
```
jsdanielh: Same as above

primitives/transaction/src/account/htlc_contract.rs
```diff
@@ -122,6 +122,47 @@ impl AnyHash {
...
+            }
+            AnyHash::Keccak256(_) => {
+                let mut hasher = Keccak256Hasher::default();
+                hasher.write_all(other.as_bytes()).unwrap();
+                hasher.write_all(self.as_bytes()).unwrap();
+                AnyHash::Keccak256(AnyHash32(hasher.finish().into()))
```
jsdanielh: Same as above

primitives/transaction/src/account/htlc_contract.rs
viquezclaudio: Because we need to preserve the variant type, Default would return Blake2b

primitives/transaction/src/account/htlc_contract.rs
viquezclaudio: Because Hasher::digest only takes a single &[u8], so it can't hash two byte slices in sequence. We need to hash self || other without allocating.

primitives/transaction/src/account/htlc_contract.rs
jsdanielh: Ah yes, ofc it's an `enum` 💡

primitives/transaction/src/account/htlc_contract.rs
jsdanielh: Ahh, sorry. Yes, I meant `chain`

primitives/account/src/account/oracle_contract.rs
```diff
@@ -222,19 +222,42 @@ impl AccountTransactionInteraction for OracleContract {
...
+                    for new_hash in &hashes {
+                        // Get the previous hash (last in the list, or zero hash if empty)
+                        let previous_hash = self.hashes.last().cloned().unwrap_or_else(|| {
+                            // Use a zero hash of the same type as the new hash
+                            new_hash.zero()
+                        });
+
+                        // Hash the previous hash and new hash using the same algorithm as the new hash
+                        // Chain: data_i = H(data_{i-1} || state_i) = H(previous_hash || new_hash)
+                        let chained_hash = previous_hash.digest(&new_hash);
```
jsdanielh: Clippy warning:
```suggestion
                        let chained_hash = previous_hash.digest(new_hash);
```

primitives/account/tests/oracle_contract.rs
```diff
@@ -88,6 +88,24 @@ fn make_hash_keccak256(value: u8) -> AnyHash {
...
+
+    // Hash the previous hash and new hash together using the same algorithm as the previous hash
+    // Chain: data_i = H(data_{i-1} || state_i) = H(previous_hash || new_hash)
+    prev.digest(new_hash)
+}
+
+fn compute_chained_hashes(hashes: &[AnyHash]) -> Vec<AnyHash> {
+    let mut result = Vec::new();
+    for hash in hashes {
+        let chained = compute_chained_hash(result.last(), hash);
```
jsdanielh: This looks a little overkill, we could have use `chain` instead

primitives/account/tests/oracle_contract.rs
```diff
@@ -327,11 +345,25 @@ fn it_implements_ring_buffer() {
...
+    // Compute expected: original chained hashes for indices 3-9, then new chained hashes
+    let expected_original = compute_chained_hashes(&initial_hashes);
+    let expected_remaining: Vec<AnyHash> = expected_original[3..].to_vec();
+    // New hashes are chained from the last remaining hash (index 9 of original)
+    let mut expected_new = Vec::new();
+    let mut prev_hash = expected_remaining.last().cloned();
+    for hash in &new_hashes {
+        let chained = compute_chained_hash(prev_hash.as_ref(), hash);
+        prev_hash = Some(chained.clone());
+        expected_new.push(chained);
```
jsdanielh: This looks a little overkilled, we could have used `chain` like it is done with `expected_all` below

primitives/transaction/src/account/htlc_contract.rs
```diff
@@ -122,6 +122,48 @@ impl AnyHash {
...
+    }
+
+    /// Hashes this hash together with another hash using the same algorithm as this hash.
+    /// Returns a new hash of the same type.
+    /// Computes H(self || other).
+    pub fn digest(&self, other: &Self) -> Self {
+        match self {
+            AnyHash::Blake2b(_) => {
+                let mut hasher = Blake2bHasher::default();
+                hasher.write_all(self.as_bytes()).unwrap();
```
jsdanielh: Isn't it better to use `chain` to avoid importing `std::io::Write`?

primitives/transaction/src/account/htlc_contract.rs
```diff
@@ -122,6 +122,42 @@ impl AnyHash {
...
+    }
+
+    /// Hashes this hash together with another hash using the same algorithm as this hash.
+    /// Returns a new hash of the same type.
+    /// Computes H(self || other).
+    pub fn digest(&self, other: &Self) -> Self {
+        let combined: Vec<u8> = self
+            .as_bytes()
+            .iter()
+            .chain(other.as_bytes())
```
jsdanielh: I was thinking on `chain` from the hasher. Something like:
`Blake2bHasher::default().chain(self.as_bytes()).chain(other.as_bytes()).finish();`

viquezclaudio: Ok there is a problem with `chain` because doing something like:
`Blake2bHasher::default().chain(self.as_bytes()).chain(other.as_bytes()).finish().into(),`

gives a compilation error because `chain<T: SerializeContent>(self, h: &T) uses a type parameter T that is Sized by default. self.as_bytes() is &[u8], so T is inferred as [u8], which is unsized, so the compiler errors.`

The fix would be to use the hasher's write API, as it currently is

jsdanielh: Ah yes, but then you could implement `SerializeContent` for `AnyHash`:
```
impl SerializeContent for AnyHash {
    fn serialize_content<W: io::Write, H: HashOutput>(&self, writer: &mut W) -> io::Result<()> {
        writer.write_all(self.as_bytes())?;
        Ok(())
    }
}
```

 And then:
```
Blake2bHasher::default()
    .chain(&self)
    .chain(&other)
    .finish()
    .into(),
```

jsdanielh (approved):
LGTM

PR #3590, opened 2026-02-02
Oracle ring buffer & interface

viquezclaudio:
Oracle ring buffer & RPC interface

Requires #3580 in order to be updated

## Pull request checklist


primitives/account/src/account/oracle_contract.rs
```diff
@@ -77,10 +85,19 @@ impl OracleContract {
         Ok(())
     }
 
-    /// Gets the hash type from the first hash in the contract, if any.
+    /// Gets the hash type from any hash in the contract, if any.
     /// Returns None if the contract has no hashes yet.
     fn get_hash_type(&self) -> Option<HashType> {
-        self.hashes.first().map(HashType::from)
+        // Check any entry in the ring buffer
```
jsdanielh: Why this change?

primitives/account/src/account/oracle_contract.rs
```diff
@@ -210,20 +228,99 @@ impl AccountTransactionInteraction for OracleContract {
...
-                            self.hashes.drain(0..hashes_to_remove).collect();
-                        removed
+                    // Get the zero hash type for chaining (using the type of the first new hash)
+                    let zero_hash = if let Some(first_hash) = hashes.first() {
+                        match first_hash {
+                            AnyHash::Blake2b(_) => AnyHash::Blake2b(AnyHash32::default()),
+                            AnyHash::Sha256(_) => AnyHash::Sha256(AnyHash32::default()),
+                            AnyHash::Sha512(_) => AnyHash::Sha512(AnyHash64::default()),
+                            AnyHash::Keccak256(_) => AnyHash::Keccak256(AnyHash32::default()),
+                        }
```
jsdanielh: I think it's easier to derive `default` for `AnyHash`

primitives/account/src/account/oracle_contract.rs
```diff
@@ -210,20 +228,99 @@ impl AccountTransactionInteraction for OracleContract {
...
+                    } else if let Some(existing_hash) = self.hashes.iter().find(|h| {
+                        // Find a non-zero hash
+                        !matches!(h, AnyHash::Blake2b(h) if h.0 == [0u8; 32])
```
jsdanielh: Why do we need to find a non-zero hash? And why are we forcing `Blake2b`?

primitives/account/src/account/oracle_contract.rs
```diff
@@ -210,20 +228,99 @@ impl AccountTransactionInteraction for OracleContract {
...
+                    }) {
+                        match existing_hash {
+                            AnyHash::Blake2b(_) => AnyHash::Blake2b(AnyHash32::default()),
+                            AnyHash::Sha256(_) => AnyHash::Sha256(AnyHash32::default()),
+                            AnyHash::Sha512(_) => AnyHash::Sha512(AnyHash64::default()),
+                            AnyHash::Keccak256(_) => AnyHash::Keccak256(AnyHash32::default()),
+                        }
                     } else {
-                        Vec::new()
+                        AnyHash::default()
```
jsdanielh: Why all these conditionals if we can simply use this?

primitives/account/src/account/oracle_contract.rs
```diff
@@ -210,20 +228,99 @@ impl AccountTransactionInteraction for OracleContract {
...
+
+                        // Save the hash that will be overwritten (for revert)
+                        // Only save if we're overwriting (i.e., latest_index >= hash_count)
+                        if self.latest_index >= self.hash_count as u64 {
+                            removed_hashes.push(self.hashes[pos].clone());
+                        }
+
+                        // Write the chained hash at the ring buffer position
+                        self.hashes[pos] = current_hash.clone();
+                        self.latest_index += 1;
```
jsdanielh: I thought this could be much simpler. I don't understand why this complex

primitives/account/src/account/oracle_contract.rs
```diff
@@ -507,3 +658,43 @@ struct ChangeOwnerReceipt {
...
+    }
+
+    /// Gets the hash at a given global index.
+    /// Returns None if the index is outside the retained window [earliest_index, latest_index).
+    pub fn get_hash_at_index(&self, index: u64) -> Option<&AnyHash> {
+        let first = self.earliest_index();
+        if index < first || index >= self.latest_index {
+            return None;
+        }
+        let pos = (index % self.hash_count as u64) as usize;
```
jsdanielh: I have my doubts with this. I think it could have been implemented by only keeping track of the latest index and the doing subtraction to get the desired element

primitives/account/src/account/oracle_contract.rs
```diff
@@ -507,3 +658,43 @@ struct ChangeOwnerReceipt {
...
+        }
+    }
+
+    /// Returns all hashes in chronological order (oldest to newest).
+    pub fn get_hashes_chronological(&self) -> Vec<AnyHash> {
+        let first = self.earliest_index();
+        let mut result = Vec::new();
+        for i in first..self.latest_index {
+            if let Some(hash) = self.get_hash_at_index(i) {
+                result.push(hash.clone());
```
jsdanielh: Using the method I described above this could be avoided

primitives/account/tests/oracle_contract.rs
```diff
@@ -88,6 +89,76 @@ fn make_hash_keccak256(value: u8) -> AnyHash {
...
+        return Vec::new();
+    }
+
+    let mut result = Vec::new();
+    let mut current_hash = previous_hash.clone();
+
+    for new_hash in hashes {
+        // Chain: data_i = H(data_{i-1} || state_i)
+        current_hash = match new_hash {
+            AnyHash::Blake2b(_) => {
```
jsdanielh: I think you implemented `digest` for `AnyHash` in your previous PR. That could be used here

rpc-interface/src/oracle.rs
```diff
@@ -0,0 +1,45 @@
...
+use nimiq_keys::Address;
+
+use crate::types::RPCResult;
+
+/// Oracle entry containing height and data
+#[derive(serde::Serialize, serde::Deserialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct OracleEntry {
+    /// The hash data for this entry
+    pub data: Blake2bHash,
```
jsdanielh: This is hardwiring the `Oracle` to `Blake2b`

rpc-interface/src/oracle.rs
```diff
@@ -0,0 +1,45 @@
...
+    ) -> RPCResult<u64, (), Self::Error>;
+
+    /// Returns the size of the sliding window (ring buffer capacity).
+    async fn get_window_size(&self, contract_address: Address) -> RPCResult<u16, (), Self::Error>;
+
+    /// Returns the latest hash-chain head.
+    async fn get_latest_data(
+        &self,
+        contract_address: Address,
+    ) -> RPCResult<Blake2bHash, (), Self::Error>;
```
jsdanielh: Same as above

jsdanielh: And why here we're not returning `OracleEntry`. Is it needed in the first place?

rpc-server/src/dispatchers/oracle.rs
```diff
@@ -0,0 +1,127 @@
...
+                ))),
+            }
+        } else {
+            Err(Error::NotSupportedForLightBlockchain)
+        }
+    }
+
+    /// Helper to convert AnyHash to Blake2bHash
+    /// Returns an error if the hash is not Blake2b
+    fn any_hash_to_blake2b(hash: &AnyHash) -> Result<Blake2bHash, Error> {
```
jsdanielh: But this is required because the interface is hardwired to `Blake2b`

rpc-server/src/dispatchers/oracle.rs
```diff
@@ -0,0 +1,127 @@
...
+        let oracle = self.get_oracle_contract(contract_address)?;
+
+        if oracle.latest_index == 0 {
+            return Err(Error::InvalidData(
+                "Oracle contract has no data (latest_index is 0)".to_string(),
+            ));
+        }
+
+        // Get the latest hash (at index latest_index - 1)
+        let latest_index = oracle.latest_index - 1;
```
jsdanielh: Wait, but shouldn't the latest index be `oracle.latest_index`?

rpc-server/src/dispatchers/oracle.rs
```diff
@@ -0,0 +1,127 @@
...
+        let latest_hash = oracle.get_hash_at_index(latest_index).ok_or_else(|| {
+            Error::InvalidData(format!("Failed to get hash at index {}", latest_index))
+        })?;
+
+        Self::any_hash_to_blake2b(latest_hash).map(|hash| hash.into())
```
jsdanielh: Why converting it to `Blake2b`? It might not be the case

rpc-server/src/dispatchers/oracle.rs
```diff
@@ -0,0 +1,127 @@
...
+    async fn get_entry(
+        &self,
+        contract_address: Address,
+        index: u64,
+    ) -> RPCResult<OracleEntry, (), Self::Error> {
+        let oracle = self.get_oracle_contract(contract_address)?;
+
+        // Check if index is within the retained window
+        let earliest = oracle.earliest_index();
+        if index < earliest || index >= oracle.latest_index {
```
jsdanielh: Shouldn't latest indext be valid?

rpc-server/src/dispatchers/oracle.rs
```diff
@@ -0,0 +1,127 @@
...
+                index, earliest, oracle.latest_index
+            )));
+        }
+
+        // Get the hash at the given index
+        let hash = oracle
+            .get_hash_at_index(index)
+            .ok_or_else(|| Error::InvalidData(format!("Failed to get hash at index {}", index)))?;
+
+        let data = Self::any_hash_to_blake2b(hash)?;
```
jsdanielh: Why converting it to `Blake2b`? It might not be the case

primitives/account/src/account/oracle_contract.rs
```diff
@@ -37,8 +40,13 @@ pub struct OracleContract {
     /// The number of hashes that can be stored. This is set at creation and determines the required deposit.
     pub hash_count: u16,
 
-    /// The stored hashes. The number of hashes must not exceed hash_count.
+    /// Fixed-size ring buffer of hashes. Always has exactly `hash_count` elements.
+    /// Access entry at global index `i` via `hashes[i % hash_count]`.
     pub hashes: Vec<AnyHash>,
+
+    /// Monotonically increasing counter tracking the total number of updates.
+    /// The last valid index is `latest_index - 1` (if `latest_index > 0`).
```
jsdanielh: Why not making `latest_index` the last valid index? At the end it is its name

primitives/account/src/account/oracle_contract.rs
```diff
@@ -507,3 +658,43 @@ struct ChangeOwnerReceipt {
...
+        } else {
+            self.latest_index - self.hash_count as u64
+        }
+    }
+
+    /// Gets the hash at a given global index.
+    /// Returns None if the index is outside the retained window [earliest_index, latest_index).
+    pub fn get_hash_at_index(&self, index: u64) -> Option<&AnyHash> {
+        let first = self.earliest_index();
+        if index < first || index >= self.latest_index {
```
jsdanielh: Why is that we made `latest_index` not valid?

primitives/account/src/account/oracle_contract.rs
viquezclaudio: This is just following Pascals [implementation](https://github.com/nimiq/interopability/blob/main/evm/oracle/contracts/2_EvmOracle.sol#L89):

primitives/account/tests/oracle_contract.rs
viquezclaudio: Yes, thats why this is a draft and is waiting the other pr to be merged

primitives/account/src/account/oracle_contract.rs
jsdanielh: Mmm, but something is off. That implementation complexity is much lower than this one. Isn't it?

primitives/account/src/account/oracle_contract.rs
```diff
@@ -210,20 +220,90 @@ impl AccountTransactionInteraction for OracleContract {
...
-                        let hashes_to_remove = new_hash_count - self.hash_count as usize;
-                        let removed: Vec<AnyHash> =
-                            self.hashes.drain(0..hashes_to_remove).collect();
-                        removed
+                    // Get the zero hash type for chaining (using the type of the first new hash, or existing contract hash type)
+                    let zero_hash = hashes
+                        .first()
+                        .map(|h| h.zero_of_same_type())
+                        .or_else(|| self.hashes.first().map(|h| h.zero_of_same_type()))
+                        .unwrap_or_else(|| AnyHash::default());
```
jsdanielh: Clippy warning:
```suggestion
                        .unwrap_or_else(AnyHash::default);
```

primitives/account/src/account/Untitled
```diff
@@ -0,0 +1 @@
+The last valid index is `latest_index - 1` (if `latest_index > 0`).
```
jsdanielh: What is this?

primitives/account/tests/oracle_contract.rs
```diff
@@ -88,6 +89,76 @@ fn make_hash_keccak256(value: u8) -> AnyHash {
     AnyHash::from(Keccak256Hasher::default().digest(&[value; 32]))
 }
 
+// Helper to compute chained hashes as the contract does: data_i = H(data_{i-1} || state_i)
+fn compute_chained_hashes(hashes: &[AnyHash]) -> Vec<AnyHash> {
+    use nimiq_transaction::account::htlc_contract::{AnyHash32, AnyHash64};
```
jsdanielh: Move this import to the beginning of the file

primitives/account/tests/oracle_contract.rs
```diff
@@ -88,6 +89,76 @@ fn make_hash_keccak256(value: u8) -> AnyHash {
...
+
+// Helper to compute chained hashes starting from a previous hash
+fn compute_chained_hashes_from_previous(
+    hashes: &[AnyHash],
+    previous_hash: &AnyHash,
+) -> Vec<AnyHash> {
+    use std::io::Write;
+
+    use nimiq_hash::{sha512::Sha512Hasher, Blake2bHasher, Hasher, Keccak256Hasher, Sha256Hasher};
+    use nimiq_transaction::account::htlc_contract::{AnyHash32, AnyHash64};
```
jsdanielh: Move imports to the beginning of the file

primitives/account/src/account/oracle_contract.rs
jsdanielh: Also `zero_hash` should only be needed when we have no entries 🤔 . Here we're always calculating it

primitives/account/src/account/oracle_contract.rs
```diff
@@ -210,20 +220,90 @@ impl AccountTransactionInteraction for OracleContract {
...
+                            .reserve_exact(self.hash_count as usize - self.hashes.capacity());
+                    }
+
+                    // Get the previous hash for chaining (latestData in Solidity)
+                    let previous_hash = if let Some(prev_index) = self.latest_index {
+                        let prev_pos = (prev_index % self.hash_count as u64) as usize;
+                        // Ensure the vec has enough entries
+                        while self.hashes.len() <= prev_pos {
+                            self.hashes.push(zero_hash.clone());
+                        }
```
jsdanielh: Why are we doing this?

primitives/account/src/account/oracle_contract.rs
```diff
@@ -210,20 +220,90 @@ impl AccountTransactionInteraction for OracleContract {
...
+
+                        // Ensure the vec has enough entries for this position
+                        while self.hashes.len() <= pos {
+                            self.hashes.push(zero_hash.clone());
+                        }
+
+                        // Save the hash that will be overwritten (for revert)
+                        // Only save if we're overwriting (i.e., index >= hash_count)
+                        if index >= self.hash_count as u64 {
+                            removed_hashes.push(self.hashes[pos].clone());
```
jsdanielh, line 254: But if we are adding the first entry, we would be saving `zero_hash`

primitives/account/src/account/oracle_contract.rs
```diff
@@ -210,20 +220,90 @@ impl AccountTransactionInteraction for OracleContract {
...
+                        }
+
+                        // Write the chained hash at the ring buffer position
+                        self.hashes[pos] = current_hash.clone();
+                    }
+
+                    // Update latest_index to the last written index
+                    if !hashes.is_empty() {
+                        self.latest_index = Some(start_index + hashes.len() as u64 - 1);
+                    }
```
jsdanielh, line 264: I still think this could be simpler. Something like:
```
let mut removed_hashes = vec![];
for (index, hash) in hashes.iter.enumerate() {
    let pos = (index + self.latest_index) % self.hash_count; 
    let prev_hash = if self.latest_index == 0 { zero_hash() } else {
        let prev_pos = (index + self.latest_index - 1) % self.hash_count;
        self.hashes[prev_pos]
    }; 
    removed_hashes.push(self.hashes[pos]);
    self.hashes[pos] = prev_hash.digest(hash);
}
```

primitives/account/src/account/oracle_contract.rs
```diff
@@ -280,18 +360,100 @@ impl AccountTransactionInteraction for OracleContract {
...
+                                            // Trim if too long (shouldn't happen, but be safe)
+                                            self.hashes.truncate(target_len);
+                                        }
+                                    }
+                                }
+                                None => {
+                                    // Underflow - set to None (empty)
+                                    self.latest_index = None;
+                                    self.hashes.clear();
+                                }
```
jsdanielh: I think this is much complex than what it should be

primitives/account/tests/oracle_contract.rs
jsdanielh: We are still not using `digest`

primitives/account/tests/oracle_contract.rs
```diff
@@ -347,25 +392,33 @@ fn it_implements_ring_buffer() {
...
+    // Check using chronological order helper
+    let expected_all = compute_chained_hashes(&initial_hashes);
+    // The new hashes are chained from the last hash in expected_all, not from zero
+    let previous_hash = expected_all.last().cloned().unwrap_or_else(|| {
+        // If expected_all is empty, use zero hash of the same type as new_hashes
+        match &new_hashes[0] {
+            AnyHash::Blake2b(_) => AnyHash::Blake2b(AnyHash32::default()),
+            AnyHash::Sha256(_) => AnyHash::Sha256(AnyHash32::default()),
+            AnyHash::Sha512(_) => AnyHash::Sha512(AnyHash64::default()),
+            AnyHash::Keccak256(_) => AnyHash::Keccak256(AnyHash32::default()),
```
jsdanielh: We could use the introduced `zero_hash_of_same_type`

primitives/account/tests/oracle_contract.rs
```diff
@@ -74,38 +76,78 @@ fn make_hash_blake2b(value: u8) -> AnyHash {
...
-    // Hash the previous hash and new hash together using the same algorithm as the previous hash
-    // Chain: data_i = H(data_{i-1} || state_i) = H(previous_hash || new_hash)
-    prev.digest(new_hash)
+    // Determine hash type from first hash
+    let zero_hash = match &hashes[0] {
+        AnyHash::Blake2b(_) => AnyHash::Blake2b(AnyHash32::default()),
+        AnyHash::Sha256(_) => AnyHash::Sha256(AnyHash32::default()),
+        AnyHash::Sha512(_) => AnyHash::Sha512(AnyHash64::default()),
+        AnyHash::Keccak256(_) => AnyHash::Keccak256(AnyHash32::default()),
+    };
```
jsdanielh: We could use the introduced `zero_hash_of_same_type`

primitives/transaction/src/account/htlc_contract.rs
```diff
@@ -142,34 +142,24 @@ impl AnyHash {
...
+            AnyHash::Sha512(_) => {
+                AnyHash::Sha512(AnyHash64(Sha512Hasher::default().digest(&combined).into()))
+            }
             AnyHash::Keccak256(_) => AnyHash::Keccak256(AnyHash32(
-                Keccak256Hasher::default()
-                    .chain(self)
-                    .chain(other)
-                    .finish()
-                    .into(),
+                Keccak256Hasher::default().digest(&combined).into(),
```
jsdanielh, line 162: Why are we reverting this? I think it was better as it was

rpc-interface/src/oracle.rs
```diff
@@ -0,0 +1,37 @@
...
+    ) -> RPCResult<u64, (), Self::Error>;
+
+    /// Returns the size of the sliding window (ring buffer capacity).
+    async fn get_window_size(&self, contract_address: Address) -> RPCResult<u16, (), Self::Error>;
+
+    /// Returns the latest hash-chain head.
+    async fn get_latest_data(
+        &self,
+        contract_address: Address,
+    ) -> RPCResult<AnyHash, (), Self::Error>;
```
jsdanielh, line 32: Now that I remember @paberr said that he intended to store the block number along with the hashes in the oracle and this to return this in this interface as well

rpc-interface/src/oracle.rs
```diff
@@ -0,0 +1,37 @@
...
+
+    /// Returns the data for a given index.
+    /// Returns an error if index is outside the currently retained window.
+    async fn get_entry(
+        &self,
+        contract_address: Address,
+        index: u64,
+    ) -> RPCResult<AnyHash, (), Self::Error>;
```
jsdanielh, line 40: Now that I remember @paberr said that he intended to store the block number along with the hashes in the oracle and this to return this in this interface as well

rpc-server/src/dispatchers/oracle.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+#[nimiq_jsonrpc_derive::service(rename_all = "camelCase")]
+#[async_trait]
+impl OracleInterface for OracleDispatcher {
+    type Error = Error;
+
+    async fn get_latest_index(&self, contract_address: Address) -> RPCResult<u64, (), Self::Error> {
+        let oracle = self.get_oracle_contract(contract_address)?;
+        Ok(oracle.latest_index.map(|i| i + 1).unwrap_or(0).into())
```
jsdanielh: Why are we adding `1` here?

rpc-interface/src/oracle.rs
```diff
@@ -0,0 +1,37 @@
...
+use nimiq_transaction::account::htlc_contract::AnyHash;
+
+use crate::types::RPCResult;
+
+#[nimiq_jsonrpc_derive::proxy(name = "OracleProxy", rename_all = "camelCase")]
+#[async_trait]
+pub trait OracleInterface {
+    type Error;
+
+    /// Returns the latest index (number of updates performed so far).
```
jsdanielh: Mmm, an `index` is zero based, while a count is not. This is confusing to me

rpc-server/src/error.rs
```diff
@@ -101,6 +101,12 @@ pub enum Error {
 
     #[error("Invalid Keccak256 parameters: {0}")]
     InvalidKeccak256Parameters(String),
+
+    #[error("Invalid address: {0}")]
+    InvalidAddress(String),
```
jsdanielh: I think this should be:

```suggestion
    InvalidAddress(Address),
```

primitives/account/src/account/oracle_contract.rs
```diff
@@ -303,18 +319,50 @@ impl AccountTransactionInteraction for OracleContract {
...
+
+                    if let Some(current_latest) = self.latest_index {
+                        if num_hashes == 0 {
+                            tx_logger.push_log(Log::OracleUpdate {
+                                contract_address: transaction.recipient.clone(),
+                                hashes,
+                            });
+                            return Ok(());
+                        }
+                        if num_hashes >= current_latest + 1 {
```
jsdanielh: Clippy warning

primitives/account/src/account/oracle_contract.rs
```diff
@@ -303,18 +319,50 @@ impl AccountTransactionInteraction for OracleContract {
...
-                        restored_hashes.append(&mut self.hashes);
-                        self.hashes = restored_hashes;
+                    let num_hashes = hashes.len() as u64;
+
+                    if let Some(current_latest) = self.latest_index {
+                        if num_hashes == 0 {
+                            tx_logger.push_log(Log::OracleUpdate {
+                                contract_address: transaction.recipient.clone(),
+                                hashes,
+                            });
```
jsdanielh: Here `hashes` is empty. What would be the value of logging this?

primitives/account/src/account/oracle_contract.rs
```diff
@@ -303,18 +319,50 @@ impl AccountTransactionInteraction for OracleContract {
...
+                        if let Some(receipt) = receipt {
+                            let update_receipt = UpdateReceipt::try_from(receipt)?;
+                            for (i, old_hash) in update_receipt.removed_hashes.iter().enumerate() {
+                                let pos =
+                                    ((start_index + i as u64) % self.hash_count as u64) as usize;
+                                self.hashes[pos] = old_hash.clone();
+                            }
+                        }
+
+                        let new_latest = current_latest - num_hashes;
```
jsdanielh, line 366: Beware, this could also go to `None`

primitives/account/src/account/oracle_contract.rs
```diff
@@ -303,18 +319,50 @@ impl AccountTransactionInteraction for OracleContract {
...
+                        self.latest_index = Some(new_latest);
+
+                        // Ring length is min(hash_count, new_latest + 1)
```
jsdanielh: Not sure why we want to optimize this. Is it necessary?

primitives/account/src/account/oracle_contract.rs
```diff
@@ -303,18 +319,50 @@ impl AccountTransactionInteraction for OracleContract {
...
+                            return Ok(());
+                        }
+
+                        let start_index = current_latest - (num_hashes - 1);
+
+                        if let Some(receipt) = receipt {
+                            let update_receipt = UpdateReceipt::try_from(receipt)?;
+                            for (i, old_hash) in update_receipt.removed_hashes.iter().enumerate() {
+                                let pos =
+                                    ((start_index + i as u64) % self.hash_count as u64) as usize;
```
paberr: **Bug: Revert misaligns removed hashes when an update straddles the `hash_count` boundary.**

The `removed_hashes` collected during commit only include entries where `index >= hash_count` (i.e., positions that actually overwrote existing data). But here in the revert, they're written back starting at `start_index`, which may be *before* the `hash_count` threshold.

**Example:** `hash_count=10`, `latest_index=8` (9 hashes written), then add 3 hashes (indices 9, 10, 11):

Forward commit:
- index=9, pos=9 → no eviction (9 < 10)
- index=10, pos=0 → `removed_hashes[0]` = old `hashes[0]`
- index=11, pos=1 → `removed_hashes[1]` = old `hashes[1]`

Revert (`current_latest=11`, `num_hashes=3`, `start_index=9`):
- `removed_hashes[0]` written to pos `(9+0)%10 = 9` — **wrong**, should go to pos 0
- `removed_hashes[1]` written to pos `(9+1)%10 = 0` — **wrong**, should go to pos 1

The removed hashes get written to the wrong positions, corrupting state. The offset `i` needs to account for the fact that the first `hash_count - start_index` positions didn't generate removed hashes.

primitives/account/src/account/oracle_contract.rs
```diff
@@ -303,18 +319,50 @@ impl AccountTransactionInteraction for OracleContract {
...
+
+                        // Ring length is min(hash_count, new_latest + 1)
+                        let target_len = if new_latest >= (self.hash_count as u64).saturating_sub(1)
+                        {
+                            self.hash_count as usize
+                        } else {
+                            (new_latest + 1) as usize
+                        };
+                        if self.hashes.len() > target_len {
+                            self.hashes.truncate(target_len);
```
paberr: **Bug: Truncation can cause out-of-bounds panic on subsequent commits.**

Once the ring buffer is initialized to `hash_count` size (on first write), it should never be truncated. But this code shrinks it when reverting to a state where fewer than `hash_count` entries had been written.

**Example:** `hash_count=10`, 8 hashes written (`latest_index=7`), add 2 more → `latest_index=9`. Receipt is `None` (no eviction, indices 8 and 9 are < 10). Revert those 2:
- `new_latest = 7`, `target_len = 8`, `hashes.truncate(8)`

Next forward commit writes to index 8 (pos 8): `self.hashes[8]` panics since vec has only 8 entries (indices 0–7).

The resize-on-first-write path (`if self.latest_index.is_none()`) won't trigger because `latest_index` is `Some(7)`. I'd suggest removing the truncation entirely — the ring buffer should remain at `hash_count` size once allocated.

primitives/account/src/account/oracle_contract.rs
```diff
@@ -210,51 +217,60 @@ impl AccountTransactionInteraction for OracleContract {
...
-                    if removed_hashes.is_empty() && previous_hash_state.is_none() {
+                    // Return receipt with removed hashes for proper revert (ring buffer)
+                    if removed_hashes.is_empty() {
                         Ok(None)
                     } else {
                         Ok(Some(
                             UpdateReceipt {
                                 removed_hashes,
-                                previous_hash_state,
+                                previous_hash_state: None,
```
paberr: `previous_hash_state` is always `None` and never read anywhere. This was used by the old drain-based implementation but serves no purpose in the ring buffer approach. Consider removing it from `UpdateReceipt` to avoid dead code confusion.

primitives/account/src/account/oracle_contract.rs
```diff
@@ -532,3 +583,87 @@ struct ChangeOwnerReceipt {
...
+    }
+}
+
+impl PartialEq for OracleContract {
+    fn eq(&self, other: &Self) -> bool {
+        self.owner == other.owner
+            && self.balance == other.balance
+            && self.hash_count == other.hash_count
+            && self.latest_index == other.latest_index
+            && self.get_hashes_chronological() == other.get_hashes_chronological()
```
paberr: **Performance:** `get_hashes_chronological()` allocates a new `Vec` on every call. This runs on every `==` comparison and on every `Ord::cmp` call.

Since two contracts with the same `latest_index` and `hash_count` will have identical ring buffer positions, you could simply compare `self.hashes == other.hashes` directly — it's semantically equivalent and allocation-free.

rpc-server/src/dispatchers/oracle.rs
```diff
@@ -0,0 +1,108 @@
...
+    async fn get_entry(
+        &self,
+        contract_address: Address,
+        index: u64,
+    ) -> RPCResult<AnyHash, (), Self::Error> {
+        let oracle = self.get_oracle_contract(contract_address)?;
+
+        // Check if index is within the retained window
+        let earliest = oracle.earliest_index().unwrap_or(0);
+        let latest = oracle.latest_index.unwrap_or(0);
```
paberr, line 102: When `latest_index` is `None` (no data written), this defaults to 0. Combined with `earliest` also defaulting to 0, the bounds check `index < 0 || index > 0` lets `index=0` through. But then `get_hash_at_index(0)` returns `None` (because `latest_index` is `None`), producing a confusing error: *"Failed to get hash at index 0"*.

Consider checking for `None` explicitly first:
```rust
let latest = oracle.latest_index.ok_or_else(|| {
    Error::InvalidData("Oracle contract has no data".to_string())
})?;
```

primitives/account/tests/oracle_contract.rs
```diff
@@ -384,8 +437,9 @@ fn it_implements_ring_buffer() {
         .expect("Failed to revert");
     // Should be back to the original 10 chained hashes
     assert_eq!(oracle_contract.hashes.len(), 10);
-    let expected_initial = compute_chained_hashes(&initial_hashes);
-    assert_eq!(oracle_contract.hashes, expected_initial);
+    assert_eq!(oracle_contract.latest_index, Some(9));
+    let chronological_after_revert = oracle_contract.get_hashes_chronological();
+    assert_eq!(chronological_after_revert, expected_all);
```
paberr, line 403: **Missing test coverage:** This test writes exactly `hash_count` (10) hashes first, so the second update starts at index 10 where all indices are `>= hash_count`. This means the revert always has removed hashes perfectly aligned with `start_index`.

Consider adding a test where the first update writes *fewer* than `hash_count` hashes (e.g., 9), then a second update straddles the boundary (e.g., 3 more hashes at indices 9, 10, 11). This is the case where the revert bug manifests — some indices generate evictions and some don't.

paberr (commented):
Review focusing on correctness of the ring buffer implementation and revert logic. Found two bugs in the revert path that can corrupt state or cause panics in edge cases, plus a few smaller issues.

primitives/transaction/src/account/htlc_contract.rs
viquezclaudio: I thought you wanted to use digest

rpc-interface/src/oracle.rs
viquezclaudio: @paberr  ?

primitives/transaction/src/account/htlc_contract.rs
jsdanielh: Ah, I know where the confusion is coming from. I was originally mixing `chain` and `digest` but in the messages PR we settled on `chain` by implementing `SerializeContent`

primitives/account/src/account/oracle_contract.rs
```diff
@@ -210,51 +217,60 @@ impl AccountTransactionInteraction for OracleContract {
...
+                        // data_i = H(data_{i-1} || state_i)
+                        current_hash = current_hash.digest(new_hash);
+                        self.hashes[pos] = current_hash.clone();
+                    }
 
-                        // Store the chained hash
-                        self.hashes.push(chained_hash);
+                    if !hashes.is_empty() {
+                        self.latest_index = Some(start_index + hashes.len() as u64 - 1);
                     }
```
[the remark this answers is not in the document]
viquezclaudio: This comment is no longer valid with the current implementation

primitives/account/src/account/oracle_contract.rs
viquezclaudio: This was indeed simplified

rpc-server/src/dispatchers/oracle.rs
```diff
@@ -0,0 +1,108 @@
...
+        if index < earliest || index > latest {
+            return Err(Error::InvalidData(format!(
+                "Index {} is outside the retained window [{}, {}]",
+                index, earliest, latest
+            )));
+        }
```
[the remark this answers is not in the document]
viquezclaudio: This is no longer valid.

primitives/account/src/account/oracle_contract.rs
viquezclaudio: This is no longer valid.

PR #3638, opened 2026-03-11
Initialize to test policy for unit albatross

ii-cruz:
The client is initialized always with default policy, despite the network ID being unit albatross, which makes the zkp verification fail when we try to run nimiq clients locally for testing.

## Pull request checklist


ii-cruz:
we removed the zkps so this is no longer relevant

PR #3653, opened 2026-03-27
Discard stale peers received via discovery

### LINKED ISSUE #3652: Discard stale peers received via discovery
I noticed some logs in my nodes for a very long time that it removes old peers in large quantities, over and over again. Plus when debugging the WebClient in testnet, I observed connection attempts to very old testnet peers I haven't had running since about a year.
So I checked the source code and, funnily enough, there was a TODO describing the exact problem. 

The problem in a greater view: House keeping removes stale peers, but nodes still accept stale peers. So even though there is house keeping, all nodes ensure that they broadcast all the stale peers before *all* nodes have the chance to have them pruned by house keeping.

I found a clean and minimal solution to this problem that I'd like to propose here. As this is a minimal change, I'll create a PR for this now.
Please review that PR: #3653

Have a nice weekend !

wille-io:

The problem in a greater view: House keeping removes stale peers, but nodes still accept stale peers. So even though there is house keeping, all nodes ensure that they broadcast all the stale peers before all nodes have the chance to have them pruned by house keeping.

Now, if a node receives a stale peer or a peer from the future (functionality adapted from house keeping) then the peer will not be added. Plus, a peer that already exists is only overwritten if the peer info is newer than the existing info.

Note: The tests were written by AI.

## Pull request checklist

### DIFF
--- a/network-libp2p/src/discovery/peer_contacts.rs
+++ b/network-libp2p/src/discovery/peer_contacts.rs
@@ -427,27 +427,40 @@ impl PeerContactBook {
             }
         }
 
-        // Reject contacts with timestamps in the future
         let current_ts = SystemTime::now()
             .duration_since(SystemTime::UNIX_EPOCH)
             .unwrap()
             .as_secs();
+
+        // Reject contacts with timestamps in the future
         if info.contact().timestamp > current_ts {
             return;
         }
 
+        // Rejects contacts that are older than the allowed age
+        if info.exceeds_age(
+            Duration::from_secs(PeerContactBook::MAX_PEER_AGE),
+            Duration::from_secs(current_ts),
+        ) {
+            return;
+        }
+
         let info = Arc::new(info);
-        let is_new = self
-            .peer_contacts
-            .insert(info.peer_id, Arc::clone(&info))
-            .is_none();
-        if is_new {
-            log::trace!(
-                peer_id = %info.peer_id,
-                services = ?info.services(),
-                addresses = ?info.contact.inner.addresses,
-                "Adding peer contact",
-            );
+        match self.peer_contacts.entry(info.peer_id) {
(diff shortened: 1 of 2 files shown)


nibhar:
Part of this was added to albatross end of last week already. 
Please rebase this PR. I will review the rebased version so we can get this merged.

wille-io:
@nibhar rebase done 👍🏼

network-libp2p/src/discovery/peer_contacts.rs
```diff
@@ -427,27 +427,40 @@ impl PeerContactBook {
...
+            }
+            std::collections::hash_map::Entry::Vacant(entry) => {
+                log::trace!(
+                    peer_id = %info.peer_id,
+                    services = ?info.services(),
+                    addresses = ?info.contact.inner.addresses,
+                    "Adding peer contact",
+                );
+                entry.insert(info);
+            }
```
nibhar, line 463: I would probably look into 
```rust
self.peer_contacts.entry(info.peer_id)
    .and_modify(|contact| { ... })
    .or_insert_with(|| { ... });
```
But that is just preference and needn't be done. Also I am aware we have the exact same construct elsewhere in this file.

nibhar (approved):
LGTM.

Note that contacts are signed and in this callsite already verified. That means that contact contents are controlled by the peer referenced in the contact and not the peer who send it.

network-libp2p/src/discovery/peer_contacts.rs
wille-io: @nibhar I understand and your preference looks cleaner. But it looks like it would require creating another Arc and / or the use of `Arc::clone`. I think that looks way uglier than using Occupied & Vacant.

jsdanielh (approved):

PR #3655, opened 2026-03-27
Fix peer contact book poisoning via future timestamps

jsdanielh:
## Summary

- Peers could inject contacts with far-future timestamps that survived housekeeping eviction indefinitely, polluting the address book and degrading peer discovery quality.
- Reject contacts with future timestamps at insertion time in both `insert()` and `insert_filtered()`.
- Treat future timestamps as exceeded in `exceeds_age()` as defense-in-depth so housekeeping evicts any that slip through.


viquezclaudio (approved):

jsdanielh:
Merged into `albatross`: 28d1bce5e7786537cb102e343b7517019b6d906d

PR #3657, opened 2026-03-27
Fix quorum bypass via out-of-range BitSet signer indices

jsdanielh:
## Summary

- The `MultiSignature` signers `BitSet` was used without validating that slot indices fall within `Policy::SLOTS`. An attacker could craft a `BitSet` with out-of-range indices to inflate the apparent signer count, bypassing the `TWO_F_PLUS_ONE` quorum check in `TendermintProof`, `SkipBlockProof`, and `DoubleVoteProof` verification.
- Add `checked_signer_slots()` in `multisig.rs` that validates all `BitSet` entries are within range, and apply it across all three proof verification paths.

### DIFF
--- a/primitives/block/src/equivocation_proof.rs
+++ b/primitives/block/src/equivocation_proof.rs
@@ -18,7 +18,7 @@ use nimiq_transaction::{
 };
 use thiserror::Error;
 
-use crate::{MacroHeader, MicroHeader};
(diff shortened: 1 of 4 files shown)


viquezclaudio (approved):

jsdanielh:
Merged into `albatross`: d02059053181ed8ddad6b59a0adfd661ef5cd823

PR #3658, opened 2026-03-27
Fix underflow panic in vesting and HTLC insufficient funds error path

jsdanielh:
## Summary

- When a vesting or HTLC contract had `total_amount > balance` (reachable via the 32-byte creation format), the `InsufficientFunds` error construction computed `self.balance - min_cap` which panics on underflow. In the block processing path this poisons the blockchain `RwLock`, making the node permanently unusable.
- Use `saturating_sub` for the error field arithmetic and add validation that `total_amount <= tx_value` during contract creation.


viquezclaudio (approved):

jsdanielh:
Merged into `albatross`: 4d01946f0b3d6c6e31786f91cdfb3eb902908da0

PR #3659, opened 2026-03-27
Fix panic on HistoryTreeProof length mismatch

jsdanielh:
## Summary

- `HistoryTreeProof::verify` used `assert_eq!` to check that `history` and `positions` vectors have equal length. A malicious peer could send a proof with mismatched lengths, panicking the verifying node.
- Replace the assert with a guard that returns `None` on mismatch, consistent with the method's documented error-returning contract.


viquezclaudio (approved):

jsdanielh:
Merged into `albatross`: 6ff0800e8e031363e787c827d8d033e5694e4e6a

PR #3660, opened 2026-03-27
Fix panic in RequestMacroChain with micro block locator

jsdanielh:
## Summary

- A peer could send a `RequestMacroChain` with a micro block hash in the locator list. The handler accepted any on-chain block without checking if it was a macro block, then called `get_macro_blocks()` which panics when given a micro block hash.
- Add a macro block type check before accepting a locator hash, skipping micro blocks and continuing the search.


viquezclaudio (approved):

jsdanielh:
Merged into `albatross`: ae6c1e92342e72f80fd12accbe66ee80dd6802ac

PR #3661, opened 2026-03-27
Fix off-by-one in proposal buffer signer bounds check

jsdanielh:
## Summary

- The bounds check used `>` instead of `>=` when comparing the proposal signer field against `num_validators()`. When `signer == num_validators()` the check passed, then `get_validator_by_slot_band()` panicked on out-of-bounds access. Any peer could trigger this via gossip with a crafted proposal.
- Change the comparison to `>=` to correctly reject boundary values.


viquezclaudio (approved):

jsdanielh:
Merged into `albatross`: 9199364b60c7acae4219800d194bbe07d2997b8c

PR #3662, opened 2026-03-27
Fix crash via invalid election macro block validators voting key

jsdanielh:
## Summary

- An untrusted peer could announce an election macro block containing an invalid compressed BLS voting key. The node called `uncompress().unwrap()` during block hashing, panicking on the invalid key.
- Make `voting_keys()` and `voting_keys_g2()` return `Result` instead of panicking. Add `validate_keys()` to `Validators` for explicit structural validation.
- Call `verify_header()` in the block queue before hashing to reject malformed blocks early.
- Handle decompression failures gracefully in `TendermintProof` and `DoubleVoteProof` verification.

Note: this PR builds on top of the BitSet signer slots fix.

### DIFF
--- a/consensus/src/sync/live/block_queue/queue.rs
+++ b/consensus/src/sync/live/block_queue/queue.rs
@@ -160,6 +160,16 @@ impl<N: Network> BlockQueue<N> {
 
         let blockchain = self.blockchain.read();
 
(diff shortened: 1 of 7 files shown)


viquezclaudio (approved):

jsdanielh:
Merged into `albatross`: e10eaebcd7774e5da6d0ff5e88ed13503474f0ff

PR #3663, opened 2026-03-27
Fix request/response codec reading entire stream before size validation

jsdanielh:
## Summary

- The `MessageCodec` read the entire stream into memory via `read_to_end()` before checking the declared message length, allowing a peer to force allocation up to `MAX_RESPONSE_SIZE` (10 MB) regardless of actual content.
- Refactor into a `read_message()` helper that reads the length prefix first, validates it against the limit, then allocates exactly the declared size.
- Add outbound size checks in `write_request()` and `write_response()` to prevent sending oversized messages.


viquezclaudio (approved):

jsdanielh:
Merged into `albatross`: c021a5337b808c73571b44999f9753051bac7508

PR #3664, opened 2026-03-27
Fix discovery handler underflow when peer sends limit=0

jsdanielh:
## Summary

- A peer sending `limit=0` during the discovery handshake caused an underflow in the periodic update path where `peer_list_limit - 1` wraps to `usize::MAX`, triggering a capacity overflow panic.
- Clamp the peer-supplied limit to our own `update_limit` on receipt, and use `saturating_sub(1)` when computing the update list size.


viquezclaudio (approved):

jsdanielh:
Merged into `albatross`: 8f60a2d75b74b55764ecf34bd4435f4961630595

PR #3665, opened 2026-03-27
Fix stale response_channels on request/response inbound failure

jsdanielh:
## Summary

- When an inbound request failed, the handler logged the error but did not remove the corresponding entry from `response_channels`. Over time this leaked memory for every failed inbound request.
- Remove the response channel on inbound failure to prevent the leak.


viquezclaudio (approved):

jsdanielh:
Merged into `albatross`: 44b15d1dabc84ce5ebb09682260ccc2a3330aa4e

PR #3666, opened 2026-03-27
Fix crash on duplicate discovery substream

jsdanielh:
## Summary

- A peer opening a second discovery substream on the same connection triggered `panic!("Inbound already connected")`. Similarly, duplicate outbound substreams or outbound substreams in unexpected state caused panics.
- Replace all three `panic!()` calls with a `fail_closed()` method that emits an error event and closes the connection gracefully.

Note: this PR builds on top of the peer contact future timestamp fix.


viquezclaudio (approved):

jsdanielh:
Merged into `albatross`: e0d4e01994f061bf41d3c2835bc74040d3c084f5

PR #3667, opened 2026-03-27
Fix panic in history index request handlers on nodes without index

jsdanielh:
## Summary

- Request handlers for transaction proofs and receipts called `history_index().unwrap()` which panics when the node runs without the history index enabled (`index_history: false`). A peer could trigger this by sending any transaction proof or receipt request.
- Check for the index with a guard clause and return an appropriate error or empty response instead of panicking.


viquezclaudio (approved):

jsdanielh:
Merged into `albatross`: 0e5c90a6c75b722f3d6091769776a4040e694dba

PR #3668, opened 2026-03-27
Fix election macro block interlink verification bypass

jsdanielh:
## Summary

- The interlink verification was only performed in the standard block push path but not in the Tendermint proposal verification path. A malicious proposer could submit an election macro block proposal with a wrong or missing interlink, which would be accepted and finalized without validation.
- Extract the interlink check into `verify_macro_block_interlink()` and apply it in both the push and proposal verification paths.


viquezclaudio (approved):

jsdanielh:
Merged into `albatross`: 9d7d17c9163384e79f61cdbbfe9853ae57bb8bf7

PR #3678, opened 2026-04-10
Remove unwrap/ safer arithmetic logic

viquezclaudio:
Remove unwrap in light blockchain
Use safer arithmetic logic for txn validity computation

These are no real issues in practice since push_zkip is only called with macro blocks and the txn validity can no longer be exploited

## Pull request checklist

### DIFF
--- a/primitives/transaction/tests/validity.rs
+++ b/primitives/transaction/tests/validity.rs
@@ -1,0 +1,56 @@
+//! Regression tests for transaction validity window arithmetic.
+
+use nimiq_keys::Address;
+use nimiq_primitives::{coin::Coin, networks::NetworkId, policy::Policy};
+use nimiq_test_log::test;
+use nimiq_transaction::Transaction;
+
+fn basic_transaction(validity_start_height: u32) -> Transaction {
+    Transaction::new_basic(
+        Address::from([1u8; 20]),
+        Address::from([2u8; 20]),
+        Coin::try_from(100u64).unwrap(),
+        Coin::try_from(0u64).unwrap(),
+        validity_start_height,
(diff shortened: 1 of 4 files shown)


primitives/transaction/tests/validity.rs
```diff
@@ -0,0 +1,73 @@
...
+#[test]
+fn is_valid_at_does_not_overflow_at_u32_max() {
+    let tx = basic_transaction(u32::MAX);
+
+    let result = panic::catch_unwind(panic::AssertUnwindSafe(|| tx.is_valid_at(u32::MAX)));
+
+    assert!(
+        result.is_ok(),
+        "is_valid_at() panicked with validity_start_height=u32::MAX"
+    );
```
jsdanielh: I'd prefer to not `catch_unwind`. At the end we are only using it for checking if it panicked with `result.is_ok()`.

primitives/transaction/tests/validity.rs
```diff
@@ -0,0 +1,73 @@
...
+    assert_eq!(
+        window,
+        Policy::transaction_validity_window() * Policy::blocks_per_batch()
+    );
+
+    // `overflow_start + window` would overflow without saturating arithmetic.
+    let overflow_start = u32::MAX - (window - 1);
+    let tx = basic_transaction(overflow_start);
+
+    let result = panic::catch_unwind(panic::AssertUnwindSafe(|| tx.is_valid_at(overflow_start)));
```
jsdanielh: I'd prefer to not `catch_unwind`. At the end we are only using it for checking if it panicked with `result.is_ok()`.

jsdanielh (approved):

PR #3690, opened 2026-04-14
Include symbols in the release-with-debug profile

jeffesquivels:


jsdanielh (approved):
LGTM

Cargo.toml
```diff
@@ -116,8 +116,9 @@ opt-level = "s"      # Optimize for size
 strip = "debuginfo"
 
 [profile.release-with-debug]
-debug = true
 inherits = "release"
+debug = 2                   # full debug info (types, locations, functions)
```
jsdanielh: Mmm, this is the same as `true` according to the documentation

PR #3702, opened 2026-04-21
metrics-server: Implement HTTP Basic Auth from config credentials

jsdanielh:
The metrics server config exposed `username`/`password` fields and the config builder wrapped them in a `Credentials`, but the credentials were never passed to `start_metrics_server` and the HTTP handler served `/metrics` without any check, giving operators a false sense of security.

Thread `Credentials` through `start_metrics_server` and validate the `Authorization: Basic` header via the `headers` crate before serving metrics. When credentials are configured, missing or invalid ones now return `401 Unauthorized` with a `WWW-Authenticate` challenge. Authentication remains optional: an unset config still allows unauthenticated access (unchanged behavior).

Also move the shared `Credentials` type from `nimiq-lib`'s config into `nimiq-utils` under a new `credentials` feature so both `nimiq-lib` and `nimiq-metrics-server` can use the same type without a circular dependency.

## Pull request checklist


jeffesquivels (approved):
LGTM

PR #3718, opened 2026-04-22
Fix rate limit pending deletion TOCTOU

### LINKED ISSUE #3701: Panic in `rate_limiting.rs`: "Tried to remove a non existing rate limit from peer_request_limits"
My validator node crashed due to a panic in the network-libp2p component. The root cause appears to be an "unreachable code" violation in the rate-limiting logic. Once this occurred, the networking task failed, leading to a cascade of SendError panics and the inability for the client to retrieve NetworkInfo.

nimiq-client Version: v1.4.0-pre

[Logs-logs-2026-04-20 20_17_01.txt](https://github.com/user-attachments/files/26906989/Logs-logs-2026-04-20.20_17_01.txt)

nibhar:
This PR fixes a very rare race condition where a peer leaving very close to the expiry of an existing rate limit may leave behind a pending deletion entry while the to be deleted entry is already removed. The main fix is just to make sure that a single timestamp is used for the comparisons, but I also did some mild refactoring to make the code easier to read and understand.

Furthermore I added a regression test which fails before the changes, but does now pass and removed the `unreachable!` macros in favor of logging errors as this kind of error does not warrant a full shutdown of the entire client, specifically because the error should be self correcting in these instances where the panic occurred.

Resolves #3701

### DIFF
--- a/network-libp2p/src/rate_limiting.rs
+++ b/network-libp2p/src/rate_limiting.rs
@@ -103,7 +103,7 @@ impl PendingDeletion {
             .expect("First item must exist");
         assert!(
             self.by_peer_and_id
-                .remove(&(expiration.peer_id.clone(), expiration.rate_limit_id.clone()))
+                .remove(&(expiration.peer_id, expiration.rate_limit_id.clone()))
                 .is_some(),
             "The pending for deletion rate limits should be consistent among them"
         );
--- a/network-libp2p/src/rate_limiting.rs
+++ b/network-libp2p/src/rate_limiting.rs
@@ -287,3 +287,121 @@ impl RateLimits {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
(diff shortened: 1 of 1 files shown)


network-libp2p/src/rate_limiting.rs
```diff
@@ -93,9 +93,21 @@ pub(crate) struct PendingDeletion {
...
+        if self.by_expiration_time.first()?.expiration_time > current_time {
+            return None;
+        }
+        let expiration = self
+            .by_expiration_time
+            .pop_first()
+            .expect("First item must exist");
+        assert!(
+            self.by_peer_and_id
+                .remove(&(expiration.peer_id.clone(), expiration.rate_limit_id.clone()))
```
jsdanielh: Clippy warning

jsdanielh (approved):
LGTM

PR #3732, opened 2026-04-28
Enhance plain transaction conversion in web client

viquezclaudio:
Remove unwraps and handle Plain transaction conversion

...
## Pull request checklist

### DIFF
--- a/web-client/src/common/transaction.rs
+++ b/web-client/src/common/transaction.rs
@@ -382,6 +382,18 @@ impl From<nimiq_transaction::Transaction> for Transaction {
 }
 
 impl Transaction {
+    fn raw_data(data: &[u8]) -> PlainRawData {
+        PlainRawData {
+            raw: hex::encode(data),
+        }
+    }
+
+    fn raw_proof(proof: &[u8]) -> PlainRawProof {
+        PlainRawProof {
+            raw: hex::encode(proof),
+        }
+    }
+
     #[cfg(feature = "primitives")]
     pub fn do_sign(
         &mut self,
@@ -459,6 +471,10 @@ impl Transaction {
         genesis_block_number: Option<u32>,
         genesis_timestamp: Option<u64>,
     ) -> PlainTransaction {
+        let raw_sender_data = Self::raw_data(&self.inner.sender_data);
+        let raw_recipient_data = Self::raw_data(&self.inner.recipient_data);
+        let raw_proof = Self::raw_proof(&self.inner.proof);
+
         PlainTransaction {
             transaction_hash: self.hash(),
             format: self.format(),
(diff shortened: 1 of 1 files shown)


web-client/src/common/transaction.rs
```diff
@@ -504,42 +512,40 @@ impl Transaction {
...
                 if self.inner.proof.is_empty() {
-                    PlainTransactionProof::Raw(PlainRawProof::default())
+                    PlainTransactionProof::Raw(raw_proof)
                 } else if self.inner.sender_type == AccountType::HTLC {
                     HashedTimeLockedContract::parse_proof(
                         &self.inner.proof,
                         !self.inner.network_id.is_albatross(),
                     )
-                    .unwrap()
+                    .unwrap_or_else(|_| PlainTransactionProof::Raw(raw_proof))
```
jsdanielh: Clippy error

web-client/src/common/transaction.rs
```diff
@@ -504,42 +512,40 @@ impl Transaction {
...
+                    .unwrap_or_else(|_| PlainTransactionRecipientData::Raw(raw_recipient_data))
                 } else if self.inner.recipient_type == AccountType::HTLC {
                     HashedTimeLockedContract::parse_data(
                         &self.inner.recipient_data,
                         !self.inner.network_id.is_albatross(),
                         genesis_block_number,
                         genesis_timestamp,
                     )
-                    .unwrap()
+                    .unwrap_or_else(|_| PlainTransactionRecipientData::Raw(raw_recipient_data))
```
jsdanielh: Clippy error

web-client/src/common/transaction.rs
```diff
@@ -504,42 +512,40 @@ impl Transaction {
...
```
jsdanielh: Clippy error

web-client/src/common/transaction.rs
```diff
@@ -467,35 +483,27 @@ impl Transaction {
...
-                    PlainTransactionSenderData::Raw(PlainRawData { raw: raw_data.raw })
+                    PlainTransactionSenderData::Raw(raw_sender_data)
                 }
             },
             data: {
                 if self.inner.recipient_type == AccountType::Staking {
-                    // Parse transaction data
-                    StakingContract::parse_data(&self.inner.recipient_data).unwrap()
+                    StakingContract::parse_data(&self.inner.recipient_data)
+                        .unwrap_or_else(|_| PlainTransactionRecipientData::Raw(raw_recipient_data))
```
jsdanielh: Clippy error

jsdanielh (commented):
Not sure about this solution: we try to parse something and if we can't, we return the raw bytes that were supposed to be parsed in the first place? To me, we should fail with proper error messages

viquezclaudio:
> Not sure about this solution: we try to parse something and if we can't, we return the raw bytes that were supposed to be parsed in the first place? To me, we should fail with proper error messages

_to_plain_transaction()_  returns a PlainTransaction, not a Result, so on parse failure we really have two choices:

- panic
- degrade gracefully

I guess we could change the API, but I dont know if we want to do that.
toPlain() is a formatting/conversion method, not a validation API. The caller still gets a complete object, and valid is already computed separately via self.verify(None).is_ok().

jsdanielh:
> > Not sure about this solution: we try to parse something and if we can't, we return the raw bytes that were supposed to be parsed in the first place? To me, we should fail with proper error messages
> 
> _to_plain_transaction()_ returns a PlainTransaction, not a Result, so on parse failure we really have two choices:
> 
>     * panic
> 
>     * degrade gracefully
> 
> 
> I guess we could change the API, but I dont know if we want to do that. toPlain() is a formatting/conversion method, not a validation API. The caller still gets a complete object, and valid is already computed separately via self.verify(None).is_ok().

Mmm, couldn't we just change the function signature and return `Result<PlainTransaction, Error>`?

PR #3771, opened 2026-05-21
Fixing web client flaky tests

ii-cruz:
Locally, the tests for the web client fail frequently. The problem is that there are tests assuming test policy while others initialize it with the mainnet policy. Since policy is a OnceCell, the nondeterministic nature of the test ordering may result in the wrong policy being initialized and thus a test failure.

This pr fixes it by initializing policy with `TEST_POLICY` and making test-related changes to hard-coded block heights.

## Pull request checklist

### DIFF
--- a/web-client/src/common/transaction.rs
+++ b/web-client/src/common/transaction.rs
@@ -1271,7 +1271,11 @@ extern "C" {
 #[cfg(all(test, feature = "client"))]
 mod tests {
     use nimiq_keys::Address;
-    use nimiq_primitives::{coin::Coin, networks::NetworkId, policy::Policy};
+    use nimiq_primitives::{
+        coin::Coin,
+        networks::NetworkId::UnitAlbatross,
+        policy::{Policy, TEST_POLICY},
+    };
     use nimiq_transaction::{
         historic_transaction::{
             EquivocationEvent, HistoricTransaction, HistoricTransactionData, JailEvent,
@@ -1285,7 +1289,7 @@ mod tests {
 
     fn make_penalize_historic_tx() -> HistoricTransaction {
         HistoricTransaction {
-            network_id: NetworkId::UnitAlbatross,
+            network_id: UnitAlbatross,
             block_number: 100,
             block_time: 1000,
             data: HistoricTransactionData::Penalize(PenalizeEvent {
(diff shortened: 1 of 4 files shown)


sisou (approved):

web-client/src/common/transaction.rs
```diff
@@ -1343,9 +1343,9 @@ mod tests {
 
     #[test]
     fn penalize_inherent_does_not_panic_on_conversion() {
-        let _ = Policy::get_or_init(nimiq_primitives::policy::TEST_POLICY);
+        let _ = Policy::get_or_init(nimiq_primitives::policy::MAINNET_POLICY);
```
jsdanielh: It should work with the test policy as well... 🤔

web-client/src/common/transaction.rs
```diff
@@ -1365,9 +1365,9 @@ mod tests {
 
     #[test]
     fn jail_inherent_does_not_panic_on_conversion() {
-        let _ = Policy::get_or_init(nimiq_primitives::policy::TEST_POLICY);
+        let _ = Policy::get_or_init(nimiq_primitives::policy::MAINNET_POLICY);
```
jsdanielh: Same as above

web-client/src/common/transaction.rs
```diff
@@ -1387,9 +1387,9 @@ mod tests {
 
     #[test]
     fn equivocation_inherent_does_not_panic_on_conversion() {
-        let _ = Policy::get_or_init(nimiq_primitives::policy::TEST_POLICY);
+        let _ = Policy::get_or_init(nimiq_primitives::policy::MAINNET_POLICY);
```
jsdanielh: Same as above

web-client/src/common/transaction.rs
ii-cruz: they do, but it makes all other tests flaky, in my machine, it consistently fails on the consensus tests because these went first and initialized the policy to mainnet first instead of the test policy

jsdanielh: Mmm, but to me no test in the suite should run with the mainnet policy. We use the test policy to speed up block generation

ii-cruz: I changed it to mainnet because i thought the webclient wanted to use the mainnet policy to be extra safe. I can gladly change them all to testnet. @sisou is that ok?

sisou: I have no preference here. Whatever works and fits best with the rest of the tests.

ii-cruz: Perfect, changed the policy used in the web client tests to test-policy instead. @jsdanielh

jsdanielh (approved):
LGTM

PR #3785, opened 2026-06-03
bridge/oracle: Fix validation-program bounds, fee model, oracle revert, and a redundant scan

jsdanielh:
A set of correctness fixes for the new Bridge and Oracle contracts, found via code review of this branch vs `albatross`. Complements #3784 (nonce revert, oversized-amount handling, validation-program creation check); the fixes here are distinct.

## Fixes

**1. Bridge: bounds checks in the ValidationProgram interpreter**
The `Load*` ops computed `offset + length` on values taken from `PushConst`. In release builds the addition can wrap, bypass the `> len` check, and panic on the out-of-bounds slice. Switched to checked arithmetic so an out-of-range offset/length is rejected cleanly. Same pattern fixed for `LoadBytes`/`LoadU64`/`LoadU32`/`LoadAddress`/`LoadEvmU64`.

**2. Bridge: align reserve with the permissionless commit path**
`can_change_balance` (reached only from `reserve_balance`) required the owner's signature, but `commit_outgoing_transaction` intentionally does not — burn-releases are permissionless, authorized by the Merkle burn-proof checked at commit. As a result the mempool rejected transactions that commit would accept. Dropped the owner-signature requirement so mempool admission and commit agree.

**3. History: avoid scanning the epoch twice in `get_keccak256_proof`**
The whole epoch's transactions were fetched twice per request (the first fetch only pulled one tx by index). Reuse the single fetch.

**4. Bridge: make the signer-pays fee model consistent and reserved**
The fee handling was only partly wired: `reserve_balance` reserved value+fee against the bridge; commit-success deducted only `value` and charged the fee to nobody; commit-failure charged the fee to the signer, but the signer was never reserved, so a transaction could pass mempool admission and then fail at commit with `InsufficientFunds`. Now:
- the bridge reserves/releases only `value`;
- the signer pays the fee on both success and failure (charged up front in `try_commit_transaction` so a signer that can't cover it fails before any other state changes; restored on revert);
- the mempool reserves the fee against the signer (`put` / `remove` / per-block rebuild).

**5. Oracle: correct revert of oversized updates**
An Update may carry more hashes than the ring buffer's `hash_count` (intended batch-fill). When a slot is overwritten more than once in a single commit, the eviction receipt recorded the intermediate value rather than the true pre-commit value, so revert restored the wrong predecessors and left the ring buffer inconsistent after a reorg. Now each slot's pre-commit value is recorded only on its first eviction, and `num_evicted` is capped at `hash_count`. Oversized updates keep working; non-oversized behavior is unchanged.

## Tests
- `reserve_balance_excludes_fee_for_bridge` — the bridge reserves value only.
- `it_reverts_oversized_update_on_populated_contract` — reverting an oversized update restores the exact pre-commit ring buffer (verified to fail against the pre-fix code).
- Full `nimiq-account` and `nimiq-mempool` suites pass; `cargo fmt` / `clippy` clean for the changed files.

## Notes
- The mempool reserves the signer fee within the bridge sender's `ReservedBalance` bucket. It aggregates multiple burn-releases through the same bridge by the same signer, but not a signer's obligations across different bridges or with its own non-bridge transactions — those cases can still over-commit a signer. Hardening that needs per-signer-bucket tracking.
- Validated with targeted tests, not full CI.


viquezclaudio:
I think we can close this PR in favor of #3784

PR #3786, opened 2026-06-04
chore: Fix clippy warnings in codebase

jsdanielh:
Fix clippy warnings in codebase.

## Pull request checklist

### DIFF
--- a/consensus/src/consensus/mod.rs
+++ b/consensus/src/consensus/mod.rs
@@ -523,17 +523,17 @@ impl<N: Network> Consensus<N> {
             #[cfg(feature = "full")]
             // The syncing process of the history node does guarantee we have the most recent zkp proof. Thus,
             // we may request it to our peers. This ensures the latest zkp replication across the different node types.
-            SyncerProxy::History(_) => {
-                // Only request a proof when we are too far from the last election head.
-                // This avoids spam requests when the proof for the latest election block is still being generated.
+            //
+            // Only request a proof when we are too far from the last election head.
+            // This avoids spam requests when the proof for the latest election block is still being generated.
+            SyncerProxy::History(_)
                 if self.zkp_proxy.get_zkp_state().latest_block.block_number()
                     < Policy::election_block_before(
                         self.blockchain.read().election_head().block_number(),
-                    )
-                {
-                    self.zkp_proxy
-                        .request_zkp_from_peers(self.sync.peers(), false);
-                }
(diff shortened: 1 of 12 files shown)


consensus/src/consensus/mod.rs
```diff
@@ -523,17 +523,17 @@ impl<N: Network> Consensus<N> {
...
-                    self.zkp_proxy
-                        .request_zkp_from_peers(self.sync.peers(), false);
-                }
+                    ) =>
+            {
+                self.zkp_proxy
+                    .request_zkp_from_peers(self.sync.peers(), false);
             }
             _ => {
                 // The Full and Light sync ensure the latest proof is already pushed, there is no need to
```
hrxi, line 539: This case is now also entered for the `History` sync after these changes. The comment should probably be adjusted accordingly.

network-libp2p/src/discovery/behaviour.rs
```diff
@@ -251,13 +251,11 @@ impl NetworkBehaviour for Behaviour {
...
                 peer_id,
-                other_established,
+                other_established: 0,
                 ..
             }) => {
-                if other_established == 0 {
-                    // This is the first connection to this peer
-                    self.connected_peers.insert(peer_id);
-                }
+                // This is the first connection to this peer
```
hrxi: The comment should probably go to the pattern `other_established: 0` above to keep its meaning.

jsdanielh: Mmm, not sure if I'm following the comment but I think it is already there

network-libp2p/src/discovery/behaviour.rs
```diff
@@ -251,13 +251,11 @@ impl NetworkBehaviour for Behaviour {
             }
             FromSwarm::ConnectionEstablished(ConnectionEstablished {
                 peer_id,
-                other_established,
+                other_established: 0,
```
hrxi:
```suggestion
                other_established: 0, // This is the first connection to this peer
```

I meant like this. To show that the comment refers to the `other_established == 0` — but feel free to ignore if that's already clear from the current comment location.

network-libp2p/src/discovery/behaviour.rs
jsdanielh: Ahh ok, changed it to use the suggested style

viquezclaudio (approved):

PR #3799, opened 2026-06-09
web-client: download wasm-opt for the host platform

jsdanielh:
The build script always fetched the `x86_64-linux` `binaryen` release, so it installed an unrunnable binary on macOS and Linux arm64. Detect OS/arch via `uname` and pick the matching asset, and keep `bin/` + `lib/` together since the macOS `wasm-opt` loads `libbinaryen.dylib` via `@loader_path/../lib`.

## Pull request checklist

### DIFF
--- a/web-client/scripts/build.sh
+++ b/web-client/scripts/build.sh
@@ -98,23 +98,40 @@ then
     BINARYEN_VERSION=129
     BINARYEN_FOLDER="binaryen-version_$BINARYEN_VERSION"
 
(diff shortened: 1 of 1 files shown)


sisou (approved):
I'm assuming you tested this on the now supported platforms, as I can't. Code-wise it looks good.

jsdanielh:
> I'm assuming you tested this on the now supported platforms, as I can't. Code-wise it looks good.

Did a run in x86 and looked good

PR #3800, opened 2026-06-09
exposing protocol upgrade version and readd txs to mempool upon protocol upgrade

ii-cruz:
- Adds the protocol version to the blockchain, account tx commit, and to the tx verification
- A new blockchain event for the protocol upgrade
- The removal of duplicate work on listeners for blockchain events. The `EpochFinalized` and `Finalized` events are emitted in addition to the `Extended` event for the election or macro block that was pushed.
- The reverification of txs in the mempool after the protocol upgrade
- Caches the current protocol version in the mempool and updates it via blockchain event

Web client breaking changes:
- Removes the `valid` field from transaction `PlainTransaction`/`PlainTransactionDetails`
- The transaction `verify` function signature now requires `protocol_version`

## Pull request checklist


web-client/src/common/transaction.rs
```diff
@@ -177,13 +177,15 @@ impl Transaction {
     /// **Throws with any transaction validity error.** Returns without exception if the transaction is valid.
     ///
     /// Throws when the given networkId is unknown.
-    pub fn verify(&self, network_id: Option<u8>) -> Result<(), JsError> {
+    pub fn verify(&self, network_id: Option<u8>, protocol_version: u16) -> Result<(), JsError> {
```
sisou: Generally it's not good practice to have a required parameter after an optional parameter. Certain JS linters even forbid it.

Can the protocol version not please have a default?

web-client/extras/tests/Transaction.test.ts
```diff
@@ -96,7 +96,9 @@ describe('Transaction', () => {
         const plain2 = tx.toPlain();
         expect(plain2.proof).toEqual(plain.proof);
         expect(plain2.size).toEqual(plain.size);
-        expect(plain2.valid).toEqual(plain.valid);
+        // `toPlain()` no longer self-verifies: a transaction's validity now depends on the
+        // protocol version and is established by consensus, so it defaults to `false` here.
+        expect(plain2.valid).toBe(false);
```
sisou: Seeing this comment, I'd rather we remove the `valid` property from the plain tx, as it doesn't mean anything anymore if it's always `false` anyway. Plus, it will definitely confuse devs to show `false` for a valid tx.

web-client/src/common/transaction.rs
```diff
@@ -548,7 +553,7 @@ impl Transaction {
                 }
             },
             size: self.serialized_size(),
-            valid: self.verify(None).is_ok(),
+            valid: false,
```
sisou: I think I would be in favor of changing the type to `valid: Option(Bool)`. Because in this case it would be good to set it to `None` (undefined in JS), which has the most fitting semantic meaning (because it really is undefined - we don't know if the tx is valid or not). 

But setting it to `false` as if it was invalid is not great.

So if we keep the `valid` property (and maybe we shouldn't), then it must be extended to be able to represent unknown validity.

web-client/src/common/transaction.rs
```diff
@@ -1343,9 +1349,9 @@ mod tests {
 
     #[test]
     fn penalize_inherent_does_not_panic_on_conversion() {
-        let _ = Policy::get_or_init(nimiq_primitives::policy::TEST_POLICY);
+        let _ = Policy::get_or_init(nimiq_primitives::policy::MAINNET_POLICY);
         let hist_tx = make_penalize_historic_tx();
-        let current_block = 200;
+        let current_block = Policy::genesis_block_number() + 1;
```
sisou, line 1344: Can the change to `Policy::genesis_block_number() + 1` not also be done with the `TEST_POLICY`? Or do these two changed lines only work together?

sisou (changes requested):

sisou (approved):
The web-client changes look good to me now.

However, IMO it should be documented in the Policy mainnet defaults, that (next time) `max_supported_version` should only be increased in the code release that has the full support for the new version (e.g. the code version that is used for the hard fork).

mempool/src/mempool.rs
```diff
@@ -394,6 +394,20 @@ impl Mempool {
         Mempool::recompute_sender_balances(affected_senders, &blockchain, &mut mempool_state);
     }
 
+    /// After a protocol upgrade there might be transactions that are no longer valid because of new rules introduced by the
+    /// protocol upgrade.
+    /// This revalidates all transactions in the mempool and removes the ones that are no longer valid.
+    pub fn readd_transactions(&self) {
```
paberr: -> `read_transactions` (typo)
actually the name is a bit weird. shouldn't it be `revalidate_transactions` or so?

web-client/src/common/transaction.rs
```diff
@@ -177,13 +175,15 @@ impl Transaction {
     /// **Throws with any transaction validity error.** Returns without exception if the transaction is valid.
     ///
     /// Throws when the given networkId is unknown.
-    pub fn verify(&self, network_id: Option<u8>) -> Result<(), JsError> {
+    pub fn verify(&self, protocol_version: u16, network_id: Option<u8>) -> Result<(), JsError> {
```
paberr, line 180: Arg-order inconsistency with the native API: `nimiq_transaction::Transaction::verify(network_id, protocol_version)` takes the network ID first, but this WASM `verify(protocol_version, network_id)` reverses them. It's easy to transpose the two when moving between the two APIs. Consider matching the native order (`verify(network_id, protocol_version)`).

web-client/src/common/transaction.rs
```diff
@@ -906,9 +905,6 @@ pub struct PlainTransaction {
     /// The transaction's serialized size in bytes. It is used to determine the fee-per-byte that this
     /// transaction pays.
     pub size: usize,
-    /// Encodes if the transaction is valid, meaning the signature is valid and the `data` and `proof` fields
-    /// follow the correct format for the transaction's recipient and sender type, respectively.
-    pub valid: bool,
```
paberr, line 911: Breaking change to the web-client output schema: removing `valid` from `PlainTransaction`/`PlainTransactionDetails` drops `valid` from the JS `tx.toPlain()` / transaction-details output. This is reasonable (it can't be computed without a protocol version in this context), but it should be called out in the changelog/release notes, since downstream JS consumers reading `.valid` will break. The same release-note note applies to the new required `protocol_version` argument added to `verify` (a breaking signature change for JS callers).

mempool/src/verify.rs
```diff
@@ -41,12 +41,12 @@ pub(crate) fn verify_tx(
...
 ) -> Result<(), VerifyErr> {
-    // 1. Verify transaction signature (and other stuff)
-    transaction.verify_mut(network_id)?;
-
-    // 2. Acquire blockchain read lock
+    // 1. Acquire blockchain read lock
     let blockchain = blockchain.read();
 
+    // 2. Verify transaction signature (and other stuff)
+    transaction.verify_mut(network_id, blockchain.state.current_version())?;
```
paberr: Reordering the verification to after `blockchain.read()` (so `current_version()` is available) means signature verification now runs while holding the blockchain read lock. Signature verification is CPU-heavy, so this lengthens the time the read lock is held and increases contention with block-push writers. Consider snapshotting `current_version()` under a short-lived read, releasing the lock, verifying, then re-acquiring for the validity-window/balance checks.

mempool/mempool-task/src/lib.rs
```diff
@@ -161,6 +161,17 @@ impl<N: Network> MempoolTask<N> {
             {
                 self.mempool.update(new_chain, old_chain);
             }
+            BlockchainEvent::ProtocolUpgrade(_, version) => {
```
paberr: Two small things on this arm:

1. Unlike the sibling arms (`Extended`/`Rebranched`/`HistoryAdopted`), this one doesn't gate on `self.consensus.is_ready_for_validation()`. It's harmless when the pool is empty/inactive, but it's inconsistent — consider gating it the same way.
2. The second `debug!("Reset all mempool transactions…")` after `readd_transactions()` is redundant and slightly misleading (nothing is "reset" at that point). One log line before the call is enough.

primitives/account/tests/staking_contract/mod.rs
```diff
@@ -21,6 +21,8 @@ use nimiq_transaction::{
 
 mod punished_slots;
 mod staker;
+mod staker_add_stake_policy;
+mod staker_tombstone_interaction;
```
paberr, line 25: Scope/size: this PR bundles a sizable staking-contract test refactor and expansion (new `staker_tombstone_interaction.rs` ~+1190 lines and `staker_add_stake_policy.rs` ~+460, plus the split of `staker.rs`). It's only tangentially related to the protocol-upgrade work — driven mostly by `BlockState::new` now requiring a protocol version — and it significantly inflates the review surface. If practical, splitting the test refactor into its own PR would make both easier to review.

web-client/src/common/transaction.rs
ii-cruz: this is on purpose because the option should always be the last parameter of a function for js

blockchain-interface/src/error.rs
```diff
@@ -13,7 +13,8 @@ pub enum ForkEvent {
 }
 
 /// Events from the blockchain.
-/// Note that `Finalized` and `EpochFinalized` will be sent **in addition** to `Extended` events.
+/// Note that `Finalized`, `EpochFinalized`, and `ProtocolUpgrade` may be emitted alongside
```
nibhar: The 'will' was correct. 'May' implies that i could be differently but it is not, they are always sent alongside `Extended`.

blockchain/src/block_production/mod.rs
```diff
@@ -339,12 +339,14 @@ impl BlockProducer {
             .punished_slots
             .next_batch_initial_punished_set(block_number, &staking_contract.active_validators);
 
+        let version = version.unwrap_or_else(|| blockchain.state().current_version());
```
nibhar, line 346: here we probably want to sanity check:
```suggestion
        let version = version.unwrap_or_else(|| blockchain.state().current_version());
        if blockchain.state().current_version() != version 
            && blockchain.state().current_version() + 1 != version {
            // error as this will be a bigger version bump than 1.
        }
```

blockchain/src/blockchain/push.rs
```diff
@@ -426,8 +427,16 @@ impl Blockchain {
             .ok();
         if is_election_block {
             this.notifier
-                .send(BlockchainEvent::EpochFinalized(block_hash))
+                .send(BlockchainEvent::EpochFinalized(block_hash.clone()))
```
nibhar: if you swap them around you can loose a clone and only clone when you need to. Notify the upgrade first with the clone and the notify then epoch finalized.

blockchain/src/blockchain/macro_sync.rs
```diff
@@ -195,8 +194,16 @@ impl Blockchain {
 
         if is_election_block {
             this.notifier
-                .send(BlockchainEvent::EpochFinalized(block_hash))
+                .send(BlockchainEvent::EpochFinalized(block_hash.clone()))
```
nibhar: Here you can also get rid of the unnecessary clone and only clone if you need to when you swap the notifiers.

blockchain/src/blockchain/wrappers.rs
```diff
@@ -170,7 +170,8 @@ impl Blockchain {
         transaction: &Transaction,
         reserved_balance: &mut ReservedBalance,
     ) -> Result<(), AccountError> {
-        let block_state = BlockState::new(self.block_number(), self.timestamp());
+        let block_state =
+            BlockState::new(self.block_number(), self.timestamp(), self.head().version());
```
nibhar, line 174: What is the semantics of this version here?

blockchain/src/blockchain_state.rs
```diff
@@ -30,6 +30,6 @@ pub struct BlockchainState {
 impl BlockchainState {
     /// The current protocol version.
     pub fn current_version(&self) -> u16 {
-        self.main_chain.head.version()
+        self.election_head.header.version
```
nibhar: This change should be inconsequential, and I actually prefer the previous version as it is the most recent block.

blockchain/tests/history_sync.rs
```diff
@@ -268,6 +283,75 @@ fn history_sync_works() {
...
+            &upgrade_history
+        ),
+        Ok(PushResult::Extended)
+    );
+    assert_eq!(
+        collect_blockchain_events(&mut receiver),
+        vec![
+            BlockchainEvent::HistoryAdopted(upgrade_hash.clone()),
+            BlockchainEvent::EpochFinalized(upgrade_hash.clone()),
+            BlockchainEvent::ProtocolUpgrade(upgrade_hash, upgrade_version),
```
nibhar, line 339: This will break if you swap the order around as I suggested earlier. A better, order agnostic, check would be to find each of the events you would expect in the vec and do not require a specific order.

blockchain/tests/inherents.rs
```diff
@@ -116,7 +116,8 @@ fn it_can_create_batch_finalization_inherents() {
             &[penalize_inherent],
             &BlockState::new(
                 Policy::blocks_per_batch() + 1 + Policy::genesis_block_number(),
-                1
+                1,
+                Policy::max_supported_version()
```
nibhar: Not so sure this is correct. Why not `blockchain.version()`? IF it should be 2 and not 1, what about in 5 forks and `max_supported_version` is 6?

blockchain/tests/inherents.rs
```diff
@@ -825,7 +826,8 @@ fn it_burns_all_rewards_when_all_slots_penalized() {
                 &[penalize_inherent],
                 &BlockState::new(
                     Policy::blocks_per_batch() + 1 + Policy::genesis_block_number(),
-                    1
+                    1,
+                    Policy::max_supported_version()
```
nibhar: Same as above.

blockchain/tests/history_sync.rs
```diff
@@ -2,19 +2,34 @@ use std::sync::Arc;
...
+    let mut events = Vec::new();
+
+    loop {
+        match receiver.try_recv() {
+            Ok(event) => events.push(event),
+            Err(TryRecvError::Empty) => return events,
+            Err(error) => panic!("unexpected broadcast error: {error:?}"),
+        }
+    }
+}
```
nibhar: If the panic is not relevant (which it does not seem to be) `.try_collect()` instead of this function should do the trick. Also duplicated code.

blockchain/tests/push.rs
```diff
@@ -15,12 +17,30 @@ use nimiq_primitives::{
...
+    let mut events = Vec::new();
+
+    loop {
+        match receiver.try_recv() {
+            Ok(event) => events.push(event),
+            Err(TryRecvError::Empty) => return events,
+            Err(error) => panic!("unexpected broadcast error: {error:?}"),
+        }
+    }
+}
```
nibhar: Duplicate code. `.try_collect()` should work here. If it does not work here, put this into test-utils or such to dedupe.

blockchain/tests/push.rs
```diff
@@ -239,6 +259,77 @@ fn simply_push_election_block(config: &BlockConfig, expected_res: &Result<PushRe
...
+
+    assert_eq!(temp_producer.push(upgrade_block), Ok(PushResult::Extended));
+    assert_eq!(
+        collect_blockchain_events(&mut receiver),
+        vec![
+            BlockchainEvent::Extended(upgrade_hash.clone()),
+            BlockchainEvent::EpochFinalized(upgrade_hash.clone()),
+            BlockchainEvent::ProtocolUpgrade(upgrade_hash, upgrade_version),
+        ],
+    );
```
nibhar: Same as already mentioned before.

blockchain/tests/push.rs
```diff
@@ -239,6 +259,77 @@ fn simply_push_election_block(config: &BlockConfig, expected_res: &Result<PushRe
...
+    };
+    let election_hash = election_block.hash();
+
+    assert_eq!(temp_producer.push(election_block), Ok(PushResult::Extended));
+    assert_eq!(
+        collect_blockchain_events(&mut receiver),
+        vec![
+            BlockchainEvent::Extended(election_hash.clone()),
+            BlockchainEvent::EpochFinalized(election_hash),
+        ],
```
nibhar, line 311: Same as before.

consensus/src/consensus/remote_event_dispatcher.rs
```diff
@@ -245,9 +245,7 @@ impl<N: Network> Future for RemoteEventDispatcher<N> {
             let mut new_blocks = vec![];
 
             match event {
-                BlockchainEvent::Extended(block_hash)
-                | BlockchainEvent::EpochFinalized(block_hash)
-                | BlockchainEvent::Finalized(block_hash) => {
+                BlockchainEvent::Extended(block_hash) => {
                     let block = self
```
nibhar, line 249: So this does change how things work. Was the previous code a bug?

consensus/src/sync/live/state_queue/mod.rs
```diff
@@ -555,12 +553,18 @@ impl<N: Network> Stream for StateQueue<N> {
...
+                }
                 BlockchainEvent::HistoryAdopted(_) => {
                     // Nothing to do for adopted history
                 }
                 BlockchainEvent::Stored(_block) => {
                     // Block has not been applied so nothing to do here.
                 }
+                BlockchainEvent::ProtocolUpgrade(..) => {
+                    // Nothing todo for protocol upgrade.
+                }
```
nibhar: inconsistent to how it was added previously. I would prefer each having their own match arm here and in the previous spot, but it should at least be done equally.

consensus/src/sync/live/state_queue/mod.rs
```diff
@@ -514,9 +514,7 @@ impl<N: Network> Stream for StateQueue<N> {
         // Reset on a rebranch (since we potentially revert to an incomplete state).
         while let Poll::Ready(Some(event)) = self.blockchain_rx.poll_next_unpin(cx) {
             match event {
-                BlockchainEvent::Finalized(_)
-                | BlockchainEvent::EpochFinalized(_)
-                | BlockchainEvent::Extended(_) => {
+                BlockchainEvent::Extended(_) => {
```
nibhar, line 517: Same question as before.

test-utils/src/blockchain.rs
```diff
@@ -259,6 +267,127 @@ pub fn validator_address() -> Address {
...
+        next_election_block_with_version(producer, blockchain, next_version),
+        next_version,
+    )
+}
+
+pub fn signal_next_protocol_version_via_tx(
+    producer: &BlockProducer,
+    blockchain: &Arc<RwLock<Blockchain>>,
+) -> u16 {
+    let next_version = blockchain.read().state.current_version() + 1;
```
nibhar, line 357: what happens when `next_version > max_supported_version`? Blockchain is a parameter and can have any possible state.

test-utils/src/blockchain.rs
```diff
@@ -259,6 +267,127 @@ pub fn validator_address() -> Address {
     Address::from(&validator_key())
 }
 
+pub fn signal_all_validators_support_for_next_version(blockchain: &Arc<RwLock<Blockchain>>) -> u16 {
+    let blockchain_read = blockchain.read();
+    let next_version = blockchain_read.state.current_version() + 1;
```
nibhar, line 272: same question as before

primitives/transaction/src/account/staking_contract/structs.rs
```diff
@@ -105,7 +105,11 @@ impl IncomingStakingTransactionData {
         Ok(Self::deserialize_all(&transaction.recipient_data)?)
     }
 
-    pub fn verify(&self, transaction: &Transaction) -> Result<(), TransactionError> {
+    pub fn verify(
+        &self,
+        transaction: &Transaction,
+        _protocol_version: u16,
```
nibhar, line 111: Seems not needed.

primitives/account/tests/staking_contract/mod.rs
```diff
@@ -157,8 +159,20 @@ fn make_delete_validator_transaction() -> Transaction {
...
+        data_store,
+        staker_active_balance,
+        Policy::max_supported_version(),
+    )
+}
+
+fn make_sample_contract_with_protocol_version(
     mut data_store: DataStoreWrite,
     staker_active_balance: Option<u64>,
+    _protocol_version: u16,
```
nibhar, line 175: This does not seem right. Why do we make a contract with a version and then never use it?

primitives/account/tests/staking_contract/mod.rs
```diff
@@ -231,6 +245,7 @@ impl ValidatorSetup {
         before_state_release_block_state: BlockState,
         state_release_block_state: BlockState,
         staker_active_balance: Option<u64>,
+        _protocol_version: u16,
```
nibhar, line 248: Why is this added when it is never used?

primitives/account/src/interaction_traits.rs
```diff
@@ -14,13 +14,15 @@ use crate::{
 pub struct BlockState {
     pub number: u32,
     pub time: u64,
+    pub protocol_version: u16,
```
nibhar, line 17: I am not so sure I like this spot to add the version too much. It seems not really related to what this was used for before. I think the version could have just stood on its own and only put where it is actually needed. What are the semantics behind putting it into the BlockState. Maybe I am just missing the follow up to this though.

primitives/account/tests/staking_contract/staker.rs
```diff
@@ -308,228 +309,6 @@ fn create_staker_works() {
     );
 }
 
-#[test]
```
nibhar, line 311: What happened here? These tests seem useful? If this is just copy over to a different file I would strongly prefer not to do this as it just introduces a huge changeset that does nothing, just bloat. These tests were alright here and other tests regarding Staker would also have been alright here.

primitives/account/tests/staking_contract/staker.rs
```diff
@@ -2299,86 +2084,6 @@ fn remove_stake_works() {
...
-fn remove_stake_from_tombstone_works() {
```
nibhar, line 2303: Same here. This change is just bloat if it is only moving them to a different file. And i would very strongly prefer not to do these kinds of changes in a multi thousand line changeset PR like this.

primitives/account/tests/staking_contract/staker_add_stake_policy.rs
```diff
@@ -0,0 +1,459 @@
...
+fn add_stake(protocol_version: u16) {
+    let env = MdbxDatabase::new_volatile(Default::default()).unwrap();
+    let accounts = Accounts::new(env.clone());
+    let data_store = accounts.data_store(&Policy::STAKING_CONTRACT_ADDRESS);
+    let block_state = BlockState::new(2, 2, protocol_version);
+    let mut db_txn = env.write_transaction();
+    let mut db_txn = (&mut db_txn).into();
+
+    let (validator_address, staker_address, mut staking_contract) =
+        make_sample_contract(data_store.write(&mut db_txn), Some(150_000_000));
```
nibhar: should this not be `make_sample_contract_with_version`, It does nothing different , as mentioned before, but this I think should still use the version given as argument probably.

consensus/src/sync/live/state_queue/mod.rs
```diff
@@ -555,12 +553,18 @@ impl<N: Network> Stream for StateQueue<N> {
...
+                    // Nothing to do for finalized batches since macro block pushes emit the `Extended` event as well.
+                }
                 BlockchainEvent::HistoryAdopted(_) => {
                     // Nothing to do for adopted history
                 }
                 BlockchainEvent::Stored(_block) => {
                     // Block has not been applied so nothing to do here.
                 }
+                BlockchainEvent::ProtocolUpgrade(..) => {
+                    // Nothing todo for protocol upgrade.
```
viquezclaudio: Nothing to do

genesis/src/networks.rs
```diff
@@ -56,6 +56,8 @@ pub struct NetworkInfo {
     network_id: NetworkId,
     name: &'static str,
     genesis: GenesisData,
+    // This should only be increased when the hard fork code is released.
```
viquezclaudio: this should only be increased when  a new hard fork code is released

mempool/src/mempool.rs
```diff
@@ -394,6 +403,20 @@ impl Mempool {
         Mempool::recompute_sender_balances(affected_senders, &blockchain, &mut mempool_state);
     }
 
+    /// After a protocol upgrade there might be transactions that are no longer valid because of new rules introduced by the
+    /// protocol upgrade.
+    /// This revalidates all transactions in the mempool and removes the ones that are no longer valid.
+    pub fn revalidate_transactions(&self) {
```
viquezclaudio, line 406: I think this could be done in-place like this:

```
pub fn revalidate_transactions(&self) {
      // Lock order matches update()/cleanup(): blockchain first, then state.
      let blockchain = self.blockchain.read();
      let network_id = blockchain.network_id;
      let version = blockchain.state.current_version();
      let mut state = self.state.write();

      // Find transactions that no longer pass verification under the new protocol version.
      let invalid: Vec<Blake2bHash> = state
          .regular_transactions
          .transactions
          .values()
          .chain(state.control_transactions.transactions.values())
          .filter_map(|tx| {
              let mut tx = tx.clone();
              tx.reset_verification_state();
              tx.verify(network_id, version).err().map(|_| tx.hash())
          })
          .collect();

      // Remove them; `remove()` releases each sender's reserved balance and prunes empty sender state.
      for hash in invalid {
          state.remove(&blockchain, &hash, EvictionReason::Invalid);
      }
  }
```

primitives/account/tests/staking_contract/staker_add_stake_policy.rs
```diff
@@ -0,0 +1,459 @@
...
+    // -----------------------------------
+    let mut staker_setup = StakerSetup::setup_staker_with_inactive_retired_balance_and_protocol(
+        ValidatorState::Active,
+        0,
+        50_000_000,
+        Policy::MINIMUM_STAKE + 1,
+        Policy::max_supported_version(),
+    );
+    assert!(staker_setup.active_stake < staker_setup.retired_stake);
+    assert!(staker_setup.active_stake < staker_setup.inactive_stake);
```
nibhar: These two seem weird to me. They make sure the constants from 2 lines above are not malformed. That should probably be done within `setup_staker_with_inactive_retired_balance_and_protocol` if that state is forbidden. If it is not forbidden, then it does not make much sense here either, as the values were chosen specifically such that these asserts never fire.

primitives/account/tests/staking_contract/staker_add_stake_policy.rs
```diff
@@ -0,0 +1,459 @@
...
+        staking_contract.active_validators.get(&validator_address),
+        Some(&Coin::from_u64_unchecked(
+            Policy::VALIDATOR_DEPOSIT + 150_000_000
+        ))
+    );
+}
+
+#[test]
+fn add_stake_works() {
+    add_stake(Policy::max_supported_version());
```
nibhar: `add_stake(version)` is never reused. Only callsite is here and the function extraction seems not very useful.

primitives/account/tests/staking_contract/staker_add_stake_policy.rs
```diff
@@ -0,0 +1,459 @@
...
+    assert_eq!(
+        staker_setup.staking_contract.balance,
+        Coin::from_u64_unchecked(
+            Policy::VALIDATOR_DEPOSIT + 50_000_000 + Policy::MINIMUM_STAKE + 1
+        )
+    );
+}
+
+#[test]
+fn add_stake_enforces_minimum_stake_works() {
```
nibhar, line 289: This test has changed compared to what was removed from `./staker.rs` It is super tedious to review that but I see no reason why it should have changed so far. 

The code block with the comment "Cannot add less than minimum stake" was removed, which to me seems like reduced code coverage.

primitives/account/tests/staking_contract/staker_add_stake_policy.rs
```diff
@@ -0,0 +1,459 @@
...
+    );
+}
+
+#[test]
+fn add_stake_works() {
+    add_stake(Policy::max_supported_version());
+}
+
+#[test]
+fn add_stake_priority_works() {
```
nibhar: What does priority mean?

nibhar: I am not quite sure about it, but this test seems very samey to `add_stake_enforces_minimum_stake_works()` Function extraction here may be a nice idea to reduce code a bit and to allow for easy addition of other active-/inactive-/retired-stake cirsumstances.

primitives/account/tests/staking_contract/staker_add_stake_policy.rs
```diff
@@ -0,0 +1,459 @@
...
+        Coin::from_u64_unchecked(Policy::VALIDATOR_DEPOSIT + Policy::MINIMUM_STAKE)
+    );
+    assert_eq!(
+        staker_setup.staking_contract.balance,
+        Coin::from_u64_unchecked(Policy::VALIDATOR_DEPOSIT + 50_000_000 + Policy::MINIMUM_STAKE)
+    );
+}
+
+#[test]
+fn add_stake_enforces_greater_than_zero_works() {
```
nibhar, line 391: This also could benefit from the function extraction. Something like
``` 
test_add_stake(
    active: u64, 
    inactive: u64, 
    retired: u64, 
    add_stake_tx: Transaction,
    version: u16
) -> (Staker, Validator)
```
might not even need the return and can assert everything within the function call.

consensus/tests/sync_utils.rs
```diff
@@ -265,3 +273,111 @@ pub async fn sync_two_peers(
         BlockchainProxy::Light(_) => {}
     }
 }
+
+pub async fn sync_two_peers_across_protocol_upgrade(sync_mode: SyncMode) {
```
nibhar, line 277: As I underastand it this will test the sync across the version 1 -> 2 upgrade. Any other upgrade would have to be added or created an additional test for. Could potentially be done generically, but for today not a priority.

lib/src/client.rs
```diff
@@ -614,6 +614,11 @@ impl Client {
         self.inner.blockchain.read().head().clone()
     }
 
+    /// Returns the blockchain protocol version
+    pub fn protocol_version(&self) -> u16 {
+        self.inner.blockchain.read().protocol_version()
```
nibhar, line 619: I don't like these short lived `.read()`s. Data consistency is a concern here.

light-blockchain/src/blockchain.rs
```diff
@@ -97,6 +97,10 @@ impl LightBlockchain {
         this.current_validators = genesis_block.validators()
     }
 
+    pub fn current_version(&self) -> u16 {
+        self.election_head.header.version
```
nibhar: Could and probably should also be head block version instead of election head.

light-blockchain/src/push.rs
```diff
@@ -208,8 +209,16 @@ impl LightBlockchain {
...
+                .send(BlockchainEvent::EpochFinalized(block_hash.clone()))
                 .ok();
+            if is_protocol_upgrade {
+                this.notifier
+                    .send(BlockchainEvent::ProtocolUpgrade(
+                        block_hash,
+                        this.current_version(),
+                    ))
+                    .ok();
+            }
```
nibhar, line 218: if you put this before the previous notifier you can clone there and only clone when necessary.

light-blockchain/src/sync.rs
```diff
@@ -81,17 +81,26 @@ impl LightBlockchain {
...
                 .ok();
+
+            if is_protocol_upgrade {
+                this.notifier
+                    .send(BlockchainEvent::ProtocolUpgrade(
+                        block_hash,
+                        this.current_version(),
+                    ))
+                    .ok();
+            }
```
nibhar, line 101: Can also reorder here for a cleaner `.clone()`.

light-blockchain/src/sync.rs
```diff
@@ -155,18 +164,25 @@ impl LightBlockchain {
...
             .ok();
-
+        if is_protocol_upgrade {
+            this.notifier
+                .send(BlockchainEvent::ProtocolUpgrade(
+                    block_hash,
+                    this.current_version(),
+                ))
+                .ok();
+        }
```
nibhar, line 185: Here, too.

light-blockchain/tests/push.rs
```diff
@@ -12,10 +12,47 @@ use nimiq_primitives::policy::Policy;
...
+    let mut events = Vec::new();
+
+    loop {
+        match receiver.try_recv() {
+            Ok(event) => events.push(event),
+            Err(TryRecvError::Empty) => return events,
+            Err(error) => panic!("unexpected broadcast error: {error:?}"),
+        }
+    }
+}
```
nibhar: Duplicate code. Look into `Stream::try_collect` or `Stream::collect`. if that does not help move this to a central place.

light-blockchain/tests/push.rs
```diff
@@ -279,6 +316,96 @@ fn simply_push_macro_block(config: &BlockConfig, expected_res: &Result<PushResul
...
+        initial_version.add(1),
+        upgrade_version,
+        "Initial blockchain should be one version bellow the targeted upgrade version"
+    );
+    let upgrade_hash = upgrade_block.hash();
+
+    assert_eq!(
+        LightBlockchain::push(
+            light_blockchain.upgradable_read(),
+            remove_micro_body(upgrade_block),
```
nibhar: Isn't this a macro block? Does this make sense?

light-blockchain/tests/push.rs
```diff
@@ -279,6 +316,96 @@ fn simply_push_macro_block(config: &BlockConfig, expected_res: &Result<PushResul
...
+            &blockchain,
+            &BlockConfig::default(),
+        )
+    };
+    let election_hash = election_block.hash();
+
+    assert_eq!(
+        LightBlockchain::push(
+            light_blockchain.upgradable_read(),
+            remove_micro_body(election_block),
```
nibhar: It is a macro block, probably does not need to remove the micro body.

light-blockchain/tests/push.rs
```diff
@@ -279,6 +316,96 @@ fn simply_push_macro_block(config: &BlockConfig, expected_res: &Result<PushResul
...
+        ),
+        Ok(PushResult::Extended),
+    );
+    assert_eq!(
+        collect_blockchain_events(&mut receiver),
+        vec![
+            BlockchainEvent::Extended(election_hash.clone()),
+            BlockchainEvent::EpochFinalized(election_hash),
+        ],
+    );
```
nibhar: Similar to before I think it would be great if this was order agnostic.

light-blockchain/tests/push.rs
```diff
@@ -279,6 +316,96 @@ fn simply_push_macro_block(config: &BlockConfig, expected_res: &Result<PushResul
...
+        Ok(PushResult::Extended),
+    );
+    assert_eq!(
+        collect_blockchain_events(&mut receiver),
+        vec![
+            BlockchainEvent::Extended(upgrade_hash.clone()),
+            BlockchainEvent::EpochFinalized(upgrade_hash.clone()),
+            BlockchainEvent::ProtocolUpgrade(upgrade_hash, upgrade_version),
+        ],
+    );
```
nibhar: Similar to before I think it would be great if this was order agnostic.

primitives/account/tests/staking_contract/staker_tombstone_interaction.rs
nibhar, line 1: Very nice set of additional tests 👍! Can probably extract some functions and make this a bit more concise and easier to add additional tests.

primitives/account/tests/staking_contract/staker_tombstone_interaction.rs
```diff
@@ -0,0 +1,1191 @@
...
+    // Test setup:
+    // -----------------------------------
+    let mut staker_setup = StakerSetup::setup_staker_with_inactive_retired_balance(
+        ValidatorState::Deleted,
+        1,
+        Policy::MINIMUM_STAKE,
+        50_000_000,
+    );
+    assert!(staker_setup.active_stake < staker_setup.retired_stake);
+    assert!(staker_setup.active_stake < staker_setup.inactive_stake);
```
nibhar: Again a couple of asserts I find weird as the constants are literally 3 lines above and handpicked to be fulfilling these.

mempool/src/mempool.rs
ii-cruz: we cannot do that because the tx upon version upgrade may change the way we reserve balances as well.

edit: we can do this after all. i'll change it

primitives/account/tests/staking_contract/staker_tombstone_interaction.rs
```diff
@@ -0,0 +1,1191 @@
...
+        .data_store(&Policy::STAKING_CONTRACT_ADDRESS);
+    let mut db_txn = staker_setup.env.write_transaction();
+    let mut db_txn = (&mut db_txn).into();
+    let staker_keypair = ed25519_key_pair(STAKER_PRIVATE_KEY);
+    let initial_staking_contract_balance = staker_setup.staking_contract.balance;
+
+    // -----------------------------------
+    // Test execution:
+    // -----------------------------------
+    // Add stake operation fails due to unexisting validator.
```
nibhar:
```suggestion
    // Add stake operation fails due to non-existent validator.
```
It is not wrong, but this is a much more commonly used word.

primitives/account/tests/staking_contract/staker_tombstone_interaction.rs
```diff
@@ -0,0 +1,1191 @@
...
+    assert_eq!(
+        staker.delegation,
+        Some(staker_setup.validator_address.clone())
+    );
+    assert_eq!(staker.active_balance, Coin::from_u64_unchecked(1));
+    assert_eq!(
+        staker.inactive_balance,
+        Coin::from_u64_unchecked(Policy::MINIMUM_STAKE)
+    );
+    assert_eq!(staker.inactive_from, Some(328));
```
nibhar: Where does the `328` come from? Should probably be computed from constants instead.

primitives/account/tests/staking_contract/staker_tombstone_interaction.rs
```diff
@@ -0,0 +1,1191 @@
...
+    assert_eq!(validator.num_remaining_stakers, 1);
+    assert_eq!(validator.remaining_stake, Coin::from_u64_unchecked(1));
+    assert_eq!(
+        staker_setup.staking_contract.balance,
+        initial_staking_contract_balance
+    );
+}
+
+#[test]
+fn add_stake_to_tombstone_does_not_legacy_work_0v() {
```
nibhar, line 142: Did anything here change, or is there another reason why this must be tested separately? I wonder if general we need to test these things for every version or just for pre and post change. Also the two tests are except for a single call identical, so can probably be simplified a lot.

primitives/account/tests/staking_contract/staker_tombstone_interaction.rs
```diff
@@ -0,0 +1,1191 @@
...
+    // -----------------------------------
+    let mut staker_setup = StakerSetup::setup_staker_with_inactive_retired_balance_and_protocol(
+        ValidatorState::Deleted,
+        1,
+        Policy::MINIMUM_STAKE,
+        50_000_000,
+        0,
+    );
+    assert!(staker_setup.active_stake < staker_setup.retired_stake);
+    assert!(staker_setup.active_stake < staker_setup.inactive_stake);
```
nibhar: Handpicked constants being asserted.

primitives/account/tests/staking_contract/staker_tombstone_interaction.rs
```diff
@@ -0,0 +1,1191 @@
...
+        .data_store(&Policy::STAKING_CONTRACT_ADDRESS);
+    let mut db_txn = staker_setup.env.write_transaction();
+    let mut db_txn = (&mut db_txn).into();
+    let staker_keypair = ed25519_key_pair(STAKER_PRIVATE_KEY);
+    let initial_staking_contract_balance = staker_setup.staking_contract.balance;
+
+    // -----------------------------------
+    // Test execution:
+    // -----------------------------------
+    // Add stake operation fails due to unexisting validator.
```
nibhar:
```suggestion
    // Add stake operation fails due to non-existent validator.
```

primitives/account/tests/staking_contract/staker_tombstone_interaction.rs
```diff
@@ -0,0 +1,1191 @@
...
+    assert_eq!(
+        staker.delegation,
+        Some(staker_setup.validator_address.clone())
+    );
+    assert_eq!(staker.active_balance, Coin::from_u64_unchecked(1));
+    assert_eq!(
+        staker.inactive_balance,
+        Coin::from_u64_unchecked(Policy::MINIMUM_STAKE)
+    );
+    assert_eq!(staker.inactive_from, Some(328));
```
nibhar: Magic number, should be calculated from constants.

primitives/account/tests/staking_contract/staker_tombstone_interaction.rs
```diff
@@ -0,0 +1,1191 @@
...
+    assert_eq!(
+        staker_setup
+            .staking_contract
+            .get_tombstone(&data_store.read(&db_txn), &validator_address2),
+        None
+    );
+}
+
+#[test]
+fn redelegate_from_tombstone_with_reactivation_works() {
```
nibhar, line 665: Likely very similar to above with a single `false` replaced with a `true`.

primitives/src/policy.rs
```diff
@@ -33,7 +33,9 @@ pub struct Policy {
     /// Genesis block number
     #[cfg_attr(feature = "ts-types", wasm_bindgen(skip))]
     pub genesis_block_number: u32,
-    /// Maximum supported protocol version
+    /// Maximum supported protocol version.
+    // This should only be increased when the hard fork code is released.
```
nibhar:
```suggestion
    // This should only be increased when a new hard fork code is released.
```

primitives/transaction/tests/staking_contract_verify.rs
```diff
@@ -580,7 +679,22 @@ fn stake() {
...
+    // Deposit as zero (pre upgrade restriction).
+    tx.value = Coin::ZERO;
+    assert_eq!(
+        AccountType::verify_incoming_transaction(&tx, 0),
+        Err(TransactionError::ZeroValue)
+    );
+    assert_eq!(
+        Err(TransactionError::ZeroValue),
+        AccountType::verify_incoming_transaction(&tx, Policy::max_supported_version()),
+    );
```
nibhar: Aesthetically a weird inversion of the arguments. Just a nitpick though

test-utils/src/block_production.rs
```diff
@@ -62,29 +62,45 @@ impl TemporaryBlockProducer {
     }
 
     pub fn new_merged() -> Self {
+        Self::new_merged_with_protocol_version(unit_genesis_block().version())
```
nibhar, line 66: Probably worth putting in a comment what version this will create.

wallet/tests/multisig.rs
```diff
@@ -75,7 +75,10 @@ pub fn it_can_create_valid_transactions() {
         )
         .unwrap();
 
-    assert!(tx.verify(NetworkId::UnitAlbatross).is_ok())
+    assert!(tx.verify(NetworkId::UnitAlbatross, 0).is_ok());
+    assert!(tx
+        .verify(NetworkId::UnitAlbatross, Policy::max_supported_version())
+        .is_ok())
```
nibhar: Asserts like this are added all throughout this PR. If the intention is to test this with every version these should be a loop. Initial version also might be 1 and not 0. I don't think we started at 0.

For the loop it would probably be worth to create some helpers that do these kinds of asserts, as after version 5 this gets very tedious. Not for this PR but for the future maybe.

web-client/src/common/transaction.rs
```diff
@@ -1345,7 +1341,7 @@ mod tests {
     fn penalize_inherent_does_not_panic_on_conversion() {
         let _ = Policy::get_or_init(nimiq_primitives::policy::TEST_POLICY);
         let hist_tx = make_penalize_historic_tx();
-        let current_block = 200;
+        let current_block = Policy::genesis_block_number() + 1;
```
nibhar, line 1344: Pretty sure this is not the same. What made that change necessary?

viquezclaudio (approved):

mempool/src/mempool.rs
```diff
@@ -394,6 +403,38 @@ impl Mempool {
...
+        let mut state = self.state.write();
```
nibhar, line 408: As @viquezclaudio commented above this should probably take at least `blockchain.read()` before the state lock. It could be argued that even an `.upgradable_read()` or `.write()` would be alright, as this should likely sequentialize the revalidation and any subsequent push. A `.read()` also does this though until the `.upgradeable_read()` is upgraded, which might be enough.

mempool/src/mempool.rs
```diff
@@ -394,6 +403,38 @@ impl Mempool {
...
+        let protocol_version = self.protocol_version();
+        let MempoolState {
+            regular_transactions,
+            control_transactions,
```
nibhar, line 412: Control transaction should get the same treatment as regular transactions.

mempool/src/mempool.rs
```diff
@@ -394,6 +403,38 @@ impl Mempool {
...
+            .transactions
+            .values_mut()
+            .chain(control_transactions.transactions.values_mut())
+            .filter_map(|tx| {
+                tx.reset_verification_state();
+                tx.verify_mut(tx.network_id, protocol_version)
+                    .err()
+                    .map(|_| tx.hash())
+            })
+            .collect();
```
nibhar, line 427: There is probably a way of collecting those that remain valid and setting them as the new set in `self`. Saves an iteration.

mempool/src/mempool.rs
```diff
@@ -629,18 +670,35 @@ impl Mempool {
...
+
     /// Checks whether a transaction has been filtered
     pub fn is_filtered(&self, hash: &Blake2bHash) -> bool {
         self.filter.read().blacklisted(hash)
     }
 
+    /// Adds a transaction hash to the mempool blacklist.
+    pub fn blacklist_transaction_hash(&self, hash: Blake2bHash) {
+        self.filter.write().blacklist(hash);
+    }
```
nibhar: I am not super sure what this is supposed to do. It would make some sense to me to add the no longer valid transactions during revalidation here, but that is not the case. If this is only for debugging/testing, this should probably not be compiled into the release binary.

nibhar (changes requested):
First and foremost I feel like this PR should have been split into multiple PR.

I would have preferred not to generalize putting the version everywhere without using it. When it is needed it can be added and if multiple PR add it in the same place they can resolve this during merging. This is just a lot of changes that do not do anything on their own and the relevance is rather unclear in this PR alone.

I think the initial version should be 1 not 0.

Quite a bit more code duplication than i think is necessary, but that can stay as is. We should down the road create some helpers though to cut down on that a little.

Other than that also a couple of suggestions and improvements that are probably not required but good to add nonetheless. Lots of them are mentioned multiple times as I encountered them, thus the comment count.

primitives/account/tests/staking_contract/mod.rs
ii-cruz: because there are no upgrade code paths on the staking contract yet.

primitives/transaction/src/account/staking_contract/structs.rs
ii-cruz: it will be needed for the staking contract stuff later.

primitives/account/tests/staking_contract/mod.rs
ii-cruz: it will be used for the create validator later on

primitives/account/tests/staking_contract/staker_add_stake_policy.rs
ii-cruz: it will be useful afterwards

consensus/src/consensus/remote_event_dispatcher.rs
ii-cruz: It was not a bug; it was redundant work. We were repeating the same operation for the same block_hash, most places where this was done, the operations were idempotent, so there would not be a bug, just slower performance. However, i think this is a bad practice since one could change the operations set and lose the idempotency and create a bug.

blockchain/src/blockchain/wrappers.rs
ii-cruz: This is always used to reserve the balance on the current head state. When we go down the call stack we can use this version to reserve the appropriate amount to the sender if it needs a version gate later on.

lib/src/client.rs
ii-cruz: this is consistant with the other getters of the client.

light-blockchain/tests/push.rs
ii-cruz: it was copy pasta

web-client/src/common/transaction.rs
ii-cruz: these tests are flaky. if the policy is initialized for the tests as sometihing else they fail. this fizes that

primitives/account/tests/staking_contract/staker_tombstone_interaction.rs
ii-cruz: shouldn't be there, removed it

primitives/account/tests/staking_contract/staker_tombstone_interaction.rs
ii-cruz: it happened before that we didn't catch a bug due to asserting against the setup variables. so we decided back then to hard code the number.

primitives/account/tests/staking_contract/staker_tombstone_interaction.rs
ii-cruz: we need to have it for both versions add stake will have two different functions, one for legacy one for current version.

primitives/account/tests/staking_contract/staker_tombstone_interaction.rs
ii-cruz: different code paths though and then different effects on the staking contract.

primitives/account/tests/staking_contract/staker_add_stake_policy.rs
ii-cruz: it will be different later when we change the add stake, so it's better to have it like this for now.

primitives/account/tests/staking_contract/staker.rs
ii-cruz: probably it would have been better to have this on the other PR for the add stake instead however i would like to have the change somewhere and i would prefer to not have to change this PR substancially again.

primitives/account/src/interaction_traits.rs
ii-cruz: The version is very much needed together with the block number; it's like the context of the actions being performed on the blockchain. Everywhere we need blockstate, we need to know the version too, and I find it cleaner to added to this struct than add a param everywhere.

mempool/src/mempool.rs
ii-cruz: i don;t think so, this way we don't have to recompute all balances since we remove the txs individually and update the relevant info. if we use some retain logic we then need to recompute balances separately.

mempool/src/mempool.rs
ii-cruz: we discussed it and we concluded that only a read on the blockchain is needed

primitives/account/tests/staking_contract/staker_add_stake_policy.rs
ii-cruz: i just checked and indeed there was a loss of the check for the invariant 1 - active + inactive must be >= minimum stake. It was an upside from moving things from #3196.  All other tests are the same and working well; they will be further adapted in the other pr.

primitives/account/tests/staking_contract/staker_add_stake_policy.rs
ii-cruz: comes from the other PR because there will be a different balance to credit the funds depending on the current staker active balances. for this pr this doesn't make a lot of sense with this name. so i changed it to `add_stake_does_not_touch_inactive_or_retired_balances`. this test makes it so that there is no current active stake and checks if the operation chooses the right balance to attribute the stake to.

primitives/account/tests/staking_contract/staker.rs
nibhar: This change should not be anywhere. It is just bloat, it does not add anything worthwhile, while removing commit history, change history and all that from these code lines. In my opinion this is a bad change and should not have been done. Not here, not in the other PR.

primitives/account/tests/staking_contract/staker_tombstone_interaction.rs
nibhar: As I understand it this magic constant was added in this PR (but is hard to see for the copy paste), so I do not quite understand the response.

In my opinion this should really just be 
```suggestion
    assert_eq!(staker.inactive_from, Some(Policy::genesis_block_number() + Policy::blocks_per_epoch()));
```
In TEST_POLICY this will be 200 and 128 respectively. If these constants in policy would ever change, this test will fail.

primitives/account/tests/staking_contract/staker_add_stake_policy.rs
```diff
@@ -0,0 +1,485 @@
...
+
+    assert_eq!(
+        staker.active_balance,
+        Coin::from_u64_unchecked(Policy::MINIMUM_STAKE)
+    );
+    assert_eq!(
+        staker.inactive_balance,
+        Coin::from_u64_unchecked(Policy::MINIMUM_STAKE)
+    );
+    assert_eq!(staker.inactive_from, Some(328));
```
nibhar: Magic Constant
```suggestion
    assert_eq!(staker.inactive_from, Some(Policy::genesis_block_number() + Policy::blocks_per_epoch()));
```

nibhar (approved):

PR #3804, opened 2026-06-10
Web client: validate `seedNodes` input instead of panicking

Eligioo:
`ClientConfiguration::seedNodes()` in the WASM web client unwrapped the result of deserializing each array element as a `String`.

This PR also fixes a similar issue in the `BlsCache`.

- `seedNodes` now returns `Result<(), JsError>` and propagates the deserialization error.
- `BlsCache::init` maps the deserialization error to `idb::Error::UnexpectedJsType("BlsKeyEntry", e.into())` so it flows through the existing `Result<(), Error>` return.

## Pull request checklist


jsdanielh (approved):
LGTM

PR #3808, opened 2026-06-11
validator: fix version-upgrade trigger

jsdanielh:
Only bump the block version on election blocks: the network rejects a version increase on checkpoint blocks, so proposing one there would halt the chain at the first checkpoint after the threshold is crossed.

Also gauge support by active stake alone, dropping the elected-slot check (a stale snapshot) so the trigger matches the set that `deactivate_unsupporting_validators` acts on.

## Pull request checklist

### DIFF
--- a/blockchain/src/blockchain/verify.rs
+++ b/blockchain/src/blockchain/verify.rs
@@ -276,6 +276,44 @@ impl Blockchain {
             return Ok(());
         }
 
+        // If this block triggers a version upgrade, it must be an election block with enough
+        // support. The support check mirrors the proposer-side check in
+        // `TendermintProtocol::create_proposal` and gates the same `VersionUpgrade` inherent applied
+        // during `commit_accounts`: without it a malicious proposer could bump the version with
+        // insufficient support and thereby deactivate most of the active validators.
+        let new_version = macro_block.header.version;
+        if Some(new_version) == self.state.current_version().checked_add(1) {
+            // The network only accepts a version increase on election blocks: proposing one on a
+            // checkpoint block would halt the chain. Reject it here too as defense in depth, even
+            // though `Block::verify_immediate_successor` already rejects a version bump on a
+            // non-election block.
+            if !macro_block.is_election() {
+                warn!(
+                    %macro_block,
+                    new_version,
+                    reason = "Version upgrade on a non-election block",
+                    "Rejecting block"
+                );
+                return Err(PushError::InvalidBlock(BlockError::InvalidVersionUpgrade));
+            }
+
+            let validators = self
+                .current_validators()
+                .expect("Current validators must be present");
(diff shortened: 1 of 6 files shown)


viquezclaudio (commented):
LGTM

paberr (commented):
I don't get why we removed the supporting slots check?
Can you explain that?

And I would also add a check on the validator side to only accept if the conditions are met to make the check symmetric (it's a nice-to-have).

jsdanielh:
> I don't get why we removed the supporting slots check? Can you explain that?

It was a misunderstanding. Added it back

> And I would also add a check on the validator side to only accept if the conditions are met to make the check symmetric (it's a nice-to-have).

Added too

blockchain/src/blockchain/verify.rs
```diff
@@ -276,6 +276,32 @@ impl Blockchain {
...
         }
 
+        // If this election block triggers a version upgrade, verify that the upgrade has enough
+        // support. This mirrors the proposer-side check in `TendermintProtocol::create_proposal`
+        // and gates the same `VersionUpgrade` inherent applied during `commit_accounts`: without it
+        // a malicious proposer could bump the version with insufficient support and thereby
+        // deactivate most of the active validators.
+        let new_version = macro_block.header.version;
+        if macro_block.is_election()
+            && Some(new_version) == self.state.current_version().checked_add(1)
```
paberr: This if looks a bit like it could go wrong in the future!
If someone produces a non-election block with version upgrade, the if is not executed.
We (hopefully) have that check elsewhere already that only election blocks can increase the version, but it doesn't hurt to turn this around here too.

So, I would do:
```
if version upgrade {
  if !is_election {
     error
  }
  ...
```

blockchain/src/blockchain/verify.rs
```diff
@@ -276,6 +276,32 @@ impl Blockchain {
...
+        {
+            let validators = self
+                .current_validators()
+                .expect("Current validators must be present");
+            let data_store = self.get_staking_contract_store();
+            let data_store_read = data_store.read(txn);
+            if !staking_contract.supports_version_upgrade(&data_store_read, validators, new_version)
```
paberr, line 305: ~~do we only want to check this or also the slots constraint?~~
Edit: I noticed that this already captures both.

blockchain/src/blockchain/verify.rs
paberr: or put a comment with the check (I think in verify immediate successor), but changing the if is safer (defense in depth)

jsdanielh: Done, changed as you suggested

viquezclaudio (approved):

paberr (approved):

PR #3821, opened 2026-06-15
Add checkpoints to pico sync fallback

jsdanielh:
Adds a hash + block-number checkpoint to the network config. Pico clients can use it as a fallback during sync; other clients only log a mismatch (to verify the checkpoint) rather than failing.

Checkpoint verification happens during the blockchain push, checkpoint fetches are deduplicated, and light-client macro block requests go through a `SyncQueue` to avoid request floods.

The mainnet checkpoint is set to election block #52401600 (2026-06-01).

Stacked on top of #remove-zkp.

### DIFF
--- a/consensus/src/sync/light/sync_requests.rs
+++ b/consensus/src/sync/light/sync_requests.rs
@@ -1,6 +1,5 @@
 use std::sync::Arc;
 
-use futures::FutureExt;
 use nimiq_block::Block;
 use nimiq_blockchain_interface::AbstractBlockchain;
 use nimiq_blockchain_proxy::BlockchainProxy;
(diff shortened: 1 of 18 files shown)


consensus/src/sync/light/sync_requests.rs
```diff
@@ -220,53 +193,48 @@ impl<TNetwork: Network> LightMacroSync<TNetwork> {
...
         }
 
-        let mut peer_requests = PeerMacroRequests::new();
+        // The highest block this peer reports: once our head reaches it, the peer is fully synced
+        // from our side. Drives the tip-follow re-request in `poll_macro_block_queue`. A checkpoint
+        // (a later, non-election macro block) is always higher than the last election block.
+        //
+        // A peer reporting a nonsensically large epoch number yields a `u32::MAX` target it has no
+        // hope of reaching, so it just sits in `waiting_macro_peers` until it disconnects. Harmless:
+        // it never gets emitted, but it also wastes nothing (it is requested no blocks).
```
viquezclaudio, line 202: "it is requested no blocks"
what does that mean?

PR #3828, opened 2026-06-17
Allow validators to signal protocol upgrades with their warm key

paberr:
## Summary

Chain version (hard-fork) upgrades are coordinated through a validator's `signal_data` field: validators encode the protocol version they support in the first two bytes, and at each election block the proposer bumps the block version once ≥80% of stake **and** slots signal support (`validator/src/tendermint.rs`, `Policy::supports_upgrade`, `StakingContract::get_supporting_stake`/`get_supporting_slots`).

Until now, `signal_data` could only be set via the `UpdateValidator` transaction, which is signed by the validator **cold (master) key** — the key operators are supposed to keep offline. Requiring the cold key for a routine, recurring operational action is poor operational hygiene.

This PR adds a dedicated **`SetSignalData`** staking transaction signed by the validator's **warm (signing) key** — the key already online in the validator process (it already authorizes `DeactivateValidator`/`ReactivateValidator`). Operators can now signal upgrades without touching the cold key. A `version: u16` convenience is provided in the transaction builder, RPC and CLI for the common "signal support for version N" case.

> **Stacked PR.** This is based on `iicruz/upgrade-mechanim-tx-verification`, which threads `protocol_version` into the transaction verify path. That branch must merge first; retarget this to `albatross` afterwards. The only commit belonging to this PR is the warm-key-signalling commit.

### Why a new dedicated transaction (rather than extending `UpdateValidator`)

`UpdateValidator` carries a **single** `proof` that authorizes the whole transaction, and it deliberately has **no** `validator_address` field — the validator is *derived* from the signer (`proof.compute_signer()` **is** the validator address, because it is signed by the cold key whose address is the validator address).

Letting the warm key authorize `signal_data` inside `UpdateValidator` would break that core invariant: the warm key's address is **not** the validator address, so the transaction would need an extra `validator_address` field plus conditional signer logic ("if only `signal_data` changed → warm key, otherwise cold key"). That is a breaking change to `UpdateValidator` itself and an ambiguous, footgun-prone signer model (one proof, two possible signers depending on which fields are populated).

A dedicated transaction is clean instead: it mirrors the existing warm-key `Deactivate`/`Reactivate` pattern exactly — it carries `validator_address` and a `proof`, and commit checks `proof.compute_signer() == Address::from(&validator.signing_key)`. It is self-contained (own verify/commit/revert), leaves `UpdateValidator` untouched, and is appended at the end of `IncomingStakingTransactionData` so existing serde discriminants are preserved. The cold-key `UpdateValidator` path keeps working for backward compatibility.

### Why the full `signal_data` payload (rather than a `version: u16`)

The transaction carries the whole field, `new_signal_data: Option<Blake2bHash>`, instead of just a `u16` version. This is for **forward-compatibility**:

- Today only the first two bytes of `signal_data` are interpreted (as a version), but the field is documented as general-purpose validator coordination. If we later attach more meaning to the remaining bytes (e.g. a bitmask of supported versions for non-sequential/multiple-proposal upgrades, or other coordination data), the **transaction format never has to change** — only the counting/proposer logic would. A `u16`-only payload would force a second transaction/format migration the moment we wanted to use more of the field.
- The sequential-only nature of upgrades today lives entirely in the proposer/counting logic (it only ever considers `current_version + 1` and reads `signal_data` as a single `u16`), **not** in the signalling payload. So the full-field payload does not change current behavior; it just avoids boxing us in.
- A `version: u16` convenience wrapper (builder/RPC/CLI) keeps usability identical for the common case — callers that just want to signal a version pass a number and never see the hash.

Trade-off accepted: the online warm key gains write access to the entire `signal_data` field (today only read as a version). This is acceptable since it is the validator's operational key, and it is the same field the cold key could already write.

### Version gating (consensus safety)

`SetSignalData` is a new consensus rule, so it is **gated by protocol version**: it is rejected in `IncomingStakingTransactionData::verify` until the chain reaches `Policy::WARM_KEY_SIGNALING_VERSION` (currently `2`). The gate uses the `protocol_version` plumbed into the verify path by the base branch.

This is consensus-safe because the authoritative caller (block verification) passes the block's own `version()`, so every node validating a block applies the same gate. A pre-activation block carrying the transaction can never be finalized, hence can never enter history-synced state — so the commit path does not need to re-check the version (consistent with how the other staking checks live in `verify`, with commit assuming `verify` passed).

**Activation bootstrap:** the *first* hard fork to `WARM_KEY_SIGNALING_VERSION` must still be signalled with the cold-key `UpdateValidator` transaction (the new transaction is invalid until the chain reaches that version). From then on, the warm key suffices for all subsequent upgrades. Note: `WARM_KEY_SIGNALING_VERSION = 2` assumes this ships as the v2 hard fork on mainnet — adjust if version numbering differs.

## Changes

- **Core:** `Policy::WARM_KEY_SIGNALING_VERSION` + `Policy::signal_data_for_version` helper; `TransactionError::InvalidForVersion`; `SetSignalData` variant with version-gated `verify`, `is_signaling`, `set_signature`; commit/revert + `SetSignalDataReceipt` (signer must equal the validator's signing key); `Log::SetSignalData`; `Transaction::related_addresses`.
- **Tooling:** transaction-builder (`new_set_signal_data` + `new_signal_version`); RPC interface + server (`create/send_set_signal_data_transaction` + `create/send_signal_version_transaction`); rpc-client CLI (`signal-version`); web-client plain-data type.
- **Tests:** verify (version gate / signature / zero-value), account commit + revert + wrong-signer rejection, and the tendermint end-to-end upgrade test (now driving the warm-key `set_signal_data` path through the version bump).

## Verification

- `cargo build --workspace` (incl. web-client) clean; `cargo fmt` (nightly) clean; no new clippy warnings.
- Green suites: `nimiq-transaction` (staking verify), `nimiq-account` (staking_contract, incl. `set_signal_data_works`), `nimiq-transaction-builder`, `nimiq-validator` (`it_triggers_version_upgrades`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sisou (changes requested):
Please add a method to create this tx type to `web-client/src/primitives/transaction_builder.rs`.

viquezclaudio:
Just a general comment:
Once #3800 is merged into albatross, this PR should be opened to **albatross-v2**

primitives/transaction/src/account/staking_contract/structs.rs
```diff
@@ -207,6 +217,19 @@ impl IncomingStakingTransactionData {
...
+                verify_transaction_signature(transaction, proof)?
+            }
+            IncomingStakingTransactionData::SetSignalData { proof, .. } => {
+                // Setting the signal data with the warm key is only allowed once the protocol has
+                // upgraded to the version that introduces this transaction. Before that the
+                // transaction must be rejected, otherwise nodes that have not upgraded yet would
+                // diverge.
+                if protocol_version < upgrades::v2::WARM_KEY_SIGNALING {
+                    warn!("SetSignalData transactions are not allowed before protocol version {}. The offending transaction is the following:\n{:?}", upgrades::v2::WARM_KEY_SIGNALING, transaction);
+                    return Err(TransactionError::InvalidForVersion);
```
jsdanielh, line 249: Use tracing style

rpc-server/src/dispatchers/consensus.rs
```diff
@@ -1071,6 +1071,113 @@ impl ConsensusInterface for ConsensusDispatcher {
...
+        signing_secret_key: String,
+        new_signal_data: Option<String>,
+        fee: Coin,
+        validity_start_height: ValidityStartHeight,
+    ) -> RPCResult<String, (), Self::Error> {
+        let secret_key = PrivateKey::deserialize_from_vec(&hex::decode(signing_secret_key)?)
+            .map_err(|_| Error::InvalidArgument("Signing Key".to_string()))?;
+        let signing_key_pair = KeyPair::from(secret_key);
+
+        // null = clear the signal data, "0x29a4b..." = set the signal data.
```
jsdanielh:
```suggestion
        // `None` = clear the signal data, "0x29a4b..." = set the signal data.
```

rpc-server/src/dispatchers/consensus.rs
```diff
@@ -1071,6 +1071,113 @@ impl ConsensusInterface for ConsensusDispatcher {
...
+        let new_signal_data: Option<Blake2bHash> = match new_signal_data {
```
jsdanielh: Couldn't a map work here?

rpc-client/src/subcommands/validator_subcommands.rs
```diff
@@ -111,6 +111,26 @@ pub enum ValidatorCommand {
         tx_commons: TxCommon,
     },
```
jsdanielh, line 124: And what happens with the `set signal data` implementation?

rpc-interface/src/consensus.rs
```diff
@@ -500,6 +500,60 @@ pub trait ConsensusInterface {
...
+        signing_secret_key: String,
+        new_signal_data: Option<String>,
+        fee: Coin,
+        validity_start_height: ValidityStartHeight,
+    ) -> RPCResult<Blake2bHash, (), Self::Error>;
+
+    /// Returns a serialized `set_signal_data` transaction that signals support for the given
+    /// protocol `version` (or clears the signal if null), signed with the validator's signing
+    /// (warm) key. You need to provide the address of a basic account (the sender wallet) to pay
+    /// the transaction fee.
```
jsdanielh: We should warn that this overrides the `signal_data`

rpc-interface/src/consensus.rs
```diff
@@ -500,6 +500,60 @@ pub trait ConsensusInterface {
...
+        sender_wallet: Address,
+        validator_address: Address,
+        signing_secret_key: String,
+        version: Option<u16>,
+        fee: Coin,
+        validity_start_height: ValidityStartHeight,
+    ) -> RPCResult<String, (), Self::Error>;
+
+    /// Sends a `set_signal_data` transaction that signals support for the given protocol `version`
+    /// to the network.
```
jsdanielh, line 552: We should warn that this overrides the `signal_data`

rpc-interface/src/consensus.rs
```diff
@@ -500,6 +500,60 @@ pub trait ConsensusInterface {
         validity_start_height: ValidityStartHeight,
     ) -> RPCResult<Blake2bHash, (), Self::Error>;
 
+    /// Returns a serialized `set_signal_data` transaction, signed with the validator's signing
+    /// (warm) key. You need to provide the address of a basic account (the sender wallet) to pay
+    /// the transaction fee. The `new_signal_data` field is interpreted as follows:
+    /// null = Clear the signal data field (set it to None).
+    /// "0x29a4b..." = Set the signal data field to Some(0x29a4b...).
+    /// Note: this transaction is only valid from protocol version `upgrades::v2::WARM_KEY_SIGNALING` onwards.
```
jsdanielh: We should warn that this overrides the `signal_version`

rpc-interface/src/consensus.rs
```diff
@@ -500,6 +500,60 @@ pub trait ConsensusInterface {
...
+        sender_wallet: Address,
+        validator_address: Address,
+        signing_secret_key: String,
+        new_signal_data: Option<String>,
+        fee: Coin,
+        validity_start_height: ValidityStartHeight,
+    ) -> RPCResult<String, (), Self::Error>;
+
+    /// Sends a `set_signal_data` transaction to the network. See
+    /// `create_set_signal_data_transaction` for the meaning of `new_signal_data`.
```
jsdanielh, line 523: We should warn that this overrides the `signal_version`

transaction-builder/src/lib.rs
```diff
@@ -1444,6 +1444,87 @@ impl TransactionBuilder {
...
+                builder.sign_with_key_pair(key_pair);
+                builder.generate().unwrap()
+            }
+            _ => unreachable!(),
+        }
+    }
+
+    /// Convenience wrapper around [`new_set_signal_data`](Self::new_set_signal_data) that signals
+    /// support for the given protocol `version` (or clears the signal if `None`), signed with the
+    /// validator's signing (warm) key.
```
jsdanielh: There should be a waning about that this is only valid in v2 as it happens in the transaction above

jsdanielh (commented):
TBH I'm not a fan of having:
- Set signal data
- Signal version

Signal version actually overrides the whole signal data even when version is only the first 2 bytes of signal data

web-client/src/primitives/transaction_builder.rs
```diff
@@ -410,6 +410,41 @@ impl TransactionBuilder {
...
+    #[wasm_bindgen(js_name = newSetSignalData)]
+    pub fn new_set_signal_data(
+        sender: &Address,
+        validator: &Address,
+        signal_data: Option<String>,
+        fee: Option<u64>,
+        validity_start_height: u32,
+        network_id: u8,
+    ) -> Result<Transaction, JsError> {
+        let native_signal_data = signal_data.map(|r| Blake2bHash::from_str(&r).unwrap());
```
jsdanielh: We shouldn't `unwrap` here but return an error

jsdanielh (changes requested):

paberr:
Addressed the review in c939a8828.

Following the discussion with @jsdanielh: rather than two parallel mechanisms where signaling a version clobbered the whole 32-byte field, the `SetSignalData` transaction now carries a `SignalDataUpdate` enum with two modes:

- `Full(Option<Blake2bHash>)` — replaces the entire `signal_data` field (or clears it with `None`).
- `Version(Option<u16>)` — a **non-destructive partial update**: it writes only the first two (protocol-version) bytes and preserves bytes `2..32`. An all-zero result canonicalizes to `None`.

So version signaling no longer overwrites unrelated signal data, and there's a single transaction type. A full bitmask was considered but dropped in favor of this (smaller tx). The merge lives in `Policy::signal_data_with_version`.

Both modes are exposed end-to-end: `signal_version` (Version) and `set_signal_data` (Full) in the builder/RPC/CLI/web-client. @sisou the web-client now has both `newSetSignalData` and the new `newSignalVersion`.

primitives/transaction/src/account/staking_contract/structs.rs
paberr: Done — switched to structured tracing fields:

```rust
warn!(
    min_protocol_version = upgrades::v2::WARM_KEY_SIGNALING,
    protocol_version,
    ?transaction,
    "SetSignalData transactions are not allowed before the warm-key signaling upgrade"
);
```

rpc-server/src/dispatchers/consensus.rs
paberr: Applied, thanks.

rpc-server/src/dispatchers/consensus.rs
paberr: Yep — replaced the `match` with `.map(...).transpose()?`:

```rust
let new_signal_data: Option<Blake2bHash> = new_signal_data
    .map(|string| {
        Blake2bHash::deserialize_from_vec(&hex::decode(string)?)
            .map_err(|_| Error::InvalidArgument("Signal Data".to_string()))
    })
    .transpose()?;
```

rpc-client/src/subcommands/validator_subcommands.rs
paberr: Good catch — there was no CLI for the full/raw path, only `signal-version`. Added a `set-signal-data` subcommand (Full mode) that takes the new signal data as hex (or clears it when omitted) and calls `create/send_set_signal_data_transaction`.

rpc-interface/src/consensus.rs
paberr: Resolved by the design change in c939a8828: `signal_version` now emits a `Version` partial update that only touches the two protocol-version bytes and preserves the rest of `signal_data` — so it no longer overrides the whole field. Updated the doc to state that explicitly (and to point at `set_signal_data` for a full replace).

rpc-interface/src/consensus.rs
paberr: Resolved by the design change in c939a8828: `signal_version` now emits a `Version` partial update that only touches the two protocol-version bytes and preserves the rest of `signal_data` — so it no longer overrides the whole field. Updated the doc to state that explicitly (and to point at `set_signal_data` for a full replace).

rpc-interface/src/consensus.rs
paberr: Updated the doc to note that `set_signal_data` (Full mode) replaces the *entire* signal data field, overwriting any signaled protocol version — and to point at `signal_version`/`create_signal_version_transaction` for updating only the version while preserving the rest.

rpc-interface/src/consensus.rs
paberr: Updated the doc to note that `set_signal_data` (Full mode) replaces the *entire* signal data field, overwriting any signaled protocol version — and to point at `signal_version`/`create_signal_version_transaction` for updating only the version while preserving the rest.

transaction-builder/src/lib.rs
paberr: Added the `# Note` block about being valid only from `upgrades::v2::WARM_KEY_SIGNALING` onwards to `new_signal_version` (matching `new_set_signal_data`), plus a note that it only updates the version bytes and preserves the rest.

web-client/src/primitives/transaction_builder.rs
paberr: Fixed — it now propagates the parse error instead of unwrapping:

```rust
let native_signal_data = signal_data
    .map(|r| Blake2bHash::from_str(&r))
    .transpose()
    .map_err(|e| JsError::new(&e.to_string()))?;
```

I also added a `newSignalVersion` method here (the Version-mode counterpart of `newSetSignalData`).

paberr:
Once @ii-cruz PR is merged, I'll also adapt the hardfork constants to the new conventions

rpc-interface/src/consensus.rs
```diff
@@ -500,6 +500,66 @@ pub trait ConsensusInterface {
...
+        &self,
+        sender_wallet: Address,
+        validator_address: Address,
+        signing_secret_key: String,
+        new_signal_data: Option<String>,
+        fee: Coin,
+        validity_start_height: ValidityStartHeight,
+    ) -> RPCResult<Blake2bHash, (), Self::Error>;
+
+    /// Returns a serialized `set_signal_data` transaction that signals support for the given
```
jsdanielh:
```suggestion
    /// Returns a serialized `set_signal_version` transaction that signals support for the given
```

rpc-interface/src/consensus.rs
```diff
@@ -500,6 +500,66 @@ pub trait ConsensusInterface {
...
+        &self,
+        sender_wallet: Address,
+        validator_address: Address,
+        signing_secret_key: String,
+        version: Option<u16>,
+        fee: Coin,
+        validity_start_height: ValidityStartHeight,
+    ) -> RPCResult<String, (), Self::Error>;
+
+    /// Sends a `set_signal_data` transaction that signals support for the given protocol `version`
```
jsdanielh:
```suggestion
    /// Sends a `set_signal_version` transaction that signals support for the given protocol `version`
```

transaction-builder/src/recipient/staking_contract.rs
```diff
@@ -117,6 +119,42 @@ impl StakingRecipientBuilder {
...
+            validator_address,
+            update: SignalDataUpdate::Full(new_signal_data),
+            proof: Default::default(),
+        });
+        self
+    }
+
+    /// Convenience wrapper around [`set_signal_data`](Self::set_signal_data) that signals support
+    /// for the given protocol `version` (or clears the signal if `None`). It needs to be signed by
+    /// the validator's signing (warm) key.
```
jsdanielh: I think the comment is stale: it is no longer a wrapper around `Self::set_signal_data`

transaction-builder/src/lib.rs
```diff
@@ -1444,6 +1445,118 @@ impl TransactionBuilder {
...
+            recipient,
+            fee,
+            validity_start_height,
+            network_id,
+        )
+    }
+
+    /// Convenience wrapper around [`new_set_signal_data`](Self::new_set_signal_data) that signals
+    /// support for the given protocol `version` (or clears the signal if `None`), signed with the
+    /// validator's signing (warm) key.
```
jsdanielh: I think this comment is stale: this is no longer a wrapper around `Self::new_set_signal_data`

web-client/proofs/proof_epoch_2.bin
jsdanielh, line 1: What is this file for? Was it added by mistake?

viquezclaudio (commented):
LGTM, I dont have any further comment apart from what JD already mentioned

web-client/proofs/proof_epoch_2.bin
paberr: true

jsdanielh (approved):
LGTM

viquezclaudio (approved):

sisou (approved):
Didn't test it, but the diff looks good.

PR #3829, opened 2026-06-17
feat(primitives): add per-fork upgrade activation registry

jsdanielh:
Adds a `policy::upgrades` module where each `vN` submodule declares the changes shipping in protocol version `N` as named constants. Call sites gate behaviour with a plain comparison:

```rust
if block.version() >= upgrades::v2::V2_CHANGE { ... }
```

Scaffolds the v2 and v3 registries.


viquezclaudio (approved):
Love it

nibhar (approved):

PR #3836, opened 2026-06-19
Add protocol-upgrade signaling dashboard (web client)

paberr:
## What

A simple, **read-only** browser dashboard that shows whether validators are signaling support for the next protocol upgrade: the two on-chain thresholds (**% of active stake** and **% of slots** signaling the target version) plus a filterable validator explorer.

It lives in `web-client/upgrade-dashboard/` as a vanilla HTML + ES-modules page on top of the `@nimiq/core` web client (light client, pico sync) — no node or RPC server required.

## Web client change

Adds `Client.getElectedValidators()` (`web-client/src/client/lib.rs`, `web-client/src/client/account.rs`), returning the current epoch's elected validators with their slot counts (`{ address, numSlots }`). The web client already exposes per-validator `signalData` and `totalStake`, but not slot counts — and slots are one of the two on-chain upgrade thresholds. Implemented via the existing `blockchain.current_validators()` accessor; purely additive.

## How the metrics mirror the chain

- **Target version** = head `version + 1`.
- A validator **supports** the upgrade iff the first two bytes (big-endian) of its `signalData` decode to the target version — the JS equivalent of `Policy::supports_upgrade`.
- **Stake** is summed over the staking contract's active validators; **slots** over the elected validators.
- Both are compared against `Policy::UPGRADE_MIN_SUPPORT` (80%), matching the upgrade trigger in `validator/src/tendermint.rs`.

## Build & run

Fully containerized (nothing installed on the host); see `web-client/upgrade-dashboard/README.md`:

```sh
docker compose -f web-client/upgrade-dashboard/docker-compose.yml run --rm build
docker compose -f web-client/upgrade-dashboard/docker-compose.yml up serve
# open http://localhost:8000/upgrade-dashboard/
```

A native `./scripts/build.sh` + static-serve path is documented too.

## Notes

- Read-only; submitting signals and historical/trend views are out of scope.
- Verified end-to-end against mainnet: pico sync establishes and the validator table + metrics populate (0% there, since no mainnet validator signals the next version — as expected).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/web-client/upgrade-dashboard/index.html
+++ b/web-client/upgrade-dashboard/index.html
@@ -1,0 +1,140 @@
+<!DOCTYPE html>
+<html lang="en-US">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <title>Nimiq Protocol Upgrade Signaling</title>
+  </head>
+  <body>
+    <script type="module" src="./dashboard.js"></script>
+
+    <header>
+      <h1>Protocol Upgrade Signaling</h1>
+      <div class="status-line">
+        <span>Consensus: <strong id="consensus" class="connecting">connecting</strong></span>
+        <span>Peers: <strong id="peers">0</strong></span>
+        <span>Block: <strong id="block">-</strong></span>
+        <span>Epoch: <strong id="epoch">-</strong></span>
+        <span>Version: <strong id="version">-</strong> &rarr; Target <strong id="target-version">-</strong></span>
+        <span class="muted">Updated: <span id="updated">never</span></span>
+      </div>
+    </header>
+
+    <!-- Overall verdict: upgrade activates only when BOTH metrics reach the threshold. -->
+    <div id="verdict" class="verdict pending">
+      Waiting for consensus&hellip;
+    </div>
+
+    <section class="metrics">
+      <div class="card">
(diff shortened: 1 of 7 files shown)


viquezclaudio:
Looks good!

Screenshot 2026-06-19 at 17 06 30

web-client/upgrade-dashboard/index.html
```diff
@@ -0,0 +1,140 @@
...
+    <script type="module" src="./dashboard.js"></script>
+
+    <header>
+      <h1>Protocol Upgrade Signaling</h1>
+      <div class="status-line">
+        <span>Consensus: <strong id="consensus" class="connecting">connecting</strong></span>
+        <span>Peers: <strong id="peers">0</strong></span>
+        <span>Block: <strong id="block">-</strong></span>
+        <span>Epoch: <strong id="epoch">-</strong></span>
+        <span>Version: <strong id="version">-</strong> &rarr; target <strong id="target-version">-</strong></span>
```
viquezclaudio: Minor cosmetic change, can the 't' from target be upper case? i.e: Target instead of target

paberr: I guess we can do that

jsdanielh (approved):
LGTM

PR #3869, opened 2026-07-13
Fix the network unreachable test panic

ii-cruz:
Some two-peer sync tests used hardcoded libp2p memory addresses. Since the memory transport registry is shared by tests running in the same binary, parallel tests could attempt to listen on the same address. This caused `Swarm::listen_on` to fail and led to subsequent network cancellation or unreachable panics before the sync assertions were reached.

This PR adds an asynchronous `Network::listen_on_address` API that waits for libp2p to report the resolved listen address. Test networks can now listen on `/memory/0`, allowing libp2p to allocate an available address, and return that address to callers for connecting peers.

The full, history, light, and pico sync tests now use these allocated addresses instead of deriving or hardcoding memory addresses. A regression test verifies that concurrent `/memory/0` listeners receive distinct, nonzero addresses.

The PR also fixes a separate deadlock exposed when history-sync tests initialize genesis data in parallel. The temporary genesis block is kept alive until the permanent decompressed BLS keys are cached, preventing its lazy public keys from being dropped while another thread is initializing the same key cache.

## Pull request checklist

### DIFF
--- a/consensus/tests/history_sync.rs
+++ b/consensus/tests/history_sync.rs
@@ -197,8 +197,8 @@ async fn sync_ingredients() {
     // Produce the blocks.
     produce_macro_blocks(&producer, &blockchain1, num_macro_blocks);
 
-    let net1: Arc<Network> =
-        TestNetwork::build_network(2, Default::default(), &mut Some(hub)).await;
+    let (net1, _): (Arc<Network>, u64) =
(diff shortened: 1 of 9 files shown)


jsdanielh (commented):
I'd prefer a random number per test. We should display this generated number though

viquezclaudio (commented):
LGTM

PR #3884, opened 2026-07-17
mempool: avoid stalling block production during upgrade revalidation

jsdanielh:
`revalidate_transactions` re-verified every transaction while holding the blockchain read and mempool write locks. Hold the state as an upgradable read (readers still run) and take the version as a parameter, so no blocking lock is held across the per-transaction checks.

## Pull request checklist

### DIFF
--- a/mempool/src/mempool.rs
+++ b/mempool/src/mempool.rs
@@ -21,7 +21,7 @@ use nimiq_transaction::{
     TransactionTopic,
 };
 use nimiq_utils::spawn;
-use parking_lot::RwLock;
+use parking_lot::{RwLock, RwLockUpgradableReadGuard};
 use tokio_metrics::TaskMonitor;
 
 #[cfg(feature = "metrics")]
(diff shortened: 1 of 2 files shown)


mempool/src/mempool.rs
```diff
@@ -403,30 +403,37 @@ impl Mempool {
...
-            control_transactions,
-            ..
-        } = &mut *state;
-
-        // Find transactions that no longer pass verification under the new protocol version.
-        let invalid: Vec<Blake2bHash> = regular_transactions
+    ///
+    /// `protocol_version` is the version to validate against: the caller passes the version of the
+    /// election block that activated the upgrade, so it is fixed and never re-read from the blockchain.
+    pub fn revalidate_transactions(&self, protocol_version: u16) {
```
nibhar, line 409: So since this version gets passed in as argument and as a a `u16` which is `Copy` there is nothing preventing the actual version number to change during this function call. The rust doc say how this should be used, but there is nothing making sure of that. When the blockchain read lock was held, the version can not possibly change as that would require a blockchain write lock.

mempool/src/mempool.rs
```diff
@@ -403,30 +403,37 @@ impl Mempool {
...
+        // Hold state as an upgradable read for the whole operation so nothing changes between
+        // deciding what is invalid and removing it. Readers (e.g. block production) still run;
+        // only writers wait. The blockchain lock is not held across the per-tx checks
```
nibhar, line 412: If the state is held to prevent writers, than we probably can just take the blockchain read. When are writes to the state taken? I would assume at least whenever a block needs to be produced or whenever a block was pushed, but i have not looked. Both of these will ultimately also acquire blockchains upgradable read, so I don't think this will do much.

nibhar (changes requested):
I don't think this changes much except it makes data inconsistencies easier to mistakenly produce and opens a door for a potential deadlock. 
The blockchain lock was held for a reason, namely that data behind that lock is depended upon and must therefore not be mutable to others.
What is the reasoning behind this change? As the state write is still prevented it does not provide much utility I think.

I would prefer not to merge this.

viquezclaudio (commented):
if it ain't broke, don't fix it..

ii-cruz (commented):
I think that the concern of the slowdown/skip blocks after upgrading the version will not be an issue in the end, since it's rather normal for tx submissions before hard forks. Even if that is not fully enforced, I would expect mempools to have few txs, and thus this wouldn't be much of a problem.
I also would prefer to improve the situation in some other way if it becomes a problem. For example, we could add a flag to the protocol upgrade version if there are changes to the way we process txs, or an enum with the tx types to be revalidated.I think that

PR #3889, opened 2026-07-20
Abort macro block proposal creation on an incomplete accounts state

jsdanielh:
Replace the `.expect` panics in `create_proposal` and `next_macro_block_proposal_with_rng` with graceful failures guarded by `accounts_complete()`, mirroring `verify_macro_block_proposal` on the verifying side. Add regression tests for both paths

## Pull request checklist

### DIFF
--- a/blockchain/src/block_production/mod.rs
+++ b/blockchain/src/block_production/mod.rs
@@ -303,6 +303,13 @@ impl BlockProducer {
         // The rng seed. We need this parameterized in order to have determinism when running unit tests.
         rng: &mut R,
     ) -> Result<MacroBlock, BlockProducerError> {
+        // The complete accounts state is needed to produce a macro block proposal. Abort
+        // gracefully otherwise, mirroring `Blockchain::verify_macro_block_proposal` on the
+        // verifying side
+        if !blockchain.accounts_complete() {
+            return Err(BlockProducerError::AccountsIncomplete);
+        }
+
         // The network ID stays unchanged for the whole blockchain.
         let network = blockchain.head().network();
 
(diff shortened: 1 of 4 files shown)


viquezclaudio (approved):

blockchain/src/block_production/mod.rs
```diff
@@ -303,6 +303,13 @@ impl BlockProducer {
         // The rng seed. We need this parameterized in order to have determinism when running unit tests.
         rng: &mut R,
     ) -> Result<MacroBlock, BlockProducerError> {
+        // The complete accounts state is needed to produce a macro block proposal. Abort
+        // gracefully otherwise, mirroring `Blockchain::verify_macro_block_proposal` on the
+        // verifying side
+        if !blockchain.accounts_complete() {
```
paberr, line 309: With this early `accounts_complete()` guard, the `.ok_or(BlockProducerError::AccountsIncomplete)?` at line 346 and the one in `next_macro_body` are now unreachable in practice.

paberr: Do we need both?

paberr (commented):
LGTM, just one minor comment

blockchain/src/block_production/mod.rs
jsdanielh: This guard is to prevent the panic that we would face calling `exercise_transactions`. At line 346 we check that the staking contract subtree is complete but the trie could still be incomplete somewhere else. This makes the `ok_or` check in line 346 redundant (we can change it for a `expect`) but I prefer to keep it for defensive future proof reasons.

PR #3890, opened 2026-07-20
Sync the mempool's protocol version on `ProtocolUpgrade` events

jsdanielh:
The version was only bumped on live election `Extended` events, so a node crossing an upgrade via history or macro sync kept verifying transactions with the pre-fork version until the next live election. Also resync on mempool activation and at elections as a backstop for dropped events

## Pull request checklist

### DIFF
--- a/mempool/mempool-task/src/lib.rs
+++ b/mempool/mempool-task/src/lib.rs
@@ -90,6 +90,10 @@ impl<N: Network> MempoolTask<N> {
             return;
         }
 
+        // A protocol upgrade event might have been dropped by a lagging event stream,
+        // so resynchronize the version used for transaction verification
+        self.mempool.resync_protocol_version();
+
         let mempool = Arc::clone(&self.mempool);
         let network = Arc::clone(&self.consensus.network);
         #[cfg(not(feature = "metrics"))]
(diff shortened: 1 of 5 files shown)


mempool/mempool-task/src/lib.rs
```diff
@@ -90,6 +90,10 @@ impl<N: Network> MempoolTask<N> {
             return;
         }
 
+        // A protocol upgrade event might have been dropped by a lagging event stream,
```
viquezclaudio, line 93: I dont understand why we need this on mempool init?

jsdanielh: This is to cover the case where the protocol upgrade event would be silently dropped if the upgrade happens during an initial sync:
- `Mempool::new` caches the version at construction: for a fresh node that's genesis (v1).
- A node then history/macro-syncs across the v2 fork. The ProtocolUpgrade event fires during that sync.
- But `blockchain_event_rx` is a `BroadcastStream` (notifier_as_stream) that silently drops events on Lagged. And a multi-epoch history sync is exactly the event flood that causes lag/drops. If the `ProtocolUpgrade` event is dropped, nothing updates the cached version.

ii-cruz (approved):
LGTM

PR #3891, opened 2026-07-21
Collapse the signal-data no-signal ambiguity on the v2 signaling paths

jsdanielh:
`signal_data` on a validator is `Option<Blake2bHash>`, so "no signal" had two encodings: `None` and `Some([0;32])`. The two are equivalent for upgrade signaling — `Policy::supports_upgrade` returns `false` for both at every real version (≥ 2) — but the code was internally inconsistent: `SignalDataUpdate::Version` collapsed an all-zero result to `None` while `Full` stored it verbatim.

Rather than enforce a single on-chain representation (which isn't consensus-relevant, and can't be done globally since the live v1 `CreateValidator`/`UpdateValidator` wire formats can't change), this removes the *type-level* ambiguity on the v2-gated `SetSignalData` paths only:

- `SignalDataUpdate::Version(Option<u16>)` → `Version(u16)`. A version update only touches the version bytes and always yields a present value, so `Version(None)` is now unrepresentable and `Full(None)` is the single way to clear the field.
- `Policy::signal_data_with_version` drops its `Option<u16>` parameter and the all-zero → `None` canonicalization; it always returns the raw bytes.
- The all-zero rejection introduced earlier is removed — `Full(Some([0;32]))` is a valid value of the field, equivalent to `None`.
- `version: u16` is threaded through the transaction builder, RPC (interface/server/CLI) and web-client; clearing now routes through `set_signal_data(None)`.

`SetSignalData` is gated on `upgrades::v2::WARM_KEY_SIGNALING` and is not yet deployed, so changing its serialization is safe. The v1 `CreateValidator`/`UpdateValidator` transactions and the `Validator` state struct are deliberately untouched — they can still hold `Some([0;32])`, which is harmless since consensus treats it identically to `None`.

## Pull request checklist

### DIFF
--- a/primitives/account/tests/staking_contract/validator.rs
+++ b/primitives/account/tests/staking_contract/validator.rs
@@ -843,7 +843,7 @@ fn signal_version_partial_update_works() {
     let tx_version = make_signed_incoming_transaction(
         IncomingStakingTransactionData::SetSignalData {
             validator_address: validator_address.clone(),
-            update: SignalDataUpdate::Version(Some(2)),
+            update: SignalDataUpdate::Version(2),
             proof: SignatureProof::default(),
         },
         0,
@@ -884,6 +884,39 @@ fn signal_version_partial_update_works() {
         .get_validator(&data_store.read(&db_txn), &validator_address)
         .expect("Validator should exist");
     assert_eq!(validator.signal_data, Some(full_signal_data));
+
+    // Withdrawing the signal uses `Version(0)` (there is no `Version(None)`): it zeroes the
+    // version bytes but keeps the field present with the rest preserved, so the validator no
+    // longer supports an upgrade.
+    let tx_withdraw = make_signed_incoming_transaction(
+        IncomingStakingTransactionData::SetSignalData {
+            validator_address: validator_address.clone(),
+            update: SignalDataUpdate::Version(0),
+            proof: SignatureProof::default(),
+        },
+        0,
+        &signing_keypair,
+    );
+    validator_setup
+        .staking_contract
+        .commit_incoming_transaction(
+            &tx_withdraw,
+            &block_state,
+            data_store.write(&mut db_txn),
+            &mut TransactionLog::empty(),
+        )
+        .expect("Failed to commit signal withdrawal transaction");
+
+    let validator = validator_setup
+        .staking_contract
+        .get_validator(&data_store.read(&db_txn), &validator_address)
+        .expect("Validator should exist");
+    let signal_data = validator
+        .signal_data
+        .expect("Signal data should still be present");
+    assert_eq!(&signal_data.as_slice()[..2], &[0u8, 0u8]);
+    assert_eq!(signal_data.as_slice()[10], 0xCD);
+    assert!(!Policy::supports_upgrade(Some(signal_data), 2));
 }
 
 #[test]
--- a/primitives/src/policy.rs
+++ b/primitives/src/policy.rs
@@ -177,18 +177,17 @@ impl Policy {
         Blake2bHash(signal_data)
     }
 
-    /// Applies a protocol-version update to an existing `signal_data` value: the `version` is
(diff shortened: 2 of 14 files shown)


primitives/transaction/src/account/staking_contract/structs.rs
```diff
@@ -249,6 +249,15 @@ impl IncomingStakingTransactionData {
                     return Err(TransactionError::InvalidForVersion);
                 }
 
+                // An all-zero hash is a non-canonical encoding of "no signal". Require `None`
+                // (i.e. `Full(None)`) so the field has a single representation on-chain, matching
+                // the `Version` update path which maps an all-zero result to `None`.
+                if let SignalDataUpdate::Full(Some(signal_data)) = update
```
paberr: Non-blocking, but I don't think this fully gives us the single on-chain representation the description promises. `signal_data` can also be written directly via `CreateValidator` (the `signal_data` field) and `UpdateValidator` (`new_signal_data`), and neither of those `verify` arms rejects an all-zero hash.

paberr: Question is if we should prohibit it there too.
If not, should we care at all?

primitives/src/policy.rs
```diff
@@ -177,18 +177,16 @@ impl Policy {
...
     }
 
-    /// Applies a protocol-version update to an existing `signal_data` value: the `version` is
-    /// encoded into the first two bytes (big-endian) while the remaining bytes are preserved.
-    /// `version == None` clears the version bytes (sets them to zero). Returns `None` if the
-    /// resulting value is all-zero, so that "no signal" has a single canonical representation.
+    /// Applies a protocol-version update to an existing `signal_data` value: `version` is encoded
+    /// into the first two bytes (big-endian) while the remaining bytes are preserved. This is a
+    /// partial update, so it always yields a present value (clearing the whole field is a separate
+    /// operation). `version == 0` zeroes the version bytes. Note that an all-zero result and a
```
paberr: The PR description qualifies this correctly with "at every real version (>= 2)", but the doc comment states it unconditionally. `supports_upgrade(Some([0; 32]), 0)` is `true` while `supports_upgrade(None, 0)` is `false`, so the two aren't equivalent at version 0. Not a concern in practice, but worth putting here in the doc

primitives/account/tests/staking_contract/validator.rs
```diff
@@ -843,7 +843,7 @@ fn signal_version_partial_update_works() {
     let tx_version = make_signed_incoming_transaction(
         IncomingStakingTransactionData::SetSignalData {
             validator_address: validator_address.clone(),
-            update: SignalDataUpdate::Version(Some(2)),
+            update: SignalDataUpdate::Version(2),
```
paberr, line 846: `Version(0)` is now the only way to withdraw a signal with the warm key, since it replaces the old `Version(None)`, but I don't see it covered at the account level; we only test `signal_data_with_version(None, 0)` in the policy unit test. Could we extend this test to commit a `Version(0)` afterwards and assert that `signal_data` stays `Some` with zeroed version bytes and the rest preserved, and that `Policy::supports_upgrade(.., 2)` is then `false`?

web-client/src/common/transaction.rs
```diff
@@ -790,7 +790,7 @@ pub struct PlainSetSignalDataData {
     /// For `full` mode: the new signal data as a hex string, or `null` to clear it. Always `null`
     /// in `version` mode.
     pub new_signal_data: Option<String>,
-    /// For `version` mode: the signaled protocol version, or `null` to clear it. Always `null` in
+    /// For `version` mode: the signaled protocol version (always present). Always `null` in
```
paberr: reads a bit contradictory: "always present" and then "always `null` in `full` mode". maybe just "the signaled protocol version; `null` in `full` mode." or so?

paberr (approved):
Generally LGTM. Only minor suggestions for docs + tests

paberr:
LGTM now

PR #3895, opened 2026-07-22
Remove dead `tree_from_chunks`

jsdanielh:
`HistoryInterface::tree_from_chunks` builds a history tree from range-proof chunks, but it has no production caller — and never had one (it has never been referenced anywhere in `consensus/` in the git history). History sync adopts epochs incrementally via `add_to_history` (`blockchain/src/blockchain/history_sync.rs`), not `tree_from_chunks`. Its prover counterpart `prove_chunk` *is* live (the `RequestHistoryChunk` handler), which is likely why the unused adopter method went unnoticed. It's a blockchain-internal trait method with no web-client/RPC/wasm exposure.

This removes it: the `HistoryInterface` trait method and all four implementations (`HistoryStore` — the real logic, and the `HistoryStoreIndex` / `HistoryStoreProxy` / `HistoryStoreMerger` delegations), plus the now-unused `RangeProof` / `PartialMerkleMountainRange` imports. Pure deletion, no behaviour change; the live history-chunk-serving path (`prove_chunk`) is untouched.

## Pull request checklist

### DIFF
--- a/blockchain/src/history/history_store_index.rs
+++ b/blockchain/src/history/history_store_index.rs
@@ -11,10 +11,7 @@ use nimiq_database::{
 use nimiq_genesis::NetworkId;
 use nimiq_hash::Blake2bHash;
 use nimiq_keys::Address;
-use nimiq_mmr::{
-    error::Error as MMRError,
-    mmr::proof::{RangeProof, SizeProof},
-};
+use nimiq_mmr::{error::Error as MMRError, mmr::proof::SizeProof};
 use nimiq_primitives::policy::Policy;
 use nimiq_transaction::{
     historic_transaction::{HistoricTransaction, HistoricTransactionData, RawTransactionHash},
@@ -451,16 +448,6 @@ impl HistoryInterface for HistoryStoreIndex {
             chunk_index,
             txn_option,
         )
-    }
-
-    fn tree_from_chunks(
-        &self,
-        epoch_number: u32,
-        chunks: Vec<(Vec<HistoricTransaction>, RangeProof<Blake2bHash>)>,
-        txn: &mut MdbxWriteTransaction,
-    ) -> Result<Blake2bHash, MMRError> {
-        self.history_store
-            .tree_from_chunks(epoch_number, chunks, txn)
     }
 
     fn get_last_leaf_block_number(&self, txn_option: Option<&MdbxReadTransaction>) -> Option<u32> {
--- a/blockchain/src/history/history_store.rs
+++ b/blockchain/src/history/history_store.rs
@@ -13,12 +13,7 @@ use nimiq_genesis::NetworkId;
 use nimiq_hash::{Blake2bHash, Hash};
 use nimiq_mmr::{
     error::Error as MMRError,
-    mmr::{
-        partial::PartialMerkleMountainRange,
-        position::leaf_number_to_index,
-        proof::{RangeProof, SizeProof},
-        MerkleMountainRange,
-    },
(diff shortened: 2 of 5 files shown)


blockchain/src/history/history_store_index.rs
```diff
@@ -459,8 +504,26 @@ impl HistoryInterface for HistoryStoreIndex {
         chunks: Vec<(Vec<HistoricTransaction>, RangeProof<Blake2bHash>)>,
         txn: &mut MdbxWriteTransaction,
     ) -> Result<Blake2bHash, MMRError> {
-        self.history_store
-            .tree_from_chunks(epoch_number, chunks, txn)
+        // The delegate below only populates the history tree and `hist_tx_table`, so unless we
```
paberr: Am I right that `tree_from_chunks` has no caller outside this module? As far as I can tell the only references in the tree are the trait declaration in `interface.rs`, the three impls, and the delegation between them, plus the new test. The actual history sync goes `push_history_sync` -> `extend_history_sync` -> `add_to_history_for_epoch`, which was already indexed.  When do we actually use this method?

blockchain/src/history/history_store_index.rs
```diff
@@ -459,8 +504,26 @@ impl HistoryInterface for HistoryStoreIndex {
...
+        // index here as well the adopted epoch would be unreachable by transaction hash and by
+        // address. Build the index in the same leaf order the wrapped store uses (chunks
+        // concatenated in order), then persist it only once the tree has been built.
+        let (hashes, addresses) = self.build_tx_index(
```
paberr: Re-deriving the leaf indices here assumes that the epoch is empty and that `chunks` is the complete epoch starting at leaf 0. If this implicit assumption is not true, it messes everything up.

blockchain/src/history/history_store_index.rs
```diff
@@ -1509,6 +1558,59 @@ mod tests {
...
+        let mut target_txn = target_env.write_transaction();
+        target
+            .tree_from_chunks(epoch, vec![(chunk.history, chunk.proof)], &mut target_txn)
+            .expect("Should adopt the epoch from chunks");
+
+        // The adopted transactions must be discoverable by hash and by address, not merely stored
+        // (both lookups read the index tables that the chunk path must now populate).
+        assert!(target
+            .get_hist_tx_by_hash(&hist_txs[0].tx_hash(), Some(&target_txn))
+            .is_some());
```
paberr: Two small things on this test:

1. `is_some()` also passes if the leaf index is wrong, since we'd just resolve to a different transaction. Given that the leaf-index alignment is exactly what this PR relies on, could we assert the resolved transaction itself, like `get_hist_tx_by_hash_works` does above?
2. We adopt a single chunk here, so the chunk-concatenation order that `build_tx_index` assumes is never actually exercised. Maybe prove two chunks (e.g. `chunk_size = 2`) and pass both?

blockchain/src/history/history_store_index.rs
paberr: Have looked through the codebase again, and I can't find a single instance where we use `tree_from_chunks`. Are we sure this isn't dead code that can be removed?!

paberr (commented):
Not sure if we need this method at all.

jsdanielh:
> Not sure if we need this method at all.

You are right. I changed it to completely remove it

viquezclaudio (approved):

paberr:
LGTM now, can be merged

PR #3898, opened 2026-07-23
tendermint: set validValue for late proposals admitted in the precommit window

nibhar:
When a proposal's 2f+1 prevote polka arrived before the proposal, state.valid was only set if the proposal became known while still in Prevote. If the node had already timed out into Precommit, valid stayed None, so a node later proposing would build a fresh proposal instead of re-proposing the polka'd value, delaying finality.

Fix: process_proposal sets valid when the admitted proposal already has a recorded 2f+1 prevote polka for its round. Idempotent; locked and votes are untouched, so safety holds.

Adds regression tests for the prevote- and precommit-window orderings.


viquezclaudio (approved):
LGTM

PR #3905, opened 2026-07-28
Don't underflow `helper_index` on crafted partial merkle proofs

viquezclaudio:
A deserialized `PartialMerkleProof` can carry an arbitrary `total_len`, so its implied tree shape may request more helper nodes than the previous result provides. `PartialMerkleProof::compute` decremented `helper_index` unchecked, panicking in debug builds (wrapping in release). Use `checked_sub` and return `InvalidPreviousProof` instead.

...
## Pull request checklist


jsdanielh (approved):
LGTM

PR #3911, opened 2026-07-30
Don't establish consensus on a head request with no responses

viquezclaudio:
When all `RequestHead` queries fail, `HeadRequests` resolves with zero known and zero unknown blocks, and the "2/3 of heads known" check (`num_known >= 2 * num_unknown`) is vacuously satisfied by `0 >= 0`, establishing consensus on zero evidence. A set of Sybil peers that pass macro sync but drop all head responses could hold a node in a false established state.

Require at least one head response before establishing. The only legitimate zero-response case is a lone seed node (`min_peers == 0`) with no peers to ask, which is identified by `num_agents == 0`. Extract the decision into `head_request_established` and unit test it.

...
## Pull request checklist

### DIFF
--- a/consensus/src/consensus/mod.rs
+++ b/consensus/src/consensus/mod.rs
@@ -178,6 +178,23 @@ pub struct Consensus<N: Network> {
     waker: Option<Waker>,
 }
 
+/// Decides whether a finished head request provides enough evidence to establish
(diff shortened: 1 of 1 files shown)


jsdanielh (approved):
LGTM

PR #3912, opened 2026-07-31
fix(tendermint): align locked-round check and de-duplicate it

nibhar:
`is_allowed_to_vote_on` used a strict `locked_round < vr`, where Algorithm of the Tendermint paper specifies `lockedRound <= vr || lockedValue = v`.

The two differ only when `locked_round == vr` and the locked hash differs from the proposal. That state requires two distinct 2f+1 prevote polkas in the same round, i.e. f+1 equivocating validators — beyond the fault threshold, and rejected outright within a single aggregate (`TendermintVerifier` returns `Forged` on overlapping contributor buckets). The deviation was also conservative: it could only prevote nil where the paper prevotes for the value, so it was never a safety concern. No behaviour change under `<= f` faults.

`propose()` already implemented the same predicate correctly with `<=`, hand-copied rather than shared. It now calls `is_allowed_to_vote_on`, so the rule exists once.

No test: the only case the two operators disagree on is unreachable without `> f` Byzantine validators, so a test would have to fabricate an aggregate that real verification rejects. Existing coverage of the reachable paths (`it_locks_and_unlocks`, `it_locks_and_rebroadcasts`, `it_does_not_unlock`) is unchanged.

### DIFF
--- a/tendermint/src/states/propose.rs
+++ b/tendermint/src/states/propose.rs
@@ -20,15 +20,18 @@ impl<TProtocol: Protocol> Tendermint<TProtocol> {
     /// will lead to a different proposal, but since the former proposal was not broadcast and was not acted on no harm is done, and the protocol
     /// is not breached.
     pub(crate) fn propose(&mut self) -> Result<Return<TProtocol>, ProtocolError> {
-        // Retrieve the set of proposals for the current round. Create the set if it does not exist yet.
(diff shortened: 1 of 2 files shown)


jsdanielh (approved):
LGTM

viquezclaudio (approved):

PR #3918, opened 2026-08-04
web-client: return `JsError` instead of panicking when signing in-staking transactions

jsdanielh:
Malformed `recipient_data` and unsupported sender types aborted the WASM runtime via `unwrap()` and `unwrap_basic()`. Propagate both as catchable errors.

## Pull request checklist

### DIFF
--- a/web-client/src/common/transaction.rs
+++ b/web-client/src/common/transaction.rs
@@ -146,6 +146,10 @@ impl Transaction {
     /// of a transaction can be different key pairs (addresses). If no inner key pair is provided, the outer
     /// key pair is used for both signatures.
     ///
+    /// Throws when the transaction's recipient data is not valid data for an incoming staking
+    /// transaction, or when an incoming staking transaction is not sent from a basic account or
+    /// a vesting contract.
+    ///
     /// ### Limitations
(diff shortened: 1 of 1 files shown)


web-client/src/common/transaction.rs
```diff
@@ -396,11 +400,17 @@ impl Transaction {
         let proof = match proof_builder {
             TransactionProofBuilder::Basic(mut builder) => {
                 builder.sign_with_key_pair(key_pair.native_ref());
-                builder.generate().unwrap().proof
+                builder
+                    .generate()
+                    .ok_or_else(|| JsError::new("Failed to sign transaction"))?
```
sisou, line 405: Is there any information in the error generated by `generate` that we could use for a more specific error message?

sisou: I opened the diff in my editor and saw that `generate()` returns an Option. So no information of why it failed. 

IMO, these `unwraps` here cannot fail, as the condition for them failing would be that the signature has not been set (correctly), but it is definitively set by the line before each `generate()`. So IMO these converting of unwrap-to-error is unnecessary here, but I see the "defense-in-depth" situation, so you have my approval.

sisou (approved):

PR #3922, opened 2026-08-10
Apply back pressure between the block queue and live sync

nibhar:
`BlockQueueProxy` forwarded queued blocks to the live sync consumer over an unbounded channel. The consumer only polls that channel while it has no push operation in flight, so any slow blockchain push leaves the background task draining the block queue at line rate into a buffer with no limit.

Announcements are the cheapest way to fill it: they are admitted on `verify_header` alone, which performs no signature check for micro blocks, and `QueuedBlock::Head` is emitted for anything whose parent is known — the per-peer limit in the buffer does not cover that path.

Use a bounded channel and reserve capacity via `PollSender::poll_reserve` before polling the queue, so a block is never taken that cannot be delivered. When the channel is full the task parks without touching the queue, which stalls the block stream and lets the network layer drop the excess, as it did before the queue moved to its own task.

Reserving before locking keeps the parked task from holding the queue lock, and the poll result is bound to a local so the lock is released before sending.

Tests cover both halves of the contract: back pressure reaches the announcing side once the consumer stalls, and no announcement is dropped on the way through.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3971,6 +3971,7 @@ dependencies = [
  "thiserror 2.0.20",
  "tokio",
  "tokio-stream",
+ "tokio-util",
  "tracing",
 ]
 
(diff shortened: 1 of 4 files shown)


viquezclaudio (approved):

jsdanielh (approved):
LGTM

PR #3938, opened 2026-08-13
Update tooling for signal upgrade with the warm key

jsdanielh:
- mktx: Add `validator set-signal-data` and `validator signal-version`, the warm-key `SetSignalData` commands every other client surface already has.
  Drop `validator signal-upgrade`: it built a cold-key `UpdateValidator` and zeroed the non-version bytes of `signal_data`. The chain is on protocol version 2, so the warm key suffices; `validator update --new-signal-data` still covers the cold-key case.
  With it gone, `version_to_signal_data` has no callers. It duplicated `Policy::signal_data_for_version`, so remove it.

- `set-signal-data` and `signal-version` warned that they are only valid once the chain reaches the protocol version that introduces them. The chain is on version 2, so the warning is noise at the point of use.
  Point `update-validator --new-signal-data` at `signal-version` instead, since it needs the validator (cold) wallet unlocked and signaling no longer does.

## Pull request checklist

### DIFF
--- a/rpc-client/src/subcommands/validator_subcommands.rs
+++ b/rpc-client/src/subcommands/validator_subcommands.rs
@@ -86,7 +86,9 @@ pub enum ValidatorCommand {
         #[clap(long)]
         new_reward_address: Option<Address>,
 
-        /// The new signal data showed by the validator.
+        /// The new signal data showed by the validator. For upgrade signaling prefer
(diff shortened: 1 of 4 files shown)


jeffesquivels (approved):
LGTM

viquezclaudio (approved):

PR #3939, opened 2026-08-14
Staking: Account for and validate inactive stake when creating stakers

ii-cruz:
`create_staker` and `revert_create_staker` only accounted for the staker's active balance when updating the staking contract balance. Although runtime `CreateStaker` transactions always use a zero inactive balance and no inactivation height, genesis configurations can provide both fields directly.

This PR includes the inactive balance during creation and reversion so the staking contract balance remains consistent with its stored staker balances. It also enforces the staker invariants requiring a non-zero inactive balance to have an inactivation height, while a zero inactive balance must have none. Invalid combinations are rejected with `InvalidForRecipient` before any state or log mutation.

Regression coverage verifies both invalid combinations, confirms that rejection leaves the staking contract, validator, staker store, and transaction log unchanged, and covers creating and reverting a staker with a non-zero inactive balance.

Current mainnet and testnet configurations are unaffected, but this hardens manual genesis configurations and prevents future networks or callers from creating inconsistent staking state.

Closes BUG-112 and COR-30 on linear

## Pull request checklist

### DIFF
--- a/primitives/account/src/account/staking_contract/staker.rs
+++ b/primitives/account/src/account/staking_contract/staker.rs
@@ -203,6 +203,13 @@ impl StakingContract {
             });
         }
 
+        // Transaction-created stakers always use (zero, None), but genesis configurations can set
+        // these fields directly. Inactive stake and its inactivation height must either both be
+        // present or both be absent (invariants 3 and 4).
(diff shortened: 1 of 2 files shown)


jsdanielh (approved):
LGTM

viquezclaudio (approved):

PR #3940, opened 2026-08-14
Staking: Enforce validator status invariants on creation

ii-cruz:
This hardens validator creation by rejecting states where a validator is active while retired or while its jail period is still in effect.

`create_validator` now evaluates validator status at an explicit block height. Runtime transactions provide the current block height, while genesis generation provides its configured genesis height. This preserves valid active validators with expired jail history and keeps `active_validators` consistent with `Validator::is_active()`.

Tests cover:

- Atomic rejection of active retired or currently jailed validators.
- Acceptance at the exact jail-release boundary.
- Valid inactive, jailed, and retired status combinations.
- Propagation of the configured genesis block height.

Current mainnet and testnet configurations are unaffected, but this prevents future networks or callers from accidentally creating an inconsistent genesis state.

Closes COR-29 on linear

## Pull request checklist

### DIFF
--- a/primitives/account/src/account/staking_contract/validator.rs
+++ b/primitives/account/src/account/staking_contract/validator.rs
@@ -170,6 +170,7 @@ pub struct Tombstone {
 impl StakingContract {
     /// Creates a new validator. The initial stake is always equal to the validator deposit
     /// and can only be retrieved by deleting the validator.
+    /// `block_number` is the height at which the supplied validator state is evaluated.
     /// This function is public to fill the genesis staking contract.
     pub fn create_validator(
         &mut self,
@@ -183,6 +184,7 @@ impl StakingContract {
         inactive_from: Option<u32>,
         jailed_from: Option<u32>,
         retired: bool,
+        block_number: u32,
         tx_logger: &mut TransactionLog,
     ) -> Result<(), AccountError> {
         // Fail if the validator already exists.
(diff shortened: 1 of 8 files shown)


primitives/account/src/account/staking_contract/validator.rs
```diff
@@ -21,6 +21,7 @@ use crate::{
     JailValidatorReceipt, Log, RetireValidatorReceipt, TransactionLog,
 };
 
+#[allow(clippy::doc_overindented_list_items)] // Used to typeset in monospace.
```
jsdanielh: Mmm I am not a fan of adding it. Why is that we use overindentation? `rustdoc` won't be correctly generated

viquezclaudio (approved):

primitives/account/src/account/staking_contract/validator.rs
```diff
@@ -49,6 +50,14 @@ use crate::{
 ///      in the first place.
 /// (**) The validator may be set to automatically reactivate itself upon inactivation.
 ///      If this setting is not enabled the state change can only be triggered manually.
+/// (***) The validator has a set of status invariants:
+///             (invariant 1) Retired validators are marked inactive:
+///                           `retired` implies `inactive_from.is_some()`
+///             (invariant 2) Validators whose jail period is still active are marked inactive:
+///                           `is_jailed(block_number)` implies `inactive_from.is_some()`
```
jsdanielh, line 56:
```suggestion
/// (***) The validator has a set of status invariants:
///            (invariant 1) Retired validators are marked inactive:
///               `retired` implies `inactive_from.is_some()`
///            (invariant 2) Validators whose jail period is still active are marked inactive:
///               `is_jailed(block_number)` implies `inactive_from.is_some()`
```

primitives/account/src/account/staking_contract/validator.rs
ii-cruz: This comes from copying from `valdiator.rs`, it's actually not needed, so i removed it. thanks!

ii-cruz: it came from copy pasting from `staker.rs`. This was not actually needed on this file so i removed it.

primitives/account/src/account/staking_contract/validator.rs
ii-cruz: I would prefer it to stay as it is for readability. This doesn't trigger any Clippy warnings, so should be fine right?

jsdanielh (approved):
LGTM

PR #3941, opened 2026-08-17
Bounds check slot numbers in ValidatorRegistry

nibhar:
ValidatorRegistry narrowed slot numbers to u16 before checking them against Policy::SLOTS. Out of range slots therefore either truncated onto an existing slot or reached Validators::get_band_from_slot untruncated and tripped its assert!:

weight returned Some(1) for any id congruent to a valid slot mod 2¹⁶, inflating the weight compared against TWO_F_PLUS_ONE.
public_key returned a real validator's voting key for a slot that does not exist, folding a non-signing validator into the aggregate public key instead of failing as UnknownSigner. public_key and signers_identity panicked on slots in 512..=65535.

All three now go through a checked_slot helper that checks the bound in usize before narrowing, the same shape as checked_signer_slots in the block crate.

Not reachable from the network today: the Checked wrapper on SerializableLevelUpdate rejects contributor bits >= Policy::SLOTS at deserialization, and both entry points (SkipBlockUpdate, TendermintUpdate) go through it. This is defense in depth for the case where a contribution reaches the registry without that check.

Tests were added in a preceding commit and cover both layers: the registry functions at the range boundaries and at truncating ids, and the Checked wrapper itself, which had no coverage.

### DIFF
--- a/validator/src/aggregation/registry.rs
+++ b/validator/src/aggregation/registry.rs
@@ -25,11 +25,18 @@ impl ValidatorRegistry {
     }
 }
 
+/// Narrows a slot number to the `u16` the `Validators` accessors take, rejecting slots that do not
+/// exist. The bound must be checked before the cast, as truncating first maps out of range slots
+/// onto existing ones. `Validators::get_band_from_slot` panics on anything this rejects.
+fn checked_slot(slot: usize) -> Option<u16> {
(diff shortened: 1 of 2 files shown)


jsdanielh (approved):
LGTM

PR #3981, opened 2026-09-21
blockchain: Always clear a reverted block from the validity store

viquezclaudio:
`ValidityStore::delete_block_transactions` returned early when the store held transactions of a single block number only (`first_bn == last_bn`). `remove_block` relies on it when a rebranch reverts a block, so in that state the reverted block's transaction hashes stayed marked as included. The node then rejected the canonical block that re-included them with `DuplicateTransaction` while every other node accepted it, leaving it stuck at that height — a fork of that node off the network. The mempool consults the same store, so the node also refused the transaction when its sender re-submitted it.

The guard is not needed: an emptied store is harmless, `first_bn` and `last_bn` fall back to 0, `update_validity_store` repopulates it with the next block, and removing an absent key is a no-op in the database layer.

Pinned by a validity-store unit test that reverts the only tracked block.

...
## Pull request checklist


jsdanielh (approved):
LGTM

